PR #4, opened 2014-07-17
Add stock_picking_mass_assign

guewen:
a module that allow to check the availability on a selection of delivery
orders, and an all the confirmed delivery orders by a scheduled action.

### DIFF
--- a/stock_picking_mass_assign/stock.py
+++ b/stock_picking_mass_assign/stock.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Author: Guewen Baconnier
+#    Copyright 2014 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import logging
+
+from openerp.osv import orm
+
+_logger = logging.getLogger(__name__)
+
+
+class stock_picking(orm.Model):
+
+    _inherit = 'stock.picking'
+
+    def check_assign_all(self, cr, uid, ids=None, context=None):
+        """ Try to assign confirmed pickings """
+        if isinstance(ids, (int, long)):
+            ids = [ids]
+        if ids is None:
+            domain = [('type', '=', 'out'),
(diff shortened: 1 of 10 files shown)


stock_picking_mass_assign/stock.py
```diff
@@ -0,0 +1,52 @@
...
+#
+##############################################################################
+import logging
+
+from openerp.osv import orm
+
+_logger = logging.getLogger(__name__)
+
+
+class stock_picking(orm.Model):
```
pedrobaeza, line 28: Better to use CapsWords class name convention: StockPicking

stock_picking_mass_assign/wizard/check_assign_all.py
```diff
@@ -0,0 +1,48 @@
...
+        if not picking_ids:
+            raise orm.except_orm(
+                _('Error'),
+                _('No selected delivery orders'))
+
+        picking_obj = self.pool['stock.picking']
+        domain = [('type', '=', 'out'),
+                  ('state', '=', 'confirmed'),
+                  ('id', 'in', picking_ids)]
+        picking_ids = picking_obj.search(cr, uid, domain,
```
pedrobaeza, line 44: You are overwritten context.get('active_ids') value always. I think you don't need to make this here. You are already searching for that domain on stock.picking method if ids argument is empty (for cron job). This is an interactive wizard.

guewen, line 44: This is:
1. to filter only on confirmed and out pickings since users can select non-confirmed pickings
2. to apply a sql `order by` so they are confirmed in the correct order

rvalyi:
Hello,

The Travis failure is unrelated. Feature seems OK. Last explanation too. CamelCase would be better for both for stock_picking and check_assign_all classes, but this is a detail. I give my approval already.

:+1:

stock_picking_mass_assign/stock.py
bwrsandman, line 28: Additionally, the filename should match the model:
- stock_picking.py

lepistone:
The 7.0 branch is now green, please rebase on 7.0 so that we can see if this branch is green as well.

Thanks!

bwrsandman:
@lepistone no need, I restarted the build ;)

lepistone:
thanks @bwrsandman I wasn't aware of that. Can you elaborate on why that works if we didn't update the proposed branch?

bwrsandman:
The proposed branch wasn't updated, but restarting the build makes travis redo the merge and refetch the test suite from `maintainer-quality-tools`, since the repo is now green and this PR doesn't break the tests, the tests remain green.

lepistone:
@bwrsandman, I found in the travis doc what I didn't know: http://docs.travis-ci.com/user/pull-requests. It says:

> Rather than test the commits from the branches the pull request is sent from, we test the merge between the origin and the upstream branch.

gurneyalex:
:+1:

pedrobaeza:
Don't merge yet, because I need to check what Guewen's says.

lepistone:
@pedrobaeza any news on that one? thanks!

lepistone:
:+1:

pedrobaeza:
OK, I have checked it and it's OK. I proceed with the merge. My two style comments can be addressed later.

PR #8, opened 2014-08-18
[fix] return list instead of set

lepistone:
This problem was caught by the tests: the set is used in a domain
and OpenERP expects a list, not a set.


rdeheele:
:+1:

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #10, opened 2014-08-19
7.0 add update date

lepistone:


stock_picking_update_date/model/picking.py
```diff
@@ -0,0 +1,139 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm, fields
+
+
+class Picking(orm.Model):
+
+    _name = 'stock.picking'
```
pedrobaeza: This is not needed, because the name is the same as inherited.

stock_picking_update_date/model/picking.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Please name it _stock_picking.py_

stock_picking_update_date/model/picking.py
```diff
@@ -0,0 +1,139 @@
...
+                'stock.move': (
+                    _move_trigger, ['date_expected'], 10
+                )
+            }),
+    }
+
+
+class PickingIn(orm.Model):
+
+    _name = 'stock.picking.in'
```
pedrobaeza: This is not needed.

stock_picking_update_date/model/picking.py
```diff
@@ -0,0 +1,139 @@
...
+                'stock.move': (
+                    _move_trigger, ['date_expected'], 10
+                )
+            }),
+    }
+
+
+class PickingOut(orm.Model):
+
+    _name = 'stock.picking.out'
```
pedrobaeza: This is not needed

pedrobaeza:
Hi, @lepistone, thanks for the PR.

Can you explain why the standard field cannot be used for this purpose?

Regards.

stock_picking_update_date/model/picking.py
bwrsandman: Adding to that, please only put one class/model per file and name the class and file after the model.

lepistone:
Thanks for reviewing @pedrobaeza and @bwrsandman. I addressed your remarks. I didn't split the classes in separate files, because they are really small and related and the resulting file is still very short.

pedrobaeza:
:+1: 

Thanks

rvalyi:
:+1:  LGTM

gurneyalex:
:+1:

guewen:
:+1:

PR #15, opened 2014-09-03
[IMP] cancel a picking dispatch : disassociate linked moves...

### LINKED ISSUE #13: cancel meaning of a picking dispatch (functional issue)
_Steps_
Create a picking dispatch
Add some moves.
Cancel the picking dispatch

_Identified behavior_
The moves and by extension the picking are cancelled too.

_Expected behavior_
In a functional context, it seems more coherent to "desassociate" moves from the cancelled dispatch, and don't change the move states.

Any opinions?

rdeheele:
instead of cancel them, fixes #13


rdeheele:
not such a good idea, see #13

PR #16, opened 2014-09-04
[IMP] from a wave, a dispatch can be directly assigned

rdeheele:
When the user asks a wave of x pickings in the wizard, it assigns the wave to himself or another user (then the action to assign the wave to someone is already done), then we can directly consider the dispatch as assigned.


lepistone:
:+1: thanks

gurneyalex:
:+1:

PR #18, opened 2014-09-04
Corrections on stock_picking_update_date:

guewen:
- The views were not loaded because the declared models were not using the
  out/in suffixes
- The date_expected field was updated to the last min_date value instead
  of the new one due to the ordering of the function fields


lepistone:
Thanks a lot for fixing that! :+1:

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #19, opened 2014-09-04
7.0: add compute date

lepistone:
This module allows to recompute the delivery dates of outgoing moves based on when we expect to receive the goods.

The logic is different for "Make To Order" and "Make To Stock" products and is described in the **openerp**.py.

In the yaml tests you can find three use cases.

Thanks!


jgrandguillaume:
Hi Leonardo,

I just finished to test this new module and it's just working like a charm ! 

Code review + test ok for me ! Thanks for the good job.

:+1:

guewen:
Seems that running several times without new 'in' moves may change the dates, just need to investigate this further.

stock_picking_compute_delivery_date/wizard/all_products.xml
```diff
@@ -0,0 +1,34 @@
...
+    </record>
+
+    <act_window name="Compute delivery dates for all products"
+      res_model="compute.all.delivery.dates.wizard"
+      src_model="product.product"
+      key2="client_action_multi"
+      view_mode="form"
+      target="new"
+      groups="base.group_no_one"
+      id="product_compute_all_delivery_dates_action"/>
```
guewen, line 26: This is misleading because it appears in 'Other options' on the products but will run on all the products, not the selected one. At least it appears only for the technical group. Is there a reason to have the entry in "Other Options" or could we just keep the menuitem.

guewen:
When a part of the moves fall in the PlanFinished trap, they keep their current 'date_expected'. I know this was on purpose, leaving room to implement that later, but I think that we should implement it now otherwise the computation will be biased: as the moves are sorted by `date_expected`, the ones that were in PlanFinished will be the first to be included the next time we run the computation, hence they will gain priority over the others and the others will be delayed. I think that we should write the last date of the plan (or last date of the plan + 1) on the moves that have not enough stock to be considered.
I will write a test and make a proposal.

lepistone:
@guewen +1

jgrandguillaume:
Hi,

Just want to put my input about last @guewen comments:
- First you're complitely right, it doesn't work now as we expect this to work
- My idea at first was to order the moves not by expected date, but by creation date of the picking

Want we want here is : First arrived, first served.

So my suggestion would be that we order the moves by creation_date as it represent who should be delivered first.

This is an easy way to solve all this what do you think ?

Regards,

Joël

lepistone:
I like your idea @jgrandguillaume, I did that. Thanks!

jgrandguillaume:
Looks good now ! Thanks for the fix. :+1:

guewen:
Using the `date` is fine because then the order will stay the same each time we launch the computation.
:+1:

PR #20, opened 2014-09-05
[ADD] add fnct_search method on related_dispatch_ids field

rdeheele:
It allows to search pickings on tree view by dispatch name.

### DIFF
--- a/picking_dispatch/picking_dispatch.py
+++ b/picking_dispatch/picking_dispatch.py
@@ -366,20 +366,8 @@ class StockPickingIn(Model):
 
     def _search_dispatch_pickings(self, cr, uid, obj, name, args,
                                   context=None):
-        dispatch_ids = []
-        picking_ids = []
-        for arg in args:
-            if isinstance(arg, tuple):
-                dispatch_ids = arg[2]
-        if dispatch_ids:
-            move_obj = self.pool['stock.move']
-            move_ids = move_obj.search(cr, uid,
-                                       [('dispatch_id', 'in', dispatch_ids)],
-                                       context=context)
-            for move in move_obj.browse(cr, uid, move_ids, context=context):
-                if move.picking_id.id not in picking_ids:
-                    picking_ids.append(move.picking_id.id)
-        return [('id', 'in', picking_ids)]
+        return super(StockPickingIn, self)._search_dispatch_pickings(
+            cr, uid, obj, name, args=args, context=context)
 
     _columns = {
         'related_dispatch_ids': fields.function(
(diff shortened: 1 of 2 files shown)


picking_dispatch/picking_dispatch.py
```diff
@@ -361,9 +397,27 @@ def _get_related_dispatch(self, cr, uid, ids, field_names, arg=None,
         return super(StockPickingOut, self)._get_related_dispatch(
             cr, uid, ids, field_names, arg=arg, context=context)
 
+    def _search_dispatch_pickings(self, cr, uid, obj, name, args,
+                                  context=None):
+        dispatch_ids = []
```
pedrobaeza: Why don't you call here to the same method you have on 'stock.picking'?

rdeheele:
thanks Pedro, I change it as you have proposed.

pedrobaeza:
:+1: 

Thanks.

lepistone:
:+1:

picking_dispatch/picking_dispatch.py
```diff
@@ -331,9 +331,27 @@ def _get_related_dispatch(self, cr, uid, ids, field_names, arg=None,
...
+                                       context=context)
+            for move in move_obj.browse(cr, uid, move_ids, context=context):
+                if move.picking_id.id not in picking_ids:
+                    picking_ids.append(move.picking_id.id)
+        return [('id', 'in', picking_ids)]
+
     _columns = {
         'related_dispatch_ids': fields.function(
-            _get_related_dispatch, method=True, type='one2many',
+            _get_related_dispatch, method=True,
```
guewen, line 353: `method=True` is deprecated

picking_dispatch/picking_dispatch.py
```diff
@@ -361,9 +385,15 @@ def _get_related_dispatch(self, cr, uid, ids, field_names, arg=None,
...
 
+    def _search_dispatch_pickings(self, cr, uid, obj, name, args,
+                                  context=None):
+        return super(StockPickingOut, self)._search_dispatch_pickings(
+            cr, uid, obj, name, args=args, context=context)
+
     _columns = {
         'related_dispatch_ids': fields.function(
-            _get_related_dispatch, method=True, type='one2many',
+            _get_related_dispatch, method=True,
```
guewen, line 395: `method=True` is deprecated

picking_dispatch/picking_dispatch.py
```diff
@@ -331,9 +331,27 @@ def _get_related_dispatch(self, cr, uid, ids, field_names, arg=None,
...
+        for arg in args:
+            if isinstance(arg, tuple):
+                dispatch_ids = arg[2]
+        if dispatch_ids:
+            move_obj = self.pool['stock.move']
+            move_ids = move_obj.search(cr, uid,
+                                       [('dispatch_id', 'in', dispatch_ids)],
+                                       context=context)
+            for move in move_obj.browse(cr, uid, move_ids, context=context):
+                if move.picking_id.id not in picking_ids:
```
guewen, line 347: `picking_ids` should be a `set` so you would not need to check for the presence of the id

lepistone:
Sorry @guewen for the bad timing for the merge. If there are no blocking issues, I'd ask @rdeheele make a new PR to address your remarks.

Thank you!

picking_dispatch/picking_dispatch.py
```diff
@@ -331,9 +331,27 @@ def _get_related_dispatch(self, cr, uid, ids, field_names, arg=None,
...
+                                  context=None):
+        dispatch_ids = []
+        picking_ids = []
+        for arg in args:
+            if isinstance(arg, tuple):
+                dispatch_ids = arg[2]
+        if dispatch_ids:
+            move_obj = self.pool['stock.move']
+            move_ids = move_obj.search(cr, uid,
+                                       [('dispatch_id', 'in', dispatch_ids)],
```
guewen, line 344: Any operator can be used when searching, not only `in`. I think they should not all translate to the domain `('dispatch_id', 'in', dispatch_ids)`. Some operators such as '<', '>' should raise an error (use a whitelist of allowed operators). `not in` or `=` should directly be used in the domain instead of `in`.

guewen:
No blocking issue but my last comment especially worth a fix because the search could return unexpected values when using another operator than `in`.

lepistone:
@guewen I think we can manage to fast-track that, like today, since these fixes would have normally gone here. Thanks

PR #24, opened 2014-09-17
[UPD] improve related_dispatch_ids fnct_search function

rdeheele:
linked to @guewen advices on #20 (thanks!):
- improves fnct_search function
- delete deprecated "method=True"

### DIFF
--- a/picking_dispatch/picking_dispatch.py
+++ b/picking_dispatch/picking_dispatch.py
@@ -333,26 +333,29 @@ class StockPicking(Model):
 
     def _search_dispatch_pickings(self, cr, uid, obj, name, args,
                                   context=None):
-        dispatch_ids = []
-        picking_ids = []
-        for arg in args:
-            if isinstance(arg, tuple):
-                dispatch_ids = arg[2]
-        if dispatch_ids:
+        if not len(args):
+            return []
+        ops = ['ilike', 'not ilike', '=', '!=']
+        picking_ids = set()
+        for field, symbol, value in args:
+            if symbol not in ops:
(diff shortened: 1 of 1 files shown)


picking_dispatch/picking_dispatch.py
```diff
@@ -333,26 +333,29 @@ def _get_related_dispatch(self, cr, uid, ids, field_names, arg=None,
...
                                   context=None):
-        dispatch_ids = []
-        picking_ids = []
-        for arg in args:
-            if isinstance(arg, tuple):
-                dispatch_ids = arg[2]
-        if dispatch_ids:
+        if not len(args):
+            return []
+        ops = ['ilike', 'not ilike', '=', '!=']
```
guewen: I guess this list should be:

```
ops = ['in', 'not in', '=', '!=']
```

rdeheele: I thought it too before I dig further: the not advanced search uses 'ilike'. When we make a search by picking dispatch, we search by dispatch name, dispatchs are found via name_search.

guewen: Oh! Ok, it makes sense. Still, 'in' and 'not in' could be added since they would also be handled properly if a list of ids is passed (example when doing a direct search on the field from a python method).

In fact, after thinking, I think that the restriction could be removed entirely: the method only dispatches the search from a model to another, so there is no reason to restrict the operators, the search() on stock.move will eventually do.

guewen:
:+1: Thanks!

gurneyalex:
:+1:

PR #25, opened 2014-09-25
Do not remove the 'min_date' field but hide it

guewen:
Because it may break other views using 'min_date' as reference


pedrobaeza:
:+1:

tafaRU:
:+1:

PR #26, opened 2014-10-11
[IMP] add the possibility to reopen a stock move from the form view.

sebastienbeau:
This is usefull when you received by using the menu Incomming Shipments and Incomming Product

### DIFF
--- a/stock_cancel/stock.py
+++ b/stock_cancel/stock.py
@@ -33,19 +33,11 @@ class stock_move(orm.Model):
     def action_revert_done(self, cr, uid, ids, context=None):
         picking_obj = self.pool['stock.picking']
         for move in self.browse(cr, uid, ids, context):
-            if picking_obj.has_valuation_moves(cr, uid, move):
-                raise orm.except_orm(
-                    _('Error'),
-                    _('Line %s has valuation moves. \
-                        Remove them first') % (move.name))
-            move.write({'state': 'draft'})
-            if move.picking_id:
-                picking_obj._reopen_picking(
-                    cr, uid, move.picking_id, context=context)
-
+            picking_obj._reopen_picking(
+                cr, uid, move.picking_id, [move], context=context)
         for (id, name) in self.name_get(cr, uid, ids):
             message = _(
-                "The stock move '%s' has been set in draft state."
+                "The stock move '%s' has been set in confirm state."
             ) % (name,)
             self.log(cr, uid, id, message)
         return True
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The problem I see with this is that you reopen entired picking although you only select one stock move. This must be at least warn to the user with `confirmation` attribute at button level.

sebastienbeau:
You're right.
After thinking twice the picking will be at the "draft" state, maybe we should also reconfirm it? What do you think?

pedrobaeza:
You can check if all moves of the picking are in draft state, and then reopen picking (if it's not already open)

sebastienbeau:
Hum
In case that I have a done picking with 2 move. If I reopen one of this two move. What is the best scenario
- I set back to draft the move I want to re-open (the second will stay at the done state) and I set back the picking to draft. Now the user will have to confirm again the picking manually
- I set back to draft the move I want to re-open (the second will stay at the done state) and I set back the picking to draft and confirm it. The re-open move and the picking are in the confirm state (the orther move is in the done state)

What do you think?

gurneyalex:
@sebastienbeau I'd go for the second option. 

Please add an automated test :smile_cat:

pedrobaeza:
Yeah, second option seems the best one.

sebastienbeau:
Hi sorry for being so long. (customer want do not want want again the feature)
So I am back.

After thinking twice, If I want to have a clean solution for cancelling a move from the incomming and the outgoing. I need to change some behaviour. Let explain.

An incoming move is always in the state assigned (during the creation from the PO it's force to be set in assigned). If I cancel a move or the full picking, the moves will be in draft state and confirming the picking will put them in "confirmed" but never in "assigned". In my case I want that the "revert" do exactly the opposite of "done" so I need them to be back to "assigned" and not in "draft" then "confirmed". The only solution I have found is to automatically confirm the picking (for "in" and "out" and force the assigned in the "in" case).

As it change the current behaviour (but I have no other clean solution) I totaly understand that this PR can be rejected. So before I do the unitest I would like to know If you are ok that the "action revert" do exactly the opposite of the "done" button and set back the picking to confirmed instead of draft. And in the case of incoming move it will set the line to "assigned" instead of "draft"

What do you think?

sebastienbeau:
@pedrobaeza @gurneyalex what do you think?

gurneyalex:
@sebastienbeau (sorry for the late answer) the solution you suggest sounds ok to me.

sebastienbeau:
@gurneyalex so we can merge it right?

pedrobaeza:
We all agree on the topic, so I merge.

PR #28, opened 2014-10-15
[IMP] select only moves with internal source location (avoid to select dropshipping moves)

rdeheele:
It makes sense to grab only moves coming from an internal source location, for example it avoids to select dropshipping moves.


lepistone:
:+1: thanks, and what a nice commit message! :innocent:

pedrobaeza:
:+1:

gurneyalex:
:+1:

PR #41, opened 2014-12-04
[FIX: product_serial] Last Location Search Function

### LINKED ISSUE #40: Last Location search fix for product_serial
The _last_location_id_search method provided for last_location field added to stock.production.lot in product_serial/stock.py currently only serve the Prodlots shortcut where it would provide the domain in the [('last_location_id', '=', 9)] format where 9 is the database ID of stock.location being search.  It does not suppoert searching by any means and hence searching through Advance Filter for Last Location generates various error.  The patch here can also accomodate both Prodlots short cuts and search through Advance Filter:

``` udiff
diff --git a/product_serial/stock.py b/product_serial/stock.py
index 79ffae1..0e8558f 100644
--- a/product_serial/stock.py
+++ b/product_serial/stock.py
@@ -366,26 +366,44 @@ class stock_production_lot(orm.Model):
         return res

     def _last_location_id_search(self, cr, uid, obj, name, args, context=None):
-        ops = ['=', ]
-        prodlot_ids = ()
-        if not len(args):
-            return []
-        prodlot_ids = []
-        for a in args:
-            operator = a[1]
-            value = a[2]
-            if operator not in ops:
-                raise orm.except_orm(
-                    _('Error !'),
-                    _('Operator %s not supported in searches for '
-                      'last_location_id (product.product).' % operator))
-            if operator == '=':
-                cr.execute(
-                    "select distinct prodlot_id "
-                    "from stock_report_prodlots "
-                    "where location_id = %s and qty > 0 ",
-                    (value, ))
-                prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
+        _not = False
+        _loc_pool = self.pool.get('stock.location')
+        _loc_ids = []
+        _is_null = False
+        for _leaf in args:
+            if _leaf[0] == 'last_location_id':
+                if isinstance(_leaf[-1], (int, long)):
+                    _val_id = _leaf[-1]
+                else: 
+                    _val = _leaf[-1]
+
+                _op = _leaf[1].strip()
+                _not = (('not' in _op) or ('!' in _op) or ('<' in _op) or
+                        ('>' in _op))
+                if isinstance(_leaf[-1], (bool)):
+                    if _leaf[1] == '=':
+                        _is_null = True
+                    else:
+                        _loc_ids.append(0)
+                elif isinstance(_leaf[-1], (int, long)):
+                    _loc_ids.append(_leaf[-1])
+                else: 
+                    _locs = _loc_pool.name_search(cr, uid, _leaf[-1], context=context)
+                    _locs = tuple([x[0] for x in _locs])
+                    _loc_ids.extend(_locs)
+
+        if _is_null:
+            cr.execute(
+                "SELECT DISTINCT prodlot_id "
+                "FROM stock_report_prodlots "
+                "WHERE location_id IS NULL AND qty > 0")
+        else:
+            cr.execute(
+                "SELECT DISTINCT prodlot_id "
+                "FROM stock_report_prodlots "
+                "WHERE location_id %s IN %%s AND qty > 0 " % (_not and 'NOT' or ''),
+                (tuple(_loc_ids), ))
+        prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
         return [('id', 'in', tuple(prodlot_ids))]

     _columns = {
```

niecw:
Last Location search function has been improved to cater for Advance Filter from search bar

closes #40

### DIFF
--- a/product_serial/stock.py
+++ b/product_serial/stock.py
@@ -368,14 +368,13 @@ class stock_production_lot(orm.Model):
     def _last_location_id_search(self, cr, uid, obj, name, args, context=None):
         _not = False
         _loc_pool = self.pool.get('stock.location')
-        _loc_ids = []
+        _loc_ids = [0]
         _is_null = False
         for _leaf in args:
             if _leaf[0] == 'last_location_id':
                 _op = _leaf[1].strip()
-                _not = (('not' in _op) or ('!' in _op) or ('<' in _op) or
-                        ('>' in _op))
-                if isinstance(_leaf[-1], (bool)):
+                _not = _op not in ('in', 'ilike', '=')
+                if isinstance(_leaf[-1], bool):
                     if _leaf[1] == '=':
                         _is_null = True
                     else:
@@ -385,8 +384,7 @@ class stock_production_lot(orm.Model):
                 else:
                     _locs = _loc_pool.name_search(cr, uid, _leaf[-1],
                                                   context=context)
-                    _locs = tuple([x[0] for x in _locs])
-                    _loc_ids.extend(_locs)
+                    _loc_ids.extend([x[0] for x in _locs])
 
         if _is_null:
             cr.execute(
@@ -398,7 +396,7 @@ class stock_production_lot(orm.Model):
                 "SELECT DISTINCT prodlot_id "
                 "FROM stock_report_prodlots "
                 "WHERE location_id %s IN %%s AND qty > 0 " %
-                (_not and 'NOT' or ''),
+                ('NOT' if _not else ''),
                 (tuple(_loc_ids), ))
         prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
         return [('id', 'in', tuple(prodlot_ids))]
--- a/product_serial/stock.py
+++ b/product_serial/stock.py
@@ -366,26 +366,44 @@ class stock_production_lot(orm.Model):
         return res
 
     def _last_location_id_search(self, cr, uid, obj, name, args, context=None):
-        ops = ['=', ]
-        prodlot_ids = ()
-        if not len(args):
-            return []
-        prodlot_ids = []
-        for a in args:
-            operator = a[1]
-            value = a[2]
-            if operator not in ops:
-                raise orm.except_orm(
-                    _('Error !'),
-                    _('Operator %s not supported in searches for '
-                      'last_location_id (product.product).' % operator))
-            if operator == '=':
-                cr.execute(
-                    "select distinct prodlot_id "
-                    "from stock_report_prodlots "
-                    "where location_id = %s and qty > 0 ",
-                    (value, ))
-                prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
+        _not = False
+        _loc_pool = self.pool.get('stock.location')
+        _loc_ids = []
+        _is_null = False
+        for _leaf in args:
+            if _leaf[0] == 'last_location_id':
+                if isinstance(_leaf[-1], (int, long)):
+                    _val_id = _leaf[-1]
+                else: 
+                    _val = _leaf[-1]
+
+                _op = _leaf[1].strip()
+                _not = (('not' in _op) or ('!' in _op) or ('<' in _op) or
+                        ('>' in _op))
+                if isinstance(_leaf[-1], (bool)):
+                    if _leaf[1] == '=':
+                        _is_null = True
+                    else:
+                        _loc_ids.append(0)
+                elif isinstance(_leaf[-1], (int, long)):
+                    _loc_ids.append(_leaf[-1])
+                else: 
+                    _locs = _loc_pool.name_search(cr, uid, _leaf[-1], context=context)
+                    _locs = tuple([x[0] for x in _locs])
+                    _loc_ids.extend(_locs)
+
+        if _is_null:
+            cr.execute(
+                "SELECT DISTINCT prodlot_id "
+                "FROM stock_report_prodlots "
+                "WHERE location_id IS NULL AND qty > 0")
+        else:
+            cr.execute(
+                "SELECT DISTINCT prodlot_id "
+                "FROM stock_report_prodlots "
+                "WHERE location_id %s IN %%s AND qty > 0 " % (_not and 'NOT' or ''),
+                (tuple(_loc_ids), ))
+        prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
         return [('id', 'in', tuple(prodlot_ids))]
 
     _columns = {
--- a/product_serial/stock.py
+++ b/product_serial/stock.py
@@ -372,11 +372,6 @@ class stock_production_lot(orm.Model):
         _is_null = False
         for _leaf in args:
             if _leaf[0] == 'last_location_id':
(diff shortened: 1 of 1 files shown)


product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
+        _not = False
+        _loc_pool = self.pool.get('stock.location')
+        _loc_ids = []
+        _is_null = False
+        for _leaf in args:
+            if _leaf[0] == 'last_location_id':
+                _op = _leaf[1].strip()
+                _not = (('not' in _op) or ('!' in _op) or ('<' in _op) or
+                        ('>' in _op))
+                if isinstance(_leaf[-1], (bool)):
```
gurneyalex: remove useless parenthesis aroung `bool`

product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
-                prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
+        _not = False
+        _loc_pool = self.pool.get('stock.location')
+        _loc_ids = []
+        _is_null = False
+        for _leaf in args:
+            if _leaf[0] == 'last_location_id':
+                _op = _leaf[1].strip()
+                _not = (('not' in _op) or ('!' in _op) or ('<' in _op) or
+                        ('>' in _op))
```
gurneyalex: I think this test would be clearer if written as: `if _op in ('not in', '!=', '<>')` (maybe adding some missing operators).

At the moment, what is matched by the test is not clear to me and I'd rather have an explicit list.

product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
+                if isinstance(_leaf[-1], (bool)):
+                    if _leaf[1] == '=':
+                        _is_null = True
+                    else:
+                        _loc_ids.append(0)
```
gurneyalex, line 381: it seems to me that this block assumes that the boolean value of _leaf[-1] is `False`, which is not guaranteed, I think.

product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
+                elif isinstance(_leaf[-1], (int, long)):
+                    _loc_ids.append(_leaf[-1])
+                else:
+                    _locs = _loc_pool.name_search(cr, uid, _leaf[-1],
+                                                  context=context)
+                    _locs = tuple([x[0] for x in _locs])
```
gurneyalex: you can simplify this a bit with `tuple(x[0 for x in _locs])`, and you could even get rid of the conversion to tuple here as it is done in the sql query below.

product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
+            cr.execute(
+                "SELECT DISTINCT prodlot_id "
+                "FROM stock_report_prodlots "
+                "WHERE location_id IS NULL AND qty > 0")
+        else:
+            cr.execute(
+                "SELECT DISTINCT prodlot_id "
+                "FROM stock_report_prodlots "
+                "WHERE location_id %s IN %%s AND qty > 0 " %
+                (_not and 'NOT' or ''),
```
gurneyalex: use the Python ternary operator : `'NOT' if _not else ''`

product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
+                (tuple(_loc_ids), ))
```
gurneyalex, line 400: you need to check that `_loc_ids` is not empty: I think this will cause an SQL exception. This can happen in case the `name_search` call returned nothing (e.g. because of a typo in the name typed by the user).

product_serial/stock.py
```diff
@@ -366,26 +366,41 @@ def _last_location_id(self, cr, uid, ids, field_name, arg, context=None):
...
+        prodlot_ids = filter(None, map(lambda x: x[0], cr.fetchall()))
```
gurneyalex, line 401: the call to filter is useless: we have no ids in the database with a value of 0 (and if we did have such ids, we probably would not want to remove them). Please use:

```
prodlot_ids = [x[0] for x in cr.fetchall()]
```

product_serial/stock.py
niecw: @gurneyalex it certainly can be written in the list like you have proposed.  However I was writing it to somewhat futureproof the implementation, so that later if they introduce new operators it can still be accomodated.  If it were implemented as a list for now it should be `if _op in ('not in', 'not ilike', '!=', '<=', '>=', '<>', '>', '<'):` or, for clarity sake it may be easier to specify the opposite: `if _op not in ('in', 'ilike', '=')`.  I'll use the last code for now (revision 1e64e15).

product_serial/stock.py
niecw, line 401: The filter is to catch lines that does not have prodlot_id as here it is operating on the search result from stock_report_prodlots view which does not guarantee that prodlot_id will always be filled.  The line does not change from it's original form.  I'm keeping it as it is for now.

product_serial/stock.py
niecw, line 381: @gurneyalex, yes it is not guaranteed if it is used by domain or other programming means.  My aim for now is to make it work from the filtering from the UI (search toolbar) which translate "is set" as `!= False` and "set" as `False`.

product_serial/stock.py
niecw: @gurneyalex do you mean `tuple(x[0] for x in _locs)`?  Anyway, I'm combining it with the line below and make it to be a list. Committed to revision 1e64e15.

product_serial/stock.py
niecw, line 400: Initializing `_loc_ids = [0]` so that it wouldn't be empty.  Committed to revision 1e64e15.

product_serial/stock.py
niecw: Committed to revision 1e64e15.

product_serial/stock.py
niecw: Committed to revision 1e64e15.

gurneyalex:
@rvayi could you look at this one?

gurneyalex:
@niecw could you add an automated test showing that this enhancement works?

gurneyalex:
@rvalyi could you give this PR a look?

yvaucher:
@niecw Your commits are missing an email. You need to configure your local git.

niecw:
Hi Yannick,

I've checked my local git configuration (git config -l), it does list 
user.email as one of the configuration item.  However I'm not sure 
whether the configuration is done.  I've rebase and amend the commits 
with the latest configuration and pushed it up to github.

yvaucher:
@niecw when amending you need to do use `--author="Name <email>"`

niecw:
@yvaucher Done.

yvaucher:
@niecw thanks it's good now your commits are correctly authored.

:+1:

gurneyalex:
:+1:

PR #42, opened 2014-12-15
8.0 stock route transit

gurneyalex:


stock_route_transit/test/stock_users.yml
```diff
@@ -0,0 +1,30 @@
...
+    password: sam
+    email: stockmanager@yourcompany.com
+-
+  I added groups for Stock Manager.
+-
+  !record {model: res.users, id: res_users_stock_manager}:
+    groups_id:
+      - stock.group_stock_manager
+-
+   Create a user as 'Stock User'
```
lepistone: this is not used.

stock_route_transit/__init__.py
```diff
@@ -0,0 +1,22 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Author: Alexandre Fayollle
```
lepistone: lll

stock_route_transit/__openerp__.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Author: Alexandre Fayollle
```
lepistone: lll

stock_route_transit/model/__init__.py
```diff
@@ -0,0 +1,22 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Author: Alexandre Fayollle
```
lepistone: lll

stock_route_transit/README.rst
```diff
@@ -0,0 +1,11 @@
+Stock Route Transit
+===================
+
+
+This addon adds Transit route configuration to warehouses.
```
lepistone: I'd love more explanations here.

lepistone:
Thanks @gurneyalex ! Globally looks good! :+1: . I fixed a minor lint issue myself.

Before merging that, I'd:
- expand the readme, at least to say what this module is supposed to do, and if the same result can be archieved with manual configuration of routes (or not)
- The test is long and it's not clear to me where we want to go. Even with this style, maybe some descriptions can be expanded where it makes sense to understand what the module does (i.e. `check the routes` -> `check that a special route "transit" has been generated` or something like that)
- A good part of the module seems to be tested, but maybe we can remove for now the missing test on the dropshipping case (not sure on this one). If you think it doesn't work yet, we can just note that in the readme.

Thanks!

lepistone:
Thanks for the improvements! It's :+1: for me.

jgrandguillaume:
Hi,

Functionnal test + code riview. LGTM :+1:  when travis green.

Thanks for this.

Joël

yvaucher:
:+1:

PR #44, opened 2014-12-24
Stock ownership availability rules

lepistone:


stock_ownership_availability_rules/README.rst
```diff
@@ -0,0 +1,7 @@
+Stock Ownership Availability Rules
+==================================
+
```
pedrobaeza, line 3: Can you please explain the purpose of the module?

lepistone, line 3: done!

stock_ownership_availability_rules/__openerp__.py
```diff
@@ -0,0 +1,34 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+{'name': 'Stock Ownership Availability Rules',
+ 'summary': 'XXX',
```
gurneyalex: Can you remove the entry or provide a meaningful summary?

stock_ownership_availability_rules/model/product.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Product(models.Model):
+    _inherit = 'product.template'
+
+    def action_open_quants(self, cr, uid, ids, context=None):
+        result = super(Product, self).action_open_quants(cr, uid, ids, context)
+        result['context'] = (
+            "{'search_default_locationgroup': 1, "
+            "'search_default_ownergroup': 1, "
+            "'search_default_internal_loc': 1}"
```
gurneyalex, line 29: curious... You can't use a real dict here?!

stock_ownership_availability_rules/model/quant.py
```diff
@@ -0,0 +1,59 @@
...
+                                   prefered_domain_list=None,
+                                   restrict_lot_id=False,
+                                   restrict_partner_id=False):
+        if domain is None:
+            domain = []
+        if prefered_domain_list is None:
+            prefered_domain_list = []
+
+        my_partner = location.company_id.partner_id
+        if restrict_partner_id == my_partner.id or not restrict_partner_id:
```
gurneyalex: Several points here:
1. i find the identifier 'my_partner' confusing in this context. 
2. nothing seems to be done about location.partner_id.id? Is it normal?
3. the condition in the test is not clear to me. I would have expected `and restrict_partner_id` instead of `or not restrict_partner_id`. Am I missing something obvious?

stock_ownership_availability_rules/model/product.py
lepistone, line 29: I think I can indeed. The string comes from the overridden method in addons/stock/product.py.

stock_ownership_availability_rules/model/quant.py
lepistone: 1. it means "my company's partner". What do you suggest?
2. No, I agree with you. If it is set, we can use that for "my_partner" as a first choice.
3. it means in english: "If there is no restriction, or if the restriction is on my_partner" which mean the same.

In general, I'm open to suggestions to refactor this.

gurneyalex: re 1. -> location_partner_id

lepistone: Refactored and docstring added.

gurneyalex: re 3: -> OK got it.  I suggest you write the condition in the same way you've explained it in english (i.e. permute the operands of `or` which sounds more natural to me). Side question: is it possible for `my_partner.id` to be `False`?

lepistone:
@gurneyalex you're right with your "side question": if a location has no partner, and no company (neither is required), what can I use as "own" stock? `self.env.user.company_id.partner_id`?

gurneyalex:
@lepistone: sounds good to me.

gurneyalex:
:+1:

stock_ownership_availability_rules/model/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
yvaucher: Missing license header

stock_ownership_availability_rules/model/product.py
```diff
@@ -0,0 +1,14 @@
+from openerp import models
```
yvaucher, line 18: Missing header

stock_ownership_availability_rules/model/quant.py
```diff
@@ -0,0 +1,64 @@
+from openerp import models, api, fields
```
yvaucher, line 18: No header

stock_ownership_availability_rules/model/quant.py
```diff
@@ -0,0 +1,64 @@
...
+
+        On the other hand, the core stock module uses the requested owner as
+        first choice, but then reserves anything else. Also, no restriction
+        means we can reserve anything, while we want to reserve only own stock.
+
+        """
+        if domain is None:
+            domain = []
+        if prefered_domain_list is None:
+            prefered_domain_list = []
```
yvaucher, line 69: ~~Is this check necessary ? Can't you just throw prefered_domain_list as it is to super ?~~

Ok `prefered_domain_list=[]` in signature in stock.py :-/

yvaucher:
Apart from missing license header LGTM :+1:

lepistone:
@yvaucher I added the missing headers. Feel free to merge :smile_cat:

yvaucher:
Thanks

PR #48, opened 2015-01-19
Customer supplier invoice on dropshipping

lepistone:
fixes https://github.com/odoo/odoo/issues/4131 with a separate addon


customer_supplier_invoice_on_dropshipping/README.rst
```diff
@@ -0,0 +1,24 @@
...
+
+The wizard does create a supplier invoice and a customer invoice, but in order
+to see them correctly, the user has to open them via the "Customer Invoices"
+and "Supplier Invoices" menu. That is because they have two different form
+views.
+
+Contributors
+------------
+
+* Leonardo Pistone <leonardo.pistone@camptocamp.com>
```
pedrobaeza: You should use OCA template in https://github.com/OCA/maintainers-tools/blob/master/template/module/README.rst

pedrobaeza:
Thanks for this very needed feature!

I see as a better name  _stock_dropshipping_dual_invoice_, to follow conventions and short it. What do you think?

Please check Travis error also.

lepistone:
Thanks as usual for reviewing @pedrobaeza!

The red build is due to `stock_route_transit` changing the default dropshipping route. Before just isolating the two modules, I'll take a moment to discuss about that with @gurneyalex.

For the title I agree `dropshipping_dual_invoice` is clearer and shorter, thanks! However I wonder: do we really need to repeat `stock_` everywhere? This module, in fact, interacts with sales, purchases, stock and accounting. Moreover, all those modules are already grouped in a repository dealing with stock.

pedrobaeza:
Yeah, I agree that we can omit stock_ part, but I put it because the module that starts all of this is called _stock_dropshipping_, and this way, all related modules are seen together in all lists (module list in Odoo, repository list, and so on).

pedrobaeza:
Thanks!

:+1:

gurneyalex:
:+1:

PR #51, opened 2015-01-21
move transport_information from OCA/vertical-ngo

yvaucher:
In order to be able to use this module for logistics purpose in a new module stock_shipment_management I move this module here to break dependency on OCA/vertical-ngo

Removed from vertical-ngo at: https://github.com/OCA/vertical-ngo/pull/113


lepistone:
:+1: thanks

rdeheele:
:+1:

PR #53, opened 2015-01-22
fix crash on module update

lepistone:
The default method is called on module update with no 'active_id' in the
context. This is a wizard that in normal use has 'active_id' in the
context.

### DIFF
--- a/stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
+++ b/stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
@@ -25,13 +25,14 @@ class StockInvoiceOnshipping(models.TransientModel):
         return self.env['account.journal'].search([('type', '=', 'sale')])
 
     def _need_two_invoices(self):
-        pick = self.env['stock.picking'].browse(self.env.context['active_id'])
-        so = pick.sale_id
-        po = pick.move_lines[0].purchase_line_id.order_id
-        if so.order_policy == 'picking' and po.invoice_method == 'picking':
-            return True
-        else:
-            return False
+        if 'active_id' in self.env.context:
+            pick = self.env['stock.picking'].browse(self.env.context['active_id'])
+            so = pick.sale_id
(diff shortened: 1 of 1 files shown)


stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
```diff
@@ -25,13 +25,14 @@ def _default_second_journal(self):
...
 
     def _need_two_invoices(self):
-        pick = self.env['stock.picking'].browse(self.env.context['active_id'])
-        so = pick.sale_id
-        po = pick.move_lines[0].purchase_line_id.order_id
-        if so.order_policy == 'picking' and po.invoice_method == 'picking':
-            return True
-        else:
-            return False
+        if 'active_id' in self.env.context:
```
pedrobaeza, line 28: Better to put `self.env.context.get('active_id')` to avoid an empty value, although the key is in the dictionary.

yvaucher:
Very strange that default value of TransientModel is being computed on update...

Anyway this fix does fix such error :+1:

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
```diff
@@ -25,13 +25,14 @@ def _default_second_journal(self):
...
+            pick = self.env['stock.picking'].browse(self.env.context['active_id'])
```
yvaucher:
```
./stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py:29:80: E501 line too long (82 > 79 characters)
```

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
lepistone, line 28: @pedrobaeza do you think we can actually get `['active_id': False]` ? If not, I prefer to crash if something wrong is passed on.

PR #55, opened 2015-01-28
new module: stock_ownership_by_move (see README)

lepistone:


stock_ownership_by_move/README.rst
```diff
@@ -0,0 +1,58 @@
...
+Installing this module, the setting to show owners will be enabled
+automatically.
+
+Usage
+=====
+
+- In a picking, you can set the owner on every line.
+- When transferring a picking, lines with different owners will not be grouped.
+- In the transfer details wizard, the owner of every line can be changed.
+- After the trasfer, the owner is also shown in the Operations tab.
```
yvaucher: s/trasfer/transfer

stock_ownership_by_move/view/picking.xml
```diff
@@ -0,0 +1,17 @@
...
+<openerp>
+  <data>
+
+    <record id="view_picking_form" model="ir.ui.view">
+      <field name="name">view_picking_form</field>
+      <field name="model">stock.picking</field>
+      <field name="inherit_id" ref="stock.view_picking_form"/>
+      <field name="arch" type="xml">
+
+        <xpath expr="//field[@name='owner_id']/.." position="replace"/>
```
yvaucher, line 11: Might be better to hide it instead of removing it. This would ease inheritance.

yvaucher:
Only few remarks to fix

guewen:
LGTM apart the things raised by @yvaucher

stock_ownership_by_move/view/picking.xml
lepistone, line 11: It is a div, containing a label, a field and a button. I don't think I can hide it.

lepistone:
@yvaucher fixed and tests added.

stock_ownership_by_move/view/picking.xml
yvaucher, line 11: https://github.com/OCA/OCB/blob/8.0/addons/account_voucher/voucher_payment_receipt_view.xml#L535

It seems possible. There are multiple div tag using attrs to set invisible

yvaucher:
@lepistone I answered your comment about the view part annihilation.

jgrandguillaume:
Tested: works perfectly. :+1:

stock_ownership_by_move/view/picking.xml
lepistone, line 11: @yvaucher I tried this https://github.com/lepistone/stock-logistics-workflow/commit/af97a90163155fb4e75959f1db1a98c8da47c269 (with `eval`) but it doesn't work (I still see the owner).

yvaucher, line 11: Ok then let's keep it that way

yvaucher:
:+1:

PR #60, opened 2015-02-02
translation refinements

rschnapka:
new templates, german translation or even added i18n-functionality alltogether

### DIFF
--- a/product_serial/i18n/de.po
+++ b/product_serial/i18n/de.po
@@ -1,0 +1,217 @@
+# Translation of OpenERP Server.
+# This file contains the translation of the following modules:
+#       * product_serial
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: OpenERP Server 7.0\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2013-10-09 09:30+0000\n"
+"PO-Revision-Date: 2015-01-31 15:17+0100\n"
+"Last-Translator: Rudolf Schnapka <schnapkar@golive-saar.de>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Language: de\n"
+"X-Generator: Poedit 1.5.4\n"
+
+#. module: product_serial
+#: field:stock.picking.prodlot.selection,last_number:0
+msgid "Last Number"
+msgstr "Letzte Nummer"
+
+#. module: product_serial
+#: selection:product.product,lot_split_type:0
+msgid "Single"
+msgstr "Einzel"
+
+#. module: product_serial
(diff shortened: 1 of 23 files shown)


product_serial/i18n/de.po
```diff
@@ -0,0 +1,221 @@
...
+msgstr "stock.picking.prodlot.selection"
+
+#. module: product_serial
+#: field:stock.production.lot,last_location_id:0
+msgid "Last location"
+msgstr "Letzter Platz"
+
+#. module: product_serial
+#: view:stock.move:0
+msgid "[('product_id','=?',product_id)]"
```
pedrobaeza: Remove this entry

product_serial/i18n/de.po
```diff
@@ -0,0 +1,221 @@
...
+
+#. module: product_serial
+#: view:stock.picking.prodlot.selection:0
+msgid "Validate"
+msgstr "Prüfen"
+
+#. module: product_serial
+#: view:stock.move:0
+msgid ""
+"onchange_lot_id(prodlot_id, product_qty, location_id, product_id, "
```
pedrobaeza, line 124: Remove this entry

product_serial/i18n/de.po
```diff
@@ -0,0 +1,221 @@
...
+"invoice from a picking. If inactive, each move line will generate one "
+"invoice line."
+msgstr ""
+"Wenn aktiv, dann werden gleiche Rechnungszeilen gruppiert/zusammengefasst, "
+"wenn eine Rechnung aus der Kommissionierung heraus erstellt wird. "
+"Anderenfalls werden je Warenbewegung einzelne Positionen angelegt."
+
+#. module: product_serial
+#: view:stock.move:0
+msgid "{'location_id':location_id, 'product_id':product_id}"
```
pedrobaeza, line 158: Remove this entry

product_serial/i18n/de.po
yvaucher: This kind of entry is in .pot file as it is exported by Odoo. So nothing against it as it would appear again next time a translation file is generated.

yvaucher:
:+1:

product_serial/i18n/de.po
pedrobaeza: Yeah, but it's a bug in the translation export that I always try to remove to avoid logic error because a bad translation

product_serial/i18n/de.po
rschnapka, line 124: reading the comments I assume you don't want to have "scripted code" that actually should not be translated (or dependant on language) in the po's even if OpnERP/ODOO export them - due to this export being a bug. Is that right? I respect that, was just wondering why the export had it in there, and would have kept those exorts in there otherwise...

pedrobaeza, line 124: Yeah, that's it. Export tool is taking this into account because it takes string values and put it here. It doesn't filter about what values to export (or it doesn't make it right).

PR #68, opened 2015-02-15
[ADD] add a new module to change the original back-order logic of Odoo

duanyp1991:
in chained locations introducing true back-to-back orders.
 Changes to be committed:

```
new file:   stock_back2back_order_proc/__init__.py
new file:   stock_back2back_order_proc/__openerp__.py
new file:   stock_back2back_order_proc/stock.py
new file:   stock_back2back_order_proc/stock_view.xml
new file:   stock_back2back_order_proc/test/test_back2back_order_proc.yml
```


stock_back2back_order_proc/__openerp__.py
```diff
@@ -0,0 +1,126 @@
...
+    'version': '1.0',
+    'category': 'Warehouse',
+    'description': """
+This module aims to change the original back-order logic of Odoo
+in chained locations introducing true back-to-back orders.
+
+
+STANDARD ODOO BACK-TO-BACK ORDER BEHAVIOR:
+Original behavior is not fully suitable to handle back-to-back backorders
+    process (check "back-to-back orders comparison.pdf"):
```
gurneyalex, line 35: where is this file?

stock_back2back_order_proc/__openerp__.py
```diff
@@ -0,0 +1,126 @@
...
+eg: Let's take the following example to understand the implemented difference:
```
gurneyalex: s/eg://

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+    _inherit = "stock.location"
+    _columns = {
+        'retention_mode': fields.selection(
+            [('retention', _('Retention Mode')), ('thru', _('Thru mode'))],
+            _('Retention Mode'),
+            required=True,
+            help=_("In 'Retention mode' the system wait for the\
+                  whole quantity before the stuff is processed.\n" +
+                   "In 'Thru mode' the shipped quantity is processed regardless\
+                  of the ordered quantity.")
```
gurneyalex: stylistic note: you don't need the backquotes or the `+` here, you can rely on the automatic joint of string litterals in Python. This is cleaner because it does not include the indentation spaces in the string

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+
+class StockLocation(orm.Model):
+    _name = 'stock.location'
+    _inherit = "stock.location"
+    _columns = {
+        'retention_mode': fields.selection(
+            [('retention', _('Retention Mode')), ('thru', _('Thru mode'))],
+            _('Retention Mode'),
+            required=True,
+            help=_("In 'Retention mode' the system wait for the\
```
gurneyalex: s/wait/waits/

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                  whole quantity before the stuff is processed.\n" +
```
gurneyalex: s/stuff is/products/

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+    _name = 'stock.picking'
+    _inherit = "stock.picking"
+
+    def get_move_chain(self, cr, uid, move_id, context=None, move_obj=False):
+        '''Recursively get the chained moves
+        @return list of the chained moves
+        '''
+        if not move_obj:
+            move_obj = self.pool.get('stock.move')
+        move_tbc = move_obj.browse(cr, uid, move_id, context, move_obj)
```
gurneyalex: what does 'tbc' mean in this context?

gurneyalex: You are passing `move_obj` to the `list_class` argument of `browse()` and I am pretty sure this is _not_ what you want to do. I am also surprised that this did not crash when you tested your code, because it really looks like a :bug:

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+
+        # If there is move_dest_id in the chain
+        if move_tbc.move_dest_id:
+            move_chain = self.get_move_chain(
+                cr, uid, move_tbc.move_dest_id.id, context)
```
gurneyalex: you are not passing `move_ob`j in your recursive call. I think this is fine and you shoud remove the argument as well as the test on line 56.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+        'retention_mode': 'retention',
+    }
+
+
+class StockPicking(orm.Model):
+    _name = 'stock.picking'
+    _inherit = "stock.picking"
+
+    def get_move_chain(self, cr, uid, move_id, context=None, move_obj=False):
+        '''Recursively get the chained moves
```
gurneyalex: Generally speaking, I'd rather have an iterative implementation of this than a recursive implementation. This is much more efficient in Python (while loop vs. function call, you can share local variables between calls, no risk of `MaxRecursionDepthExceeded` exceptions, and Python has no tail optimisation for recursion)

In this specific case, recursion might be ok, because the chains are generally short, but this is something you may want to keep in mind.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+
+        # If there is move_dest_id in the chain
+        if move_tbc.move_dest_id:
+            move_chain = self.get_move_chain(
+                cr, uid, move_tbc.move_dest_id.id, context)
+        else:
+            move_chain = []
+
+        move_chain.append(move_tbc)
+        return move_chain
```
gurneyalex, line 78: you are returning a list of browse records, not a list of ids, so this method must not be exposed to xmlrpc. You should rename it `_get_move_chain`

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+class StockPicking(orm.Model):
+    _name = 'stock.picking'
+    _inherit = "stock.picking"
+
+    def get_move_chain(self, cr, uid, move_id, context=None, move_obj=False):
+        '''Recursively get the chained moves
+        @return list of the chained moves
+        '''
+        if not move_obj:
+            move_obj = self.pool.get('stock.move')
```
gurneyalex: use `self.pool['stock.move']`

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+        for move in all_moves:
+            all_chained_moves.extend(
+                self.get_move_chain(cr, uid, move.id, context))
+
+        for move in all_chained_moves:
+            if not move.picking_id:
+                # TODO handle this kind of exception.
+                raise orm.except_orm(
+                    _('Data Error'),
+                    _('There is no picking relates to this move: ') +
```
gurneyalex: s/relates/related/

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+            return name
+
+        for move in all_moves:
+            all_chained_moves.extend(
+                self.get_move_chain(cr, uid, move.id, context))
+
+        for move in all_chained_moves:
+            if not move.picking_id:
+                # TODO handle this kind of exception.
+                raise orm.except_orm(
```
gurneyalex: do we really want to completely stop all processing in this case?

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                    _('Data Error'),
+                    _('There is no picking relates to this move: ') +
+                    '%s' % move.name)
```
gurneyalex: you should keep the `%s` inside the translated string, as it allow translators to put the translation elsewhere than in the end of the string.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+            if move.picking_id.id and (move.picking_id.id not in new_picks):
```
gurneyalex: you can get rid of the first part of the test: you've just raised an exception in that case.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                pick_tbc = self.browse(cr, uid, move.picking_id.id, context)
+                new_note = ((pick_tbc.note if pick_tbc.note else '') +
```
gurneyalex: use `pick_tbc.note or u''` here

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                            ' Copy of stock.pick[%d].') % move.picking_id.id
```
gurneyalex: this string needs to be translated. 

and you probably need to add a `\n` somewhere.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                        'name': _get_sequence_name(sequence_obj, pick_tbc),
+                        'move_lines': [],
+                    })
+                new_picks[move.picking_id.id] = new_pick_id
+
+        return new_picks
+
+    def copy_move_chain(
+            self, cr, uid, move_id, product_qty, new_picks, context=None, move_obj=False):
+        '''Recursively copy the chained move until a location in retention mode or the end.
```
gurneyalex: same remark about recursion.

gurneyalex: s/move/moves/

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+        # If there is move_dest_id in the chain and the current location is in thru mode,
+        # we need to make a copy of that, then use it as new move_dest_id.
+        if move_tbc.move_dest_id and \
+                move_tbc.location_dest_id.retention_mode == 'thru':
+            move_dest_id = self.copy_move_chain(
+                cr, uid, move_tbc.move_dest_id.id, product_qty, new_picks, context, move_obj)
+
+        my_picking_id = new_picks.get(move_tbc.picking_id.id, False)
+
+        new_note = ((move_tbc.note if move_tbc.note else '') + ' Copy of stock.move[%d].') % move_id
```
gurneyalex: same remarks as on lines 99-100

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+
+        # Create the move_history_ids (child) if there is.
+        if move_dest_id:
+            move_obj.write(
+                cr, uid, [new_move_id],
+                {'move_history_ids': [(4, move_dest_id)]})
+        return new_move_id
+
+    def update_move_chain_pick(self, cr, uid, move_id, vals, new_picks, context=None):
+        '''Recursively update the new chained move with the new related picking
```
gurneyalex: recursively you say? I don't think so :smile_cat:

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+            move_obj.write(cr, uid, [move_tbu.id], vals, context)
+
+            if not move_tbu.move_dest_id or move_tbu.location_dest_id.retention_mode != 'thru':
+                break
+
+            move_tbu = move_tbu.move_dest_id
+        return True
+
+    def update_move_chain(self, cr, uid, move_id, vals, context=None):
+        '''Recursively update the old chained move by the first move id
```
gurneyalex: nope, this is not a recursive function.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                          delivery moves with product_id, product_qty, uom
+        @return: Dictionary of values
+        """
+        if context is None:
+            context = {}
+        else:
+            context = dict(context)
+        res = {}
+        move_obj = self.pool.get('stock.move')
+        uom_obj = self.pool.get('product.uom')
```
gurneyalex: `self.pool[model.name]`

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+
+    def has_production_mode(self, cr, uid, all_moves, context=None):
+        for move in all_moves:
+            if self.check_production_node_move_chain(cr, uid, move, context):
+                return True
+        return False
+
+    # FIXME: needs refactoring, this code is partially duplicated in stock_move.
+    # do_partial()!
+    def do_partial(self, cr, uid, ids, partial_datas, context=None):
```
gurneyalex, line 239: needs fixing: this code does not call the super() implementation of do_partial
- this needs to be noted in the docstrving (and in the module documentation)
- I assume most of the code below is copied from the original implementaiton. Please highlight the changed parts with notes in comments.

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+        '''
+        move_tbc = self.browse(cr, uid, move_id, context)
+        move_dest_id = False
+        # If there is move_dest_id in the chain and the current location is in thru mode,
+        # we need to make a copy of that, then use it as new move_dest_id.
+        if move_tbc.move_dest_id and move_tbc.location_dest_id.retention_mode == 'thru':
+            move_dest_id = self.copy_move_chain(
+                cr, uid, move_tbc.move_dest_id.id, product_qty, context)
+
+        new_note = ((move_tbc.note if move_tbc.note else '') + ' Copy of stock.move[%d].') % move_id
```
gurneyalex: same remarks as on line 99-100

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                            context=context)
+        return res
+
+
+class StockMove(orm.Model):
+    _name = "stock.move"
+    _inherit = "stock.move"
+
+    def copy_move_chain(self, cr, uid, move_id, product_qty, context=None):
+        '''Recursively copy the chained move until a location in retention mode or the end.
```
gurneyalex: recursion again...

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+                'price_unit': move_tbc.price_unit,
+                'auto_validate': False
+            })
+        # Create the move_history_ids (child) if there is.
+        if move_dest_id:
+            self.write(cr, uid, [new_move_id], {'move_history_ids': [(4, move_dest_id)]})
+        return new_move_id
+
+    def update_move_chain(self, cr, uid, move_id, vals, context=None):
+        '''Recursively update the chained move by the first move id
```
gurneyalex: no recursion in this method AFAICT

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+        until a location in retention mode or the end.
+
+        @return True if ok.
+        '''
+
+        if isinstance(move_id, list):
+            move_id = move_id[0]
```
gurneyalex: please `assert len(move_id) == 1`

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+        ids = [move_id]
```
gurneyalex: remove this

stock_back2back_order_proc/stock.py
```diff
@@ -0,0 +1,552 @@
...
+
+        while move_tbu.move_dest_id and move_tbu.location_dest_id.retention_mode == 'thru':
+            ids.append(move_tbu.move_dest_id.id)
+            move_tbu = move_tbu.move_dest_id
+
+        self.write(cr, uid, ids, vals, context)
+        return True
+
+    def do_partial(self, cr, uid, ids, partial_datas, context=None):
+        """ Makes partial pickings and moves done.
```
gurneyalex, line 432: same remark as for the previous `do_partial` reimplementation

stock_back2back_order_proc/stock_view.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+		<record model="ir.ui.view" id="view_location_form_back2back">
+		    <field name="name">stock.location.form.back2back</field>
+            <field name="model">stock.location</field>
```
gurneyalex, line 6: please fix intentation

gurneyalex:
Other points to fix
- [ ] travis is red
- [ ] missing `i18n/stock_back2back_order_proc.pot`

stock_back2back_order_proc/stock.py
LeartS: relate**s** -> relate**d**

yvaucher:
@duanyp1991 Don't forget to settup your email on your local git config so Github will recognize your commits

stock_back2back_order_proc/__openerp__.py
duanyp1991, line 35: Thanks for your advise,Alexandre! Sorry for this late reply since it was during Chinese New Year holiday. I'll start working on this today.:)

stock_back2back_order_proc/stock.py
duanyp1991: @gurneyalex , For performance reason? Do you think we should use this for all cases?

stock_back2back_order_proc/stock.py
duanyp1991: You are right, we might not need to block whole process.

gurneyalex:
@duanyp1991 can you rename the file en_US.po to stock_back2back_order_proc.pot?

gurneyalex:
@duanyp1991 do you have still interest in this PR?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #70, opened 2015-02-17
transit and shipment collaboration

gurneyalex:
- reinit the transit picking types when changing the config of a warehouse (resetting possible editions made by the user which would break the config)
- fix the location of the Buy pull route
- shipment: fix transit location detection by using the `usage` field of the location


yvaucher:
:+1:

lepistone:
:+1:

yvaucher:
Needs a rebase

Issue #59, opened 2015-02-02
8.0: stock_cancel breaks stock levels


alexis-via:
Here is the scenario with an up-to-date code on odoo v8:
- install the stock_cancel module
- we will use the demo product "USB Headset", which doesn't have any stock move in the demo data
- On the product form of USB headset, click on "Update" next to the Quantity on hand and set a qty of 10.
- Create a new sale order with 1 sale order line with 2 USB headset. Confirm it and transfer the picking. Now, the "Quantity on hand" for USB Headset = 8.
- On the picking, click on the "Re-open" button. The "Quantity on hand" for USB Headset = 8 ; it should be 10 ! It's a bug.

alexis-via:
If you continue the scenario and you re-transfer the picking, the "Quantity on hand" for USB Headset = 6 (it should be 8).

pedrobaeza:
Please directly disable the module until we fix this.

alexis-via:
@pedrobaeza I don't have commit rights, so I leave you do that. We urgently need to fix the bad data to restore good stock levels (I'm working on this at the moment, help appreciated) ; and we also need to fix the bug of course.

alexis-via:
The issue is clearly linked to the fact that we don't take into account the stock.quant object when we re-open a picking. But it's still unclear to me what a stock.quant represent in v8. Do you know ?

pedrobaeza:
OK, I'll do it in a couple of hours and investigate the best way to do it.

pedrobaeza:
I have put it as not installable in https://github.com/OCA/stock-logistics-workflow/commit/533bb1a3e3d845a84c3f897cb490df02fb11a71f.

alexis-via:
@pedrobaeza 
OK, good. Now, we urgently need to find a way to fix the bad stock.quants. I started to work on it, but it's really not easy at all. I dream that we could delete all stock quants and re-generate them all, but I'm not sure we can do that.

pedrobaeza:
Yeah, in a first glance, I see that the problem is the propagation of the data across stock.quants. Maybe first restrict this cancel only for the base case (this stock hasn't been used)?

jgrandguillaume:
Hi,

I don't think we can ever re-create all the quants because they don't contain only quantity information, but prices that are stored only there. So if you drop them, you lose this valuable information (and may other).

This is why we should not be able to cancel a picking IMO. It was ok before quants appear, but now, it'll introduce quite a lot of trouble.

[...]

PR #72, opened 2015-02-19
[FIX] stock_cancel breaks stock levels

eLBati:
Fix for https://github.com/OCA/stock-logistics-workflow/issues/59

### DIFF
--- a/stock_cancel/stock.py
+++ b/stock_cancel/stock.py
@@ -50,7 +50,7 @@ class stock_picking(models.Model):
             picking.move_lines.write({'state': 'draft'})
             picking.state = 'draft'
             for move in picking.move_lines:
-                for quant in picking.move_lines.quant_ids:
+                for quant in move.quant_ids:
                     if move.location_dest_id.id == quant.location_id.id:
                         quant.location_id = move.location_id.id
             if picking.invoice_state == 'invoiced' and not picking.invoice_id:
--- a/stock_cancel/stock.py
+++ b/stock_cancel/stock.py
@@ -23,7 +23,7 @@
 #
 ##############################################################################
 
-from openerp import models, api, _, workflow, exceptions
+from openerp import models, api, _, exceptions
 
 
 class stock_picking(models.Model):
@@ -55,11 +55,6 @@ class stock_picking(models.Model):
                         quant.location_id = move.location_id.id
             if picking.invoice_state == 'invoiced' and not picking.invoice_id:
                 picking.invoice_state = '2binvoiced'
-            # Deleting the existing instance of workflow
-            workflow.trg_delete(
-                self._uid, 'stock.picking', picking.id, self._cr)
-            workflow.trg_create(
-                self._uid, 'stock.picking', picking.id, self._cr)
             picking.message_post(
                 _("The picking has been re-opened and set to draft state"))
         return
--- a/stock_cancel/stock.py
+++ b/stock_cancel/stock.py
@@ -49,6 +49,10 @@ class stock_picking(models.Model):
                     _('Picking %s has invoices!') % (picking.name))
             picking.move_lines.write({'state': 'draft'})
             picking.state = 'draft'
+            for move in picking.move_lines:
+                for quant in picking.move_lines.quant_ids:
+                    if move.location_dest_id.id == quant.location_id.id:
+                        quant.location_id = move.location_id.id
             if picking.invoice_state == 'invoiced' and not picking.invoice_id:
                 picking.invoice_state = '2binvoiced'
             # Deleting the existing instance of workflow
--- a/stock_cancel/tests/__init__.py
+++ b/stock_cancel/tests/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    Copyright (C) 2013 Agile Business Group sagl (<http://www.agilebg.com>)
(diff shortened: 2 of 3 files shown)


stock_cancel/stock.py
```diff
@@ -49,6 +49,10 @@ def action_revert_done(self):
                     _('Picking %s has invoices!') % (picking.name))
             picking.move_lines.write({'state': 'draft'})
             picking.state = 'draft'
+            for move in picking.move_lines:
+                for quant in picking.move_lines.quant_ids:
+                    if move.location_dest_id.id == quant.location_id.id:
+                        quant.location_id = move.location_id.id
             if picking.invoice_state == 'invoiced' and not picking.invoice_id:
                 picking.invoice_state = '2binvoiced'
             # Deleting the existing instance of workflow
```
alexis-via: There is no more workflow on stock.picking in v8, so I guess we can delete the 2 lignes trg_delete and trg_create.

stock_cancel/stock.py
```diff
@@ -49,6 +49,10 @@ def action_revert_done(self):
                     _('Picking %s has invoices!') % (picking.name))
             picking.move_lines.write({'state': 'draft'})
             picking.state = 'draft'
+            for move in picking.move_lines:
+                for quant in picking.move_lines.quant_ids:
+                    if move.location_dest_id.id == quant.location_id.id:
```
alexis-via, line 54: and move.quantity == quant.quantity ? I'm afraid it way not always be the case, because the quant for the transfer of this picking may have add "added" to another quant instead of be created from scratch. In this case, it won't work, no ?
You also need "else raise...."

stock_cancel/tests/test_cancel_picking.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2013 Agile Business Group sagl (<http://www.agilebg.com>)
```
gurneyalex: for a new file you probably want to change this.

stock_cancel/tests/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2013 Agile Business Group sagl (<http://www.agilebg.com>)
```
gurneyalex: new file, this needs updating

stock_cancel/tests/test_cancel_picking.py
```diff
@@ -0,0 +1,59 @@
...
+            'product_id': self.productA.id,
+            'product_uom_qty': 1,
+            'product_uom': self.productA.uom_id.id,
+            'picking_id': picking_out.id,
+            'location_id': self.stock_location,
+            'location_dest_id': self.customer_location})
+        picking_out.action_confirm()
+        picking_out.action_assign()
+        picking_out.do_transfer()
+        quants = self.StockQuantObj.search([
```
lepistone, line 41: Couldn't you use the available quantity of your product? (It should then count the quants for you behind the scenes)

stock_cancel/tests/test_cancel_picking.py
```diff
@@ -0,0 +1,59 @@
...
+            ('product_id', '=', self.productA.id),
+            ('location_id', '=', self.stock_location)
+            ])
+        total_qty = [quant.qty for quant in quants]
+        self.assertEqual(
```
lepistone, line 46: Comparing floats: it's probably safer to use assertAlmostEqual (on the other hand, I don't think we need odoo-style manual rounding).

stock_cancel/tests/test_cancel_picking.py
```diff
@@ -0,0 +1,59 @@
...
+            sum(total_qty), -1,
+            'Expecting -1 Unit , got %.4f Unit on location stock!' % (
```
lepistone, line 48: assertEqual already gives a nice error message with the expected and actual value by default. You can probably skip yours completely.

stock_cancel/stock.py
```diff
@@ -49,6 +49,10 @@ def action_revert_done(self):
                     _('Picking %s has invoices!') % (picking.name))
             picking.move_lines.write({'state': 'draft'})
             picking.state = 'draft'
+            for move in picking.move_lines:
+                for quant in picking.move_lines.quant_ids:
```
lepistone: I think you want `for quant in move.quant_ids` otherwise it would crash on a picking with 2 lines.

lepistone:
I added a few local remarks, but I am also not convinced that this will actually work. The `quant_ids` of the move are the reserved quants if you actually reserve them. But:
- an incoming shipments makes no reservations, but it does create quants
- same thing when you force availability
- merges and splits of quants could take place.

I am not sure about the details. Maybe some manual testing of those cases (and maybe more) can shed some light. For the moment I'm :triangular_flag_on_post: 
Thanks!

lepistone:
Hi @eLBati, do you want to keep this one open?

pedrobaeza:
We have another proposal for this. I will publish it in short.

eLBati:
So, let's close this for now

yvaucher:
@pedrobaeza any news about your proposal ?

pedrobaeza:
We are redoing this in https://github.com/OCA/stock-logistics-workflow/pull/145

PR #76, opened 2015-02-27
fix dropshipping route with transit

gurneyalex:
we need a push rule to ensure that dropshipping PO going through the outgoing
transit location do arrive at the customer location


lepistone:
:+1:

yvaucher:
:+1:

Issue #79, opened 2015-03-03
[8.0] stock_picking_invoice_link Error on Invoice creation


hurrinico:
When i create 1 invoice from 2 or more pickings, on invoice's picking_lines appear only one picking.
I will be providing a PR that fixes this issue.

yvaucher:
Fixed by #80

PR #80, opened 2015-03-03
[FIX] FIxes bug #79

hurrinico:
In order to solve https://github.com/OCA/stock-logistics-workflow/issues/79 i create this PR

### DIFF
--- a/stock_picking_invoice_link/stock.py
+++ b/stock_picking_invoice_link/stock.py
@@ -34,6 +34,7 @@ class stock_move(orm.Model):
         inv_line_id = super(stock_move, self)._create_invoice_line_from_vals(
             cr, uid, move, invoice_line_vals, context=context)
         move.write({'invoice_line_id': inv_line_id})
+        move.picking_id.invoice_id = invoice_line_vals['invoice_id']
         return inv_line_id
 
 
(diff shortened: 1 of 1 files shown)


eLBati:
:+1:

lmignon:
@hurrinico It should be nice to have a unittest that demonstrate your fix..

stock_picking_invoice_link/stock.py
```diff
@@ -34,6 +34,7 @@ def _create_invoice_line_from_vals(
         inv_line_id = super(stock_move, self)._create_invoice_line_from_vals(
             cr, uid, move, invoice_line_vals, context=context)
         move.write({'invoice_line_id': inv_line_id})
+        move.picking_id.write({'invoice_id': invoice_line_vals['invoice_id']})
```
pedrobaeza: You can do a shorter form:

```
move.picking_id.invoice_id = invoice_line_vals['invoice_id']
```

hurrinico:
@lmignon i'm quite busy at the moment, so i can't do the unittest by now.

lepistone:
Hi @hurrinico. Thanks for the fix!

 Looking at the code I see a line added and a method deleted, but I do not understand how this fixes the bug. A agree we need a test here.

Also, the commit messages say "fix" 12 times, but instead it would be great to have only one commit with a message that actually says why you are doing this. Thanks!

chafique-delli:
I tested the fix, it fixes the bug that I also found. :+1:

guewen:
I guess a method has been renamed. Would be cool you if could you squash your commits and put an explanation in the commit message.

hurrinico:
@lepistone @guewen @chafique-delli i've sqhashed all the commit in one and rebased the brach

Issue #74, opened 2015-02-24
stock_ownership_availability_rules disables stock accounting


yostashiro:
No journal entry gets generated for stock valuation from receipt/delivery after installing this module even when the owner of the stock is the company.

According to stock_account module, accounting valuation entry is designed not to be generated in case a quant has an owner.
https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock_account.py#L113

yostashiro:
A module to fix the issue (temporary use): https://github.com/rfhk/awo-custom/tree/master/stock_account_adjust

lepistone:
Thanks @yostashiro for the report!

Indeed the module makes the owner required, using the current company as owner by default, so that line needs to be changed to something like 

``` python
if quant.owner_id and quant.owner_id != current_company.partner_id
```

However, this is a long method, and to change that line, we'll probably have to duplicate the whole method.

Any ideas? (ping @gurneyalex)

lepistone:
I can't think of a more clever fix than to redefine the whole method with that line changed. If you could make a pull request towards `stock-ownership-availability-rules` I'd be glad to review it. Thanks!

yostashiro:
Hi @lepistone,
I could not think of a better way to handle the issue.  Appreciate your review.

gurneyalex:
@lepistone maybe a PR on odoo could be made for this?

(cc @jgrandguillaume)

pedrobaeza:
Added a warning in the README in 523555fd72d381ec7ca59b0900580cda122aa13a

PR #82, opened 2015-03-04
[FIX] enable stock accounting if stock owner = company

yostashiro:
This is to fix the issue #74.

### DIFF
--- a/stock_ownership_availability_rules/model/quant.py
+++ b/stock_ownership_availability_rules/model/quant.py
@@ -21,6 +21,37 @@ from openerp import models, api, fields
 class Quant(models.Model):
     _inherit = 'stock.quant'
 
+    # this method is written in old api style as
+    # "Company = self.env['res.company']" would throw an error for not having
+    # attribute 'env' when called from a method with old api
+#     def _get_company_partner_id(self, cr, uid, location, context=None):
+# #         if not context:
+# #             context = {}
+#         Company = self.pool.get('res.company')
+#         res = (location.company_id.partner_id.id or
+#             Company.browse(
+# #                 Company._company_default_get(cr, uid, 'stock.quant', context=context)
+# #                 Company._company_default_get('stock.quant')
+#                 self.env['res.company']._company_default_get('stock.quant')
+#             ).partner_id.id
+#         )
+#         return res
+
+    def _get_company_partner_id(self, location):
+#         if not context:
+#             context = {}
+        Company = self.env['res.company']
+        res = (location.company_id.partner_id.id or
+            Company.browse(
+#                 Company._company_default_get(cr, uid, 'stock.quant', context=context)
+#                 Company._company_default_get('stock.quant')
+                Company._company_default_get('stock.quant')
+            ).partner_id.id
+        )
+        return res
+
+    
+    
     @api.model
     def create(self, vals):
         """Set the owner based on the location.
@@ -29,15 +60,11 @@ class Quant(models.Model):
 
         """
         if not vals.get('owner_id'):
-            Company = self.env['res.company']
             location = self.env['stock.location'].browse(vals['location_id'])
 
             vals['owner_id'] = (
                 location.partner_id.id or
-                location.company_id.partner_id.id or
-                Company.browse(
-                    Company._company_default_get('stock.quant')
-                ).partner_id.id
+                self._get_company_partner_id(location)
             )
 
         return super(Quant, self).create(vals)
--- a/stock_ownership_availability_rules/model/quant.py
+++ b/stock_ownership_availability_rules/model/quant.py
@@ -21,36 +21,15 @@ from openerp import models, api, fields
 class Quant(models.Model):
     _inherit = 'stock.quant'
 
-    # this method is written in old api style as
-    # "Company = self.env['res.company']" would throw an error for not having
-    # attribute 'env' when called from a method with old api
-#     def _get_company_partner_id(self, cr, uid, location, context=None):
-# #         if not context:
-# #             context = {}
-#         Company = self.pool.get('res.company')
-#         res = (location.company_id.partner_id.id or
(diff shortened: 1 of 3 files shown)


lepistone:
Thanks for your contribution @yostashiro ! Actually, giving a look, we compute the default owner from the location like this:
https://github.com/OCA/stock-logistics-workflow/blob/8.0/stock_ownership_availability_rules/model/quant.py#L35-L40

In your code, you know the location (`quant.location_id`), so it would be most correct to use the same logic for your comparison. Extra points if you extract that to a separate method that takes a location as parameter, and use it both times.

Unfortunately, the OCA branches are automatically linted (while odoo's aren't), so for your travis build to pass you lint all the method (including splitting long lines). Look at the travis build for details.

Also, could you please mark with commentaries the start and end of your modification? That will ease maintenance and distinguish between commentaries that were already in the original.

Thanks for your work!

yostashiro:
Hi @lepistone , thanks for your advice.  I am trying to create a method for getting company partner id which should be called from both methods.  However, the method throws an error when it's called from `_account_entry_move` while it works fine when it's called from `create` method.

The new method looks like this:

```
    def _get_company_partner_id(self, location):
        Company = self.env['res.company']
        res = (location.company_id.partner_id.id or
            Company.browse(
                Company._company_default_get('stock.quant')
            ).partner_id.id
        )
        return res
```

The error when called from `_account_entry_move` is `AttributeError: 'stock.quant' object has no attribute 'env'`.  The line `Company = self.env['res.company']` is triggering this error.  Do you how to get around this issue?  `_account_entry_move`, being copied from the official code, uses old API, and that I guess is the reason.

lepistone:
@yostashiro try to decorate your method with `@api.model`.

lepistone:
@yostashiro doc here: https://www.odoo.com/documentation/8.0/reference/orm.html#module-openerp.api

yostashiro:
Thanks for your advice, @lepistone .  Could you please help review the PR?  The cause of the problem that I mentioned earlier was that I was calling the new method from the wrong object in `_account_entry_move`.  The issue was fixed after correcting this part.  I was not sure if adding `@api.model` to the new method would change anything but I added that anyway.

The change in `_account_entry_move` now looks a bit more bloated than we had originally thought as I thought using `q.location_id` was not necessarily correct and should use an internal location involved in the stock move.

lepistone:
@yostashiro I think `q.location_id` is already the location where the quant is, and normally you have quants on internal locations only.

Moreover, what we are trying to do, is to identify quants that have the default owner, which is broadly speaking **us** (i.e. the location's partner if any, otherwise the partner of the location's company if any, otherwise the default company's partner). That is computed in the default method you are factoring out, using the quant's location.

(I might of course be wrong in this reasoning, say so if you have a case where it doesn't make sense)

stock_ownership_availability_rules/model/quant.py
```diff
@@ -29,15 +39,11 @@ def create(self, vals):
 
         """
         if not vals.get('owner_id'):
-            Company = self.env['res.company']
             location = self.env['stock.location'].browse(vals['location_id'])
 
             vals['owner_id'] = (
                 location.partner_id.id or
```
lepistone, line 45: I would put this line inside the method, and give the method a more generic name like `_get_default_owner`.

stock_ownership_availability_rules/model/stock_account.py
```diff
@@ -0,0 +1,122 @@
...
+        company_from = location_obj._location_owner(cr, uid, location_from,
+            context=context)
+        company_to = location_obj._location_owner(cr, uid, location_to,
+            context=context)
+
+        if move.product_id.valuation != 'real_time':
+            return False
+        for q in quants:
+            # if q.owner_id:  # REPLACED FROM
+            if q.owner_id and q.owner_id.id != \
```
lepistone, line 70: Now that I think of it, owner_id is now required, so the first part is always true.

lepistone:
A last remark: travis for some reason did not build on your last commit. It happens sometimes, but it should build on your next push.

yostashiro:
Thanks for your comments, @lepistone .

> I think q.location_id is already the location where the quant is, and normally you have quants on internal locations only.

This is not quite true according to my understanding.  A quant is created at the destination location of stock move regardless of the type of the location.  `q.location_id` in `_account_entry_move` is the destination location, which means the customer location in case of outgoing picking.  Then it is not correct to use this location to identify the partner id of the company.

> I would put this line inside the method, and give the method a more generic name like _get_default_owner.

I think this will make the method unusable for `_account_entry_move` in which what we need to do is to identify the partner id of the company, which is not equal to identifying the default owner.

Please let me know if I'm misunderstood anything.

lepistone:
@yostashiro you're right, quants are created on customer locations, too. So I am not sure what is the condition on ownership to decide to not make any valuation entries. I'd love to hear some opinions here, anyone?

gurneyalex:
@yostashiro could you add an automated test?

yostashiro:
@gurneyalex sorry about my late reply.  I'm trying to figure out how to create an automated test (I have not done this before).  Appreciate your pointers in the mean time.

yostashiro:
I now realize that obviously stock_account should be added as a dependency for this change.  I wonder if it is the right direction to take since some users may not use stock accounting at all.  I personally think it might be better to have a separate module to go with stock_ownership_availability_rules, which can be installed as necessary (i.e. stock accounting is used).  Any thoughts?

gurneyalex:
@yostashiro the way to solve this is to create a separate module depending on stock_ownership_availability_rules and stock_account and with auto_install = True in **openerp**.py : that way it will be pulled in when both its dependencies are installed.

yostashiro:
Thanks @gurneyalex .  I have closed this PR for now.  I will later try proposing a separate module as per your advice when I find time.

PR #83, opened 2015-03-05
fix two bugs in shipment plans

lepistone:


stock_shipment_management/model/shipment_plan.py
```diff
@@ -207,11 +207,12 @@ def _get_arrival_picking_ids(self):
...
     @api.one
-    @api.depends('departure_picking_ids.volume')
+    @api.depends('departure_move_ids.product_id',
+                 'departure_move_ids.product_id.volume')
     def _compute_volume(self):
-        volume = 0
-        for picking in self.departure_picking_ids:
-            volume += picking.volume or 0.0
+        volume = 0.0
+        for move in self.departure_move_ids:
```
pedrobaeza, line 215: You can do: `sum(x.product_id.volume for x in self.departure_move_ids)`

There's no care about null value, because they are returned already as 0.0.

gurneyalex:
:+1:

lepistone:
thanks for the fixup @gurneyalex :+1:

stock_shipment_management/tests/test_shipment_creation.py
```diff
@@ -97,6 +101,12 @@ def test_create_shiping_from_picking(self):
                           1)
         self.assertEquals(shipment.arrival_picking_count,
                           1)
+        self.assertEquals(shipment.volume,
```
lepistone: I'd add `Almost` and remove `s` (I know, there are `s`'s above)

lepistone:
minor remarks, but :+1: thanks for the tests!

lepistone:
... and if change 'needs fixing' with 'bug' you get the german flag (but you might mess up the order).

stock_shipment_management/tests/test_shipment_creation.py
```diff
@@ -38,7 +38,11 @@ def setUp(self):
 
         SO = self.env['sale.order']
         SOL = self.env['sale.order.line']
-
+        PROD = self.env['product.product']
+        prod_id = self.ref('product.product_product_33')
```
lepistone: What about `prod = self.env.ref['blah']` and getting rid of `PROD`.

lepistone:
thanks for the changes :+1:

lepistone:
All is fixed now, I think we can merge.

yvaucher:
:+1:

PR #85, opened 2015-03-06
stock_shipment_management: fix transit location detection

gurneyalex:
The code would not work with incoming transit locations.


yvaucher:
Fixed in #70

PR #86, opened 2015-03-09
on ship_transit, output location is not active

gurneyalex:


lepistone:
:+1:

yvaucher:
:+1:

guewen:
It should probably be updated to use `_location_used()` [0] as it has be done upsteam [1]

[0] https://github.com/odoo/odoo/blob/8.0/addons/stock/stock.py#L3027-L3034
[1] https://github.com/odoo/odoo/blob/8.0/addons/stock/stock.py#L3036-L3058

gurneyalex:
part of https://github.com/OCA/stock-logistics-transport/pull/7

PR #87, opened 2015-03-09
Add stock_picking_package_preparation

guewen:
This module adds a new document in the Warehouse menu: **Package
Preparation**. It works on a selection of pickings and allows to add
them all in the same package, then to transfer all the pickings at once.

A possible use case is to put all the selected pickings on a pallet.


stock_picking_package_preparation/model/stock_picking_package_preparation.py
```diff
@@ -0,0 +1,247 @@
...
+        readonly=True,
+        copy=False,
+    )
+    partner_id = fields.Many2one(
+        comodel_name='res.partner',
+        string='Partner',
+        required=True,
+        states={'done': [('readonly', True)],
+                'in_pack': [('readonly', True)],
+                'cancel': [('readonly', True)]},
```
gurneyalex: suggestion: extract this dictionary in a module level dict and reuse it.

gurneyalex:
:+1: nice addon

yvaucher:
Just added bug tracker link otherwise :+1:

PR #91, opened 2015-03-17
Only add one message per computation on pickings

mdietrichc2c:
This PR is to "clean up" after another PR I made, which added logging messages after stock_picking_compute_delivery_date's cron on each picking and product.
The issue is that, for an open stock.picking with many lines, a messages was logged at each calculation, for each product; this can lead to >800 messages.

Only one message is kept per calculation.


stock_picking_compute_delivery_date/model/picking.py
```diff
@@ -113,12 +124,17 @@ def compute_mto_delivery_dates(self, cr, uid, product, context=None):
...
                 message = _("Scheduled date update to %s from %s") \
                     % (new_date_str, move_in.picking_id.name)
-                # Post message on stock picking
-                self.message_post(cr, uid, move_out.picking_id.id,
-                                  body=message, context=context)
+                # Add message on stock picking:
+                # - if it does not exist, add the current one
+                # - if it does, overwrite it, and keep the "last" one.
+                if 'messages' not in context:
+                    context['messages'] = {}
```
yvaucher: Please don't assign context it's dangerous, your changes on a mutable will be done for all calling methods.

Plus this won't work when it will be ported to v8.0 api as context will be a frozendict.

mdietrichc2c:
@yvaucher thanks; I modified the method to use a variable instead of the context

yvaucher:
@mdietrichc2c Thanks for the changes :+1:

gurneyalex:
:+1:

stock_picking_compute_delivery_date/model/picking.py
```diff
@@ -216,7 +232,11 @@ def compute_all_delivery_dates(self, cr, uid, use_new_cursor=True,
                     'Could not update delivery date for product %s',
                     product.name)
 
+        # Publish messages on pickings
+        self.picking_notify(cr, uid, messages, context=context)
```
guewen: If picking notify fails and `use_new_cursor` is True, `cr` won't be rollbacked nor closed. Should be in a try: except:

mdietrichc2c:
@guewen fixed now; the method is now encapsulated in a try: except:

max3903:
:+1:

PR #97, opened 2015-04-06
[IMP] 'stock_picking_mass_assign' add possibility to force availability and to deliver many pickings;

legalsylvain:
Hi all,
I was searching for a tool to check availability of many pickings. I found 'stock_picking_mass_assign' and I improved it to give the possibility to : 
- mass force availability;
- mass deliver;
- directly display after the invoices wizard;

I let by default the same behaviour. I just added options (by checkbox), unchecked by default.

Thanks for your review.
CC : @guewen

### DIFF
--- a/stock_picking_mass_assign/i18n/de.po
+++ b/stock_picking_mass_assign/i18n/de.po
@@ -1,20 +1,24 @@
 # Translation of OpenERP Server.
 # This file contains the translation of the following modules:
-# 	* stock_picking_mass_assign
+#	* stock_picking_mass_assign
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: OpenERP Server 7.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2014-07-17 09:00+0000\n"
-"PO-Revision-Date: 2015-02-01 13:42+0100\n"
-"Last-Translator: Rudolf Schnapka <rs@techno-flex.de>\n"
+"POT-Creation-Date: 2015-04-06 11:29+0000\n"
+"PO-Revision-Date: 2015-04-06 11:29+0000\n"
+"Last-Translator: <>\n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
-"Content-Transfer-Encoding: 8bit\n"
-"Language: de\n"
-"X-Generator: Poedit 1.5.4\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: stock_picking_mass_assign
+#: view:stock.picking.check.assign.all:0
+msgid "Apply"
+msgstr "Apply"
 
 #. module: stock_picking_mass_assign
 #: view:stock.picking.check.assign.all:0
@@ -23,7 +27,7 @@ msgstr "Abbrechen"
 
 #. module: stock_picking_mass_assign
 #: model:ir.actions.act_window,name:stock_picking_mass_assign.action_check_assign_all
-#: view:stock.picking.check.assign.all:0
+#: field:stock.picking.check.assign.all,check_availability:0
 msgid "Check Availability"
 msgstr "Verfügbarkeit prüfen"
 
@@ -33,16 +37,14 @@ msgid "Check the availability of the selected delivery orders"
 msgstr "Verfügbarkeit für ausgewählte Lieferaufträge prüfen"
 
 #. module: stock_picking_mass_assign
-#: code:_description:0
-#: model:ir.model,name:stock_picking_mass_assign.model_stock_picking
-#, python-format
-msgid "Picking List"
-msgstr "Kommissionierschein"
+#: field:stock.picking.check.assign.all,create_invoice:0
+msgid "Create Invoices/Refunds"
+msgstr "Erstelle Rechnung / Gutschrift"
 
 #. module: stock_picking_mass_assign
-#: view:stock.picking.check.assign.all:0
-msgid "or"
-msgstr "oder"
+#: field:stock.picking.check.assign.all,process_picking:0
+msgid "Deliver"
+msgstr "Ausliefern"
 
 #. module: stock_picking_mass_assign
 #: code:_description:0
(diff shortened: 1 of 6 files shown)


stock_picking_mass_assign/i18n/de.po
```diff
@@ -1,20 +1,24 @@
...
-"Content-Transfer-Encoding: 8bit\n"
-"Language: de\n"
-"X-Generator: Poedit 1.5.4\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: stock_picking_mass_assign
+#: view:stock.picking.check.assign.all:0
+msgid "Apply"
+msgstr "Apply"
```
pedrobaeza: Not translated. Better to leave it empty

stock_picking_mass_assign/i18n/de.po
```diff
@@ -33,6 +37,40 @@ msgid "Check the availability of the selected delivery orders"
...
+#. module: stock_picking_mass_assign
+#: field:stock.picking.check.assign.all,force_availability:0
+msgid "Force Availability"
+msgstr "Erzwinge Verfügbarkeit"
+
+#. module: stock_picking_mass_assign
+#: code:addons/stock_picking_mass_assign/wizard/check_assign_all.py:63
+#, python-format
+msgid "No selected delivery orders"
+msgstr "No selected delivery orders"
```
pedrobaeza: The same

stock_picking_mass_assign/wizard/check_assign_all.py
```diff
@@ -27,22 +27,99 @@ class check_assign_all(orm.TransientModel):
...
+                'view_type': 'form',
+                'view_mode': 'form',
+                'res_model': 'stock.invoice.onshipping',
+                'type': 'ir.actions.act_window',
+                'target': 'new',
+                'context': ctx,
+            }
+        else:
+            # close the wizard
+            return {'type': 'ir.actions.act_window_close'}
```
pedrobaeza: This is not needed on v7

gurneyalex:
:+1: for merging once pedro's remarks have been taken into account.

legalsylvain:
Hi @gurneyalex, @pedrobaeza, 
Thanks for your review.
Just one minor question. I installed this updated module on production database some days ago, and the name of the button "Check Availability" is not very clear now. I would prefer "Mass Process", because now you can "check availability", "force availability", "mass deliver" and "mass invoice".
But as this module is used in production in other OCA deployment, I don't know if it's a good idea to change the name of this button, because if someone else do an update, the behaviour of Odoo will change for end users.

What's your PoV ? CC : @guewen

Note : @gurneyalex : '7.0' is better a milestone than a label. It isn't ?

stock_picking_mass_assign/wizard/check_assign_all.py
legalsylvain: I don't know, it was here before ;-) 
https://github.com/OCA/stock-logistics-workflow/pull/97/files#diff-67e417441c98c7a244d9a98d7694a2f6L48
a function called by a button must return something, but maybe return True is enough....

pedrobaeza:
Well, I see no special problem on renaming. Maybe a better strategy would be to have one action entry for each entry, that calls the same wizard, but with a context key that tells what operation to do. 

This way, you don't lose "Check availability" button, but add other options.

stock_picking_mass_assign/wizard/check_assign_all.py
pedrobaeza: This is inheritance of 6.1, where it was needed. For v7, you can even not return nothing (that it's to say, implicit `return None`), and the action would be to close the dialog.

legalsylvain: Thanks. I'm waiting for other PoV of contributors about other point, and I'll do all the changes.

legalsylvain:
@pedrobaeza It makes sense ! 
@other, other point of view ?

legalsylvain:
Other point : somebody can say me what I need to fix (related to the error on runbot) ?
https://runbot.odoo-community.org/runbot/build/2926102
In log, I don't see where is the problem :
https://runbot.odoo-community.org/runbot/static/build/2926102-97-33987a/logs/job_20_test_all.txt

stock_picking_mass_assign/wizard/check_assign_all.py
legalsylvain: In fact, I wrote a "return True". Otherwise, I have a warning in a log.
openerp.osv.osv: The method check of the object stock.picking.check.assign.all can not return `None` !

legalsylvain:
@gurneyalex , @pedrobaeza : Done.

pedrobaeza:
:+1:

gurneyalex:
@legalsylvain I'm not too keen on the use of 7.0 / 8.0 as milestones for OCA projects (although I don't object if people want to use them). 

The reason is, IMO, a milestone is something you want to reach for you project, not a target in a different project. You create a milestone, group tickets belonging to that milestone, implement them, and once all the tickets are done, you tag and release. Then you move on to the next milestone. This is the standard workflow for many projects. 

When you set up 7.0 and 8.0 milestones on OCA projects, you are really telling which odoo branch you are targetting (and therefore in which project branch you are proposing to merge the feature), which means that these milestones will never be reached.

gurneyalex:
- no objection to renaming if you feel like it
- runbot is failing for reasons unrelated to your PR (warnings for deprecated fields in views). Normally https://github.com/OCA/stock-logistics-workflow/pull/99 should fix this and I don't understand why it does not... 

:+1:

legalsylvain:
Hi, @gurneyalex. thanks for your explanation about milestones. I understand your PoV. My opinion is : Github milestones are unused for the time being for OCA / Odoo projects, so why not use them ? like we managed milestones / series in LP : https://launchpad.net/openerp-french-localization/+series) But this is out of the scope of this PR. Maybe we could talk about that on odoo mailing list ?

> no objection to renaming if you feel like it

Finally, I pushed a solution without renaming button, but creating two new ones, as @pedrobaeza proposed.

Thanks for your review.

Regards.

legalsylvain:
There is 2 approves since 2 weeks, could we consider to merge this PR ?
kind regards.

PR #98, opened 2015-04-06
[FIX] stock_dropshipping_dual_invoice: Default value for sale journal

pedrobaeza:
To avoid an error when having more than one sale journal.


gurneyalex:
:+1:

pedrobaeza:
Please fast-track as this is blocking.

yvaucher:
:+1:

PR #99, opened 2015-04-08
remove deprecated type field on ir.ui.view

gurneyalex:
these lines cause the runbot runs to end in Warning state, and the PRs not to be green when they should.


pedrobaeza:
:+1:

yvaucher:
:+1:

PR #100, opened 2015-04-21
[ADD] stock_account_fix

### LINKED ISSUE #74: stock_ownership_availability_rules disables stock accounting

yostashiro:
Once again, this is to fix #74 .


gdgellatly:
Hi,

I see @lepistone suggested this may be necessary and someone else said a PR on Odoo is a better idea.  I would much rather see this on OCB/Odoo because in essence the Odoo code is wrong.  They are using owner_id as a proxy for ! company_id which is a shortcut at best.

Cedric-Pigeon:
@yostashiro Do you plan to achieve this PR ? otherwise we will close it in 3 days. Thanks.

pedrobaeza:
Closing due to inactivity.

PR #101, opened 2015-04-21
add module picking_line_sale_link that creates link between parent lines...

chafique-delli:
... orders and child in stock move for bundle product

### DIFF
--- a/picking_line_sale_link/sale.py
+++ b/picking_line_sale_link/sale.py
@@ -31,33 +31,20 @@ class SaleOrderLine(orm.Model):
                                           'Children Line'),
         }
 
-
-class SaleOrder(orm.Model):
-    _inherit = 'sale.order'
-
-    def copy(self, cr, uid, id, default=None, context=None):
+    def copy_data(self, cr, uid, id, default=None, context=None):
         if default is None:
             default = {}
-        default = {'order_line': False}
-        new_order_id = super(SaleOrder, self).copy(cr, uid, id,
-                                                   default=default,
-                                                   context=context)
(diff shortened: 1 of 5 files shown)


chafique-delli:
Runbot test failed because it try to install module 'product_customer_code_picking' that depends on module 'product_customer_code'.But the latter module is not available in this repository.

picking_line_sale_link/sale.py
```diff
@@ -0,0 +1,63 @@
...
+    def copy(self, cr, uid, id, default=None, context=None):
+        if default is None:
+            default = {}
+        default = {'order_line': False}
+        new_order_id = super(SaleOrder, self).copy(cr, uid, id,
+                                                   default=default,
+                                                   context=context)
+        order_line_model = self.pool.get('sale.order.line')
+        order = self.browse(cr, uid, id, context=context)
+        for origin_order_line in order.order_line:
```
guewen: It seems to me it can be done in `sale_order_line.copy_data()` instead of `sale_order.copy()`?

picking_line_sale_link/stock.py
```diff
@@ -0,0 +1,52 @@
...
+            cr, uid, [('sale_line_id', 'in', ids)], context=context)
+        return move_ids
+
+    _columns = {
+        'sale_parent_line_id': fields.related(
+            'sale_line_id',
+            'line_parent_id',
+            type='many2one',
+            relation='sale.order.line',
+            string='Parent Product',
```
guewen: Parent Product? Or Parent Sale Line?

PR #104, opened 2015-04-30
in quant view, filter the unreserved ones

lepistone:
For this to work properly, that filter needs to be AND with the other on
internal locations. Rather surprisingly, this is done with a separator
between the buttons.

### DIFF
--- a/stock_ownership_availability_rules/model/product.py
+++ b/stock_ownership_availability_rules/model/product.py
@@ -26,6 +26,7 @@ class Product(models.Model):
         result['context'] = (
             "{'search_default_locationgroup': 1, "
             "'search_default_ownergroup': 1, "
-            "'search_default_internal_loc': 1}"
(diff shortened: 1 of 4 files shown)


yvaucher:
:+1:

gurneyalex:
:+1:

stock_ownership_availability_rules/model/product.py
```diff
@@ -26,6 +26,7 @@ def action_open_quants(self, cr, uid, ids, context=None):
         result['context'] = (
             "{'search_default_locationgroup': 1, "
             "'search_default_ownergroup': 1, "
-            "'search_default_internal_loc': 1}"
+            "'search_default_internal_loc': 1, "
+            "'search_default_without_reservation': 1}"
```
gurneyalex, line 30: I believe you can use a real dict here, rather than a string serialized one

PR #106, opened 2015-05-05
Migration of stock_picking_deliver_uos

tafaRU:


stock_picking_deliver_uos/wizard/stock_transfer_details.py
```diff
@@ -0,0 +1,80 @@
...
+    _inherit = 'stock.transfer_details_items'
+
+    product_uos_qty = fields.Float(
+        'Quantity (UOS)',
+        digits_compute=dp.get_precision('Product Unit of Measure'))
+    product_uos = fields.Many2one(
+        "product.uom", string='Product UOS', readonly=True)
+
+    def onchange_product_uos_qty(
+            self, cr, uid, ids, product_uos_qty, packop_id, context=None
```
florian-dacosta, line 69: Any reason you do not use the new api for this onchange?

florian-dacosta:
The module handles only the case when 1 stock move leads to 1 pack operation. I think you should add a note in the readme explaining it.
Also I would add a warning in the onchange in case there are more than 1 stock.move for 1 operation, so the user knows that changing the uos_quantity won't have any effect.

Else it is ok for me, thanks for the migration

stock_picking_deliver_uos/wizard/stock_transfer_details.py
eLBati, line 69: Because of incompatibility with YAML test I think

tafaRU, line 69: > Because of incompatibility with YAML

@florian-dacosta I confirm what @eLBati has written.

tafaRU:
@florian-dacosta thanks for your review! My last commit would address your comment, please have a look at it.

florian-dacosta:
:+1:

stock_picking_deliver_uos/wizard/stock_transfer_details.py
eLBati, line 69: @yvaucher this is what I was speaking about this morning

eLBati:
:+1:

yvaucher:
I restarted travis as tests were fixed in #111

stock_picking_deliver_uos/wizard/stock_transfer_details.py
yvaucher, line 69: @eLBati ok (sorry I missed that message) it's ok to keep things such as onchange in old api. Just hoping this will disapear in v9

yvaucher:
@eLBati Can you amend or rebase ? so it will use the propper travis config and create the test jobs we want.

tafaRU:
@eLBati @yvaucher I'm doing the rebase.

tafaRU:
@yvaucher just rebased.

yvaucher:
@tafaRU thanks

tafaRU:
@yvaucher looks green and it's ready to be merged. Thanks in advance.

PR #111, opened 2015-05-20
separate the picking_dispatch tests

gurneyalex:
picking dispatch adds a dependency on delivery
and the presence of this module breaks the tests of stock_ownership_by_move
(but the module works fine with delivery installed)

PR #113, opened 2015-05-24
[IMP] stock_dropshipping_dual_invoice: Allow to invoice more than one dropshipping pickings + returning view

pedrobaeza:
This PR contains 2 improvements:
- Allow to invoice more than one dropshipping pickings, making previously a grouping by customer that allows to call with the context correctly.
- Return a list view only with the created invoices (with no possibility to go to form view, so the data is correct).

cc @lepistone

### DIFF
--- a/stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
+++ b/stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
@@ -1,5 +1,7 @@
 #    Author: Leonardo Pistone
 #    Copyright 2015 Camptocamp SA
+#    Contributor: Pedro M. Baeza
+#    Copyright 2015 Serv. Tecnol. Avanzados
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
@@ -33,7 +35,6 @@ class StockInvoiceOnshipping(models.TransientModel):
             po = pick.move_lines[0].purchase_line_id.order_id
             if so.order_policy == 'picking' and po.invoice_method == 'picking':
                 return True
-
         return False
 
     @api.depends('journal_type', 'need_two_invoices')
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Added a fix in the view cc @antespi

pedrobaeza:
Added conditions to put correctly journals when a line is "over-transferred" (you transfer a quantity bigger than the expected). The new line is created without reference to purchase_line_id, and it is put at position 0, so current checks failed.

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
```diff
@@ -21,6 +23,23 @@
 class StockInvoiceOnshipping(models.TransientModel):
     _inherit = "stock.invoice.onshipping"
 
+    @api.model
+    def _get_journal_type(self):
+        res_ids = self.env.context.get('active_ids', [])
+        pickings = self.env['stock.picking'].browse(res_ids)
+        pick = pickings and pickings[0]
```
lepistone: Not sure, but maybe you can get `context['active_id']` and skip this line?

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
```diff
@@ -21,6 +23,23 @@
...
+        if pick.move_lines:
+            src_usage = pick.move_lines[0].location_id.usage
+            dest_usage = pick.move_lines[0].location_dest_id.usage
+            if src_usage == 'supplier' and dest_usage == 'customer':
+                moves = pick.move_lines.filtered('purchase_line_id')
+                moves = moves and moves[0]
```
lepistone: Maybe also `pick.mapped('move_lines.purchase_line_id.order_id.invoice_method')`? Not sure, but that might be enough.

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
```diff
@@ -30,10 +49,11 @@ def _need_two_invoices(self):
             pick = self.env['stock.picking'].browse(
                 self.env.context['active_id'])
             so = pick.sale_id
-            po = pick.move_lines[0].purchase_line_id.order_id
-            if so.order_policy == 'picking' and po.invoice_method == 'picking':
-                return True
-
+            moves = pick.move_lines.filtered('purchase_line_id')
```
lepistone: same (maybe)

lepistone:
Note: the test failure is unrelated and is being handled in #111

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
pedrobaeza: Yes, this was verbatim copy of previous code. I change it.

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
pedrobaeza: This is an annoying thing I have to introduce, because in some moment after the launch of the new API, they have added an ensure_one() in the field get method, so now we cannot access first record of the recordset with dot notation when there are more than one record.

With your expression, it will get a list of values, but it will be empty in case there is no records, so we have to test also if the list is empty.

yvaucher:
#111 to be merged first to fix the tests it was just updated due to a change in Odoo code base

yvaucher:
To be rebased this will also trigger the fixed travis tests

pedrobaeza:
Rebase and with some improvements in getting only one record in recordsets. Please check @lepistone and @antespi

lepistone:
:+1:

yvaucher:
:+1:

PR #116, opened 2015-06-01
Improve README.rst. Follow template recommendation

Doscaal:
Follow template recommendation and add some missing contributors


max3903:
:+1:

yvaucher:
:+1:

pedrobaeza:
This module is for now not valid, because it creates inconsistent data. Does it worth to fix the README?

florian-dacosta:
In my opinion, it can't hurt, so :+1:

pedrobaeza:
OK, let's go then

PR #122, opened 2015-06-18
Add index on dispatch_id on move as it will be used each time a picki…

yvaucher:
…ng dispatch is opened

As #121


guewen:
:+1:

lmignon:
:+1:

lepistone:
thanks :+1:

PR #123, opened 2015-06-19
try to reserve a dispatch in bulk

lepistone:
This should optimize the common scenario where a dispatch is created to
manage thousands of moves with the same product and location. In that
case, _product_reserve is for every move.

If the bulk reservation fails, we fall back to the old method.


lepistone:
ping @guewen

lepistone:
and ping @yvaucher too :smile_cat:

yvaucher:
:+1:

jgrandguillaume:
Hi,

That's just amazingly fast now ! Thanks. :+1: 

Regards,

Joël

PR #130, opened 2015-07-17
[FIX] version in bug tracker section by using that of the module

tafaRU:


eLBati:
For a bug report, I think it's more useful to know the odoo version, rather than the module version

tafaRU:
@eLBati ok, I'll fix https://github.com/tafaRU/stock-logistics-warehouse/commit/fc9bd61545b82495ba76975dbf7f878be2637ab0 then.

PR #131, opened 2015-07-17
[FIX] version in bug tracker section by using that of the module

tafaRU:


eLBati:
See https://github.com/OCA/stock-logistics-workflow/pull/130#issuecomment-122550267

PR #135, opened 2015-07-21
[IMP] order picking dispatchs

rdeheele:
It sorts the dispatch records in a descending order by date and id.


pedrobaeza:
Maybe isn't better to use planned_date?

yvaucher:
@pedrobaeza I see no planned_date field I think the date is meant to be used as planned date?

 :+1:

pedrobaeza:
Sorry, I got confused with stock objects.

:+1:

rdeheele:
@yvaucher thanks!

PR #137, opened 2015-07-24
8.0 Add module stock_disallow_negative

alexis-via:
See the module description for more information... but the name "speaks by itself".

I'd like to get some feedback on the technical implementation of the "disallow negative stock level" feature. The current implementation is extremely simple: it adds a constraint on the "qty" field of the stock.quant object. It is so simple that I'm afraid I may have "missed something"...


gurneyalex:
Is there not a risk to block things such as returns, or inventories? At the moment this module breaks the Travis build, so some more work is required. 

It would be nice to have a few automated tests for this :smile_cat:

pedrobaeza:
https://github.com/OCA/sale-workflow/tree/8.0/sale_exception_nostock isn't a better option?

alexis-via:
@pedrobaeza it's not the same thing at all ! My module only depends on stock and prevent the stock from going negative. It's not what sale_exception_nostock is doing.

stock_disallow_negative/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+##############################################################################
+
+
+{
+    'name': 'Stock Disallow Negative',
+    'version': '0.1',
+    'category': 'Inventory, Logistic, Storage',
+    'license': 'AGPL-3',
+    'summary': 'Disallow negative stock levels by default',
+    'description': '',
```
pedrobaeza: Remove this key

jgrandguillaume:
Hi,

I think making inventory or delivery from supplier may break this constraints, do you made any tests with:
- Returns
- Inventory where you lost goods and no one are anymore in the stocks
- Simple supplier delivery

Regards,

Joël

alexis-via:
Just a quick note to say that I'm using this module in production since October 1st 2015 with my customer who has the biggest warehouse/stock management in Odoo v8. It has been a big help and it helps detect errors in stock rapidly.

clonedagain:
@lbellier you may want to review this, as you wrote a similar module?

lbellier:
Hi,
I Agree @jgrandguillaume and @gurneyalex, unit tests will be welcome...
i read quickly the code and i think this module is incomplete. Also, you can add the serial numbers too ?
Is there an exception raised on location who have not 'internal' type, like 'production', 'inventory', 'supplier', 'customer', etc... ?

@clonedagain, I wrote a module who raise exception if a propagated negative quant is created, but maybe i'm wrong...

nhomar:
Hello.

This modules is good to trace when a negative comes but som little features are needed.
1. Unittest are necesary.
2. Consumable must be checked also, sadly those products need some control also.
3. It should be configurable in order to not stop the workflow, it means, a little checkbox in the product "allow negative quants" (False by default) can help to unblock the flow temporally to continuew the process in some cases.
4. A little script which fix such negative whith known cases will be welcome also, it means, that in the case a negative wuant will come we can not stop indefinitivally a  workflow we ned to include some known cases to unblock and/or fix the quant on that moment.

For the 4 cases the 4th is one evolutive case the other 3 will help to audit the process alive.

Good job BTW, conceptually it is a +1 and it can be merged when finished.

What is the roadmad to finish this module? in order to offer help.

Redards.

nhomar:
@jgrandguillaume 

Dude can you extend this use case you mentioned?:
- Inventory where you lost goods and no one are anymore in the stocks

If you share some numbers we can work on the test and the right solution, but on your set of cases that's the one I do not get and why negative stocks should be allowed.

Regards.

alexis-via:
As this module fits perfectly my needs and the needs of my 2 customers that use it in production on v8, I don't plan to do more developments on this module. The experience is very positive and it helps detect errors in stock very early, when it's still easy to understand the cause of the error and fix it.

Of course, this module is usefull only for some deployments, not for all deployments. For example, if you have a POS, you should not use this module (or at least not on the products that are sold in the POS), because when you have the customer in front of you with the product in this hands and the stock level for that product is 0, then you should let the stock level go negative and fix it later (the only problem is that "fix it later" often becomes "fix it never !"). When you do e-commerce or industry manufacturing, if you consider that having good stock levels in your ERP is very important and you can afford to block an ERP process if you have a wrong stock level, then it is a good thing to install this module : it will help to keep "good data" in your ERP and will oblige your staff to fix wrong stock levels as soon as it is detected by the ERP.

nhomar:
@alexis-via 

Just to be clear, I am not asking more developments on your side, if you are agreed with the topics I mentioned we can join the efforts here, and have only 1 version of this topic (with which I am absollutly agreed).

We can do the improvements in 2 steps:

1.- First make the behave i mentioned (if you agree) and we can do it, make a PR to your branch and then we can merge to OCA.

2.- We can merge as it is and we finish.

What I am concern about is that even if it is functionally perfect we need to try to avoid introduce this kind of blocking modules without at least unittest (well IMHO).

We have another option:

I can give my +1 here even if I think those know issues will be present if you at least put that description you mentionesd in the README of the module and we responsibly inform the user which will install it and we simply in our side develop another version.

We solve it already making a 5 line server action directly on the DB which we simply disable from the support team once it happens in order to have the use case where it brings the negative value.

What should be sad, is that akretion and vauxoo will loss an important point of convergence here which can be discussed publically.

PErsonally I prefer converge with you here, we use such Action server I mentioned in 3 huge customers and 6 dbs with more than 10k SKU each, then I think dedicate a little more of time here with a really good set of use cases on unittest will help everybody to achieve the goals you mentioned.

Oversimplify this problem with a simple contraint which will block the process IMHO is too dangerous at least on our environments, and separate the effort is not good for everybody.

A third option is:

I can make an separate approach on my side and simply put it in our own repository but again, I prefer converge in terms of conceptual use cases as I asked to @jgrandguillaume to answer.

What do you think? we try to converge? or simply you want merge this and we can make a separate apporach?

lbellier:
@alexis-via 
I have tried your module, and it's true it helpful to keep a clean stock. No problems with 'inventory' or 'production' locations.
Maybe you can add the serial number (when traceability is enabled) in the error message to help users ?

alexis-via:
@lbellier I added lot number in error message, as you suggested.

lbellier:
Great ! it will be helpful !!! :+1:

Cedric-Pigeon:
@alexis-via Hello, do you plan to achieve this PR ? Is it still usefull ? Otherwise it will be closed in 3 days. Thanks.

rousseldenis:
@alexis-via I close this one as there is an alternative module merged. That one is used to in other versions

PR #129, opened 2015-07-10
Port stock_picking_invoice_link to the new API (and other improvements)

tafaRU:

PR #138, opened 2015-07-28
add _link_invoice_to_picking method

chafique-delli:
I added the _link_invoice_to_picking method to be able to inherit in another module and so updating of other fields on stock move.

### DIFF
--- a/stock_picking_invoice_link/stock.py
+++ b/stock_picking_invoice_link/stock.py
@@ -39,7 +39,7 @@ class stock_move(orm.Model):
             self, cr, uid, move, invoice_line_vals, context=None):
         inv_line_id = super(stock_move, self)._create_invoice_line_from_vals(
             cr, uid, move, invoice_line_vals, context=context)
-        self._link_invoice_to_picking(
+        return self._link_invoice_to_picking(
             cr, uid, move, inv_line_id, invoice_line_vals, context=context)
 
 
--- a/stock_picking_invoice_link/stock.py
+++ b/stock_picking_invoice_link/stock.py
@@ -29,13 +29,18 @@ class stock_move(orm.Model):
             'account.invoice.line', 'Invoice Line', readonly=True),
     }
 
+    def _link_invoice_to_picking(
+            self, cr, uid, move, inv_line_id, invoice_line_vals, context=None):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This can be done directly in `_create_invoice_line_from_vals`. The only piece you need is to return inv_line_id (what has to be done to not break inheritance in any case).

eLBati:
@chafique-delli I agree you have to return `inv_line_id`.

@pedrobaeza I don't understand what you mean with

> This can be done directly in _create_invoice_line_from_vals

it is done in `_create_invoice_line_from_vals` yet, without the change of this PR

pedrobaeza:
No, @chafique-delli is creating a new method `_link_invoice_to_picking`, so that's why I say that this new method is unneeded.

sebastienbeau:
May you need some additionnal information. We are working on a module that management holding flow. A picking when is deliveried is invoiced by the current company and by the holding. The holding and the current company can have different rule for sending the invoice (I mean sometime the holding can invoice the customer first and sometime the current company invoice the holding before...). We are reusing the invoicing existing code the only trouble we have is at the end of the process this module will always link the invoice (current company or holding to the picking) in our case we want to write in the field "holding_inoice_id" instead of invoice_id. This is why we need to extract the "link" method between the picking and the invoice. The holding invoicing module is here : https://github.com/OCA/multi-company/pull/14

eLBati:
@pedrobaeza ok so you say this PR is unneeded. Right

eLBati:
@pedrobaeza @sebastienbeau that being said, I am not against the new method `_link_invoice_to_picking`, but `_create_invoice_line_from_vals` must keep returning `inv_line_id`

sebastienbeau:
@eLBati done ;)

tafaRU:
It needs a rebase since https://github.com/OCA/stock-logistics-workflow/pull/129 has been merged.

Cedric-Pigeon:
@chafique-delli Hello, do you plan to achieve this PR ? Is it still usefull ? Otherwise it will be closed in 3 days. Thanks.

rousseldenis:
@chafique-delli I close this one. Please feel free to reopen it if you want

PR #139, opened 2015-08-03
8.0 stock picking reorder lines

damdam-s:
add module 'stock_picking_reorder_lines' migrated to v8 from OCA/sale-workflow

### DIFF
--- a/stock_picking_reorder_lines/stock.py
+++ b/stock_picking_reorder_lines/stock.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+#
+#    Author: Alexandre Fayolle
+#    Copyright 2013 Camptocamp SA
+#
+#    Author: Damien Crier
+#    Copyright 2015 Camptocamp SA
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+
+from openerp import models, api, fields
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+    _order = 'date_expected desc, sequence, id'
+
+    @api.model
+    def _get_sequence(self):
+        last_sequence = 10
+        return last_sequence
+
+    sequence = fields.Integer(default=_get_sequence)
+
+    @api.model
+    def _get_invoice_line_vals(self, move, partner, inv_type):
+        res = super(StockMove, self)._get_invoice_line_vals(move,
+                                                            partner,
+                                                            inv_type)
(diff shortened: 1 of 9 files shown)


pedrobaeza:
This already has @gurneyalex :+1: in old PR https://github.com/OCA/sale-workflow/pull/183

qtheuret:
:+1:

stock_picking_reorder_lines/stock.py
```diff
@@ -0,0 +1,45 @@
...
+from openerp import models, api, fields
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+    _order = 'date_expected desc, sequence, id'
+
+    @api.model
+    def _get_sequence(self):
+        last_sequence = 10
```
pedrobaeza: This still lacks the query for seeing last assigned sequence.

qtheuret: There is no sequence defined for a particular stock picking, so how to compute the next sequence (last sequence +1, +5, +10…) ? If you need to put +1 on the last sequence for each new stock move, the default sequence should be 1, not 10.

gurneyalex: @qtheuret it is not needed as we also order by id, so we can work with a single value for default. When reordering with the user interface, all the sequences are recomputed. In the main flow, the sequence is set from the SO.

gurneyalex:
:+1:

stock_picking_reorder_lines/stock.py
pedrobaeza: That's not true. Let me put you an example to see the problem:
- You add 15 lines. All lines will have sequence = 10.
- You use the widget handle to reorder one of the lines. The lines now have 1, 2, 3... 15 as sequence.
- Add a new line. The new line will appear at the bottom, but the sequence will be 10. You click on Save, and the line will be in the middle of the other lines.

This is a general problem in all the models that use a sequence. A better solution for me would be to put at least a higher number that is difficult to reach (10000 for example), but it's still not perfect.

gurneyalex: I would have thought the add at bottom would set the sequence properly. I'm in favor of setting the default to 1000, unless @damdam-s has a better solution.

damdam-s: @pedrobaeza , @gurneyalex : solution provided in the last commit

pedrobaeza:
Thanks :+1:

qtheuret:
Thanks :+1:

pedrobaeza:
Travis tests are failing, so I remove my thumb until this get solved.

stock_picking_reorder_lines/test/invoice_from_picking.yml
```diff
@@ -0,0 +1,65 @@
...
+
+-
+  I create a picking with 1 move of 2 products from stock to customer
+-
+  !record {model: stock.picking, id: outgoing_shipment_main_warehouse1}:
+    picking_type_id: stock.picking_type_out
+    invoice_state: '2binvoiced'
+    origin: 'outgoing shipment'
+    partner_id: base.res_partner_6
+    date: !eval "'%s-%s-15' %((datetime.now()-timedelta(days=60)).year, (datetime.now()-timedelta(days=60)).month)"
```
StefanRijnhart: This line makes the tests fail:

```
ValueError: time data '2015-6-15' does not match format '%Y-%m-%d %H:%M:%S'
```

damdam-s:
@pedrobaeza update :)

pedrobaeza:
Thanks :+1: 

Merging...

damdam-s:
thanks

PR #141, opened 2015-08-10
[FIX] stock_dropshipping_dual_invoice: Don't detect MTO orders as dual invoiceable

pedrobaeza:
MTO orders passes the same checks as drop-shipping orders, so we need to add one more to avoid dual invoicing in these cases.

@lepistone, please check.

### DIFF
--- a/stock_dropshipping_dual_invoice/tests/test_need_two_invoices.py
+++ b/stock_dropshipping_dual_invoice/tests/test_need_two_invoices.py
@@ -73,14 +73,32 @@ class TestNeedTwoInvoices(TransactionCase):
         }).create({})
         self.assertIs(False, wizard._need_two_invoices())
 
+    def test_so_mto_need_one_invoice(self):
+        self.so.order_policy = 'picking'
+        self.sol.route_id = self.mto_route.id
+        self.so.action_button_confirm()
+        procurements = self.so.procurement_group_id.procurement_ids
+        procurements.run()
+        po = procurements.filtered('purchase_id')[:1].purchase_id
+        self.assertTrue(po)
+        po.invoice_method = 'picking'
+        po.signal_workflow('purchase_confirm')
+        picking = po.picking_ids
+        self.assertEqual(1, len(picking))
+        picking.action_done()
+        wizard = self.Wizard.with_context({
+            'active_id': picking.id,
+            'active_ids': [picking.id],
+        }).create({})
+        self.assertIs(False, wizard._need_two_invoices())
+
     def setUp(self):
         super(TestNeedTwoInvoices, self).setUp()
         self.Wizard = self.env['stock.invoice.onshipping']
-
         customer = self.env.ref('base.res_partner_3')
         product = self.env.ref('product.product_product_36')
         dropship_route = self.env.ref('stock_dropshipping.route_drop_shipping')
-
+        self.mto_route = self.env.ref('stock.route_warehouse0_mto')
         self.so = self.env['sale.order'].create({
             'partner_id': customer.id,
         })
--- a/stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
+++ b/stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
@@ -48,9 +48,13 @@ class StockInvoiceOnshipping(models.TransientModel):
                 self.env.context['active_id'])
             so = picking.sale_id
             moves = picking.move_lines.filtered('purchase_line_id')[:1]
-            return (so.order_policy == 'picking' and
+            if (so.order_policy == 'picking' and
                     moves.purchase_line_id.order_id.invoice_method ==
-                    'picking')
+                    'picking'):
+                # We still need to investigate if this an MTO or drop-shipping
+                p_type = picking.picking_type_id
+                return (p_type.default_location_src_id.usage == 'supplier' and
+                        p_type.default_location_dest_id.usage == 'customer')
         return False
 
     @api.depends('journal_type', 'need_two_invoices')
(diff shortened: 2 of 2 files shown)


lepistone:
@pedrobaeza oh, I see: in MTO, the incoming picking has both the `picking.sale_id` and the `move.purchase_line_id` non empty.

Ok, but I'd like a test for this case, and a commentary that says that (what I wrote above).

Thanks!

pedrobaeza:
@lepistone, done!

stock_dropshipping_dual_invoice/tests/test_need_two_invoices.py
```diff
@@ -73,14 +73,30 @@ def test_so_on_delivery_need_one_invoice(self):
         }).create({})
         self.assertIs(False, wizard._need_two_invoices())
 
+    def test_so_mto_need_one_invoice(self):
+        self.so_mto.order_policy = 'picking'
```
lepistone: You can reuse the same `self.so` here and just do `self.so.route_id = ...`. It will be a new, draft one because every test is in its own transaction.

stock_dropshipping_dual_invoice/tests/test_need_two_invoices.py
```diff
@@ -73,14 +73,30 @@ def test_so_on_delivery_need_one_invoice(self):
...
+        self.assertTrue(po)
+        po.invoice_method = 'picking'
+        po.signal_workflow('purchase_confirm')
+        picking = po.picking_ids[:1]
+        picking.action_done()
+        wizard = self.Wizard.with_context({
+            'active_id': picking.id,
+            'active_ids': [picking.id],
+        }).create({})
+        self.assertFalse(wizard._need_two_invoices())
```
lepistone: (minor) the assertion `assertIs(False)` is stronger because it checks for exactly `False`. On the other hand, `assertFalse` checks for something that is `False`-ish.

I do that since when I made the following mistake:
`self.assertTrue(x.my_method)`
forgetting to put the parenthesis to actually call the method. The above test passes, while `self.assertIs(True, x.my_method)` fails as it should.

stock_dropshipping_dual_invoice/wizard/stock_invoice_onshipping.py
```diff
@@ -48,9 +48,13 @@ def _need_two_invoices(self):
                 self.env.context['active_id'])
             so = picking.sale_id
             moves = picking.move_lines.filtered('purchase_line_id')[:1]
-            return (so.order_policy == 'picking' and
+            if (so.order_policy == 'picking' and
                     moves.purchase_line_id.order_id.invoice_method ==
-                    'picking')
+                    'picking'):
+                # We still need to investigate if this an MTO or drop-shipping
```
lepistone: Can you please say in the comment that we do that because a MTO incoming picking is associated both to a sale order and to purchase order lines?

stock_dropshipping_dual_invoice/tests/test_need_two_invoices.py
```diff
@@ -73,14 +73,30 @@ def test_so_on_delivery_need_one_invoice(self):
...
+    def test_so_mto_need_one_invoice(self):
+        self.so_mto.order_policy = 'picking'
+        self.so_mto.action_button_confirm()
+        procurements = self.so_mto.procurement_group_id.procurement_ids
+        procurements.run()
+        po = procurements.filtered('purchase_id')[:1].purchase_id
+        self.assertTrue(po)
+        po.invoice_method = 'picking'
+        po.signal_workflow('purchase_confirm')
+        picking = po.picking_ids[:1]
```
lepistone: This is personal opinion, but I prefer how I did in the other test. The assertion that the picking is one is crucial: if something goes wrong with the validation of the order, there will be no picking, and I'd like to catch that explicitly. (IIRC I made this mistake while writing the test (i.e. my order was wrong, and there was no picking).

`[:1]` means we accept zero, one, or more pickings, and want to get the first one, or none if there is none. This sounds like a complex behavior with many edge cases, while the only case we want is exactly one picking (not zero, not many).

pedrobaeza: That's because both pickings are linked to the purchase order, not just the purchase one. I can keep the assert to assure that there's one picking, but I cannot remove this condition. Just one reminder: if `po.picking_ids` is an empty recordset, `po.picking_ids[:1]` will be also an empty recordset

lepistone: I know `[:1]` will accept an empty recordset: In fact I'd like to make that more explicit and remove hidden edge cases.

Thanks for your comment: I didn't realize we actually have two pickings. In that case, I'd love to have an explicit assertion that they are two, and then take one.

More importantly, if there are two and we want one, I guess we want to choose the right one, not just pick the first. What do you think? That might apply to the production code as well.

pedrobaeza: Testing it, I got confused with procurements (that indeed they are 2, and with `filtered` I get the right one), so I'm coming back again to the old form you put.

pedrobaeza: Anyways, sometimes I don't like to be so strict and only test what the module/method does, not expected behaviour in the rest of the flow, because there can be other modules that modifies this flow and screw up these tests. This one, can be an exception due to the exceptional use case that we are testing...

pedrobaeza:
Changes done. Thanks for your suggestions.

yvaucher:
:+1:

lepistone:
thanks :+1:

PR #143, opened 2015-08-12
[ADD] add module stock_backorder_strategy

laetitia-gangloff:
Define on stock_picking_type how to manage backorder.
create : default behavior
no_create : no backorder created
cancel : backorder is cancelled

### DIFF
--- a/stock_picking_backorder_strategy/README.rst
+++ b/stock_picking_backorder_strategy/README.rst
@@ -1,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Picking backordering strategies
+===============================
+
+In types of operation add a backorder strategy to be able to cancel created
+backorder or to avoid to create a backorder.
+
+Possible strategies are:
+
+* Create: it is the default behaviour, backorder is created as usual
+* No Create: no backorder are created
+* Cancel: backorder is cancelled
+
+Installation
+============
+
+To install this module, you need to:
+
+* Click on install button
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* define backorder strategy on picking type (Warehouse > Configuration > Types of Operation)
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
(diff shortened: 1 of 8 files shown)


stock_picking_backorder_strategy/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Picking backordering strategies
+===============================
+
+In types of operation add a backorder strategy to be able to cancel created
+backorder or to avoid to create a backorder.
+
+Possible strategies are :
```
yvaucher: Nitpicking of nitpicking: `:` doesn't have a space before it in english. In opposition to french for which it needs to be there.

Same for the 3 next lines

yvaucher:
:+1:

laetitia-gangloff:
What can I do for pylint ?

stock_picking_backorder_strategy/models/stock.py:31: [W0102(dangerous-default-value), StockPicking._create_backorder] Dangerous default value [] as argument (https://travis-ci.org/OCA/stock-logistics-workflow/jobs/75288135#L177)

yvaucher:
@laetitia-gangloff Well I checked it. It would have no side effect as if this argument `backorder_moves` gets it's default mutable value `[]` in official code the variable is check with `if not` and thus it set it to value of a field from picking.

I forgot pylint wouldn't let this pass, so I think you can simply put None instead of [] as default value.

If you want to do more, you can do as we usually did in old API with context.

laetitia-gangloff:
@yvaucher thank you !

I put None, and test it like with context in the old API.

stock_picking_backorder_strategy/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": "Picking backordering strategies",
+    "version": "0.1",
+    'author': "Acsone SA/NV, Odoo Community Association (OCA)",
+    "category": "Other",
```
pedrobaeza: Put "Warehouse Management"

pedrobaeza:
One last thing, but :+1:

lmignon:
:+1:  Thank you @laetitia-gangloff

PR #145, opened 2015-08-26
[MOD] <stock_cancel> reverting quant

oihane:


stock_cancel/models/stock.py
```diff
@@ -47,15 +48,27 @@ def action_revert_done(self):
             if picking.invoice_id:
                 raise exceptions.Warning(
                     _('Picking %s has invoices!') % (picking.name))
+            for move in picking.move_lines:
```
pedrobaeza, line 51: You can save one loop with: `for quant in picking.mapped('move_lines.quant_ids'):`

stock_cancel/models/stock.py
```diff
@@ -47,15 +48,27 @@ def action_revert_done(self):
             if picking.invoice_id:
                 raise exceptions.Warning(
                     _('Picking %s has invoices!') % (picking.name))
```
pedrobaeza: Better "Picking has been invoiced"

stock_cancel/models/stock.py
```diff
@@ -47,15 +48,27 @@ def action_revert_done(self):
...
+                                  'to them.')
+                                % (picking.name))
+                        quant_obj.quants_move(
+                            [(quant, quant.qty)], move, move.location_id,
+                            location_from=move.location_dest_id,
+                            lot_id=quant.lot_id.id, owner_id=quant.owner_id.id)
+                quant_obj.quants_unreserve(move)
             picking.move_lines.write({'state': 'draft'})
             picking.state = 'draft'
             if picking.invoice_state == 'invoiced' and not picking.invoice_id:
```
pedrobaeza: You don't need to check again `not picking.invoice_id`, as this is a warning condition

stock_cancel/tests/test_stock_cancel.py
```diff
@@ -0,0 +1,62 @@
...
+            })],
+        })
+        self.picking_in.action_confirm()
+        self.picking_in.do_prepare_partial()
+        self.picking_in.do_transfer()
+
+    def test_transferred_stock_picking_cancel(self):
+        self.assertEqual(self.picking_in.state, 'done')
+        self.picking_in.action_revert_done()
+        self.assertEqual(self.picking_in.state, 'draft')
```
pedrobaeza, line 52: Test here also the stock level and quant locations

stock_cancel/models/stock.py
oihane, line 51: No, I can't take a look to line 64

wdbaruni:
Any updates with this fix?

stock_cancel/models/stock.py
```diff
@@ -46,16 +47,32 @@ def action_revert_done(self):
                     % (picking.name))
             if picking.invoice_id:
                 raise exceptions.Warning(
-                    _('Picking %s has invoices!') % (picking.name))
+                    _('Picking %s has been invoiced') % (picking.name))
+            for move in picking.move_lines:
+                quants = move.quant_ids.filtered(
```
pedrobaeza, line 52: Why do you need to make this filter? With it, you're discarding associated quants that can be considered as a stopper for making the cancel, don't you think?

stock_cancel/models/stock.py
```diff
@@ -46,16 +47,32 @@ def action_revert_done(self):
...
+                    lambda q: q.location_id == move.location_dest_id)
+                if quants.filtered(lambda q: q.qty != move.product_uom_qty):
```
clonedagain, line 54: What if a single move touched several quants? You'll refuse to undo it won't you?

stock_cancel/tests/test_stock_cancel.py
```diff
@@ -0,0 +1,74 @@
...
+        self.picking_in = self.picking_model.create({
+            'partner_id': self.partner_delta.id,
+            'picking_type_id': self.picking_type_in.id,
+            'move_lines': [(0, 0, {
+                'name': self.product_1.name,
+                'product_id': self.product_1.id,
+                'product_uom_qty': 288,
+                'product_uom': self.product_1.uom_id.id,
+                'location_id': self.supplier_location.id,
+                'location_dest_id': self.stock_location.id,
```
clonedagain, line 33: You should add several lines with the same product and different lots

stock_cancel/models/stock.py
clonedagain, line 52: Also it won't work in multi-location databases, where the actual location of the quant is decided using pack operations not stock moves (using manual entry and automatic put-away strategy).
Usually you have:
- move dest = the warehouse's location
- pack operation dest = the actual place where the goods were put.

stock_cancel/tests/test_stock_cancel.py
```diff
@@ -0,0 +1,74 @@
...
+                'name': self.product_2.name,
+                'product_id': self.product_2.id,
+                'product_uom_qty': 188,
+                'product_uom': self.product_2.uom_id.id,
+                'location_id': self.supplier_location.id,
+                'location_dest_id': self.stock_location.id,
+            })],
+        })
+        self.picking_in.action_confirm()
+        self.picking_in.do_prepare_partial()
```
clonedagain, line 44: Here you must simulate what happens when users manually put away the goods in another location than the one on the stock move.

clonedagain:
:-1: at the moment I'm not convinced what you're trying to achieve is feasible, or at least reasonable.
There is a way to amend a picking in the standard, so I think this is too risky for the little good it does.

Please add tests to prove me wrong, I'll be glad to be mistaken. Messing with quants is not something we should do without **very** extensive testing.

Cedric-Pigeon:
@oihane  Hello, do you plan to achieve this PR ? Is it still usefull ? Otherwise it will be closed in 3 days. Thanks.

oihane:
@Cedric-Pigeon you can close it, I have not done anything since 2015 with this PR.

PR #158, opened 2015-09-09
Add workflow trigger on picking after assigning move

mdietrichc2c:
This PR is to add a needed `wkf_trigger` on picking after the mass assign on stock moves; it was called before in https://github.com/OCA/OCB/blob/7.0/addons/stock/stock.py#L2229 , but that is not the case anymore.

### DIFF
--- a/picking_dispatch/picking_dispatch.py
+++ b/picking_dispatch/picking_dispatch.py
@@ -205,6 +205,9 @@ class PickingDispatch(Model):
             ('state', 'in', ('confirmed', 'waiting')),
         ], context=context)
 
+        # In order to avoid calling _product_reserve on each stock move,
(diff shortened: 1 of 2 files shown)


gurneyalex:
:+1: an automated test would be nice and would have caught the bug.

guewen:
:+1:

gurneyalex:
@mdietrichc2c  could you rebase your PR on top of https://github.com/OCA/stock-logistics-workflow/pull/164 and modify the tests I added there so that they fail without your patch?

mdietrichc2c:
@gurneyalex done: I add a check on the picking (assigned after all the moves are assigned). It fails without the fix, and works with it.

lepistone:
:+1: thanks!

PR #140, opened 2015-08-03
add context propagation

lepistone:

PR #162, opened 2015-09-17
[FIX] commit 2861d1c0b608b4a610e783640cfa2f7381c6311f, remove context…

vrenaville:
This patch fix issue introduced in PR #140 
- the context is not present in V7


laetitia-gangloff:
:+1: 

but is it not possible to ask odoo to add context in stock_move.action_assign ?

guewen:
:+1:

yvaucher:
:+1: As @laetitia-gangloff said would be great to have it in kwargs plus propagate it.

https://github.com/odoo/odoo/blob/7.0/addons/stock/stock.py#L775

Thus I merge it.

PR #169, opened 2015-10-21
fix: no name attribute on field definitions

lepistone:
Sometimes, this seems to cause frontent crashes when creating
preparations. No idea why other times it seems to work fine.

ping @guewen @fclementic2c


pedrobaeza:
:+1:

tafaRU:
:+1:

lepistone:
I found out what is "sometimes": when the field is shown in a view.

guewen:
:+1:

PR #194, opened 2016-01-21
[8.0] [ADD] stock_account_purchase_line

JordiBForgeFlow:
This module will add the purchase order line to invoice lines and to account
 move lines. The ultimate goal is to establish the purchase order line as
 one of the key fields to reconcile the Goods Received Not Invoiced accrual
 account.
# Usage

The purchase order line will be automatically copied to invoice lines and
account move lines.
- If an invoice is created from a purchase order, the po line is copied
  directly to the invoice, and to the account move line upon validation.
- If an invoice is created from a stock picking, the po line is copied from
  the stock move to the invoice, and to the account move line upon validation.
- When a stock move is validated and generates an account move, the po line is
  copied to the account move line.


stock_account_purchase_line/README.rst
```diff
@@ -0,0 +1,69 @@
...
+ move lines. The ultimate goal is to establish the purchase order line as
+ one of the key fields to reconcile the Goods Received Not Invoiced accrual
+ account.
+
+
+Usage
+=====
+The purchase order line will be automatically copied to invoice lines and
+account move lines.
+* If an invoice is created from a purchase order, the po line is copied
```
elicoidal: s/po/PO

JordiBForgeFlow:
@elicoidal @pedrobaeza Would you think that this is the right name to a module that adds the po line to account move line?

pedrobaeza:
My bet is for _account_move_line_purchase_info_, but take care of the performance in so critic element.

JordiBForgeFlow:
@pedrobaeza Your proposal LGTM. What do you mean by 'take care of the performance in so critic element'? Do you mean by account move lines? 

Being able to add the purchase order line to account moves is essential to the perpetual inventory system (real time inventory valuation), to reconcile the Goods Received Not Invoiced (GRNI) account. I see no alternative to this at all. Even when this impacts performance, if the PO line is not in the AML, it is not possible to reconcile the GRNI account, which makes perpetual inventory system very hard to maintain.

pedrobaeza:
OK, I see that is not a computed field, so no problem.

JordiBForgeFlow:
@pedrobaeza Is this the right repo, or should it be placed under purchase-workflow? Not sure about it..., but certainly the dependencies are stock, purchase, account.

pedrobaeza:
Well, I don't think so. I would prefer purchase-workflow, because purchase order line is not a stock model.

elicoidal:
Name LGTM

JordiBForgeFlow:
I'm closing this one, and opening in purchase-workflow. https://github.com/OCA/purchase-workflow/pull/207

PR #195, opened 2016-01-22
[8.0] stock_ownership_availability_rules : product change quantity

laetitia-gangloff:
The problem is the following :
- use link "update quantity" on product form, set a quantity
- reuse link "update quantity" on product form , set a quantity

The final quantity is the sum instead of the new quantity.

The cause is :
- the created inventory line don't contains the link to the owner,
- all quants  have an owner
- inventory line compute theorical quantity based on quants owner

depends :
- [x] https://github.com/odoo/odoo/pull/10598

### DIFF
--- a/stock_ownership_availability_rules/model/__init__.py
+++ b/stock_ownership_availability_rules/model/__init__.py
@@ -2,3 +2,4 @@
 
 from . import quant
 from . import product
+from . import stock_change_product_qty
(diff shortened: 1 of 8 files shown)


gurneyalex:
if your Odoo PR gets ignored, it's going to be hard to merge this one... 

Can some monkey patching be considered? I'd also welcome an OCB PR so that we can restrict the need of monkey patching to upstream only. Also may I suggest a PR against odoo master ? This has better chances of being accepted...

sbidoul:
odoo/odoo#10598 is merged with a better hook, this PR now needs to be adapted

Cedric-Pigeon:
@sbidoul @gurneyalex PR adapted regarding hook accepted by Odoo.

gurneyalex:
:+1: thanks !

rousseldenis:
:+1:

PR #200, opened 2016-02-04
[FIX] enforce lot locking when using quick-create in web client

clonedagain:
In the web client, if you type a new lot name and select "create" from the pop-up menu, the client calls the `create()` method with only the name attribute. All the other fields must be passed in the context.
As a consequence, stock_lock_lot silently failed to block the lots because it didn't find the product.
Fix it by looking for a product id in both the create values and the context.


clonedagain:
@pedrobaeza @oihane, @lbellier and I fixed a bug in stock_lock_lot which let users bypass the blocking of new lots by using the quick-create feature of the web client.
Can you review this please ?

(EDIT: make it clearer that the fix is by Loïc and me. Amended the commit in that sense.)

lbellier:
I have tested it (and debug it). :+1:

pedrobaeza:
:+1:

oihane:
LGTM :+1:

PR #201, opened 2016-02-08
[9.0][port] stock_split_picking

yvaucher:
TODO: fix tests


stock_split_picking/README.rst
```diff
@@ -40,6 +63,8 @@ Maintainer
 
 This module is maintained by the OCA.
 
-OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
 
 To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: https

stock_split_picking/README.rst
```diff
@@ -18,8 +20,29 @@ This module only needs `stock` module.
 Known issues / Roadmap
 ======================
 
-* When splitting a picking in an unassigned state, wizard won't
-  be auto completed with picking lines.
+* Spliting picking with packs is not implemented
+* It is not possible to the official method `do_split` of stock picking
```
elicoidal: "It is not possible to super the official method `do_split` of stock picking"

jgrandguillaume:
Hi,

On a functional point of view, it works like a charm :+1: 

Regards

JordiBForgeFlow:
@Eficent Can you review @gmefcent?

stock_split_picking/views/stock_picking.xml
```diff
@@ -7,8 +7,11 @@
       <field name="inherit_id" ref="stock.view_picking_form" />
       <field name="arch" type="xml">
         <field name ="state" position="before">
-          <button name="split_process"
-                  states="draft,confirmed,assigned,partially_available"
+          <xpath expr="//field[@name='do_prepare_partial]/..'" position="attributes">
+            <attribute name="attrs">"{'invisible': [('state', 'in', ('draft'))]}"</attribute>
```
gmeficent: There is some issue with the view.
image

gmeficent (commented):
Functional review on local :+1: Good work.
Maybe the picking lines ought to be automatically refreshed  when clicking 'split'.
[MIG] Next step would be porting yml tests to python and updating xml tags.

Cedric-Pigeon:
@yvaucher Hi, do you plan to fix this PR ?

yvaucher:
@Cedric-Pigeon I'll have a look at it today

AaronHForgeFlow:
Hi @yvaucher Thank you for this awesome contribution :1st_place_medal:  Are you planning to work on this?

yvaucher:
@aheficent It seems this is an issue of test isolation
stock_no_negative module is adding constraint and make the tests fail. Thus I'll try to add an entry in travis to separate test and see where it lead us.

pedrobaeza:
It's better to inhibit stock_no_negative behavior when testing other modules. See https://github.com/OCA/bank-payment/blob/f8683dd7b0d313ebef31008224781b47cfc4b8a9/account_payment_transfer_reconcile_batch/models/payment_order.py#L28 for an example on inhibition.

yvaucher:
Ok then no time yet to make the change inside `stock_no_negative` or search in existing PR if it was already done as I think there was already some other PR with the same issue. Maybe I can do it later this week.

oihane:
Hi @yvaucher 
I was testing this module in a local instance because there is an issue for version 8.0 that a splitted picking writes down date_done field. I do not work with version 9.0 so maybe is that I missunderstood but here you have some screenshoots.

fireshot capture 6 - wh_out_00008 - odoo_ - http___localhost_8969_web
First one this is the new picking, you can see that the operation has 100 units (originally there were 180 units)

fireshot capture 5 - wh_out_00005 - odoo_ - http___localhost_8969_web
This one is the splitted picking, operation continues with 180 units (I think it should be 80 units)

fireshot capture 7 - wh_out_00005 - odoo_ - http___localhost_8969_web
As you can see here stock.move has only 80 units.

But as I said before I really didn't work with v9.0

oihane:
Also, why don't you use this https://github.com/OCA/OCB/blob/9.0/addons/stock/stock.py#L1724 ?

yvaucher:
@oihane please have a look at the readme for why we don't use do_split:

https://github.com/OCA/stock-logistics-workflow/pull/201/files#diff-38e0562143fc92864a575aa0bacc8658R24

oihane:
OK, sorry I didn't read that I was focused on why in v8.0 it writes down date_done.

rousseldenis:
@yvaucher There is a duplicate of this https://github.com/OCA/stock-logistics-workflow/pull/381

As it's a backport of v10, do you mind if that one superseed this ?

yvaucher:
let's close this one in favor of #381

PR #203, opened 2016-02-10
[8.0] stock_ownership_availability_rules: update restrict_partner_id

laetitia-gangloff:
When search prefered quants, there is some cases where location is not attached to a company or a partner. For example the location "Customers".


gurneyalex:
:+1: 

I'm fast tracking this as it is a trivial fix with a test.

PR #207, opened 2016-03-10
Correctly group move IDs depending on product and location

mdietrichc2c:
Issue : all the moves in picking dispatches were set as "available" if a combination of product/location was available. This only sets as "available" the ones used for `_product_reserve`.

PR #208, opened 2016-03-10
Correctly group move IDs depending on product and location

mdietrichc2c:
8.0 version of PR https://github.com/OCA/stock-logistics-workflow/pull/207


jcoux:
:+1:

lepistone:
:+1: thanks!

cyrilgdn:
After some tests, checking availability doesn't seems to work in V8, the action_assign_moves call _product_reserve on stock.location but this method does not exists anymore!

I will create a new PR to revert the PR #123 (migration of the dispatch optimisation from V7 to V8) but i think this one should be cancel.

gurneyalex:
revert PR is #210

PR #210, opened 2016-03-11
Revert PR #123 - Dispatch optimisation is not V8 compatible

cyrilgdn:
This reverts commit 241e63ba0144f18ffe5aef4cb994ad476866e2bc.

The action_assign_moves method call _product_reserve on stock.location but this method does not exists in V8.


lepistone:
Absolutely! I made this gross hack in v7, but it indeed doesn't work in v8. Thanks!

:+1:

PR #213, opened 2016-03-17
[8.0][FIX] product_unique_serial:Change on check quantity unicity

P4R:
Check quantity unicity was calculated using quanty on hand, now it check qty using the sum of quant's lot.
Moreover I remove the parameter lot_id in mothods: **check_after_action_done** and **check_unicity_qty_available**.

### DIFF
--- a/product_unique_serial/models/stock_move.py
+++ b/product_unique_serial/models/stock_move.py
@@ -29,10 +29,10 @@ class StockMove(models.Model):
     _inherit = 'stock.move'
 
     @api.model
-    def check_after_action_done(self, operation_or_move, lot_id):
+    def check_after_action_done(self, operation_or_move):
         super(StockMove, self).\
-            check_after_action_done(operation_or_move, lot_id)
-        return self.check_unicity_qty_available(operation_or_move, lot_id)
+            check_after_action_done(operation_or_move)
+        return self.check_unicity_qty_available(operation_or_move)
 
     @api.multi
     def check_unicity_move_qty(self):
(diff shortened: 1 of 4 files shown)


product_unique_serial/models/stock_move.py
```diff
@@ -52,26 +69,23 @@ def check_unicity_move_qty(self):
...
 
     @api.model
-    def check_unicity_qty_available(self, operation_or_move, lot_id):
+    def check_unicity_qty_available(self, operation_or_move):
         """
-        Check quantity on hand to verify that has qty = 1
+        Check quantity on lot to verify that has qty = 1
         if 'lot unique' is ok on product
         """
+        lot_id = operation_or_move.lot_id
```
pedrobaeza: Call it `lot`, as it points to a recordset

P4R: Done. I check all test in local and all is correct, except test 4, but I think that it's wrong because we can have the same product with diferent serial numbers.

pedrobaeza: If you consider it correct, then you should change the test accordingly

P4R: I'm waitting your opinions, but in my opinion it can be removed because this case is correct.

pedrobaeza: OK for me

P4R: Done. Now it's correct but travis shows two erros, I execute the test in local and it not happens.

gurneyalex:
@P4R can you give a look at the travis failure (https://travis-ci.org/OCA/stock-logistics-workflow/jobs/117753640#L1025)?

P4R:
@gurneyalex I check it in local, If I install only product_unique_serial it pass all tests, but if I install all repository modules for test all like Travis the test fails. I think that problem is an incompatible module. Someone can help me to review this?

gurneyalex:
@P4R https://github.com/diagramsoftware/stock-logistics-workflow/pull/1 shoud fix this by splitting the tests Can you check ?

P4R:
@gurneyalex I merged your pull request and removed space in travis.yml because test didn't exclude this module. Now test is failling in Odoo installation. Can you force to reexecute the test?

P4R:
@gurneyalex Now all tests have passed. Thanks for your help.

gurneyalex:
:+1:

elicoidal:
:+1:

lasley:
👍

PR #215, opened 2016-03-30
[IMP] Restrict blocking/unblocking to a specific group

clonedagain:
This adds an option in stock_lock_lot to restrict blocking/unblocking to a specific group


clonedagain:
@pedrobaeza I think we discussed the idea some time ago, it seems I forgot to make a proper PR at that time.

stock_lock_lot/i18n/stock_lock_lot.pot
```diff
@@ -151,3 +151,20 @@ msgstr ""
 msgid "or"
```
pedrobaeza: Please remove this file, as Transifex exporter regenerate it automatically and it can confuse.

stock_lock_lot/models/stock_production_lot.py
```diff
@@ -48,7 +48,10 @@ def onchange_product_id(self):
 
     @api.multi
     def button_lock(self):
-        '''Lock the lot if the reservations allow it'''
+        '''Lock the lot if the reservations and permissions allow it'''
```
pedrobaeza: Use double quote `"""` for docstrings

pedrobaeza:
Thanks :+1:

stock_lock_lot/README.rst
```diff
@@ -11,6 +11,15 @@ or not.
...
 field "Block new Serial Numbers/lots".
 
+Configuration
+=============
+
+To allow a user to block or unblock a Lot:
+
+#. open the user settings (menu "Configuration > Users")
+#. in the "Warehouse" section, check the box
+   "Can block/unblock Serial Numbers/Lots"
```
elicoidal: s/Can/Allow

stock_lock_lot/models/stock_production_lot.py
```diff
@@ -43,12 +43,15 @@ def _get_locked_value(self):
...
     @api.onchange('product_id')
     def onchange_product_id(self):
-        '''Instruct the client to lock/unlock a lot on product change'''
+        """Instruct the client to lock/unlock a lot on product change"""
         self.locked = self._get_product_locked(self.product_id)
 
     @api.multi
     def button_lock(self):
-        '''Lock the lot if the reservations allow it'''
+        """"Lock the lot if the reservations and permissions allow it"""
```
elicoidal, line 51: Should not the ACL managed at the form or workflow level?

stock_lock_lot/security/stock_lock_lot_security.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="0">
+        <record id="group_lock_lot" model="res.groups">
+            <field name="name">Can block/unblock Serial Numbers/Lots</field>
```
elicoidal: s/Can/Allow

elicoidal:
@clonedagain Thanks for the contribution!

stock_lock_lot/models/stock_production_lot.py
clonedagain, line 51: We tried setting the perms client-side, but then they could be bypassed RPC.
We tried setting them at ORM level, but then the field was invisible to users who had no right to change it.
So as a last resort we enforced it this way.

elicoidal, line 51: Understand: not being the most technical guy I cannot help you here find a better solution (if there is any).

clonedagain:
@elicoidal, I've changed as you suggested and added a test to improve Coveralls score.

clonedagain:
Can this be merged please?

PR #216, opened 2016-03-30
Enforce strict locking of lots (vs forced availability)

clonedagain:
Optional ~~module~~ feature to strictly enforce the blocking of lots defined in `stock_lock_lot`, by refusing to move force-assigned goods.
A checkbox lets users ~~install this module~~ disable this strict checking from the warehouse.

### DIFF
--- a/stock_lock_lot/views/res_config_view.xml
+++ b/stock_lock_lot/views/res_config_view.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="view_stock_configuration" model="ir.ui.view">
+            <field name="name">Stock settings: Strict locking</field>
+            <field name="model">stock.config.settings</field>
+            <field name="inherit_id" ref="stock.view_stock_config_settings" />
+            <field name="arch" type="xml">
+                <data>
+                    <xpath expr="//group[last()]" position="after">
+                        <group>
+                            <label for="id" string="Lot blocking" />
+                            <div>
+                                <field name="stock_lock_lot_strict" class="oe_inline" />
+                                <label for="stock_lock_lot_strict" />
+                            </div>
+                        </group>
+                    </xpath>
+                </data>
+            </field>
+        </record>
+    </data>
+</openerp>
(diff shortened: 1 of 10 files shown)


clonedagain:
@pedrobaeza @lbellier it looks like I forgot to make a PR for this one too.

stock_lock_lot/views/res_config_view.xml
```diff
@@ -0,0 +1,25 @@
...
+        <record id="view_stock_configuration" model="ir.ui.view">
+            <field name="name">Stock settings: Strict locking</field>
+            <field name="model">stock.config.settings</field>
+            <field name="inherit_id" ref="stock.view_stock_config_settings" />
+            <field name="arch" type="xml">
+                <data>
+                    <xpath expr="//group[last()]" position="after">
+                        <group>
+                            <label for="id" string="Lot blocking" />
+                            <div>
```
pedrobaeza: Why 2 nested `<div>`?

pedrobaeza:
Isn't better to embed this functionality in the same module instead of creating a new one? The configuration can remain the same, but writing a parameter instead of installing a module.

clonedagain:
I like the separation of concerns very-small modules give (I focus best this way), but I can always change that if it's blocking.

pedrobaeza:
Well, in this case I see very little added functionality, so it makes sense to have both merged in one module (to not overhead this little functionality with manifest, README, views, models...).

clonedagain:
OK why not, do you have an example of how to read the config from the main code please?

elicoidal:
LGTM (no opinion on the granularity question though)

pedrobaeza:
@clonedagain, here you have an example: https://github.com/OCA/manufacture/pull/108/files#diff-f8b1a7cda1ede9e1055014b4a12dfc2eR16

clonedagain:
@pedrobaeza I've squashed the module into stock_lock_lot as you suggested

pedrobaeza:
:+1:

clonedagain:
Rebased to resolve conflicts.

lbellier:
LGTM :+1:

PR #224, opened 2016-04-14
Log the reason when a lot is automatically locked

clonedagain:
This is useful when custom submodules can lock lots for other reasons.

Additional changes : fix the default value for the blocked checkbox (no big deal, it was overwritten anyway) ; migrate create to api v8 ; use AccessError when permissions are wrong.

### DIFF
--- a/stock_lock_lot/models/stock_production_lot.py
+++ b/stock_lock_lot/models/stock_production_lot.py
@@ -26,19 +26,32 @@ class StockProductionLot(models.Model):
         @param product: browse-record for product.product
         @return True when the category of the product or one of the parents
                 demand new lots to be locked"""
+        return self._get_lock_reason(product) != 'none'
+
+    def _get_lock_reason(self, product):
+        """Reason to create locked
+
+        @param product: browse-record for product.product"""
         _locked = product.categ_id.lot_default_locked
         categ = product.categ_id.parent_id
         while categ and not _locked:
             _locked = categ.lot_default_locked
             categ = categ.parent_id
-        return _locked
+        return _locked and 'category' or 'none'
 
     @api.one
     def _get_locked_value(self):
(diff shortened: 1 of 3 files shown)


clonedagain:
@lbellier: I think this fits what you requested.
@pedrobaeza: Another patch to stock_lock_lot, you may be interested.

stock_lock_lot/models/stock_production_lot.py
```diff
@@ -26,19 +26,32 @@ def _get_product_locked(self, product):
...
+        """Reason to create locked
+
+        @param product: browse-record for product.product"""
         _locked = product.categ_id.lot_default_locked
         categ = product.categ_id.parent_id
         while categ and not _locked:
             _locked = categ.lot_default_locked
             categ = categ.parent_id
-        return _locked
+        return _locked and 'category' or 'none'
```
pedrobaeza, line 40: Why this?

clonedagain, line 40: I'm moving the logic from `_get_product_locked` to `_get_lock_reason`, to provide a specific reason instead of just yes/no.
If the reason is 'none', the lot is unblocked ; for all other values the lot is blocked.

This module can set only 'category' or 'none' as reasons, but custom modules can add more.

For example we have a custom module which can block lots for products in a specific state, and users want to know exactly why a lot is blocked: https://github.com/numerigraphe/numerigraphe-addons/blob/8.0-lock-lot-reason/stock_lock_lot_firstuse/stock.py#L30

pedrobaeza:
OK, I see, thanks.

:+1:

yvaucher:
:+1: Needs rebase

pedrobaeza:
Please rebase

clonedagain:
Rebased.

PR #225, opened 2016-04-18
[NEW][8.0] stock_lot_scrap: New module to scrap complete lot.

carlosdauden:


stock_lot_scrap/README.rst
```diff
@@ -0,0 +1,62 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================
+Scrap Complete Lot
+==================
+
+This module adds a button in *Production Lot/Serial Number* view form to
+*Scrap* all products contained.
```
pedrobaeza: s/products/quants

stock_lot_scrap/README.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Usage
+=====
+
+Go to *Warehouse > Traceability > Serial Numbers*, and enter in view form mode.
+Click button *Scrap*
+
+
+Configuration
```
pedrobaeza, line 13: This section goes above "Usage" section

stock_lot_scrap/models/stock_scrap_lot.py
```diff
@@ -0,0 +1,56 @@
...
+            'picking_id': picking.id,
+            'scrapped': True,
+        })
+        return res
+
+    @api.multi
+    def action_scrap_lot(self):
+        self.ensure_one()
+        if not self.quant_ids:
+            raise Warning(_('Product list must be defined.'))
```
pedrobaeza: Better to put: `The lot is empty.`

stock_lot_scrap/models/stock_scrap_lot.py
```diff
@@ -0,0 +1,56 @@
...
+        move_obj = self.env['stock.move']
+        strap_location_id = self.env.ref('stock.stock_location_scrapped').id
+        picking = self.env['stock.picking'].create({
+            'origin': 'Lot: %s' % self.name,
+            'picking_type_id': self.env.ref('stock.picking_type_internal').id,
```
pedrobaeza: This doesn't server in a multi-warehouse environment. You have to get quants warehouse(s), and make a picking per warehouse, so it's possible to return more than one picking. If you have only one picking, return picking form view. Otherwise, return tree+form view.

carlosdauden:
@pedrobaeza changes done

stock_lot_scrap/models/stock_scrap_lot.py
```diff
@@ -4,14 +4,22 @@
...
     @api.multi
+    def _prepare_picking_vals(self, warehouse):
+        return {
+            'origin': _('Lot: %s') % self.name,
+            'picking_type_id': warehouse.int_type_id.id or self.env.ref(
+                'stock.picking_type_internal').id,
+        }
+
+    @api.multi
     def _prepare_move_vals(self, picking, quant, strap_location_id):
```
pedrobaeza: s/strap_location_id/scrap_location_id (replace it in the 4 places that it appears)

pedrobaeza:
Minor comment, but :+1:

pedrobaeza:
Please add Spanish translation.

andhit-r:
@carlosdauden scrap button seems overlap with another button

overlap_scrap_button

stock_lot_scrap/models/stock_scrap_lot.py
```diff
@@ -0,0 +1,75 @@
...
+            warehouse = quant.history_ids[-1:].picking_id.picking_type_id.\
+                warehouse_id
+            if warehouse.id not in warehouse_ids:
+                warehouse_ids.append(warehouse.id)
+                picking = picking_obj.create(self._prepare_picking_vals(
+                    warehouse))
+                pickings |= picking
+            move = move_obj.create(self._prepare_move_vals(
+                picking, quant, scrap_location_id))
+            quant.reservation_id = move.id
```
pedrobaeza: You have to make confirm and click assign on each move.

stock_lot_scrap/views/sock_scrap_lot_view.xml
```diff
@@ -0,0 +1,19 @@
...
+<openerp>
+<data>
+
+    <record id="view_production_lot_form" model="ir.ui.view">
+        <field name="name">stock.production.lot.scrap.form</field>
+        <field name="model">stock.production.lot</field>
+        <field name="inherit_id" ref="stock.view_production_lot_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@class='oe_button_box oe_right']" position="before">
+                <header>
```
pedrobaeza: This defines a new header because the view doesn't contain it, but it can be define in another module (like the one that blocks lots). You can overwrite fields_view_get and see if there's already a header, and if not, create it. See this code for something similar: https://github.com/OCA/crm/pull/38/files#diff-df883f43bfa35a2a786daa20cc741e74R27

pedrobaeza:
@andhit-r I have pushed a new version that injects the button and the header if not present to avoid this overlay.

:+1: for the rest of the PR that I have reviewed.

andhit-r:
@pedrobaeza now looks good :+1: 

OOT:
My concern is the overlapping will happen again depend on sequence of installation. e.g if this module installed and then user install stock_lock_lot then the button will overlap again. This due stock_lock_lot has a header tag. I haven't test this but i think in theory this will happen.

pedrobaeza:
@andhit-r, that won't happen, at least with my module. It evaluates at execution time (when loading the view) if there's a header in the view (including all the inheritances), and if not, then it adds it. Check the code in `fields_view_get` to see how it works.

andhit-r:
@pedrobaeza i've tested it. My concern is wrong. I installed stock_lot_scrap first then i installed stock_lock_lot. No overlapping. Thank you for your explanation

pedrobaeza:
I merge then.

PR #227, opened 2016-05-03
disable the exclude/include build rules

gurneyalex:
at least until the modules are ported to 9.0


pedrobaeza:
:+1: Fast-tracking as this is basic. The ideal when ported is to use context variables to inhibit the change behaviors when testing other modules, and now that workflows have been dropped is even easier.

PR #228, opened 2016-05-04
stock_split_picking : open action instead of create wizard

laetitia-gangloff:
If the wizard is created before open the view, it can be garbage collected and user can not finish his work.

Use case :
User open the wizard
Then the user "wait" a loong time
And when the user come back he is not able to finish.


pedrobaeza:
:+1:

leemannd:
:+1:

gurneyalex:
:+1:

PR #185, opened 2015-12-10
[FIX] stock_split_picking: Fixed "split" when spliting move lines in stock.picking

maurochip:
Fixed a bug in the function `split` which ignored all the records in
`pack_operation_ids` except the first. Fixes issue #175.

PR #230, opened 2016-05-09
[FIX] stock_split_picking: Fixed "split" when spliting move lines in stock.picking

laetitia-gangloff:
Superseed of #185

### DIFF
--- a/stock_split_picking/__openerp__.py
+++ b/stock_split_picking/__openerp__.py
@@ -31,7 +31,8 @@
  'data': ['view/stock_partial_picking.xml'],
  'demo': [],
  'test': ['test/test_picking_split.yml',
-          'test/test_assigned_picking_split.yml'],
+          'test/test_assigned_picking_split.yml',
+          'test/test_picking_split_two_move_lines.yml'],
  'installable': True,
  'auto_install': False,
  }
--- a/stock_split_picking/model/stock.py
+++ b/stock_split_picking/model/stock.py
@@ -64,8 +64,6 @@ class StockMove(models.Model):
         moves = move + new_move
         if move.reserved_availability > move.product_qty:
             moves.do_unreserve()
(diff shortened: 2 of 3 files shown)


hbrunn:
:+1:

yvaucher:
I tested on runbot and I still get strange results when having 2 lines on my picking.

When trying:
I add 400 M-Wir to my inventory
I add 50 Ice creams to my inventory
I create a picking with 50 M-Wir and 200 Ice creams
I mark as todo
As I have 50 available for each
I want to split 50 M-Wir and 50 Ice creams (all what is available) to be able to send them which is what the wizard propose me by default.

I end up with:
- a picking with 50 Ice creams (available) which can be transfered
- a picking with 50 M-Wir (available) and 150 Ice creams (unavailable) (M-Wir were not pushed in the right picking)

Would this be an issue with another module ?

hbrunn:
So the amounts are correct (that's what this PR is supposed to fix, so: yay!), but you'd expect a different distribution? I think that needs to be addresses as a feature of the splitting wizard.

yvaucher:
@hbrunn The amounts are correct in total but not in the right picking. It is strange that selecting in the wizard.
- 20/100 product A 
- 100/100 product B

gives me a picking
- 80/100 product A
- 100/100 product B

And a second picking with
- 20/100 product A

Still hard for the user to predict what the wizard will do.

Cedric-Pigeon (approved):

yvaucher:
@Cedric-Pigeon That was 1 year ago but I'm still concerned about my last comment. Did you take it into account ?

Cedric-Pigeon:
@yvaucher No I did not test it. So far I am not using that module anymore as I am currently migrating my project to 10.0. Perhaps the best is to open an issue to keep track.

PR #233, opened 2016-06-08
[IMP] stock_picking_invoice_link: Many2many instead of many2one for invoice references

pedrobaeza:
For dropshipping pickings with module _stock_dropshipping_dual_invoice_, you generate
two invoices for the same picking, so the many2one only reflects the last generated
invoice. With this change, now both are linked.


pedrobaeza:
cc @Tecnativa

stock_picking_invoice_link/migrations/8.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# © 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, SUPERUSER_ID
+from openupgradelib import openupgrade
```
eLBati, line 6: @pedrobaeza to correctly run these scripts, I only need to install openupgradelib and normally update the module (without using odoo server from openupgrade itself) right?

eLBati:
LGTM :+1: thanks

stock_picking_invoice_link/migrations/8.0.2.0.0/post-migration.py
pedrobaeza, line 6: Yeah, that's it. That was the main purpose when extracting openupgradelib from the full OpenUpgrade distribution.

gurneyalex, line 6: @pedrobaeza don't you need to declare the external dependency in **openerp**.py ?

pedrobaeza, line 6: No, because this is only for migrations, not for new installs.

gurneyalex:
:+1:

PR #238, opened 2016-06-17
[ADD][9.0] New module stock_picking_rate

lasley:
This module introduces a concept of rate quotes and purchasing on Stock Pickings. This is one of the modules yielded from the discussion in OCA/carrier-delivery#90.
# Stock Picking Rate

This module adds the concept of delivery rate quotes & purchasing
on Stock Pickings

This is meant to be combined with an external connector with a carrier
supplier in order to bring in rate quotes, although they can be created &
edited manually if your workflow calls for it.
# Usage

Dispatch rates can be added in Stock Picking
## Purchase Rate

The Purchase Rate wizard can be accessed by clicking the green check next
to the rate quote in a `stock.picking`.

Completing this wizard will create `purchase.order` for your rate.
If using a connector, this is the point in which a shipment purchase would
be triggered.


rousseldenis (approved):
:+1: Tested on runbot. Seems great to use!

pedrobaeza:
Stock picking dispatch has been renamed to stock batch picking I'm afraid...

lasley:
Alright I'm just going to drop the `dispatch` suffix then, because the naming doesn't work with `stock_batch_picking`. It's not actually dependent, I was just trying to make the namespaces match

@t3ddftw - Take note of this change & please update in dependent connector

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,121 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api
```
susport: Order by name the import

stock_picking_rate/tests/test_stock_picking_rate_purchase.py
```diff
@@ -0,0 +1,192 @@
...
+        return self.Wizard.create(vals)
+
+    def test_default_rate_ids(self):
+        """ It should default to active rate ids """
+        exp = [r.id for r in self.rate_ids]
+        rec_id = self.Wizard.with_context(
+            active_ids=exp,
+            active_model=self.Wizard._name,
+        ).create({})
+        res = [r.id for r in rec_id.rate_ids]
```
woodbrettm, line 25: @lasley when you have a chance can you take a look at this? It's returning an empty array, which is causing another test to fail as well.

lasley: cc @tedsalmon

lasley: Also this shoudl be `rec_id.rate_ids.ids`. But it would still be empty

woodbrettm:
Woot the build is passing locally now. Although I think Travis is failing from something unrelated?

lasley:
Yes as long as OCA/OCB build passes, you can assume odoo/odoo error is - https://odoo-community.org/groups/contributors-15/contributors-49663?mode=thread&date_begin=&date_end=

woodbrettm:
Updated

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @api.multi
+    @api.depends('state')
+    def _compute_date_purchased(self):
+        for rec_id in self:
+            if rec_id.state == 'purchase' and not rec_id.date_purchased:
+                rec_id.date_purchased = fields.Datetime.now()
+
+    @api.multi
+    def _compute_is_purchased(self):
```
obulkin: Seems like there should be a `depends` here.

stock_picking_rate/tests/test_stock_picking_rate_purchase.py
```diff
@@ -0,0 +1,195 @@
...
+    def setUp(self):
+        super(TestStockPickingRate, self).setUp()
+        self.rate_ids = [self.new_record(), self.new_record()]
+        self.Wizard = self.env['stock.picking.rate.purchase']
+        self.wizard_vals = {
+            'rate_ids': [(6, 0, [r.id for r in self.rate_ids])],
+        }
+
+    def new_wizard(self, vals=None):
+        vals = self.wizard_vals if vals is None else {}
```
obulkin: This method is not getting used anywhere. Also, the first line seems wrong. As it stands, if the call includes an argument for `vals`, that argument won't be used and `vals` will simply be set to an empty dictionary.

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @api.multi
+    def action_purchase(self):
+        wizard_id = self.env['stock.picking.rate.purchase'].create({
+            'rate_ids': [(6, 0, [r.id for r in self])],
+        })
+        return wizard_id.action_show_wizard()
+
+    @api.multi
+    def buy(self):
```
obulkin, line 110: Seems like overkill to wrap this up in a method. Is this done so that it can be overloaded later?

obulkin (changes requested):
I'm seeing a few spots that may need changes, but generally, it LGTM.

stock_picking_rate/models/stock_picking_rate.py
lasley: Exactly, it's for overload. We now have a lot of these in v10 [API interface](https://github.com/odoo/odoo/blob/10.0/addons/delivery/models/delivery_carrier.py#L142)

stock_picking_rate/tests/test_stock_picking_rate_purchase.py
lasley: Remove if not used

woodbrettm:
@obulkin Thanks! Updated.

tedsalmon (approved):
👍  LGTM! There was an issue with files having the executable bit set, but I took care of that :)

obulkin (approved):
Looks like my comments were addressed, so this should be good to go now. 👍

lasley:
@OCA/logistics-maintainers - Can a PSC please review this one? It's been sitting for a while & I would like to see merged so we can start forward port.

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,122 @@
...
+    is_purchased = fields.Boolean(
+        compute='_compute_is_purchased',
+    )
+
+    @api.multi
+    @api.depends('state')
+    def _compute_date_purchased(self):
+        for rec_id in self:
+            if rec_id.state == 'purchase' and not rec_id.date_purchased:
+                rec_id.date_purchased = fields.Datetime.now()
```
pedrobaeza, line 82: Isn't better to just write the date on the state change?

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,122 @@
...
+    list_rate = fields.Float(
+        digits=dp.get_precision('Product Price'),
+    )
+    list_rate_currency_id = fields.Many2one(
+        string='List Rate Currency',
+        comodel_name='res.currency',
+    )
+    delivery_days = fields.Integer()
+    date_delivery = fields.Datetime(
+        string='Est Delivery Date',
```
pedrobaeza, line 56: s/Est/Est.

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,122 @@
...
+    @api.multi
+    def action_purchase(self):
+        wizard_id = self.env['stock.picking.rate.purchase'].create({
+            'rate_ids': [(6, 0, [r.id for r in self])],
+        })
+        return wizard_id.action_show_wizard()
+
+    @api.multi
+    def buy(self):
+        self.state = 'purchase'
```
pedrobaeza, line 111: Here is where you should write the date

stock_picking_rate/models/stock_picking_rate.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @api.multi
+    def _expire_other_rates(self):
+        """ Expires rates in picking that are not record """
+        for rec_id in self:
+            rate_ids = rec_id.picking_id.dispatch_rate_ids.filtered(
+                lambda r: r.id != rec_id.id
```
pedrobaeza, line 118: You can compare directly records without `.id`

stock_picking_rate/tests/test_stock_picking_rate_purchase.py
pedrobaeza: Please change this

pedrobaeza (changes requested):
The module name misses the main concept for this: delivery. The rate is for delivery, not for other thing, so the module should be called __stock_picking_delivery_rate__ IMO, and it should be hosted not here, but on https://github.com/oca/delivery-carrier

lasley:
@bmw95 - please see to Pedro's comments & move this PR to the correct repo when done

@tedsalmon - please take note for your dependent module(s)

lasley:
Closing in favor of OCA/delivery-carrier#119

PR #239, opened 2016-06-21
[ADD][9.0] Create new module stock_picking_tracking

lasley:
Stock Picking Tracking
======================

This module provides an Event tracking object for Stock Pickings, which
can provide for details such as delivery tracking information.

Installation
============

* Install module

Configuration
=============

None

Usage
=====

Tracking information is located in the ``Tracking`` tab of Stock Pickings

To see a data sample:
- Inventory => [Operations] All Transfers => WH/OUT/00001 => Tracking page

Let me know if you have any suggestions!


lasley:
@t3ddftw please review

tedsalmon (approved):
👍  Looks good to me!

lasley:
Would a PSC mind reviewing when they get a chance please?

stock_picking_tracking/README.rst
```diff
@@ -0,0 +1,72 @@
...
+   :alt: License: AGPL-3
+
+======================
+Stock Picking Tracking
+======================
+
+This module provides an Event tracking object for Stock Pickings, which
+can provide for details such as delivery tracking information.
+
+Installation
```
pedrobaeza: Remove empty sections

stock_picking_tracking/models/stock_picking_tracking_group.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api
+from .stock_picking_tracking_event import OPERATION_STATES
```
pedrobaeza: As you want to share the same states across several objects, this is not the best strategy, as on inheritance you will lose the same states on all places. Better to use this technique: https://github.com/OCA/sale-workflow/blob/eb94d71ba9f75a0f84c9435b6dac10c010a35aa7/sale_order_type/models/sale_order_type.py#L20 (it also handles translations)

stock_picking_tracking/models/stock_picking_tracking_group.py
```diff
@@ -0,0 +1,140 @@
...
+            if location:
+                name += ' {location}'.format(location=location)
+            if rec_id.ref:
+                name += ' - ' + rec_id.ref
+            res.append((rec_id.id, name))
+        return res
+
+    @api.multi
+    def _process_event(self, **kwargs):
+        """ Receive & process event, handing back vals for creation
```
pedrobaeza: s/handing/handling

stock_picking_tracking/models/stock_picking_tracking_location.py
```diff
@@ -0,0 +1,41 @@
...
+    _name = 'stock.picking.tracking.location'
+    _description = 'Stock Picking Tracking Location'
+
+    event_ids = fields.One2many(
+        string='Events',
+        comodel_name='stock.picking.tracking.event',
+        inverse_name='location_id',
+    )
+    city = fields.Char(required=True)
+    zip = fields.Char(required=True)
```
pedrobaeza: Don't use standard `zip` string, which is use by Python.

pedrobaeza (changes requested):

stock_picking_tracking/models/stock_picking_tracking_group.py
lasley: Handing better described what's happening than Handling, but I think "returning" is the more appropriate term instead? I'll change to that.

Also the returns string of this docblock is wrong, I'll update that too!

lasley:
Thanks for the review @pedrobaeza - all points fixed up

@tedsalmon - please make sure a7f2ac1 ends up in your #327 (once build passes anyways)

stock_picking_tracking/models/stock_picking_tracking_group.py
```diff
@@ -122,13 +129,14 @@ def create_event(self, message, state, location_id,
...
 
         Returns:
-            stock.picking.tracking.event Recordset of new events
+            StockPickingTrackingEvent: Recordset of new events that were
+                created.
         """
         event_obj = self.env['stock.picking.tracking.event']
         res = event_obj.browse()
-        for rec_id in self:
+        for recorc in self:
```
pedrobaeza: s/recorc/record

pedrobaeza:
Should I preserve both commits on merge?

lasley:
Nah no worries - let's squash up. @tedsalmon can just rebase onto the merged commit instead

lasley:
😆  sorry for my dartboard testing w/ Travis here, I admittedly didn't want to fire up my v9.

We should be 🍏  this time around 🚀

LoisRForgeFlow (approved):
This is a good one! :+1:

LoisRForgeFlow:
What could be good is a menuitem in setting or wherever to access to `stock.picking.tracking.location`. Because I think is good to be able to configure some of them previously, if I'm not wrong you only can create them on the go (from the stock.picking) now, right?

lasley:
@lreficent - Yeah the creation of tracking locations is meant to be done by a connector library, but I don't see the harm in having a menu for them. I added it in the Inventory => Configuration menu.

Cedric-Pigeon (approved):

Cedric-Pigeon:
@pedrobaeza I think your changes request has been included.

pedrobaeza (approved):

PR #240, opened 2016-06-29
[IMP] Check that stock move_lines are in invoice lines

antespi:
Improve `test_invoice_create`

Checks that stock move_lines at stock picking are in invoice lines of invoice created


pedrobaeza:
As this is a revision of tests and all is OK in Travis/Runbot, I fast-track.

PR #241, opened 2016-06-30
stock_batch_picking

cyrilgdn:
Reworked of picking_dispatch (in V8 or less) to match Odoo 9 new picking with pack operations.


stock_batch_picking/models/res_company.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# © 2012-2014 Alexandre Fayolle, Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+    default_picker_id = fields.Many2one(
```
lasley: Can we have this at the warehouse level instead? Having this at the company level will be an issue when multiple warehouse at separate locations come into play.

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+        readonly=True,
+        string='Related pack operations',
+        compute='_compute_pack_operation_ids'
+    )
+
+    pack_operation_product_ids = fields.One2many(
+        'stock.pack.operation',
+        string='Related pack operations',
+        compute='_compute_pack_operation_product_ids',
+        inverse=lambda *args, **kwargs: None,
```
lasley, line 92: What is the effect of this? Wouldn't we just want to disallow edits?

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @api.depends('picking_ids')
+    def _compute_pack_operation_pack_ids(self):
+        for batch in self:
+            batch.pack_operation_pack_ids = batch.picking_ids.mapped(
+                'pack_operation_pack_ids'
+            )
+
+    def get_not_empties(self):
+        """ Retourne all batches in this recordset
```
lasley: s/Retourne/Return/

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+        """ Check if batches states must be changed based on pickings states.
+
+        If all pickings are canceled, batch must be canceled.
+        If all pickings are canceled or done, batch must be done.
+        If all pickings are canceled or done or *expected_state*,
+            batch must be *expected_state*.
+
+        :return: True if batches states has been changed.
+        """
+        expected_states = {'done', 'cancel'}
```
lasley, line 156: This is not a valid dictionary creation. Are you going for a list?

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+        if expected_state is not None:
+            expected_states.add(expected_state)
+
+        all_good = True
+        for batch in self.filtered(lambda b: b.state not in expected_states):
+            states = set(batch.mapped('picking_ids.state'))
+            if not states or states == {'cancel'}:
```
lasley: Invalid dictionary creation

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+                batch.state = 'cancel'
+            elif states == {'done'} or states == {'done', 'cancel'}:
```
lasley: Invalid dictionary creation

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+                all_good = False
+
+        return all_good
+
+    @api.multi
+    def action_cancel(self):
+        """ Call action_cancel for all batches pickings
+        and set batches states to cancel too.
+        """
+        if not self.picking_ids:
```
lasley: It is possible for `self` to be multiple records here, so `self.picking_ids` is not always valid. I recommend either looping the recordset then operating on that, or a `self.ensure_one()`

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -0,0 +1,213 @@
...
+        """ Retourne all batches in this recordset
+        for which picking_ids is not empty.
+
+        :raise UserError: If all batches are empty.
+        """
+        if not self.mapped('picking_ids'):
+            if len(self) == 1:
+                message = _('This Batch has no picking')
+            else:
+                message = _('These Batches have no picking')
```
lasley: s/picking/pickings

stock_batch_picking/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+    def do_transfer(self):
+        """In addition to what the method in the parent class does,
+        Changed batches states to done if all picking are done.
+        """
+        result = super(StockPicking, self).do_transfer()
+        self.mapped('batch_picking_id').verify_state()
+
+        return result
+
+    def force_transfer(self, force_qties=True):
```
lasley: IMO `force_qties` is going to be really hard to remember, as `qties` is not a real abbreviation - `force_qty` would be better

stock_batch_picking/report/batch_report.py
```diff
@@ -0,0 +1,119 @@
...
+import logging
+import time
+from os.path import commonprefix
+from openerp.report import report_sxw
+from openerp import pooler, models
+
+_logger = logging.getLogger(__name__)
+
+
+class BatchAgregation(object):
```
lasley: s/BatchAgregation/BatchAggregation/

stock_batch_picking/report/batch_report.py
```diff
@@ -0,0 +1,119 @@
...
+    def batch_notes(self):
+        return self.batch_id.notes or u''
+
+    def __hash__(self):
+        return hash(self.batch_id.id)
+
+    def __eq__(self, other):
+        return self.batch_id.id == other.batch_id.id
+
+    def iter_locations(self):
```
lasley: Possible to get a doc block here explaining the return? I'm having trouble following exactly what this method interface is

stock_batch_picking/report/batch_report.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+            if p_code not in product_qty:
+                product_qty[p_code] = operation.product_qty
+            else:
+                product_qty[p_code] += operation.product_qty
+        for p_code in sorted(products):
+            yield products[p_code], product_qty[p_code], carrier[p_code]
+
+
+class PrintBatch(report_sxw.rml_parse):
```
lasley: Can you please split this into its own file? Makes things much more manageable from a high level, and is noted in contributing doc - https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming

lasley:
Thanks for the submission @cyrilgdn - comments inline. 

IMO there is a lot of critical logic being introduced here without test coverage. An example would be the `BatchAggregation` model w/ the syntax errors in the dictionary creations. 

@pedrobaeza - Any idea why there is no coverage.io for this repo? Travis config looks fine, is it maybe just not configured in coveralls? It would definitely help us in identifying the test coverage gaps in this PR.

cyrilgdn:
@lasley In Python, {'done', 'cancel'} is a set, not a dictionary. It's similar to set(['done', 'cancel']).

The coverage is 100% except for report logic. It was copied and adapt from picking_dispatch, I don't really know how to test it.

stock_batch_picking/models/stock_batch_picking.py
cyrilgdn: Indeed, I will add a test  and fix it.

stock_batch_picking/models/stock_batch_picking.py
cyrilgdn, line 92: I should have added (and will add) a comment for that, without this inverse (but do nothing), the modifications on pack operation (in views, I put qty_done field editable) will not be saved.

lasley:
Whoah TIL on the set shorthand - Neat - Thanks!

Understood, I won't force the testing then - although I would also be more than happy to assist if you had some specific questions on it.

stock_batch_picking/models/stock_batch_picking.py
lasley, line 92: Oh ok this makes sense. I had a feeling there was some sort of logical design decision behind this; a comment would be great!

stock_batch_picking/views/stock_batch_picking.xml
```diff
@@ -0,0 +1,148 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- stock.batch.picking form view -->
+    <record model="ir.ui.view" id="stock_bacth_picking_form">
+        <field name="name">stock.bacth.picking.form</field>
```
yvaucher: s/bacth/batch/

stock_batch_picking/views/stock_batch_picking.xml
```diff
@@ -0,0 +1,148 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- stock.batch.picking form view -->
+    <record model="ir.ui.view" id="stock_bacth_picking_form">
```
yvaucher: typo in xmlid too :-/

Did you read any H. P. Lovecraft lately ?

yvaucher:
Can't we retrieve part of the translations ?

cyrilgdn:
Updated with remarks taking in account. (typo, some doc, translations, ...)

rafaelbn:
Please @carlosdauden review, thanks!

stock_batch_picking/models/stock_batch_picking.py
yvaucher, line 156: This is a set.

Thus adding `expected_state` ensure uniqueness.

yvaucher:
:+1:

stock_batch_picking/models/res_company.py
lasley: IMO this is a blocking detail as it does not play nicely with multiple warehouses that would have default pickers.

pedrobaeza: I agree. I'll label this as "Needs fixing" until then

cyrilgdn:
@lasley @pedrobaeza default_picker_id moved to stock.warehouse

stock_batch_picking/wizard/batch_picking_creator.py
```diff
@@ -13,6 +13,25 @@ class StockBatchPickingCreator(models.TransientModel):
     _name = 'stock.batch.picking.creator'
     _description = 'Batch Picking Creator'
 
+    def _default_picker_id(self):
```
lasley: Please put this below field declarations per [model attribute ordering in Contrib docs](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field)

stock_batch_picking/wizard/batch_picking_creator.py
```diff
@@ -27,7 +46,7 @@ class StockBatchPickingCreator(models.TransientModel):
 
     picker_id = fields.Many2one(
         'res.users', string='Picker',
-        default=lambda self: self.env.user.company_id.default_picker_id,
+        default=_default_picker_id,
```
lasley: Please utilize lambda declaration for defaults in order to allow for overloading via inherited models & for proper method/field ordering: `lambda s: s.default_picker_id()`

lasley:
Thanks for the update @cyrilgdn - few incredibly minor comments inline then we'll be GTG 👍

cyrilgdn:
@lasley Updated with remark

lasley:
Looking forward to using this module, thanks @cyrilgdn :+1:

gurneyalex:
:+1:

stock_batch_picking/wizard/batch_picking_creator.py
```diff
@@ -0,0 +1,81 @@
...
+        """ Return default_picker_id from the main company warehouse
+        except if a warehouse_id is specified in context.
+        """
+        warehouse_id = self.env.context.get('warehouse_id')
+        if warehouse_id:
+            warehouse = self.env['stock.warehouse'].browse(warehouse_id)
+        else:
+            warehouse = self.env['stock.warehouse'].search([
+                ('company_id', '=', self.env.user.company_id.id)
+            ], limit=1)
```
carlosdauden: You can remove after lines adding [:1] here:

warehouse = self.env['stock.warehouse'].search([
        ('company_id', '=', self.env.user.company_id.id)
    ], limit=1)**[:1]**

stock_batch_picking/wizard/batch_picking_creator.py
```diff
@@ -0,0 +1,81 @@
...
+
+            if warehouse:
+                warehouse = warehouse[0]
+            else:
+                warehouse = False
+
+        return warehouse.default_picker_id if warehouse else False
```
carlosdauden: After changes return **warehouse.default_picker_id** directly

stock_batch_picking/report/batch_report.py
```diff
@@ -0,0 +1,133 @@
...
+        default_code
+
+        locations: a tuple (source_location, dest_location)
+        """
+        products = {}
+        product_qty = {}
+        carrier = {}
+        operations = self.operations_by_loc[locations]
+        for operation in operations:
+            p_code = operation.product_id.default_code
```
carlosdauden: Seems to me to done on purpose, but, why it is not grouped by product id?
If several products have not established default_code, total quantity (all products) is printed in only one line with the description of the first product.
You can sort by default_code anyway.

stock_batch_picking/wizard/batch_picking_creator.py
```diff
@@ -43,14 +43,9 @@ def _default_picker_id(self):
         else:
             warehouse = self.env['stock.warehouse'].search([
                 ('company_id', '=', self.env.user.company_id.id)
-            ], limit=1)
+            ], limit=1)[:1]
```
lasley: Aren't we achieving the desired result already with the `limit=1`? The result of this should be either a singleton or an empty recordset which would eval to False.

guewen: yep, `[:1]` is useless

cyrilgdn: @lasley  Right! And I changed it (again...)

As it was a very very little change, can we merge this PR ?

stock_batch_picking/report/batch_report.py
pedrobaeza: You still need to fix this point

stock_batch_picking/report/batch_report.py
pedrobaeza: And this one

stock_batch_picking/wizard/batch_picking_creator.py
pedrobaeza: Here the use is correct

carlosdauden: forgiveness , seeing the lines later I did not notice at limit param

stock_batch_picking/wizard/batch_picking_creator.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        warehouse_id = self.env.context.get('warehouse_id')
+        if warehouse_id:
+            warehouse = self.env['stock.warehouse'].browse(warehouse_id)
+        else:
+            warehouse = self.env['stock.warehouse'].search([
+                ('company_id', '=', self.env.user.company_id.id)
+            ], limit=1)
+
+            if warehouse:
```
pedrobaeza: This should be removed

stock_batch_picking/report/batch_report.py
cyrilgdn: @lasley @pedrobaeza Just to understand, I should have 2 or 3 files ?
One file for the report_sxw class, one for the AbstractModel and what about the BatchAggregation helper class ?

stock_batch_picking/wizard/batch_picking_creator.py
cyrilgdn: @pedrobaeza Already removed as you commented an outdated diff ;)

stock_batch_picking/report/batch_report.py
pedrobaeza: One file per class, except nested models like `sale.order` and `sale.order.line`, that can be in the same file.

cyrilgdn: Even for no model classes (like BatchAggregation which has no meaning alone) ?

stock_batch_picking/report/batch_report.py
cyrilgdn: Ok, I will try to fix this as soon as I can...

It's True that, as I'm still not comfortable with reports, I've just migrated the picking_dispatch report file and made the little adaptations :) 
I did not pay attention to existing bugs, my bad.

stock_batch_picking/report/batch_report.py
pedrobaeza: Yes, even for abstract models.

cyrilgdn:
@lasley @pedrobaeza @carlosdauden Fixes are done. But can someone tell me how to fix the new pylint-odoo error (old-api7-method-defined) in this case?

lasley:
I'm as stumped as you on that one @cyrilgdn - I see no examples of an api decorated `__init__` in Odoo code. @moylop260 I think we may need an exception in our linter here.

moylop260:
FYI the red of travis is because pylint show:

``` bash
stock_batch_picking/report/batch_report.py:7: [W0403(relative-import), ] Relative import 'print_batch', should be 'stock_batch_picking.report.print_batch'
stock_batch_picking/report/print_batch.py:9: [W0403(relative-import), ] Relative import 'batch_aggregation', should be 'stock_batch_picking.report.batch_aggregation'
```

The line [stock_batch_picking/report/batch_report.py:7](https://github.com/cyrilgdn/stock-logistics-workflow/blob/50658bcf68e120d24194e8973e6b5c167a6023f4/stock_batch_picking/report/batch_report.py#L7)
You should change it:
`s/from print_batch import PrintBatch/from . print_batch import PrintBatch/g`  (Notice the prefix dot)

The line [stock_batch_picking/report/print_batch.py:9](https://github.com/cyrilgdn/stock-logistics-workflow/blob/50658bcf68e120d24194e8973e6b5c167a6023f4/stock_batch_picking/report/print_batch.py#L9)
You should change it:
`s/from batch_aggregation import BatchAggregation/from . batch_aggregation import BatchAggregation/g` (Notice the prefix dot)

It because you have this library in the same path and avoid errors with libraries named with the same "virtual path" the dot use in the "virtual path" and `here path`

moylop260:
> I'm as stumped as you on that one @cyrilgdn - I see no examples of an api decorated `__init__` in Odoo code. @moylop260 I think we may need an exception in our linter here.

Maybe we could use the new api for reporting using `AbstractModel`
You could find examples in branch master of odoo/odoo with:
`rgrep "AbstractModel" addons/*/report --include=*.py`

Good examples:
- [addons/account/report/account_aged_partner_balance.py:10](https://github.com/odoo/odoo/blob/621a97dd44405343983b689fd3e5b3d2fbc078ff/addons/account/report/account_aged_partner_balance.py#L10)
- [addons/hr_holidays/report/holidays_summary_report.py:9](https://github.com/odoo/odoo/blob/621a97dd44405343983b689fd3e5b3d2fbc078ff/addons/hr_holidays/report/holidays_summary_report.py#L9)
- [addons/hr_payroll/report/report_contribution_register.py:11](https://github.com/odoo/odoo/blob/621a97dd44405343983b689fd3e5b3d2fbc078ff/addons/hr_payroll/report/report_contribution_register.py#L11)

cyrilgdn:
Checks green!

@moylop260 With the new api reporting system, we have to override render_html and redefine the entire docargs dict (with docs, doc_ids, etc....) ??
In old api we just have to update the localcontext dict...

stock_batch_picking/i18n/stock_batch_picking.pot
```diff
@@ -0,0 +1,395 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove POT file

moylop260: If the guideline changed after https://github.com/OCA/maintainer-tools/issues/97 could add it in our guidelines?

pedrobaeza: My opinion on this have changed since that issue, because I found somebody translating from an outdated POT file.

cyrilgdn:
@pedrobaeza @moylop260 POT file removed.

cyrilgdn:
Can we merge this PR?

yvaucher:
Thanks @cyrilgdn 
Merging

PR #243, opened 2016-07-06
[ADD][9.0] Create new module stock_quant_pack_dimension

woodbrettm:
Hi there, 

Fairly simple module that adds dimensions to packages for shipping/delivery. It extends sale_stock with length, width, height, and volume calculated in m/cm.

Cheers!
Brett
@lasley


pedrobaeza:
Why don't call it _stock_quant_package_dimension_? This abbreviation is not helpful.

pedrobaeza:
Isn't better to put this data in product.packaging?

lasley:
@pedrobaeza - Our use case is in the shipping itself, although adding this data to product.packaging in another module may be beneficial as well. In order to purchase a shipping label in the US a concept of length, width, and height has to be provided along with the standard weight.

We will also be making a module for templating of ordering of the packaging itself in order to represent the shipping supplies that are available for use/purchase.

Please correct me if I'm wrong, but I do not think that we could fulfill our use case with this data exclusively in product.packaging. Don't the product packages add into quant packages for picking and eventual shipment?

pedrobaeza:
If the dimensions are going to be different for each quant package that has the same product.packaging, then the module is fine, but what is usual is to put this information at product.packaging level, and then simply put related fields at stock.quant.package level.

lasley:
What about in the event that different types of products and product packaging are shipped in the same quant package? Is the strategy to put all similar product packages in separate quants, then pair the quant packages w/ parent_id? Seems a bit burdensome IMO

pedrobaeza:
But the objective of a package is to model a physical package that has an specific dimension no matter the content of it. For example, a 30x20x20 box is going to always have these dimensions, no matter if you fill it half or full.

lasley:
That's the objective of the quant package right? This was the logic we were following by adding the dimensions at the quant package level, because the physical box is what gets moved around and eventually shipped.

I envisioned a product package as the package that a product is sold in, such as a 10 pack or whatever. Multiple unknown sets of product packages combine into a quant package in order to be shipped out, which is where this module is coming in.

I see how adding this dimension to product packaging would help in terms of workflow, as this would unlock the ability to automatically determine our quant package sizes.

I think I'm missing something though, because your 30x20x20 box example is how I want this to work - it's the way that makes sense. I don't see any sort of template mechanism for any sort of packaging allowing these standard sizes to come in. I was going for something similar here which we're planning on refactoring soon - https://github.com/OCA/carrier-delivery/pull/90/files#diff-dd6df438d8028a1d2e37ed0a275db68bR8

pedrobaeza:
I have told you bad, Dave. My memory sometimes is not so good.

Indeed, product packaging is more intended to create sale packagings or logical packs of sale. For logistic purposes, we have in Odoo standard the model product.ul, that already contains this info. We have the module https://github.com/odoomrp/odoomrp-wip/tree/8.0/stock_picking_package_info that summarizes this info at picking level, and also this one: https://github.com/odoomrp/odoomrp-utils/tree/8.0/product_packaging_views for adding views that helps you to setup these packages.

With this, I think you don't need anything more. Please check.

lasley:
Hah I'm with you there on the memory Pedro, no worries!

Your module seems to hit my use case. Looks like the `product.ul` no longer exists in v9 & no comparable data exists. Any idea if there is a new strategy here, or if this is simply a matter of having to forward-port a model?

This epic is leading up towards an external shipping connector that I would like to bring over to OCA, so we need to make sure to use all OCA modules in order to comply with guidelines IIRC. Do you mind if we upgrade your modules to v9 & bring to OCA?

lasley:
Hah I'm with you there on the memory Pedro, no worries!

Your module seems to hit my use case. Looks like the `product.ul` no longer exists in v9 & no comparable data exists. Any idea if there is a new strategy here, or if this is simply a matter of having to forward-port a model?

This epic is leading up towards an external shipping connector that I would like to bring over to OCA, so we need to make sure to use all OCA modules in order to comply with guidelines IIRC. Do you mind if we upgrade your modules to v9 & bring to OCA?

pedrobaeza:
Yes, of course. The idea is to bring to OCA one day, but as you can see, there are too many open fronts to work on, and my team can't attend everything right now, but as always, I'll be around and help you in reviews and so on.

lasley:
Sounds like a plan, thanks for the insight. 

@bmw95 - LABS-193 to supersede this, please close.

woodbrettm:
Nice! Yeah it seemed that product.ul fulfilled what was done in this module. I'll close the pull request.

Cheers

PR #244, opened 2016-07-07
[9.0] Port stock_picking_mass_assign

mdietrichc2c:


stock_picking_mass_assign/__openerp__.py
```diff
@@ -1,55 +1,21 @@
...
+    'website': 'http://www.camptocamp.com',
+    'data': [
+        'data/cron_data.xml',
+        'wizard/check_assign_all_view.xml',
+    ],
+    "test": [
+        'test/test_check_assign_all.yml',
+    ],
+    'installable': True,
+    'auto_install': False,
```
lasley: No need to define this key; it is default & not required by any automation

stock_picking_mass_assign/data/cron_data.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <record id="ir_cron_check_assign_all" model="ir.cron">
+        <field name="name">Check Availability of Delivery Orders</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">hours</field>
+        <field name="numbercall">-1</field>
+        <field eval="False" name="doall"/>
```
lasley: [Please put name attribute before the value](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#format)

stock_picking_mass_assign/models/stock.py
```diff
@@ -0,0 +1,31 @@
...
+                      ('state', '=', 'confirmed')]
+            pickings = self.search(domain, order='min_date')
+
+        for picking in pickings:
+            try:
+                picking.action_assign()
+            except Exception:
+                # ignore the error, the picking will just stay as confirmed
+                name = picking.name
+                _logger.info('error in action_assign for picking %s',
```
lasley: IMO `_logger.exception` would be more appropriate here. Also, please add translation method call

stock_picking_mass_assign/wizard/check_assign_all.py
```diff
@@ -1,135 +1,76 @@
...
-    }
-
-    def check(self, cr, uid, ids, context=None):
-        context = context or {}
-        picking_obj = self.pool['stock.picking']
-        process_wizard_obj = self.pool['stock.partial.picking']
-        picking_ids = context.get('active_ids')
-        wizard_id = ids[0]
+    @api.multi
+    def _default_force_availability(self):
```
lasley, line 14: [Please put these methods below the field declarations](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field)

lasley:
@mdietrichc2c thank you for the submission. Few comments inline but otherwise code 👍

mdietrichc2c:
@lasley thanks for the review. I changed all the points except the last one : the reason is that the 2 methods defined before the fields are used in those same fields (as default values), hence it is better to define them before.

lasley:
Hi @mdietrichc2c - Thanks for the fixes. ~~I am blocking for the field declarations however, because this is in violation of the contribution guidelines for field ordering - https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field~~

```
In a Model attribute order should be

Private attributes (_name, _description, _inherit, ...)
Default method and _default_get
Fields declarations
Compute and search methods in the same order than field declaration
Constrains methods (@api.constrains) and onchange methods (@api.onchange)
CRUD methods (ORM overrides)
Action methods
And finally, other business methods.
```

lasley:
Oops I'm dumb my bad I should learn to read the docs I'm quoting 👍

yvaucher:
:+1:

PR #245, opened 2016-07-07
[ADD] website_stock_picking: Create new module

lasley:
Not sure if this is the right repo, but this is a PoC adding a basic stock picking front end in website. The goal of this project will mainly be to support a mobile workflow using a barcode scanner, something that is now locked behind Enterprise licensing in v9.

This isn't done by any means, but you should get the general picture by what I have here. I didn't want to bother thinking of a good spot to put the link yet, so it's on the main page `Pickings`. 

The listing has a simple search and kanban-ish listing of the pickings matching the search. Click in and you can update the done quantities.

Not done yet:
- Validating the picking
- Split packages
- Probably a lot of other things

Let me know thoughts, if any!


lasley:
Moving to OCA/web

PR #247, opened 2016-07-18
[ADD][9.0] stock_picking_package_info

lasley:
This module displays in the picking_ids, all packages appearing in operations and
the package provides the following information:
- Width
- Height
- Length
- Total estimated weight
- Total estimated net weight
- Permitted volume
- Total volume charge
- Empty package weight
- Real weight

Logic was provided by @odoomrp, just upgraded to v9

Few notes, making this a WIP if someone will please tag
- Save for later method was removed & logic for `create_all_move_packages` was moved to the `process` method of the wizard. From that I can tell, this allows us to not need to save due to more streamlined v9 wizard, but I could be wrong and would love a comment from @pedrobaeza  on that 😉 
- Tests are failing, and more need to be added


lasley:
@pedrobaeza - Can you assist with the missing CLA? I don't want to bother fixing up all the tests and stuff until I know we're good in that respect 😉

pedrobaeza:
It's covered by Avanzosc's CLA. cc @anajuaristi

lasley:
Hooray thanks @pedrobaeza ! I'll swoop these tests up in the next few days then

.travis.yml
```diff
@@ -41,4 +41,4 @@ script:
   - travis_run_tests
 
 after_success:
-  - travis_after_tests_successs
+  - travis_after_tests_success
```
lasley, line 44: Just to bring attention to this - we will remove before merge, it's just so we can identify the coverage gaps while working. A rebase onto current 9 contained a not fun amount of conflict work, so this seemed like a good interim solution.

lasley:
@t3ddftw did you have to make any changes to this while functionally testing during implementation in laslabs/odoo-connector-easypost#2?

tedsalmon:
Hey @lasley,

Yes, these are the changes I made:
- Add name_get to `product.packaging.template` to return dimensions in name
- Fix spelling of 'excluding' in help of `stock.quant` -> `weight_net`
- Remove requirement of `product_pack_tmpl_id` in `stock.quant.package`
- Add Packaging Template to product_packaging_view
- Update invisible attribute in stock_picking_view.xml to use `package_ids` instead of outdated `package` attribute

lasley:
Excellent, please add relevant logic into this PR so we can review 😄

tedsalmon:
@lasley,

The changes have already been dropped in and squashed :)

.travis.yml
lasley: Rebase onto current 9.0 so this commit is removed, PR was merged that included it

stock_picking_package_info/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+
+{
+    "name": "Stock Picking Package Info",
+    "version": "9.0.1.0.0",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "author": "OdooMRP team, "
+              "AvanzOSC, "
+              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
```
lasley: Change this to Tecnativa - correct me if I'm wrong @pedrobaeza

stock_picking_package_info/models/product_packaging.py
```diff
@@ -0,0 +1,41 @@
...
+        ondelete='restrict',
+        required=True,
+    )
+    ean = fields.Char(
+        string='EAN',
+        size=14,
+        help='The EAN code of the package unit',
+    )
+    code = fields.Char(
+        string='Code',
```
lasley: This string is implicit, remove

stock_picking_package_info/models/product_packaging.py
```diff
@@ -0,0 +1,41 @@
...
+
+from openerp import fields, models
+
+
+class ProductPackaging(models.Model):
+    _inherit = 'product.packaging'
+    _inherits = {'product.packaging.template': 'product_pack_tmpl_id'}
+
+    qty = fields.Integer(
+        string='Package by layer',
```
lasley: s/layer/Layer

stock_picking_package_info/models/product_packaging_template.py
```diff
@@ -0,0 +1,99 @@
...
+        ('pack', 'Pack'),
+        ('box', 'Box'),
+        ('pallet', 'Pallet'),
+    ],
+        required=True,
+    )
+    product_tmpl_id = fields.Many2one(
+        comodel_name='product.template',
+        string='Product',
+        select=1,
```
lasley: select is deprecated, `index=True`

stock_picking_package_info/models/stock_picking.py
```diff
@@ -0,0 +1,126 @@
...
+        compute='_calc_picking_package_info_ids'
+    )
+    num_packages = fields.Integer(
+        string='Number of Packages',
+        store=True,
+        compute='_calc_picking_package_info_ids',
+    )
+
+    @api.multi
+    @api.depends('pack_operation_ids', 'pack_operation_ids.result_package_id')
```
lasley: This also depends on `pack_operation_ids.package_id`

stock_picking_package_info/models/stock_picking.py
```diff
@@ -0,0 +1,126 @@
...
+    def _calc_picking_packages(self):
+        for rec_id in self:
+            rec_id.packages = (rec_id.pack_operation_ids.mapped(
+                'result_package_id') | rec_id.pack_operation_ids.mapped(
+                'package_id'))
+
+    @api.multi
+    @api.depends('pack_operation_ids', 'pack_operation_ids.result_package_id',
```
lasley: This also depends on `pack_operation_ids.package_id`, `pack_operation_ids.quant_ids`

stock_picking_package_info/models/stock_picking.py
```diff
@@ -0,0 +1,126 @@
...
+                                 precision_rounding=rounding) > 0:
+                    if forced_qties.get(move.product_id):
+                        forced_qties[move.product_id] += forced_qty
+                    else:
+                        forced_qties[move.product_id] = forced_qty
+            for vals in picking._prepare_pack_ops(
+                    picking, picking_quants, forced_qties):
+                domain = [('picking_id', '=', picking.id)]
+                if vals.get('lot_id', False):
+                    domain += [('lot_id', '=', vals.get('lot_id'))]
```
lasley: This attr is established to exist in the if statement. It would be cleaner and more efficient to use dict lookup here instead of .get

stock_picking_package_info/models/stock_picking.py
```diff
@@ -0,0 +1,126 @@
...
+                if vals.get('product_id', False):
+                    domain += [('product_id', '=',
+                                vals.get('product_id'))]
```
lasley: This attr is established to exist in the if statement. It would be cleaner and more efficient to use dict lookup here instead of .get

stock_picking_package_info/models/stock_picking.py
```diff
@@ -0,0 +1,126 @@
...
+                packs = pack_op_obj.search(domain)
+                if packs:
+                    qty = sum([x.product_qty for x in packs])
+                    new_qty = vals.get('product_qty', 0) - qty
+                    if new_qty > 0:
+                        vals.update({'product_qty': new_qty})
```
lasley: Dict assignment would be more efficient here

stock_picking_package_info/models/stock_picking_package_weight_lot.py
```diff
@@ -0,0 +1,32 @@
...
+from openerp.addons import decimal_precision as dp
+
+
+class StockPickingPackageWeightLot(models.Model):
+    _name = 'stock.picking.package.weight.lot'
+    _description = 'Stock Picking Package Weight Total'
+    _order = 'sequence'
+
+    sequence = fields.Integer()
+
```
lasley: Remove blank line

stock_picking_package_info/models/stock_quant_package.py
```diff
@@ -0,0 +1,198 @@
...
+    @api.multi
+    @api.depends('quant_ids', 'quant_ids.total_volume')
+    def _compute_total_volume(self):
+        for rec_id in self:
+            rec_id.total_volume = 0
+            for quant_id in rec_id.quant_ids:
+                rec_id.total_volume += quant_id.total_volume
+                _logger.debug('-' * 20)
+                _logger.debug('Total Volume')
+                _logger.debug(rec_id.total_volume)
```
lasley: Bit heavy on the logging, let's combine these statements

stock_picking_package_info/reports/label_creator_palet_report.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
lasley: Use `<odoo>` tag & remove data

stock_picking_package_info/views/product_packaging_template_view.xml
```diff
@@ -0,0 +1,120 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright (C) 2016-TODAY LasLabs, Inc. [https://laslabs.com]
```
lasley: Use new header & remove perpetual copyright. Also, did we create this file? Shouldn't there be another entity copyright here?

stock_picking_package_info/views/product_packaging_template_view.xml
```diff
@@ -0,0 +1,120 @@
...
+    @author Dave Lasley <dave@laslabs.com>
+    @license AGPL-3
+-->
+
+<odoo>
+    <data>
```
lasley: Remove data tag

stock_picking_package_info/views/product_packaging_view.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright (C) 2016-TODAY LasLabs, Inc. [https://laslabs.com]
```
lasley: Use new header & remove perpetual copyright. Also, did we create this file? Shouldn't there be another entity copyright here?

stock_picking_package_info/views/product_packaging_view.xml
```diff
@@ -0,0 +1,56 @@
...
+    @author Dave Lasley <dave@laslabs.com>
+    @license AGPL-3
+-->
+
+<openerp>
```
lasley: Use odoo tag, remove data

stock_picking_package_info/views/product_template_view.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright (C) 2016-TODAY LasLabs, Inc. [https://laslabs.com]
+    @author Dave Lasley <dave@laslabs.com>
+    @license AGPL-3
+-->
+
+<openerp>
```
lasley: Use odoo tag

stock_picking_package_info/views/product_template_view.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright (C) 2016-TODAY LasLabs, Inc. [https://laslabs.com]
```
lasley: Use new header & remove perpetual copyright. Also, did we create this file? Shouldn't there be another entity copyright here?

stock_picking_package_info/views/stock_picking_view.xml
```diff
@@ -0,0 +1,85 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright (C) 2016-TODAY LasLabs, Inc. [https://laslabs.com]
```
lasley: Use new header & remove perpetual copyright. Also, did we create this file? Shouldn't there be another entity copyright here?

stock_picking_package_info/views/stock_picking_view.xml
```diff
@@ -0,0 +1,85 @@
...
+    @author Dave Lasley <dave@laslabs.com>
+    @license AGPL-3
+-->
+
+<openerp>
```
lasley: odoo tag, remove data

stock_picking_package_info/views/stock_quant_package_view.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright (C) 2016-TODAY LasLabs, Inc. [https://laslabs.com]
```
lasley: Use new header & remove perpetual copyright. Also, did we create this file? Shouldn't there be another entity copyright here?

stock_picking_package_info/views/stock_quant_package_view.xml
```diff
@@ -0,0 +1,73 @@
...
+    @author Dave Lasley <dave@laslabs.com>
+    @license AGPL-3
+-->
+
+<openerp>
```
lasley: odoo tag, remove data

lasley (commented):
Thanks @t3ddftw & @bmw95 - Comments inline

stock_picking_package_info/models/stock_quant_package.py
woodbrettm: I think those may have been added for testing purposes. Can probably just be removed.

stock_picking_package_info/views/product_packaging_template_view.xml
tedsalmon: For the record, these files did not have a copyright on them previously. [See this link.](https://github.com/odoomrp/odoomrp-wip/blob/8.0/stock_picking_package_info/views/stock_quant_package_view.xml)

.travis.yml
tedsalmon: This is actually caught up with 9.0 already.

tedsalmon:
Hey @lasley,

All the requested changes have been made and squashed.

Thanks! 😄

lasley:
@t3ddftw - Please don't squash your changes after reviews. I have to review all of this code over again now (35 files), instead of just the diff.

.travis.yml
lasley: You didn't rebase onto 9.0 like I asked. Please do. This is in the diff & it should not be

lasley:
@t3ddftw - please look into Runbot warnings:

```
2016-10-19 15:05:39,859 3050 WARNING 3270174-247-066d5a-all openerp.models.schema: Table `product_packaging`: column `qty` has changed type (DB=float8, def=integer), data moved to column `qty_moved0`
2016-10-19 15:05:40,011 3050 INFO 3270174-247-066d5a-all openerp.sql_db: Programming error: column "name" named in key does not exist
, in query ALTER TABLE "product_packaging_template" ADD CONSTRAINT "product_packaging_template_name_uniq" UNIQUE(name)
2016-10-19 15:05:40,012 3050 WARNING 3270174-247-066d5a-all openerp.models.schema: Table 'product_packaging_template': unable to add 'UNIQUE(name)' constraint !
 If you want to have it, you should update the records and execute manually:
ALTER TABLE "product_packaging_template" ADD CONSTRAINT "product_packaging_template_name_uniq" UNIQUE(name)
```

tedsalmon:
Runbot warnings have been fixed. I have also updated the tests for stock_picking_package_preparation to be compatible with this module.

Thanks!

lasley:
Runbot 🔴 

```
 _auto_init Table `product_packaging`: column `qty` has changed type (DB=float8, def=integer), data moved to column `qty_moved0`
```

tedsalmon:
@lasley,

Fixed - my apologies, I thought I had knocked that one out. All warnings gone and all checks are green.

Thanks!

lasley (commented):
Thanks @t3ddftw 👍

stock_picking_package_info/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+
+{
+    "name": "Stock Picking Package Info",
+    "version": "9.0.1.0.0",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "author": "OdooMRP team, "
+              "AvanzOSC, "
+              "Tecnativa - Pedro M. Baeza, "
```
pedrobaeza: Please put only `Tecnativa`

stock_picking_package_info/models/product_packaging_template.py
```diff
@@ -0,0 +1,100 @@
...
+from openerp.addons import decimal_precision as dp
+
+
+class ProductPackagingTemplate(models.Model):
+    _name = 'product.packaging.template'
+    _description = 'Product Packaging Template'
+    _rec_name = 'packaging_template_name'
+
+    packaging_template_name = fields.Char(required=True)
+    type = fields.Selection([
```
pedrobaeza: Don't use a Python reserved word. Please use something like `tmpl_type`

stock_picking_package_info/models/product_product.py
```diff
@@ -0,0 +1,17 @@
...
+
+from openerp import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    weight_net = fields.Float(
+        string='Net Weight',
+        related='product_tmpl_id.weight_net',
```
pedrobaeza: We should have this weight at product variant level, not at template level, but when setting at template level, it should be propagated to the variants.

pedrobaeza (changes requested):
As the module also provides information for quants and product, it should be renamed to `stock_package_info`

stock_picking_package_info/models/product_packaging_template.py
lasley: IMO abbreviations should be avoided too.

FYI type is not a reserved word, it's a builtin function - https://docs.python.org/3/reference/lexical_analysis.html#keyword

pedrobaeza: `tmpl` is a common abbreviation you find across Odoo, but OK then to put `template_type`. The word is not a keyword, but a function name that is in the base: http://stackoverflow.com/questions/10568087/is-it-safe-to-use-the-python-word-type-in-my-code

lasley: Builtin methods are technically a part of the `__builtins__` namespace though, so aren't reserved in all namespaces like keywords.

That said, I'm good with the more descriptive term. I think `package_type` might fit better in the context though, but we can leave that to @t3ddftw to figure out 😛

stock_picking_package_info/__openerp__.py
anajuaristi: @pedrobaeza
Why only tecnativa if module was developed by odoomrp team and avanzosc.
Maybe you mean to erase your own name from authors since now your company is Tecnativa. Is this? 
Thank you for clarifying

pedrobaeza: Of course, Ana. I'm talking about replacing `Tecnativa - Pedro M. Baeza` by only `Tecnativa`. No need to be picky about this.

anajuaristi: As I thougth on the first term.
It was just to clarify because your request was a bit confusing.
Thank you!!

stock_package_info/models/product_product.py
```diff
@@ -0,0 +1,28 @@
...
+    weight_net = fields.Float(
+        string='Net Weight',
+        related='product_tmpl_id.weight_net',
+        help='Weight of contents excluding package weight',
+    )
+
+    @api.multi
+    def write(self, vals):
+        weight_net = vals.get('weight_net')
+        should_cascade = vals.pop('no_cascade', False)
```
tedsalmon: I'm open to suggestions here as this seems a bit hacky to me. This is my fix for the infinite recursion caused by the fact that the given product is itself a variant of its variants.

woodbrettm: I came across some code implemented for a similar issue when migrating [medical_pathology](https://github.com/OCA/vertical-medical/blob/9.0/medical_pathology/models/medical_pathology_category.py). Not sure if that'll help though in this case.

stock_package_info/models/product_product.py
```diff
@@ -0,0 +1,28 @@
...
+        if weight_net and not should_cascade:
+            self.mapped('product_variant_ids').write({
```
lasley:
```
for record in self:
    variants = record.mapped('product_variant_ids').filtered(
        lambda r: r.id != record.id
    )
    variants.write({...
```

Will solve your infinite recursion

pedrobaeza: You can omit `id` part:

```
    variants = record.mapped('product_variant_ids').filtered(
        lambda r: r != record
    )
```

lasley (commented):
Thanks for the updates @t3ddftw 👍

stock_package_info/models/product_packaging_template.py
```diff
@@ -0,0 +1,100 @@
...
+    ],
+        required=True,
+    )
+    product_tmpl_id = fields.Many2one(
+        comodel_name='product.template',
+        string='Product',
+        index=True,
+        ondelete='cascade',
+    )
+    length = fields.Float(
```
lasley, line 31: Turns out this col name can cause issues. See https://github.com/odoo/odoo/issues/13780#issuecomment-260872554

pedrobaeza: No, that's for JS, not Python.

lasley: Oh got it, thanks @pedrobaeza - I was thinking that the field names would transfer over to JS at some point in a widget or something. Sweet, this was kind of worrisome.

pedrobaeza: Oh, wait, my fault. That's exactly the problem... so you have to change it indeed.

lasley: Although I'm wondering if we should bite the bullet in this particular instance. Renaming this will create a conflict with the `product.packaging` model that it is templating- https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/delivery/models/product_packaging.py#L12

lasley: Although we should change these all to integer fields in order to match the types in the model. Carriers don't accept fractions of measurements on the dimensions anyways.

pedrobaeza: This happens on one2many lists that contains a column with this name. Maybe we can avoid the problem if this is not the case.

lasley: Sweet yeah I think we'll be safe here then as I don't see us presenting this data in that format.

tedsalmon: Hey all,

In https://github.com/OCA/stock-logistics-workflow/pull/247/commits/8151c7b882fe48911e6995e993e383baf7d4ab29 I updated all the measurement field names by prepending `pack_` to them as I thought only updating the `length`  field name would create a weird incongruity in naming. Does anyone oppose this change?

Thanks!

lasley:
@t3ddftw - Are we good for a final review on this?

stock_package_info/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# Copyright 2015 Serv. Tec. Avanzados - Pedro M. Baeza
+# Copyright 2015 AvanzOsc
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import models
+# from . import wizard
```
lmignon, line 8: @lasley You comment the import  but the wizard still exist into the code.

lmignon (changes requested):

lasley:
So this one doesn't really work that well, and will require a lot of work IMO. We're abandoning for a more simple approach that fits our needs of being able to dispatch things with a package.

The main reason for my thoughts here is that too many concepts are being blended into `product.packaging` and `stock.quant.packaging`. It complicates everything, and really doesn't play nicely with pre-existing databases - as indicated by the test failures here about the UOMs.

We will be losing a lot of the volume calculation and stuff here, but that's out of scope for my needs - sorry. Our new solution is in https://github.com/OCA/delivery-carrier/pull/141

PR #250, opened 2016-08-24
[FIX] Partially revert a83b66705e6d49ce33293c7b7164bd440da255b1

codingforfun:
- Signature of function in inherited model doesn't match parent class
  function signature. 

Code broken because of:

```
TypeError: _partial_move_for() got an unexpected keyword argument
'context'
```

### DIFF
--- a/stock_move_backdating/wizard/stock_partial_picking.py
+++ b/stock_move_backdating/wizard/stock_partial_picking.py
@@ -39,10 +39,10 @@ class stock_partial_picking(osv.TransientModel):
     _inherit = 'stock.partial.picking'
     name = 'stock.partial.picking'
 
-    def _partial_move_for(self, cr, uid, move, context=None):
+    def _partial_move_for(self, cr, uid, move):
         partial_move = \
             super(stock_partial_picking, self)._partial_move_for(
-                cr, uid, move, context=context)
+                cr, uid, move)
         partial_move.update({'date_backdating': move.date_backdating},)
         return partial_move
 


gurneyalex:
ping @max3903 @ehdem you were involved in a83b66705e6d49ce33293c7b7164bd440da255b1 maybe you'd like to chime in :smile_cat:

lasley:
I feel like this should be fixed in the superclass. Isn't context a required aspect of the old api method signature? Admittedly I don't know if the superclass is core, just my immediate thought by looking at this problem/diff

codingforfun:
> I feel like this should be fixed in the superclass.

Would be better of course but since it's not the case in the core and they probably won't fix it I don't see the point in enforcing this inside a community model if that means that the community model is permanently functional broken then. 

> Isn't context a required aspect of the old api method signature?

Probably should be, but in fact the core has plenty of functions that do not use the full method signature. Especially functions marked as private with `_` may have arbitrary signatures from what I see in the core code.

lasley (approved):
Thanks for the elaboration @codingforfun - Didn't realize the super was core, apologies for that.

👍

pedrobaeza (approved):

tafaRU:
Strange, the parent class method is https://github.com/odoo/odoo/blob/7.0/addons/stock/wizard/stock_partial_picking.py#L157, isn'it?
Please also see https://github.com/odoo/odoo/commit/c6a5be0a195c71c7ded9cc863815d115259abfd9

lasley (changes requested):
@codingforfun - Can you please provide the full traceback from the error? @tafaRU comments allude to there being another module in play here.

codingforfun:
Ok. You are right. This has been long fixed in the core in https://github.com/odoo/odoo/commit/c6a5be0a. 
It seams I used a not up to date patched version of the core while trying this in a customer project. 
I should have cross checked this better. Sorry for the inconvenience.

PR #251, opened 2016-08-26
[9.0] [MIG] stock_picking_show_return

pedrobaeza:


pedrobaeza:
cc @Tecnativa @elicoidal

stock_picking_show_return/README.rst
```diff
@@ -0,0 +1,50 @@
...
+Display returns on stock pickings
+=================================
+
+This module adds one tab in the pickings form view to display the returns
+pickings related to the current one.
+
+Usage
+=====
+
+#. Go to any picking that has a return, and see the new tab "Returns" appears.
```
elicoidal: Always strange to have a list with only one item. Maybe you can split the lines or simply do not create list

stock_picking_show_return/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
elicoidal, line 1: More a question out of curiosity: our template states a header with license+copyright but a lot of people do not include one for the init.py file. In your case it is incomplete (no copyright).
I never block for this but am curious about what should be enforced...

elicoidal:
some details (non blocking)
:+1:

stock_picking_show_return/README.rst
pedrobaeza: That's not true. You can return several times the same picking.

stock_picking_show_return/__init__.py
pedrobaeza: I usually remove persons here, as there's no real code.

stock_picking_show_return/README.rst
elicoidal: lol sorry but I think you misunderstood my comment.
I was talking that you do not write a list if there is only one item in it (English writing not python code) ;)

pedrobaeza: OK, now I have understood.

pedrobaeza: Changed!

stock_picking_show_return/README.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Usage
+=====
+
+#. Go to any picking that has a return
+#. See the new tab "Returns" appears.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/154/8.0
```
yajo: 9.0

stock_picking_show_return/views/stock_picking_view.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
```
yajo: `<odoo>`?

stock_picking_show_return/views/stock_picking_view.xml
```diff
@@ -0,0 +1,18 @@
...
+
+        <record id="view_picking_form" model="ir.ui.view">
+            <field name="model">stock.picking</field>
+            <field name="inherit_id" ref="stock.view_picking_form"/>
+            <field name="arch" type="xml">
+                <notebook position="inside">
+                    <page string="Returns" attrs="{'invisible': [('returned_ids', '=', [])]}">
```
yajo: Shouldn't it be `False` instead of `[]`? :confused:

stock_picking_show_returns/__openerp__.py
```diff
@@ -1,44 +0,0 @@
-# -*- coding: utf-8 -*-
```
yajo, line 1: Isn't it better to keep old module name? Although it has an `s` at the end, you afford some migration issues

stock_picking_show_return/views/stock_picking_view.xml
pedrobaeza: No, it doesn't work with that, so take note of this

stock_picking_show_returns/__openerp__.py
pedrobaeza: No, this module is migrated without 's' since 8.0. See the branch.

lasley:
👍  outside of what has already been said

Cedric-Pigeon:
:+1: Good for me. Thanks

pedrobaeza:
Merging with enough positive reviews and all comments attended

PR #253, opened 2016-09-06
[FIX] never create pickings for services

dcorio:
while it's fine (at least to me) to add services to the package preparation (e.g.: packaging tasks, delivery options, ...), we must not create pickings for these lines.

### DIFF
--- a/stock_picking_package_preparation_line/models/stock_picking_package_preparation_line.py
+++ b/stock_picking_package_preparation_line/models/stock_picking_package_preparation_line.py
@@ -207,7 +207,8 @@ class StockPickingPackagePreparation(models.Model):
                 # ----- If line has 'move_id' this means we don't need to
                 #       recreate picking and move again
                 if (line.product_id and
-                        line.product_id.type != 'service' and not line.move_id):
+                        line.product_id.type != 'service' and not
+                        line.move_id):
                     move_data = line.get_move_data()
                     move_data.update({
                         'partner_id': package.partner_id.id,
--- a/stock_picking_package_preparation_line/models/stock_picking_package_preparation_line.py
+++ b/stock_picking_package_preparation_line/models/stock_picking_package_preparation_line.py
@@ -206,7 +206,8 @@ class StockPickingPackagePreparation(models.Model):
             for line in package.line_ids:
                 # ----- If line has 'move_id' this means we don't need to
                 #       recreate picking and move again
-                if line.product_id and not line.move_id:
+                if (line.product_id and line.product_id.type != 'service'
+                        and not line.move_id):
                     move_data = line.get_move_data()
                     move_data.update({
                         'partner_id': package.partner_id.id,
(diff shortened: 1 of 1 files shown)


stock_picking_package_preparation_line/models/stock_picking_package_preparation_line.py
```diff
@@ -206,7 +206,8 @@ def action_put_in_pack(self):
             for line in package.line_ids:
                 # ----- If line has 'move_id' this means we don't need to
                 #       recreate picking and move again
-                if line.product_id and not line.move_id:
+                if (line.product_id and line.product_id.type != 'service'
```
pedrobaeza: How can you have a service product in the picking?

dcorio: package != picking

pedrobaeza: From the main module description

> This module adds a new document in the Warehouse menu: Package Preparation. It works on a selection of pickings and allows to add them all in the same package, then to transfer all the pickings at once.
> 
> A possible use case is to put all the selected pickings on a pallet.

So I think it's very related.

dcorio: correct, the fix i'm proposing is for the stock_picking_package_preparation_line that comes with the following description:
"This module adds a new class on Package Preparation to manage details related with stock move or not"

with this module we can add "details" to the package preparation document, but we do not want to create pickings related to these "details" if the added product is a service.

pedrobaeza: OK, I won't block because it's harmless, but an stock.move can also only be associated to a non service product AFAIK

dcorio: i always considered that a bug. you should not be able to have a stock of intangible stuff.
Anyway, the decision is up to the authors of the module :-)

eLBati:
LGTM :+1:

tafaRU:
@dcorio, please fix the flake8 error: https://travis-ci.org/OCA/stock-logistics-workflow/jobs/157886936#L447

lasley (changes requested):
Looks like we still have a flake issue https://travis-ci.org/OCA/stock-logistics-workflow/jobs/158412160#L445

eLBati:
@lasley it seems that now travis is complaining about problems unrelated to this PR

lasley:
@eLBati - looks like there is still a lint coming from this module https://travis-ci.org/OCA/stock-logistics-workflow/jobs/185698857#L335

```
stock_picking_package_preparation_line/models/ir_config.py:25: [R8110(old-api7-method-defined), StockPickingPackagePreparationLineSettings.default_get] Method defined with old api version 7
```

eLBati:
@lasley right, but it does not come from the diff of this PR

dcorio:
i can fix those issues within this pr anyway

lasley:
Oh I see, you're right @eLBati. 

@dcorio would you mind fixing the one lint that is related to this module? I'm good with ignoring the other ones because entirely unrelated.

lasley (approved):
Approved with or without the lint changes; the same tests are failing on 8.0 branch.

Cedric-Pigeon (approved):

pedrobaeza:
What about the Travis status?

lasley:
Travis was fixed upstream. I rebuilt and this now looks good to go, merging.

PR #199, opened 2016-02-02
Porting stock_picking_package_preparation* to 9

eLBati:
Modules
- stock_picking_package_preparation
- stock_picking_package_preparation_line

PR #254, opened 2016-09-07
DONT MERGE PR199 update

gurneyalex:
this is an update to try and see why #199 is red on runbot.


lasley:
@gurneyalex - Are we good to close this, or are you still testing?

gurneyalex:
@lasley thanks for the head up.

PR #256, opened 2016-09-10
[FIX] Make check in quant level so that it works in manufacturing as well

levkar:
This is simpler way to have the same functionality and works well with any kind of move (such as manufacturing moves).


stock_no_negative/models/stock_quant.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class StockQuant(models.Model):
+    _inherit = 'stock.quant'
+
+    @api.model
+    def _quant_create(self, qty, move, lot_id=False, owner_id=False,
+                      src_package_id=False, dest_package_id=False,
+                      force_location_from=False, force_location_to=False):
+        if move.product_id.check_no_negative \
```
pedrobaeza, line 31: You are not checking here if qty is negative.

pedrobaeza:
You have to rebase your branch.

stock_no_negative/models/stock_quant.py
levkar: _quant_create is called when negative otherwise _quant_split is called. 

yes it is not self explanatory but it works like a charm.

levkar:
Yes I admit, I am very bad at using git.

stock_no_negative/models/stock_quant.py
pedrobaeza: OK, I see, please document it then properly, indicating that always that this method is called and the location is internal, a negative quant is created.

pedrobaeza:
I have checked today that `_quant_create` is also called when receiving materials, so you must check quantity variable

pedrobaeza:
Why don't you simply use the approach on https://github.com/OCA/stock-logistics-workflow/pull/137?

levkar:
move.location_id.usage == 'internal' prevents it in material receive, but I agree with you that #137  seems to be a cleaner solution. 

stock_no_negative and stock_disallow_negative both doing the same thing. Do you think it is a good idea to copy the same thing from stock_disallow_negative into stock_no_negative? I know it is not a good idea to have two modules doing the same thing but perhaps it's better to remove stock_no_negative and use stock_disallow_negative.

Let me know what you think so that i can proceed.

pedrobaeza:
The other PR can be disregarded with this one. They serve the same purpose, so better to include that here that is already in the repository.

Cedric-Pigeon:
@levkar Hello, do you still plan to work on this PR ? Is it still useful ?

levkar:
3 PR's are very related, 

1) this one (we are in production with this and very happy with it)
2) PR #258 (see pedro's comment) (i get lost in the conversation)
3) PR #137 (very similar solution and PR #258 is included by design)

If you can decide on how to proceed I would be happy to help. Hopefully we can close all three PR's.

levkar:
I am closing this PR due to lack of interest on this module and existing alternatives #137.

A backport of #275 would be a better choice anyway.

PR #257, opened 2016-09-10
[IMP] No negative by default

levkar:
Stock will be no negative by default. If someone installs this module, then it is better to have this as default behavior.


levkar:
made a mistake. will make another pr

PR #258, opened 2016-09-10
[IMP] No negative by default

levkar:
It is better to have no negative by default if this module is installed

### DIFF
--- a/stock_no_negative/models/product_template.py
+++ b/stock_no_negative/models/product_template.py
@@ -19,14 +19,20 @@
 #
 ##############################################################################
 from openerp import fields, models
+from openerp.tools import config
 
 
 class ProductTemplate(models.Model):
     _inherit = 'product.template'
 
+    # for not to break other tests
+    def _get_default_no_negative_policy(self):
+        return not (config['test_enable'] and
+                    not self.env.context.get('check_no_negative'))
+
     check_no_negative = fields.Boolean(
         'Check no negative',
-        default=True,
+        default=_get_default_no_negative_policy,
         help='If this field is True can not move this'
              ' product in negative quantity available in'
              ' the internal location source')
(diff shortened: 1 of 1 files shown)


pedrobaeza:
WIth this change, you are breaking current tests for other modules, as now you activate the control in all the products. You should inhibit the check when making other tests, but enable it for testing your own module. Look here to see an example: https://github.com/OCA/bank-payment/blob/ea8ec6a7d388d5b5c98dcd1002ad1b766a47dee7/account_payment_transfer_reconcile_batch/models/payment_order.py#L28-L29

levkar:
Thanks for the hint!

pedrobaeza:
I prefer that you include the test environment check on the logic itself, not in the default value of the boolean, so that you can decide on each moment how to behave, not only at the beginning. As you are changing also the logic in other PR, maybe you can include both changes in the same PR.

stock_no_negative/models/product_template.py
```diff
@@ -19,13 +19,20 @@
...
     _inherit = 'product.template'
 
+    # for not to break other tests
+    def _get_default_no_negative_policy(self):
+        return not (config['test_enable'] and
+                    not self.env.context.get('check_no_negative'))
+
     check_no_negative = fields.Boolean(
         'Check no negative',
+        default=_get_default_no_negative_policy,
```
lasley, line 35: Use lambda to allow inheritance `lambda s: s._get_default_no_negative_policy()`

lasley: Although I agree with [@pedrobaeza comment](https://github.com/OCA/stock-logistics-workflow/pull/258#issuecomment-246260658) which would remove this entirely

Cedric-Pigeon:
@levkar  Hello, do you plan to achieve this PR ? Is it still usefull ? Thanks

levkar:
I am closing this pull request since there is an alternative #137 and it works as intended in this PR. Moreover stock_disallow_negative module replaces stock_no_negative module on V10 #275 which makes investment to this module meaningless.

PR #260, opened 2016-09-15
[ADD] Add Spanish (AR) translation on module stock_split_picking.

gabrielo77:
Well, that's it. 

Thank you !


Cedric-Pigeon:
@pedrobaeza I think it can be merged without any risk.

pedrobaeza (approved):
Well, we prefer this to be done on Transifex, but nevertheless...

PR #246, opened 2016-07-15
9.0 porting stock_picking_invoice_link

eLBati:
As the invoincing flow has chenged in odoo version 9,
invoices linked to pickings (and vice versa), are shown independently from
the invoicing method

PR #261, opened 2016-09-20
9.0 stock picking invoice link

hendrixcosta:


hendrixcosta:
#246

mileo:
Replace #246 :+1:

eLBati (approved):
Thanks

pedrobaeza (approved):

jbaudoux:
The port of this module doesn't allow anymore to find out the related moved that are invoiced. If you have 2 deliveries, it seems by reading the code, that the 2 invoices will point to the 2 deliveries and not the related one.
I'll make a PR that fix this.

PR #269, opened 2016-10-18
[8.0] Update sale workflow to be compliant with stock_auto_move features

Cedric-Pigeon:
If you define a route allowing to execute the delivery move of sale order automatically, the sale order will never be in state 'done'.
The reason is that the workflow does not expect this case.
The goal of this module is to add a transition in the sale workflow to manage such case.


pedrobaeza (approved):
At least we don't have workflows in v10...

rousseldenis (approved):
:+1: code review

adrienpeiffer (approved):

PR #270, opened 2016-10-19
10.0 mig porting stock picking package preparation

archetipo:


archetipo:
Replaced with:
https://github.com/OCA/stock-logistics-workflow/pull/271

PR #272, opened 2016-10-20
[9.0][FIX] stock_deposit: Fix domain in views for deposit quants

sergio-teruel:
@pedrobaeza I have been fix this, its a minimal change
cc @Tecnativa

### DIFF
--- a/stock_deposit/models/product.py
+++ b/stock_deposit/models/product.py
@@ -5,6 +5,7 @@
 from openerp import api, fields, models
 from openerp.tools.float_utils import float_round
 from openerp.models import expression
+from openerp.tools.safe_eval import safe_eval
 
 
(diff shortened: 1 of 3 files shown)


stock_deposit/models/product.py
```diff
@@ -22,11 +23,17 @@ def _compute_deposit_available(self):
     )
 
     def _exclude_deposit_location_action(self, res_action):
-        ctx = res_action['context'].copy()
+        ctx = safe_eval(res_action['context'])
```
pedrobaeza, line 26: @Cedric-Pigeon, this has to be eval because domain can come as a text.

pedrobaeza (approved):

pedrobaeza:
I think we can fast-track this patch to avoid errors.

PR #275, opened 2016-11-17
[10.0] Module stock_no_negative

alexis-via:
This is in fact a port to v10 of my PR #137 but changing the name to stock_no_negative, because there is already an OCA module with that name in older versions.

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# © 2015-2017 Akretion (http://www.akretion.com)
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, api, _
+from odoo.exceptions import ValidationError
+from odoo.tools import float_compare
+
+
+class StockQuant(models.Model):
+    _inherit = 'stock.quant'
+
+    @api.multi
+    @api.constrains('product_id', 'qty')
+    def check_negative_qty(self):
+        p = self.env['decimal.precision'].precision_get(
+            'Product Unit of Measure')
+        for quant in self:
+            if (
+                    float_compare(quant.qty, 0, precision_digits=p) == -1 and
+                    quant.product_id.type == 'product' and
+                    not quant.product_id.allow_negative_stock and
+                    not quant.product_id.categ_id.allow_negative_stock):
+                msg_add = ''
+                if quant.lot_id:
+                    msg_add = _(" lot '%s'") % quant.lot_id.name_get()[0][1]
+                raise ValidationError(_(
+                    "You cannot validate this stock operation because the "
+                    "stock level of the product '%s'%s would become negative "
+                    "(%s) on the stock location '%s' and negative stock is "
+                    "not allowed for this product.") % (
+                        quant.product_id.name, msg_add, quant.qty,
+                        quant.location_id.complete_name))
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -5,7 +5,7 @@
 
 from odoo import models, api, _
 from odoo.exceptions import ValidationError
-from odoo.tools import float_compare
+from odoo.tools import config, float_compare
 
 
 class StockQuant(models.Model):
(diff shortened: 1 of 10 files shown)


stock_no_negative/views/product.xml
```diff
@@ -0,0 +1,34 @@
...
+
+<odoo>
+
+    <record id="view_template_property_form" model="ir.ui.view">
+        <field name="name">stock_no_negative.product.template.form</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="stock.view_template_property_form" />
+        <field name="arch" type="xml">
+            <field name="categ_id" position="after">
+                <field name="allow_negative_stock"
```
JordiBForgeFlow, line 16: I think that this field should only be accessible to a specific security group. You basically prevent negative stocks, but allow any user to override this rule just going to the product and changing it. What do you think?

JordiBForgeFlow:
We'll add test cases to this PR, and also backport to 9.0

stock_no_negative/views/product.xml
alexis-via: Only few users should have write access on products... as a lot of parameters on products are sensitive, not only this one !

JordiBForgeFlow: Agreed!

sebalix (approved):
:+1: 
One note, the `stock_no_negative` module from 8.0 has its field named `check_no_negative` which should be converted to the new `allow_negative_stock`, but this should be done in the v9.0 migration (PR #288).

stock_no_negative/models/stock_quant.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+# © 2015-2016 Akretion (http://www.akretion.com)
+# @author Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, api, _
```
rousseldenis: Change import order

rousseldenis (approved):
:+1: Minor comment. Tested on runbot.

stock_no_negative/models/stock_quant.py
alexis-via: Why should I change import order ?

rousseldenis: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

alphabetically ordered. Details but convention ;-)

sebalix:
Is this module ready to be merged?

stock_no_negative/models/stock_quant.py
alexis-via: Arf !!! Incredible ! Don't count on me to apply these crazy coding conventions !

alexis-via:
@sebalix yes, of course. I've been using it in production for several months.

Cedric-Pigeon:
@pedrobaeza this one can be merged

pedrobaeza:
Maybe you can clean a bit the commits?

yvaucher:
Should this module be tested separately or other module's tests should ensure non zero stock level ?

yvaucher:
See comment https://github.com/OCA/stock-logistics-workflow/pull/341#issuecomment-310319638

levkar:
Works like a charm

rousseldenis:
Setup : https://github.com/akretion/stock-logistics-workflow/pull/3

MiquelRForgeFlow:
We have arrived to the conclusion after some analysis, for the good health of the repository, that the better approach to proceed with this topic is the following: In the _config.settings_ of Stock, you have to select an option to activate the stock_no_negative after the installation of the module. In the Readme is put a configuration note that after the installation, you have to select that option if you want to use the module.

We have developed a patch with this approach for v9, and we will open the PR soon. If you like, you may want to cherry-pick it here.

JordiBForgeFlow:
To add to the point of @mreficent , the tests of module 'stock_no_negative' will activate this setting, and then proceed to do the tests. The change in config.settings should create an entry in ir_config_parameter with name 'stock_no_negative' and value 'True'. 

If the parameter does not exist, or is False, then it is assumed that the negative stocks are allowed, regardless if the products have been set with allow_negative = False.

pedrobaeza:
That's not needed: you just need to deactivate the check if in test mode and you are not explicitly testing this module (which is done through a context key). Check https://github.com/OCA/bank-payment/blob/43f8723f8cb4d8defc50d53e83f8b523a7560f00/account_payment_transfer_reconcile_batch/models/bank_payment_line.py#L28 for an example on how to do it

MiquelRForgeFlow:
@pedrobaeza we will do that way now, thanks!

MiquelRForgeFlow:
Done. @alexis-via please, can you cherry-pick commit 73fa87128b160a4f211800b8404110e76e5b3bfd from https://github.com/OCA/stock-logistics-workflow/pull/357?

pedrobaeza:
Better wait first to have it finished (although of course it works)

pedrobaeza:
Now you can cherry-pick 73fa871

levkar:
Actually disallowing negative stock by default makes more sense in othis module. In v8, we had serious problems with customers because of this behavior and finally we wrote another auto-install module which disallows it by default. 

There was a PR of a new module named stock_disallow_negative but later author decided to use this module name and cancelled that PR. Same rationale.

I would strongly suggest to disallow by default.

pedrobaeza:
But it's already the default behavior in this module AFAIK.

levkar:
Sorry, I get it wrong. It's just for tests. No problem.

pedrobaeza:
@alexis-via would you be able to clean the commits and cherry-pick 73fa871 for having this module ready to be merged, or should we superseed it?

pedrobaeza (approved):
I have fixed remaining comments (rebase + squash + cherry-pick), so I merge this.

PR #278, opened 2016-11-21
9.0 mig stock disable force availability button

nicomacr:


stock_disable_force_availability_button/views/stock_view.xml
```diff
@@ -11,27 +11,5 @@
                 </button>
             </field>
         </record>
-
-        <record model="ir.ui.view" id="stock_move_disable_button_form_view">
```
pedrobaeza: Are you sure these buttons are not available in other form?

nicomacr: yes

pedrobaeza: OK, thanks for the confirmation

pedrobaeza (approved):

jjscarafia (approved):

Emiq2 (approved):

PR #279, opened 2016-11-30
[FIX] Conversion of utc datetime to naive date can mean incorrect dat…

gdgellatly:
…e on account moves when backdating

Was just experimenting with this module.  To be honest this is a problem outside of this module but seeing I can fix it here at least I will.  When using this module it accepts a datetime entry.

This could be in the format 1/11/2016 or 1/11/2016 09:00:00 or 1/11/2016 14:00:00.

For only the last one will the resulting accounting entries be dated 1/11/2016 if your timezone is UTC +13.

There is a secondary problem however, this module relies on propagating the new move date via context, however the context passed that sets the date is not the context passed which creates the account move.  I'm a little unsure whether to fix here or not.

In a further module, I have overridden create to take the date of the lines as the date of the account move if it is not set, but can happily pull into here.  The other solution is OCB to propogate the correct context, but then there is a disparity in behaviour between OCB and main.

### DIFF
--- a/stock_move_backdating/stock.py
+++ b/stock_move_backdating/stock.py
@@ -22,6 +22,7 @@
 
 from openerp.osv import fields, orm
 from datetime import datetime
+from dateutil import tz
 from openerp.tools.translate import _
 from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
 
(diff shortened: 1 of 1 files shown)


Cedric-Pigeon:
@gdgellatly Hello, do you plan to achieve this PR ? Is it still useful ? Thanks.

gdgellatly:
@Cedric-Pigeon May as well close it.  It seems that the patch works fine on odoo main branch but causes a different modules tests to fail on OCB.  We've used in production now for 6 months and its fine but I have no time to look at it and moving to 10.0.  I doubt anybody else is still on 7.0 and cares about it.

Without this patch, then stock moves that occur on a different UTC day to the actual day end up having accounting entries on different dates.

PR #283, opened 2016-12-28
[9.0][FIX] stock_batch_picking: Remove duplicate view

carlosdauden:
The view is exactly the same as the original so it does not contribute anything.
Also complicates extending the original view.

@Tecnativa


pedrobaeza (approved):

pedrobaeza:
@cyrilgdn, you were the original developer. Can you check?

yvaucher (approved):

yvaucher:
@carlosdauden Thanks for the fix

PR #286, opened 2017-01-02
[MIG][10.0] stock_picking_show_returns

fmdl:


pedrobaeza:
This doesn't work either. The problem is that you are taking the number in plural, but the module was renamed a long time ago to singular: `stock_picking_show_return`

pedrobaeza:
And you don't need to close PRs. You just need to reuse the same branch name.

fmdl:
I am not change the name of module, it is with "s", please see below : https://github.com/OCA/stock-logistics-workflow/tree/10.0/stock_picking_show_returns

regards

pedrobaeza:
In v8 and v9 the name is without s. That's why you don't import the commit history:

https://github.com/OCA/stock-logistics-workflow/tree/8.0/stock_picking_show_return
https://github.com/OCA/stock-logistics-workflow/tree/9.0/stock_picking_show_return

pedrobaeza:
The v9 version is going to work as is in v10.

PR #291, opened 2017-01-12
[9.0][IMP] stock_deposit: Set partner equal to owner for deposits pickings

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/stock_deposit/models/stock_quant.py
+++ b/stock_deposit/models/stock_quant.py
@@ -22,3 +22,21 @@ class StockQuant(models.Model):
             domain = self._block_deposit_quants(domain)
         return super(StockQuant, self).quants_get(
             qty, move, ops, domain, removal_strategy)
+
+    @api.multi
+    def write(self, vals):
+        if 'reservation_id' in vals:
+            for quant in self:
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Check Travis status

stock_deposit/models/stock_quant.py
```diff
@@ -22,3 +22,21 @@ def quants_get(self, qty, move, ops=False, domain=None,
             domain = self._block_deposit_quants(domain)
         return super(StockQuant, self).quants_get(
             qty, move, ops, domain, removal_strategy)
+
+    @api.multi
+    def write(self, vals):
+        if 'reservation_id' in vals:
```
pedrobaeza, line 28: You should do this in quant_reserve method

sergio-teruel: Yes you are rigth... But this method is not inheritable....

pedrobaeza: OK, I see that there's no other possibility.

pedrobaeza (approved):

carlosdauden (approved):

PR #294, opened 2017-01-27
[9.0][WIP] stock_account_deposit: Invoice picking deposits

sergio-teruel:
This module depends on stock_deposit and allow to invoice quants that it has in a deposit location.

cc @Tecnativa


stock_account_deposit/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
```
cubells, line 2: I think that would be better in all files:

Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>

but is a minor change.

cubells (approved):
LGTM
:+1: 

Code review and tested.

pedrobaeza (approved):

PR #295, opened 2017-01-31
[9.0] Migration of stock_picking_invoice_link module - WIP

jbaudoux:
This module makes the link between the invoice and the related deliveries that are invoiced.From version 9.0, you invoice always from the sale order based on quantites to invoice. With this module, you can know exactly what delivery is in the invoice.

### DIFF
--- a/stock_picking_invoice_link/__openerp__.py
+++ b/stock_picking_invoice_link/__openerp__.py
@@ -2,6 +2,7 @@
 ##############################################################################
 #
 #    Copyright (C) 2013-15 Agile Business Group sagl (<http://www.agilebg.com>)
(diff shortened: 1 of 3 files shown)


pedrobaeza:
The module is already merged: https://github.com/OCA/stock-logistics-workflow/pull/261

jbaudoux:
@pedrobaeza I just saw it when doing this PR. I had an old branch of this projet. Stupid ;-)
I'll write another PR because the port of the implementation is wrong according to me.

pedrobaeza:
OK, no problem. I haven't tried the module myself, so any bug fixing is welcome.

PR #302, opened 2017-02-13
[10.0] Add module product_expiry_simple

alexis-via:
See README for more details about the module.

### DIFF
--- a/product_expiry_simple/models/stock_production_lot.py
+++ b/product_expiry_simple/models/stock_production_lot.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2017 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+
+
+class StockProductionLot(models.Model):
+    _inherit = 'stock.production.lot'
+    _rec_name = 'display_name'
+
+    expiry_date = fields.Date(string='Expiry Date')
+    display_name = fields.Char(
+        compute='compute_display_name_field',
+        string='Lot/Serial Number Display', store=True, readonly=True)
+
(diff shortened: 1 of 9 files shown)


product_expiry_simple/models/stock_production_lot.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo import models, fields, api
+
+
+class StockProductionLot(models.Model):
+    _inherit = 'stock.production.lot'
+    _rec_name = 'display_name'
+
+    expiry_date = fields.Date(string='Expiry Date')
+    display_name = fields.Char(
+        compute='compute_display_name_field',
```
simahawk, line 14: Just a nitpicking request: change compute method name respecting [contrib rule](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field) for such field: `Compute Field: the compute method pattern is _compute_<field_name>`

simahawk (changes requested):
Just code review, no test.
LGTM but the small change for computing method ;)

simahawk (commented):
@alexis-via BTW what's the goal of the module and what you achieve really installing it? I understood it from the code but the README should explain what you get, not just how is different from the official one, isn't it? :)

product_expiry_simple/models/stock_production_lot.py
alexis-via: I named it that way on purpose: if you call the method compute_display_name, you have a warning in the logs of odoo v10 (this is something specific to display_name).

alexis-via:
@simahawk I improved the README as suggested

cdubuit (approved):

product_expiry_simple/models/stock_production_lot.py
simahawk: I see. Good to know :)

simahawk (approved):

PR #303, opened 2017-02-14
[8.0] stock_picking_invoice_link : better picking form view

alexis-via:
On the form view of pickings, invoice_ids is currently in the 1st column of the "Other Information" tab...  but it takes too much space ! I propose to put that field in a dedicated "Invoices" tab.


elicoidal (approved):
:+1:

sebalix (changes requested):
:+1: Can you bump the version number to `8.0.2.1.0` please?

pedrobaeza:
A new tab for only that is not convenient. It was like that before, and it was changed because of this.

:-1: for me

alexis-via:
OK, i'll do that in another module then.

PR #307, opened 2017-03-07
[MIG] stock_batch_picking: Migrated to 10.0 - [WIP]

piqoni:
Work in progress.

### DIFF
--- a/stock_batch_picking/__manifest__.py
+++ b/stock_batch_picking/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Stock batch picking',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     'author': "Camptocamp, Odoo Community Association (OCA)",
     'maintainer': 'Camptocamp',
     'category': 'Inventory',
--- a/stock_batch_picking/__manifest__.py
+++ b/stock_batch_picking/__manifest__.py
@@ -23,7 +23,7 @@
         'wizard/batch_picking_creator_view.xml',
         'security/ir.model.access.csv',
     ],
-    'installable': False,
+    'installable': True,
     'auto_install': False,
     'license': 'AGPL-3',
     'application': False
--- a/stock_batch_picking/models/product_product.py
+++ b/stock_batch_picking/models/product_product.py
@@ -2,7 +2,7 @@
 # © 2012-2014 Alexandre Fayolle, Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp import fields, models
+from odoo import fields, models
 
 
 class Product(models.Model):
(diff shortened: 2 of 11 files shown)


piqoni:
I get the following error when I try to install it.
```
Field `processed_boolean` does not exist.
```
http://pastebin.com/raw/nqJxw08d

leorochael:
>     Field `processed_boolean` does not exist.

The XML file mentioned in that error message tries to add a `processed_boolean` field into a tree view for the field `pack_operation_pack_ids`.

Likely the co-model of the `pack_operation_pack_ids` field doesn't have this field anymore.

Removing the line below from the file `stock_batch_picking/views/stock_batch_picking.xml` probably will let you make progress:

                                    <field name="processed_boolean"/>

piqoni:
Removed it. Now, what about the following, would that be the same case:
```
ParseError: "Error while validating constraint

Element '<field name="description_purchase">' cannot be located in parent view

Error context:
View `product.normal.form`
[view_id: 701, xml_id: n/a, model: product.product, parent_id: 307]
None" while parsing addons/stock_batch_picking/views/product_product.xml:3, near
```

leorochael:
@piqoni, hard to be sure without seeing the whole error, but from the message, it looks like there is an xpath in that xml file referencing a field in the parent view that is no longer present in the parent view.

The previous error was about a missing field in the model, so slightly different. But the temporary solution would be the same: remove the offending line, or check to see if the field was renamed.

piqoni:
Renamed `processed_boolean` to `is_done` since it was renamed: https://github.com/odoo/odoo/blob/9ec95b2a682ca885c0796c8e299ac720e8b7e5d8/addons/stock/models/stock_pack_operation.py#L37

`description_purchase` field has been removed from `product/views/product_views.xml` - so I used `description_sale` which exists in that view.

piqoni:
**At this point the module installs without errors** - but there are some issues/errors when using the addon which I haven't investigated fully yet but I will. If unsuccessful, I'll post them here.  

Meanwhile, I've a question regarding development, because I want to make sure I've correctly set up my local environment for SLW development. Currently, I work the module directly from odoo's `addon` directory, and then I have to copy it to my SLW repository and commit the changes. This feels wrong. Another way I tried was to add the SLW directory as a `--addons-path` to my Odoo server, but the odoo failed miserably (because it tried to install other modules of SLW as well, while I want a way to just add the stock_batch_picking directory at the moment).

leorochael:
The short and hackish answer I can give is: make a symbolic link from the odoo addons directory to your module.

There are probably a lot more cleaner answers than this, but I'll leave it to others.

piqoni:
Thanks, by the way do you have to reload manually the Odoo server to reflect each python code change? 
I started odoo-bin using `--dev=all` (which includes `reload` at the docs). No luck :( (I edit files with Pycharm, Sublime but I don't think it's related with the editors)

leorochael:
I never tried using `--dev=reload` (much less `--dev=all`).

What I recommend, specially if you're using PyCharm, is to create a launch configuration for odoo with `-u stock_batch_picking` and just clicking the `restart` button on the process whenever you save a file. This will  pick up changes in both Python files (by restarting) and data files, including views (because of the `-u` switch).

rousseldenis:
@piqoni What's the status of this ?

rousseldenis:
@piqoni

piqoni:
@rousseldenis  Sorry I have missed your previous message. The company I work for used a customized module developed by a partner, so I stopped moving this to 10.0 version. People more familiar to the module, should find it easier to migrate it since they are also familiar with the requirements of it.

PR #308, opened 2017-03-08
[9.0][IMP] stock_picking_transfer_lot_autoassign: Allow autocomplete units from ordered qty in backorder wizard when the picking has products without lots tracking

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):
Code and functional review

chienandalu (approved):
tested on runbot :+1:

luismontalba (approved):
Functionality tested on runbot

PR #310, opened 2017-03-10
[PORT] Port stock_split_picking to 10.0

aisopuro:
Make some changes to the basic workflow, as the picking workflow was
changed in 9.0. There is no longer a separate wizard for processing
the packing operations, rather they are visible on the picking
form. As such, we emulate the normal backorder creation as
closely as possible, while not actually creating a backorder.


lasley:
Hi @aisopuro - Thanks for the PR! Looks like [Travis is failing](https://travis-ci.org/OCA/stock-logistics-workflow/builds/214183531) though

aisopuro:
Yeah, seems like. I'll try to make time some day, but if anyone wants to fix the problems they're welcome to do that, don't know when I'll have time to work on this again.

Cedric-Pigeon:
To be closed, duplicate with #335

PR #314, opened 2017-03-17
[10.0] [MIG] stock_picking_show_return

pedrobaeza:

PR #315, opened 2017-03-18
[10.0] [MIG] stock_picking_show_return

fmdl:


pedrobaeza:
I have already make the PR in https://github.com/OCA/stock-logistics-workflow/pull/314. Please don't make it again and see the differences with this one (you are still not preserving commit history).

PR #316, opened 2017-03-22
[10.0] Foward-Port new module stock_picking_rate

tedsalmon:
This is a forward-port of #238

From that PR:
This module introduces a concept of rate quotes and purchasing on Stock Pickings. This is one of the modules yielded from the discussion in OCA/delivery-carrier#90.

# Stock Picking Rate

This module adds the concept of delivery rate quotes & purchasing
on Stock Pickings

This is meant to be combined with an external connector with a carrier
supplier in order to bring in rate quotes, although they can be created &
edited manually if your workflow calls for it.

# Usage

Dispatch rates can be added in Stock Picking

## Purchase Rate

The Purchase Rate wizard can be accessed by clicking the green check next
to the rate quote in a stock.picking.

Completing this wizard will create purchase.order for your rate.
If using a connector, this is the point in which a shipment purchase would
be triggered.

cc @lasley


lasley:
@tedsalmon - see https://github.com/OCA/stock-logistics-workflow/pull/238#issuecomment-288545688

tedsalmon:
Ah - I'll close this since we're moving this module. Thanks!

PR #320, opened 2017-04-03
[9.0][IMP] Delete zero operations on batch picking

jjscarafia:
In odoo, when you confirma  picking, al operations with qty zero are deleted.
We fix stock_batch_picking so that it works same way.

NOTE: still default behavior if you have all qtys with zero, then everything is transfered


pedrobaeza:
@carlosdauden please check

manuelfcordoba:
+1

lasley (approved):

Cedric-Pigeon (approved):

lmignon (approved):

PR #328, opened 2017-05-08
Improve batch picking report

guewen:
* Show display name of the product
  * So the variants are displayed correctly
  * It is now consistent with the picking reports too
* Allow to define what aggregation class to use
* Display warehouse description area only if set otherwise a weird empty grey
  rectangle is displayed under every operation.


pedrobaeza (approved):

simahawk (approved):

sebalix (approved):

lmignon (approved):
:+1:

PR #335, opened 2017-05-25
[MIG][10.0] stock_split_picking module

cubells:
- Update README
- Update manifest file
- Add tests
- Update views

cc @Tecnativa


stock_split_picking/__manifest__.py
```diff
@@ -1,37 +1,25 @@
...
- 'auto_install': False,
- }
+# Copyright 2013-2015 Camptocamp SA - Nicolas Bessi
+# Copyright 2013-2015 Camptocamp SA - Guewen Baconnier
+# Copyright 2013-2015 Camptocamp SA - Yannick Vaucher
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Split picking',
+    'summary': 'Split a picking in two unconfirmed pickings',
```
pedrobaeza: s/unconfirmed/not transferred

stock_split_picking/__manifest__.py
```diff
@@ -1,37 +1,25 @@
...
+    'version': '10.0.1.0.0',
+    'category': 'Warehouse Management',
```
pedrobaeza: Has this category the same name in v10?

stock_split_picking/README.rst
```diff
@@ -1,45 +1,77 @@
...
 This module only needs `stock` module.
 
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to **Inventory** dashboard and open any picking.
+#. If picking state is **available** you can see an split button.
+#. Enter quantity to done in every pack operation line of that picking.
```
pedrobaeza: "quantity to done in every pack operation line" is not a user term. Use better the name of the page ("Operations") and the name of the field ("Done"): `On the "Operations" tab, fill the field "Done" to the quantity you want to split for each line.`

stock_split_picking/models/stock.py
```diff
@@ -0,0 +1,58 @@
...
+            ctx = {
+                'active_model': self._name,
+                'active_ids': self.ids,
+                'active_id': len(self.ids) and self.ids[0] or False,
+                'do_only_split': True,
+                'default_picking_id': len(self.ids) and self.ids[0] or False,
+            }
+
+            view = self.env.ref('stock.view_backorder_confirmation')
+            return {
```
pedrobaeza: Instead of showing the wizard, just create it on the background and call the same method as if you click on "Create backorder" button. For user security, add to the "Split" button a confirm attribute.

pedrobaeza (changes requested):

stock_split_picking/tests/test_stock_split_picking.py
```diff
@@ -0,0 +1,72 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import SavepointCase
+from odoo.exceptions import UserError
+
+
+class TestStockSplitPicking(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
```
Cedric-Pigeon, line 12: Please use self in place of cls. I find it a little bit perturbing.

Cedric-Pigeon (changes requested):
Small remark otherwise LGTM

stock_split_picking/tests/test_stock_split_picking.py
lmignon: @Cedric-Pigeon I's a class method :smirk: Therefore, we must use `cls`

Cedric-Pigeon: @lmignon Ok, I missed the decorator

stock_split_picking/__manifest__.py
lmignon: @pedrobaeza  your are right in Odoo 10 the name of the category is Inventory. https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/module/module_data.xml#L39 @cubells Can you take into account the comments from @pedrobaeza?

stock_split_picking/__manifest__.py
```diff
@@ -1,37 +1,25 @@
...
+    'license': 'AGPL-3',
+    'website': 'http://www.camptocamp.com',
+    'depends': [
+        'stock',
+    ],
+    'data': [
+        'views/stock_partial_picking.xml',
+    ],
+    'installable': True,
+    'auto_install': False,
```
lmignon: Tuou can remove the `auto_install`key.

stock_split_picking/models/stock.py
```diff
@@ -0,0 +1,58 @@
...
+                'view_mode': 'form',
+                'res_model': 'stock.backorder.confirmation',
+                'views': [(view.id, 'form')],
+                'view_id': view.id,
+                'target': 'new',
+                'context': ctx,
+            }
+
+
+class StockMove(models.Model):
```
lmignon: @cubells Can you put each model in its own module according to the OCA guidelines?
StockPicking -> stock_picking.py
StockMove -> stock_move.py

stock_split_picking/models/stock.py
```diff
@@ -0,0 +1,58 @@
...
+    @api.multi
+    def split(self, qty, restrict_lot_id=False, restrict_partner_id=False):
+        for pack in self.picking_id.pack_operation_ids:
+            if pack.qty_done > 0:
+                new_move_id = super(StockMove, self).split(
+                    qty,
+                    restrict_lot_id=restrict_lot_id,
+                    restrict_partner_id=restrict_partner_id,
+                )
+                return new_move_id
```
lmignon: @cubells Are your sure. A `return` into the middle of a loop?

lmignon (changes requested):
Thank you for your work. Can you take into account the comments from pedro so we can move forward.

cubells:
All is done. 
@Cedric-Pigeon @lmignon can you retest?

Cedric-Pigeon (approved):

pedrobaeza (approved):

PR #322, opened 2017-04-10
10.0 mig stock_picking_invoice_link

eLBati:
Based on https://github.com/OCA/stock-logistics-workflow/pull/296
This is WIP, waiting for the destiny of https://github.com/OCA/stock-logistics-workflow/pull/296

PR #337, opened 2017-06-07
[IMP] stock_picking_invoice_link: add a button to open invoices

sebalix:
As it is done in the `sale` module: https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L487

### DIFF
--- a/stock_picking_invoice_link/models/stock_picking.py
+++ b/stock_picking_invoice_link/models/stock_picking.py
@@ -17,20 +17,12 @@ class StockPicking(models.Model):
     invoice_id = fields.Many2one(
         comodel_name='account.invoice', string='Invoice',
         compute="_compute_invoice_id")
-    invoice_exists = fields.Boolean(
-        u"Invoiced", compute='_compute_invoice_exists')
 
     @api.multi
     @api.depends('invoice_ids')
     def _compute_invoice_id(self):
         for picking in self:
             picking.invoice_id = picking.invoice_ids[:1]
-
-    @api.multi
-    @api.depends('invoice_ids')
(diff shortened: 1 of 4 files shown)


stock_picking_invoice_link/models/stock_picking.py
```diff
@@ -17,9 +17,35 @@ class StockPicking(models.Model):
     invoice_id = fields.Many2one(
         comodel_name='account.invoice', string='Invoice',
         compute="_compute_invoice_id")
+    invoice_exists = fields.Boolean(
```
pedrobaeza: Although not exactly accurate, I will use picking invoice_state for showing the button.

sebalix: The only drawback I can see is showing a button on the form while there is no related invoice if this module has been installed on an existing database. Do you prefer removing this field anyway?

pedrobaeza: It's for not adding another field to the model. I think we can live with the button leading to an empty invoice view.

sebalix: Done.

pedrobaeza (approved):

sebalix:
Test added, all is green now.

tafaRU (approved):
Thanks!

sebalix:
2 approvals + 5 days, this PR can be merged please?

sebalix:
@tafaRU Thank you!

tafaRU:
@sebalix, could you also please propose these changes to 9 version and to https://github.com/OCA/stock-logistics-workflow/pull/322?

PR #341, opened 2017-06-20
[9.0][FIX] stock_account_deposit: Fix invalidate option "only regularize" is this module is installed

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/stock_account_deposit/wizards/stock_quant_wizard.py
+++ b/stock_account_deposit/wizards/stock_quant_wizard.py
@@ -108,6 +108,8 @@ class StockQuantWizard(models.TransientModel):
 
     @api.multi
     def action_apply(self):
+        if self.quants_action == 'regularize':
+            return super(StockQuantWizard, self).action_apply()
         if self.quants_action == 'invoice':
             picking_ids = self._regularize_quants(
                 self.env.context['active_ids'])


pedrobaeza (approved):

yvaucher:
Please fix the test error:

```
2017-06-20 11:37:41,015 5753 ERROR openerp_test openerp.addons.stock_account_deposit.tests.test_stock_account_deposit: ` ValidationError: (u"You cannot validate this stock operation because the stock level of the product 'Product - Deposit - Test' would become negative on the stock location 'Physical Locations/WH-TEST/Stock' and negative stock is not allowed for this product.", None)
```

To do so you need to add the following in .travis.yml

```
- TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE=stock_no_negative
- TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE=stock_no_negative
- TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE=stock_no_negative
- TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE=stock_no_negative
```

Or ensure that your test comply with the stock_no_negative module constraint if you can.

pedrobaeza:
I think it's better to disable stock_no_negative warning on test mode if not testing the module itself, similar as it's done in https://github.com/OCA/bank-payment/blob/8.0/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

MiquelRForgeFlow:
@sergio-teruel  please, rebase the PR 😉

sergio-teruel:
@mreficent rebase done..

MiquelRForgeFlow (approved):

PR #350, opened 2017-07-28
[9.0][ADD] stock_scrap

LoisRForgeFlow:
Backports the [funcionality](https://github.com/odoo/odoo/blob/10.0/addons/stock/models/stock_scrap.py) of Odoo v10 standard to v9.

Usage
=====

To use this module, you need to:

#. Go to *Inventory > Scrap*.
#. Create a record specifying the product and quantity to scrap.
#. Click on save button.


JordiBForgeFlow:
@mreficent can you review?

LoisRForgeFlow:
@darshan-serpent @nikul-serpentcs could you review this one please?

stock_scrap/models/stock_scrap.py
```diff
@@ -0,0 +1,176 @@
...
+        'stock.picking', 'Picking', states={'done': [('readonly', True)]})
+    location_id = fields.Many2one(
+        'stock.location', 'Location', domain="[('usage', '=', 'internal')]",
+        required=True, states={'done': [('readonly', True)]},
+        default=_get_default_location_id)
+    scrap_location_id = fields.Many2one(
+        'stock.location', 'Scrap Location',
+        default=_get_default_scrap_location_id,
+        domain="[('scrap_location', '=', True)]",
+        states={'done': [('readonly', True)]})
```
serpentcs-dev1: Can you set the scrap location required to True?

Otherwise, it shows the mandatory field validation error while creating a move.

serpentcs-dev1 (approved):
Code & functional review. LGTM :+1: 

minor change needed

MiquelRForgeFlow (approved):

stock_scrap/models/stock_scrap.py
```diff
@@ -0,0 +1,176 @@
...
+            preferred_domain2 = [('reservation_id', '=', False)]
+            preferred_domain3 = [
+                '&',
+                ('reservation_id', 'not in', self.picking_id.move_lines.ids),
+                ('reservation_id', '!=', False)]
+            return [preferred_domain, preferred_domain2, preferred_domain3]
+
+    @api.multi
+    def action_get_stock_picking(self):
+        action = self.env.ref('stock.action_picking_tree_all').read([])[0]
```
rousseldenis, line 160: Maybe adding some security check here on group_stock_user (public method)

stock_scrap/models/stock_scrap.py
```diff
@@ -0,0 +1,176 @@
...
+        action['domain'] = [('id', '=', self.picking_id.id)]
+        return action
+
+    @api.multi
+    def action_get_stock_move(self):
+        action = self.env.ref('stock.act_product_stock_move_open').read([])[0]
```
rousseldenis, line 166: Same as above

rousseldenis (changes requested):
Code review. Minor comments.

Could you please add tests ?

LoisRForgeFlow:
@rousseldenis Hi and thanks for your review! Tests are added as you can see :wink: 

We discussed about the changes you propose and we think that the best approach is to keep it as it is in standard odoo v10 (just the minimun changes to make it work in v9) to avoid migration issues. This means that any modification to this module should be done in another module, and such module could be migrated to v10 and apply to stock_scrap of standard odoo without any problem. Do you agree?

LoisRForgeFlow:
As a result of what I stated above I would need to remove https://github.com/OCA/stock-logistics-workflow/pull/350/commits/5bb4d97f9a37486f82d881dc6c13e5d2481e5aea and https://github.com/OCA/stock-logistics-workflow/pull/350/commits/704f37f8ac967177ed3ba0c228c0ef0ca37cb265

EDIT: done.

rousseldenis:
@lreficent Yes it sounds good to me. Thanks!

LoisRForgeFlow:
@rousseldenis I have applied the changes, could you update your review?

LoisRForgeFlow:
@rousseldenis ping

LoisRForgeFlow:
@rousseldenis your review is blocking, could you please update it?

stock_scrap/models/stock_scrap.py
rousseldenis: @lreficent What about this?

LoisRForgeFlow: I don't see here the comment but I'm pretty sure that @rousseldenis asked for a strict backport of the functionality as it is in v10. That's why this is not readonly, actually I have a custom module to apply that change and some other to this module... 

Anyway odoo has fixed this https://github.com/odoo/odoo/commit/4d86cc78f16271a077b1456a3adcf6b2b3372390 some time after my PR so I think I can add it now also here.

rousseldenis (approved):

LoisRForgeFlow:
Thanks to all reviewers, I'm going to rebase and squash commits.

PR #351, opened 2017-07-31
[MIG] stock_picking_customer_ref: Migration to 10.0

chienandalu:
Stock picking customer reference

This module displays the sale reference/description in the pickings. It also
allows you to filter and search for this new field.

Usage
=====

Edit the client reference in a sale order with pickings. It will show up in
every picking associated.

cc @Tecnativa


stock_picking_customer_ref/README.rst
```diff
@@ -0,0 +1,60 @@
...
+Stock picking customer reference
+================================
+
+This module displays the sale reference/description in the pickings. It also
+allows you to filter and search for this new field.
+
+Usage
+=====
+
+Edit the client reference in a sale order with pickings. It will show up in
```
pedrobaeza, line 15: Make a bit better usage steps.

stock_picking_customer_ref/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Use new header templates

stock_picking_customer_ref/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2015 Tecnativa - Pedro M. Baeza
+# Copyright 2015 AvanzOSC - Ana Juaristi
+# Copyright 2015 AvanzOSC - Mikel Arregi
+# Copyright 2015 AvanzOSC - Alfredo de la Fuente
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Stock Picking Customer Ref',
+    'version': '10.0.1.0.0',
+    'category': 'Warehouse Management',
```
pedrobaeza: I think the category is now simply called `Warehouse`

stock_picking_customer_ref/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+    def write(self, vals):
+        picking_obj = self.env['stock.picking']
+        # The field "client_order_ref" of pickings, is a calculated field. It
+        # can not be put into the field dependence "sale_id".
+        for sale in self:
+            if (vals.get('client_order_ref', False) and
+                    sale.procurement_group_id):
+                cond = [('group_id', '=', sale.procurement_group_id.id)]
+                pickings = picking_obj.search(cond)
+                pickings.write({
```
pedrobaeza: Why writing this if it's a computed field?

stock_picking_customer_ref/models/stock_picking.py
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2015 AvanzOSC - Alfredo de la Fuente
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.one
```
pedrobaeza: Make it `@api.multi`

stock_picking_customer_ref/models/stock_picking.py
```diff
@@ -0,0 +1,23 @@
...
+    @api.depends('group_id')
+    def _calculate_client_order_ref(self):
+        sale_obj = self.env['sale.order']
+        self.client_order_ref = ''
```
pedrobaeza: No need of assigning default value.

pedrobaeza (changes requested):

stock_picking_customer_ref/models/sale_order.py
chienandalu: Yes, the problem here seems to be that the compute field in picking is only going to be triggered when the picking is created and the ref already existed. So the aproach should be refocused to either:

- Get rid of the compute field and catch client ref on the picking creation.
- Get rid of the sale methods and try to compute depending on another field (incomplete approach?)

Your opinion is welcome.

LoisRForgeFlow (approved):
Tested on runbot + code reveiw :+1:

stock_picking_customer_ref/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.multi
+    def write(self, vals):
+        # The field "client_order_ref" of pickings, is a calculated field. It
```
pedrobaeza: Why not? A computed field can be also a dependency.

stock_picking_customer_ref/models/sale_order.py
pedrobaeza: I don't see the problem. You can specify more dependencies (even from computed fields).

chienandalu:
@pedrobaeza Right, it was unnecesary to catch write_order write() :+1:

pedrobaeza (approved):

luismontalba (approved):
Tested on runbot

pedrobaeza:
@chienandalu squash for merging

chienandalu:
@pedrobaeza Done

PR #353, opened 2017-08-01
[ADD] stock_picking_send_by_mail module: migration to 9.0

cubells:
- Update README file
- Update views
- Add tests
- Update translation

cc @Tecnativa


stock_picking_send_by_mail/README.rst
```diff
@@ -0,0 +1,56 @@
...
+you to send the picking order by email.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to a Delivery Order.
+* Click on the button Send by Email.
+* Select the template: Stock Picking - Send by Email.
```
pedrobaeza: The template should be automatically selected, isn't it?

stock_picking_send_by_mail/README.rst
```diff
@@ -0,0 +1,56 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of
```
pedrobaeza: Put project

stock_picking_send_by_mail/i18n/es.po
```diff
@@ -0,0 +1,157 @@
...
+"            </div>\n"
+"        %endif\n"
+"        <p></p>\n"
+"    </div>\n"
+"</div>\n"
+"            "
+
+#. module: stock_picking_send_by_mail
+#: model:email.template,report_name:stock_picking_send_by_mail.email_template_stock_picking
+msgid "${(object.name or '').replace('/','_')}"
```
pedrobaeza, line 135: Update translation, as in v9 this shouldn't appear.

stock_picking_send_by_mail/models/stock_picking.py
```diff
@@ -0,0 +1,42 @@
...
+# Copyright 2015 - Sandra Figueroa Varela
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import _, api, models
+
+
+class StockPicking(models.Model):
+    _name = "stock.picking"
+    _inherit = ['stock.picking', 'mail.thread']
```
pedrobaeza: Why do you inherit from `mail.thread`? It's not needed for sending mails (you can send it as mail.message). Anyway, it can be convenient to have a thread for following the conversation, but in that case, you should rename the module to `stock_picking_mail` for being appropriated for the new features and add the chatter to the form.

pedrobaeza (changes requested):

stock_picking_send_by_mail/i18n/es.po
cubells: I've updated and it appears...

stock_picking_send_by_mail/data/stock_picking_mail.xml
```diff
@@ -0,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+
+        <record id="email_template_stock_picking" model="mail.template">
+            <field name="name">Stock Picking - Send by Email</field>
+            <field name="email_from">${(object.user_id.email or '')|safe}</field>
```
pedrobaeza: This should be sent to the customer (partner_id), not to the responsible.

stock_picking_send_by_mail/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/153/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/stock-logistics-warehouse/issues>`_. In case of
```
pedrobaeza: workflow, not warehouse

pedrobaeza (changes requested):
You are not adding `user_id` to any view, but I'm not sure about introducing that field. I prefer to remove it and let the template with the standard picking fields. Isn't any other field for

pedrobaeza (approved):
@rafaelbn @chienandalu please review

chienandalu (approved):
:+1:

rafaelbn (approved):
tested 👍

PR #356, opened 2017-08-29
[9.0][ADD] delivery_weight_uom_not_required

MiquelRForgeFlow:
Delivery Weight UoM Not Required

This module defines the Unit of Measure of the Weight (field *weight_uom_id*) as not required.

----------------------------------
Explanation of the needs of this module:
- The `stock_batch_picking` module, already merged, has as dependency the Odoo `delivery` module.
- The `delivery` module adds `weight_uom_id` as a new required field in `stock.move` and `stock.picking` models.
- Some modules, like `stock_deposit` or `stock_account_deposit`, already merged, depends from `stock` and `account` but not from `sale` nor `delivery` modules.
- Problem detected: When having installed in local both `stock_deposit` and `stock_batch_picking`, the tests of `stock_deposit` are failing when creating a move or a picking. The cause is that in this tests of `stock_deposit`,  the "self._fields" doesn't have the _weight_uom_id_ and thus the creation query outputs an IntegrityError. This problem is a known problem of dependencies as explained here: https://github.com/odoo/odoo/issues/1053.


AaronHForgeFlow (approved):
:+1: I agree

rousseldenis:
@mreficent Is 'weight_uom_id' field not used in code without testing False value ?

MiquelRForgeFlow:
Right, maybe a test should be added, but I don't know if this module is needed by someone. I don't remember why I created it.

rousseldenis:
@mreficent Do you plan to add test or is this finished ?

MiquelRForgeFlow:
I think it's is finished. However, without a test, it's difficult to say. And no, I won't add a test because I think I don't need this module anymore. @aheficent what do you think?

rousseldenis (approved):
Code review

LoisRForgeFlow (approved):

PR #357, opened 2017-08-30
[9.0][FIX] greenifying stock-workflow repository

MiquelRForgeFlow:
This PR includes:
- Fixes the tests of stock_deposit and stock_account_deposit modules.
- Improves test for stock_picking_invoice_link module.
- Improves the stock_no_negative module with the approach that @pedrobaeza suggested. With this improvement, the negative stock is enabled when testing other modules.

cc: @jbeficent

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -5,7 +5,7 @@
 
 from openerp import api, models, _
 from openerp.exceptions import ValidationError
-from openerp.tools import float_compare
+from openerp.tools import config, float_compare
 
 
 class StockQuant(models.Model):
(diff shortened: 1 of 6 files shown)


stock_no_negative/models/stock_quant.py
```diff
@@ -16,19 +16,24 @@ class StockQuant(models.Model):
...
-                    "on the stock location '%s' and negative stock is not "
-                    "allowed for this product.") % (
-                        quant.product_id.name, msg_add,
-                        quant.location_id.complete_name))
+        check_negative_qty = (
+            config['test_enable'] and self.env.context.get(
+                'test_stock_no_negative')) or (
+            'test_enable' not in config.options)
+        if check_negative_qty:
+            for quant in self:
```
pedrobaeza: Instead of having such a big diff, do better:
```
if not check_negative_qty:
    return
...
```

pedrobaeza (changes requested):

stock_no_negative/models/stock_quant.py
MiquelRForgeFlow: Done

pedrobaeza (approved):
As this is critical for greenifying the branch, I fast-track it.

PR #358, opened 2017-09-05
[9.0][MIG] Migration of module stock_auto_move to v9.0

zakiuu:

PR #359, opened 2017-09-05
[9.0][IMP]stock_auto_move: auto move from push rules

zakiuu:
With this improvement, if the push rule is set to auto confirm, all the generated moves will be automatically processed as soon as the products are available in the source location.

As a use case to explain more the feature, In a reception on two steps, if the push rule is set to auto confirm, all the moves created in the second step will be processed automatically as soon as the first step is done.

### DIFF
--- a/stock_auto_move/__openerp__.py
+++ b/stock_auto_move/__openerp__.py
@@ -12,7 +12,7 @@
         'data/stock_auto_move_data.xml',
         'views/stock_move.xml',
         'views/procurement_rule.xml',
-
+        'views/stock_location_path.xml',
     ],
     'demo': [
         'demo/stock_auto_move_demo.xml',
--- a/stock_auto_move/models/stock_location_path.py
+++ b/stock_auto_move/models/stock_location_path.py
@@ -16,12 +16,6 @@ class StockLocationPath(models.Model):
     )
 
     @api.model
-    def _apply(self, rule, move):
-        """Set auto move to the new move created by push rule."""
-        move.auto_move = rule.auto_confirm
-        return super(StockLocationPath, self)._apply(rule, move)
-
-    @api.model
     def _prepare_push_apply(self, rule, move):
         new_move_vals = super(StockLocationPath, self)._prepare_push_apply(
             rule, move)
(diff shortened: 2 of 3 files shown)


zakiuu:
this PR depends on https://github.com/OCA/stock-logistics-workflow/pull/358. and thanks to @Cedric-Pigeon for sharing the code

stock_auto_move/models/stock_move.py
```diff
@@ -0,0 +1,36 @@
...
+
+    _inherit = 'stock.move'
+
+    auto_move = fields.Boolean(
+        "Automatic move",
+        help="If this option is selected, the move will be automatically "
+        "processed as soon as the products are available.")
+
+    @api.multi
+    def action_assign(self):
```
rousseldenis: Missing `no_prepare=False` in args and ... in super()

stock_auto_move/models/stock_move.py
```diff
@@ -0,0 +1,36 @@
...
+        # We create packing operations to keep packing if any
+        todo_pickings.do_prepare_partial()
+        moves.action_done()
+
+    @api.multi
+    def _change_procurement_group(self):
+        automatic_group = self.env.ref('stock_auto_move.automatic_group')
+        moves = self.filtered(
+            lambda m: m.auto_move and m.group_id != automatic_group)
+        moves.write({'group_id': automatic_group.id})
```
rousseldenis: Functionally, why do we change the procurement group here ?
Moreover, why isn't it configurable ?

rousseldenis (changes requested):

stock_auto_move/models/stock_move.py
zakiuu: @rousseldenis any functional improvement of the module I've to do it in a separate PR, this PR is only a migration to v9

zakiuu: @rousseldenis issue fixed in #364

lmignon (approved):
@zakiuu z Your changes in https://github.com/OCA/stock-logistics-workflow/pull/359/commits/9c02939de9073e535d5ba98941d36f4ff18fe6b6 LGTM. We must wait for the merge of #358

zakiuu:
ping @lmignon @rousseldenis @Cedric-Pigeon

ThomasBinsfeld (approved):

zakiuu:
@rousseldenis can you approve?

rousseldenis (approved):
LGTM. Code review

@zakiuu Maybe a test on push_apply function should be great

ThomasBinsfeld (approved):

LoisRForgeFlow (approved):
Code review

PR #363, opened 2017-09-08
[10.0][FIX]stock_ownership_availability_rules: in case of return, virtual location id does not have any company

Cedric-Pigeon:
On a return, the "Reserve" feature can be useless because virtual locations (like Customers) does not have any company_id and no partner_id, so picking stay in Waiting state because quant owns to the company partner.


rousseldenis (approved):

rousseldenis:
@carlosdauden @lmignon

pedrobaeza:
Please change module version number

Cedric-Pigeon:
@pedrobaeza I missed this one, done!

pedrobaeza (approved):

PR #180, opened 2015-11-20
Added stock_auto_move module

npiganeau:
# Automatic Move Processing

This modules adds the possibility to have move automatically processed as soon as the products are available in the
move source location.

It also adds the possibility to define the move as being automatic in a procurement rule.

Automatic moves are triggered by previous move when the move is chained or by the scheduler otherwise.

Note that automatic moves are given a procurement group name "Automatic", whatever the user or the procurement rule selects.

PR #364, opened 2017-09-10
[9.0][FIX]stock_auto_move: assign automatic procurement to only auto moves without proc.

zakiuu:
**Actual behavior:** With the `stock_auto_move` module, moves that are marked _automatic_ are processed automatically without the interaction of the users. In the mean time, those moves are assigned to an `automatic procurement` blindly. The author of the module has a reason doing so, and it has been explained in #180.
**The issue:** Procurement group have a strong _functional semantic_ (For example: when doing a reception in two steps, a procurement group is generated to tie up the 2 created picking for each one of the steps.).

This PR attends to fix this issue, by adding an _automatic procurement_ to only moves that aren't assigned to a procurement group.

### DIFF
--- a/stock_auto_move/models/stock_location_path.py
+++ b/stock_auto_move/models/stock_location_path.py
@@ -16,12 +16,6 @@ class StockLocationPath(models.Model):
     )
 
     @api.model
-    def _apply(self, rule, move):
-        """Set auto move to the new move created by push rule."""
-        move.auto_move = rule.auto_confirm
-        return super(StockLocationPath, self)._apply(rule, move)
-
-    @api.model
     def _prepare_push_apply(self, rule, move):
         new_move_vals = super(StockLocationPath, self)._prepare_push_apply(
             rule, move)
--- a/stock_auto_move/models/stock_location_path.py
+++ b/stock_auto_move/models/stock_location_path.py
@@ -1,15 +1,29 @@
 # -*- coding: utf-8 -*-
 # © 2014-2015 NDP Systèmes (<http://www.ndp-systemes.fr>)
 
-from openerp import api, models
+from openerp import api, fields, models
 
 
 class StockLocationPath(models.Model):
 
     _inherit = 'stock.location.path'
 
+    auto_confirm = fields.Boolean(
+        help="If this option is selected, the generated moves will be "
+             "automatically processed as soon as the products are available. "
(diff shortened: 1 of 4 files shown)


zakiuu:
This PR is based on #358

lmignon (approved):
@zakiuu changes in https://github.com/OCA/stock-logistics-workflow/pull/364/commits/a274769ce37bd5cfd477fc7dfbd1d65179a50ac1 LGTM... We must wait for the merge of #358

gurneyalex:
@npiganeau can you check this PR?

stock_auto_move/models/stock_location_path.py
```diff
@@ -1,15 +1,23 @@
...
 class StockLocationPath(models.Model):
 
     _inherit = 'stock.location.path'
 
+    auto_confirm = fields.Boolean(
+        help="If this option is selected, the generated moves will be "
+             "automatically processed as soon as the products are available. "
+             "This can be useful for situations with chained moves where we "
+             "do not want an operator action."
+    )
```
npiganeau, line 16: Although I do understand the main subject of the PR and agree with it, I do not understand why you modify the `stock.location.path` code logic. It seems to me unrelated to the subject of this PR. Maybe it could be in another PR ?

stock_auto_move/views/stock_location_path.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="model">stock.location.path</field>
+        <field name="inherit_id" ref="stock.stock_location_path_form"/>
+        <field name="arch" type="xml">
+            <field name="auto" position="after">
+                <field name="auto_confirm"/>
+            </field>
+        </field>
+    </record>
+
+</odoo>
```
npiganeau, line 18: Same remark here

stock_auto_move/models/stock_location_path.py
zakiuu: Hi @npiganeau thanks for the review, but this PR is based on #359  and that's the reason why you see those commits here. I had to do so because the migration of the module wasn't merged yet

npiganeau: Ok, I understand.

ThomasBinsfeld (approved):

Jortolsa-S73 (approved):

PR #361, opened 2017-09-06
[9.0][IMP]stock_pack_operation_auto_fill: fill operations automatically

zakiuu:
Check issue https://github.com/OCA/stock-logistics-workflow/issues/360

PR #366, opened 2017-09-13
[10.0][MIG] stock_picking_transfer_lot_autoassign: Migration to 10.0

rgarnau:
cc @jbeficent @etobella


stock_picking_transfer_lot_autoassign/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    @api.model
```
rousseldenis: Even if it was done like this in v9 module, we cannot do that as there is an inconsistency between self value (which is a single recordset) and the 'model' decorator.

I think we don't have to decorate (like in stock module)

stock_picking_transfer_lot_autoassign/wizards/stock_backorder_confirmation.py
```diff
@@ -0,0 +1,21 @@
...
+class StockBackorderConfirmation(models.TransientModel):
+    _inherit = 'stock.backorder.confirmation'
+
+    @api.multi
+    def process_assign_ordered_qty(self):
+        """Autocomplete product qty with producto ordered qty only for
+        products without lots track
+        """
+        self.ensure_one()
+        for pack_operation in self.pick_id.pack_operation_ids:
```
rousseldenis: `for pack_operation in self.pick_id.pack_operation_ids.filtered(lambda o: not o.pack_lot_ids):`

rousseldenis (changes requested):
Code review. Minor comments

stock_picking_transfer_lot_autoassign/models/stock_picking.py
pedrobaeza: This has been perverted in the migration. In v9 the decorator has all the sense due to standard signature:

https://github.com/odoo/odoo/blob/7eb7e9fd6e72764f28df19ab8803e1cceff38b4d/addons/stock/stock.py#L1204

And the inherited method adapted to new API:

https://github.com/OCA/stock-logistics-workflow/blob/c466ffc62170b6b28f059e6ad6fba85624f8bbaf/stock_picking_transfer_lot_autoassign/models/stock_picking.py#L12

with `picking` as the argument and without using `self`

rousseldenis: Ok, so no decorator here.

pedrobaeza: Yes, as `@api.multi` is now the default.

rousseldenis: Be careful that in the super, Odoo consider self as a single recordset

rgarnau:
Thanks @rousseldenis - I've introduced the changes.

rousseldenis (approved):

stock_picking_transfer_lot_autoassign/models/stock_picking.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
```
rousseldenis: @rgarnau Sorry, missed this : remove api

rgarnau: done :+1:

stock_picking_transfer_lot_autoassign/README.rst
```diff
@@ -1,5 +1,5 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :target: http://www.gnu.org/licenses/agpl
```
JordiBForgeFlow: replace with https

stock_picking_transfer_lot_autoassign/wizards/stock_backorder_confirmation.py
```diff
@@ -2,20 +2,18 @@
...
 
-from openerp import api, models
+from odoo import api, models
 
 
 class StockBackorderConfirmation(models.TransientModel):
     _inherit = 'stock.backorder.confirmation'
 
     @api.multi
     def process_assign_ordered_qty(self):
```
JordiBForgeFlow: @sergio-teruel What is the purpose of this feature? I do not understand it. To me the backorder wizard is not the place to assign automatically all quantities to the existing picking. In any case the new module `stock_pack_operation_auto_fill` does what you seem to require more elegantly. https://github.com/OCA/stock-logistics-workflow/pull/361

I propose to remove this feature from the current module.

JordiBForgeFlow (changes requested):
This module works best when used in combination of stock_pack_operation_auto_fill. Not sure if both modules should be merged together.

stock_picking_transfer_lot_autoassign/wizards/stock_backorder_confirmation.py
JordiBForgeFlow: @sergio-teruel what do you think about this?

rousseldenis:
@rgarnau What is the status of this ?

rgarnau:
@rousseldenis - Migrating this module we saw that it had a feature that from our point of view, is not really related to the lot autoassign purpose and it is in another module already. That's why we are waiting for @sergio-teruel to check if it's really like that.
What's your opinion about it?

rousseldenis:
@rgarnau Yes, stock_pack_operation_auto_fill will populate operations with lot if I remember

rgarnau:
@jbeficent - attended review. The StockBackorderConfirmation has been removed as it is no longer useful.

JordiBForgeFlow (approved):

JordiBForgeFlow:
@rousseldenis @Cedric-Pigeon This one looks good to me. Can you check from your side?

PR #367, opened 2017-09-14
[9.0][FIX]stock_auto_move: fix action_done for automatique moves with lots

zakiuu:
The **current behavior**: the `stock_auto_move` module allows to tag a movement to be processed automatically. In a situation of chained movements (ex: reception in two steps) with product tracking set to `Serial` or `Lot`, the automatic movements can't be processed, and a message error is raised to the user: `You have a difference between the quantity on the operation and the quantities specified for the lots.`
Technically speaking **this issue comes from** the fact there is no quantity set on the `pack_lot_ids` of the operations linked to the movement `linked_move_operation_ids`.

**In this PR** I've tried to **fix this issue** by providing a method that auto fills the quantity on the `pack_lot_ids`, and also fill the `qty_done` of the operations using a new method that I've added called `_auto_fill_pack_lot_ids_qty` defined on the `stock.pack.operation` model. This method is called from the `action_done` of the `stock.move` model.


zakiuu:
this PR is based on #359

stock_auto_move/models/stock_move.py
```diff
@@ -24,6 +24,21 @@ def action_assign(self, no_prepare=False):
...
+    def action_done(self):
+        """
+        Override this method, to automatically fill the quatities
+        of lots used in operations linked to the automatic movements.
+        """
+        auto_moves = self.filtered(
+            lambda m: m.state == 'assigned' and m.auto_move and
+            m.product_id.tracking != 'none')
+        auto_moves_operations = \
+            auto_moves.mapped('linked_move_operation_ids').\
```
lmignon: @zakiuu `auto_moves.mapped('linked_move_operation_ids.operation_id')`

lmignon:
@zakiuu You changes makes sens. Can you check to the Travis error

zakiuu:
closed by #373

PR #370, opened 2017-09-21
[MIG] stock_lot_scrap: Migration to 10.0

chienandalu:
Scrap Complete Lot
==================

This module adds a button in *Production Lot/Serial Number* view form to
*Scrap* all quants contained.

cc @Tecnativa


pedrobaeza (changes requested):
v10 includes now a new object stock.scrap that you have to fill and call the corresponding method (do_scrap) instead of doing this manually. Please change the flow to accommodate it to this new feature.

pedrobaeza:
Does this the scrap itself or only redirects to the view of scraps?

chienandalu:
It takes every quant in the lot and scraps it. The `create` method in `stock.scrap` has a call to the `do_scrap` method.

pedrobaeza:
OK, cool. I'll try when runbot finishes.

rousseldenis:
IMHO, the idea should be to improve stock.scrap form to fill in the quantity to scrap as in the standard scrap form :

- You select the product, and automatically the available lots are selectable
- You cannot enter quantity superior to the available lot quantity for the location you scrap

The process introduced by this module (through production.lot form) should stay in addition.

@chienandalu What do you think about ?

image

pedrobaeza:
Well, the idea is good, but we don't plan to work more on this. We can put it on the roadmap any way. What do you think?

rousseldenis:
@pedrobaeza Ok, we should try not to reinvent the wheel (even if stock.scrap is new and Odoo makes us sometimes schyzophrenic on changes between versions...)

pedrobaeza (approved):

chienandalu:
@rousseldenis But actually I've adapted the migration to the new scrap model and its methods which should be the focus for the migration, I think.

LoisRForgeFlow (approved):
Code review, LGTM. Adapting the module to `stock.scrap` seems enough to me for the migration.

rousseldenis:
@chienandalu I've just tested it and I think the Scrap button should have a 'Are you sure?' confirmation as the whole quantity is directly scraped. What do you think ?

chienandalu:
@rousseldenis I added the confirmation dialog although I had to move the button to the button box and avoid all the dynamic header part... I'm not sure it's fine like this.

pedrobaeza:
@chienandalu for having a confirmation dialog you just need to add an attribute `confirm` and write the text you want to put :scream: Example: https://github.com/odoo/odoo/blob/d6d901c7bc23c7aaafd09c2abccfbe6a7c963b70/addons/account_voucher/views/account_voucher_views.xml#L220

Please undo last commit and do that.

chienandalu:
Oh my... :flushed: I guess I learnt it the hard way. Changed.

PR #373, opened 2017-09-29
[9.0][REF]stock_auto_move: fix action_assign()

zakiuu:
**Use case**: doing a reception operation in two steps where the moves on
the second step are set to be automatic (using the push rule in this
case), and the reception is done partially.
**Expected Result**: When doing a first reception step partially a back
order is created, and we expect the second step to have a back order for
the remaining qty as well.
**Current behavior**: The correct move is processed automatically but no
backorder is created. The reason for that behavior is we processing the
move directly (i.e calling `action_done` of the `stock.move`) without
going through the normal process (i.e processing the second step
picking).
**The solution to the problem**: The processing the of the automatic
stock move should follow the same behavior as done from the user interface.
I've added a new function in `stock.picking` model called
`_transfer_pickings_with_auto_move` to simulate the same behavior this
function is called inside the `action_assign` of the stock move instead
of calling the `action_done` of the move.

### DIFF
--- a/stock_auto_move/models/stock_move.py
+++ b/stock_auto_move/models/stock_move.py
@@ -13,15 +13,36 @@ class StockMove(models.Model):
         help="If this option is selected, the move will be automatically "
         "processed as soon as the products are available.")
 
+    @api.model
+    def _get_auto_moves_by_pickings(self, auto_moves):
+        """ Group moves by picking.
+        @param auto_moves: stock.move data set
+        @return dict dict of moves grouped by pickings
+        {stock.picking(id): stock.move(id1, id2, id3 ...), ...}
+        """
+        auto_moves_by_pickings = dict()
+        for move in auto_moves:
+            if move.picking_id in auto_moves_by_pickings:
+                auto_moves_by_pickings[move.picking_id] |= move
+            else:
+                auto_moves_by_pickings.update({move.picking_id: move})
+        return auto_moves_by_pickings
+
     @api.multi
     def action_assign(self, no_prepare=False):
-        super(StockMove, self).action_assign(no_prepare=no_prepare)
-        # Transfer all pickings which have an auto move assigned
-        moves = self.filtered(lambda m: m.state == 'assigned' and m.auto_move)
-        todo_pickings = moves.mapped('picking_id')
-        # We create packing operations to keep packing if any
-        todo_pickings.do_prepare_partial()
-        moves.action_done()
+        res = super(StockMove, self).action_assign(no_prepare=no_prepare)
+
+        # Select auto moves that are assigned
+        auto_moves = self.filtered(
+            lambda m: m.state == 'assigned' and m.auto_move)
+
+        # group the moves by pickings
+        auto_moves_by_pickings = self._get_auto_moves_by_pickings(auto_moves)
+
+        # process the moves by creating backorders
+        self.env['stock.picking']._transfer_pickings_with_auto_move(
+            auto_moves_by_pickings)
+        return res
 
     @api.multi
     def _change_procurement_group(self):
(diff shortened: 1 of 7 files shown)


zakiuu:
ping @lmignon @sbidoul @npiganeau @rousseldenis

stock_auto_move/models/stock_move.py
```diff
@@ -13,21 +13,54 @@ class StockMove(models.Model):
...
-        todo_pickings = moves.mapped('picking_id')
-        # We create packing operations to keep packing if any
-        todo_pickings.do_prepare_partial()
-        moves.action_done()
+        res = super(StockMove, self).action_assign(no_prepare=no_prepare)
+
+        # Use a variable in the context to whether check
+        # the auto moves or not otherwise we will have an
+        # infinite loop
+        if self.env.context.get('no_check_auto_moves'):
```
rousseldenis: @zakiuu Can you explain the case as there is maybe another solution than using a context parameter?

stock_auto_move/models/stock_picking.py
```diff
@@ -0,0 +1,77 @@
...
+        for picking in auto_moves_by_pickings:
+            if len(picking.move_lines) == len(auto_moves_by_pickings[picking]):
+                # Create immediate transfer wizard so it will fill the qty_done
+                # on the auto move linked operation
+                wizard = self.env['stock.immediate.transfer'].create(
+                    {'pick_id': picking.id})
+                wizard.process()
+            else:
+                # create back order with the auto moves
+                backorder = picking.copy(
```
rousseldenis: @zakiuu The workflow we discussed was quite right, but actually the good way is to create backorder with the rest of moves (like in standard).

For this, simply call the _create_backorder(backorder_moves) method on picking with backorder_moves = picking_moves - auto_run_moves (it will reassign the original picking and the operations should have been recreated in a good way)

rousseldenis (changes requested):
Code review. Seems to be a litlle bit changed.

stock_auto_move/models/stock_move.py
zakiuu: @rousseldenis no need for the context anymore by fixing the second comment

zakiuu:
ping @rousseldenis

stock_auto_move/models/stock_move.py
```diff
@@ -30,10 +30,23 @@ def _get_auto_moves_by_pickings(self, auto_moves):
...
     @api.multi
     def action_assign(self, no_prepare=False):
-        res = super(StockMove, self).action_assign(no_prepare=no_prepare)
+
+        already_assigned_moves = self.filtered(
+            lambda m: m.state == 'assigned')
+
+        # If all the moves have been already been assigned return the
+        # super method
+        if len(already_assigned_moves) == len(self):
```
rousseldenis: This is not necessary as the result recordset here below (not_assigned_auto_move) will be void.
Maybe you don't want to enter the below functions but that's to be put in balance with the amount of code

(I don't like several calls to the same function in a 'same function')

zakiuu: @rousseldenis thanks for the review, and it's fixed :)

lmignon (approved):
Code + Functional

LoisRForgeFlow (approved):
Tested :+1: Thanks for the contribution!

ThomasBinsfeld (approved):

LoisRForgeFlow:
@rousseldenis could you update your review?

ThomasBinsfeld:
Ping @rousseldenis

ThomasBinsfeld (approved):

rousseldenis (approved):

ThomasBinsfeld:
This one is ready to be merged @pedrobaeza

PR #388, opened 2017-12-06
[RFC|ADD] stock_cannabis: Logistics for cannabis stock & sales

lasley:
This module will add logic and workflows required for the stock and sale of cannabis products. For the most part this is an RFC, but I already somewhat created a data structure while scoping, so this is a PR instead. I'm not really sure if this is the proper repo, but here we are anyways!

This module is scoped to provide the ability for:
* Attributes on product specific to cannabis (strain for the most part)
* Harvest, production, and use by dates on stock lots (TBD whether any of this exists elsewhere - I feel like it should though)
* Cultivator information on stock lots
* Storage of test results for stock lots, as well as the certification number for verification 
* Add a report to stock pickings to print the test result label for items in the picking 

A nice to have in the scope:
* Storage of chain of custody information for lots

This module is not scoped for:
* Production workflows (converting a plant to a product)
* Medical workflows

My hope is to maybe extract the specialized needs into something a bit more generic. For example (I put the cannabis industry use case as first point of every section):

* Products and their genealogy:
    * Plants and their lineage
    * Horses (and other show animals) and their pedigree
    * Agriculture and GMO or organic assertions/pedigrees
* Product Lots that contain test results with possible expiration dates:
    * Cannabinoid profiles
    * Food industry testing (USDA, etc.)
    * Electronics (Underwriter Labs, internal QA, etc.)
* Production object lifecycle management (an object is being produced, and you must manage events in its life)
    * Plant measurements over time, soil tests, etc.
    * Livestock weights over time, vaccination records, medicine records, blood tests, etc.
* Object movement over time, with chain of custody management
    * Plants being moved to different rooms depending on grow cycle
    * Livestock being rotated amongst pastures
    * Evidence of a crime

Overall I think those four points nail it. I'm still thinking though.

Questions:
- [ ] Is there already some module with logic for expiration dates?
- [ ] Is there already some module with logic for tracking product lots at certain stages (eg harvest, production)?

### DIFF
--- a/stock_cannabis/README.rst
+++ b/stock_cannabis/README.rst
@@ -1,0 +1,68 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl.html
+   :alt: License: LGPL-3
+
+==============
+Stock Cannabis
+==============
+
+Adds support for stock and sales of Cannabis products.
+
+Configuration
+=============
+
+
+Usage
+=====
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/154/10.0
+
+Known Issues / Roadmap
+======================
+
+Known Issues
+------------
+
+
+Roadmap
+-------
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues 
+<https://github.com/OCA/stock-logistics-workflow/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first, 
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Dave Lasley <dave@laslabs.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/stock_cannabis/README.rst
+++ b/stock_cannabis/README.rst
@@ -30,6 +30,7 @@ Known Issues
 Roadmap
 -------
 
+* More in depth test results
 
 Bug Tracker
 ===========
--- a/stock_cannabis/__init__.py
+++ b/stock_cannabis/__init__.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from . import controllers
+from . import exceptions
+from . import models
+from . import wizards
--- a/stock_cannabis/__init__.py
+++ b/stock_cannabis/__init__.py
@@ -2,7 +2,4 @@
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
-from . import controllers
-from . import exceptions
 from . import models
-from . import wizards
--- a/stock_cannabis/__manifest__.py
+++ b/stock_cannabis/__manifest__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Product Cannabis',
+    'summary': 'Adds support for stock and sales of Cannabis products.',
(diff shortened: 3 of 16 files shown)


hbrunn:
do you have a customer for this? We never happened to get into the Amsterdam coffeeshop business, but then weird local laws also make large scale cannabis storage a difficult thing to do legally, and small shops don't really need an ERP

lasley:
We have a few actually. Mainly dispensaries, but we're looking into the requirements for a production facility as well.

Recreational is on the upswing here, but it's typically combined with medical so the companies have some pretty stringent reporting requirements on that end. A small dispensary in my neighborhood nets over a million (USD) a year on medical, and has advised that their business has essentially quadrupled since recreational hit in. The production facility is talking multi-million/yr profit, with the real money in cultivation because there's been massive consolidation.

The legality is questionable, but not on our end. It's actually kind of a shit situation for the customers though - they can't use banks or insurances or anything like that because it is federally illegal, and any company moving money across multiple states (basically all banks and insurance companies) are subject to federal jurisdiction. 

Storage is on them for both product and money during all aspects of the pipeline, with taxation happening retroactively (both for sales tax and wholesale tax) based on whether the final product sale ended up being for recreational or medicinal. 

Combine the above with the fact that the product is high value, and the money is all cash (unless they decide to store in crypto) - even the smallest players here are using ERPs and POS systems.

_Edit; words_

lasley:
An interesting note: the cultivation facilities out here basically all have giant vaults in them, and double as money storage facilities for them and their clients.

hbrunn:
I started an uninteresting review, but I'm going to discard it because it's not interesting. But thanks for the description of how it goes over there, that's pretty similar to the half baked legislative situation they have in the Netherlands for decades already. How it can be a surprise that the fringes overlap with crime is a mystery to me. But well, we'll get there eventually. Interesting point that IT helps here. My guess would be that given the toleration strategy as it's called here is so old, they've developed ways to cope with the situation that just works, so the demand for change is low, whereas you guys just started and had to figure out a solution with today's means.

lasley:
Interesting to know that we're not alone here with our half baked legislations. I'm figuring something will be done at the federal level here once a high percentage of our states hop on board. I think we just recently broke the 50% mark for states allowing it medically, with only 4-5 allowing it recreationally. But until then it's an interesting grey area, to say the least.

> they've developed ways to cope with the situation that just works, so the demand for change is low, whereas you guys just started and had to figure out a solution with today's means.

Yup basically. In this day and age it's almost cheaper to spin up an ERP & POS than it is to make processes around paper. That, and basically every dispensary wants "giant TVs with pictures of the products and menus on every wall". Oh, retail 🚀

hbrunn:
curiously, when doing my newspaper round, [this](https://translate.google.nl/translate?sl=auto&tl=en&js=y&prev=_t&hl=en&ie=UTF-8&u=https%3A%2F%2Fnos.nl%2Fnieuwsuur%2Fartikel%2F2206295-onderzoek-naar-nederlands-drugsbeleid-jarenlang-gemanipuleerd.html&edit-text=&act=url) popped up. Dutch state television, it's very improbable that this is a fake. And google's translation is impressively good.

lasley:
Hahaha wow this is exactly the stuff that happens here too! Crime is good for politics - helps suppress the vote 😑

lasley:
Noting this so I don't lose it - seems there's [an open database for this stuff](https://developers.cannabisreports.com/v1.0/docs). Hooray standardization!

rousseldenis:
@lasley IMHO, as it is a very specialized application, couldn't it be in a 'vertical' such repo or...

Moreover, you should so have potentially thousands of such modules here for product specialization (coffee, computers, vegetables, and, ....) ...

@lreficent What do you think about ?

LoisRForgeFlow:
I see your point @rousseldenis but since this seems (ATM) a small vertical I am not sure if it would need a new repo (I recall I don't know the @lasley plans here). 

Yet as you say, there might be room for many "mini" vertical specializations, so what comes to my mind is: why not a repo to group these small verticals? stock-logistics-verticals or so?

lasley:
@rousseldenis I think I agree with @lreficent here. Verticals aren't necessarily the easiest thing to maintain, nor is it easy to get contributors on things so niche and specialized. We've already implemented a bunch of custom stuff to fit the customer needs, and now I'm just toiling over the best way to mesh it all for public adoption.

My hope is to maybe extract the specialized needs into something a bit more generic. For example (I put the cannabis industry use case as first point of every section):

* Products and their genealogy:
    * Plants and their lineage
    * Horses (and other show animals) and their pedigree
    * Agriculture and GMO or organic assertions/pedigrees
* Product Lots that contain test results with possible expiration dates:
    * Cannabinoid profiles
    * Food industry testing (USDA, etc.)
    * Electronics (Underwriter Labs, internal QA, etc.)
* Production object lifecycle management (an object is being produced, and you must manage events in its life)
    * Plant measurements over time, soil tests, etc.
    * Livestock weights over time, vaccination records, medicine records, blood tests, etc.
* Object movement over time, with chain of custody management
    * Plants being moved to different rooms depending on grow cycle
    * Livestock being rotated amongst pastures
    * Evidence of a crime

Overall I think those four points nail it. I'm still thinking though.

_Edit: The above were good points, so I added to main description_

yvaucher:
> Is there already some module with logic for expiration dates?


@lasley for product expiration dates you have a base module in Odoo 11.0:

- https://github.com/odoo/odoo/tree/11.0/addons/product_expiry
Otherwise you have an OCA module that replaces the core module:
- https://github.com/OCA/stock-logistics-workflow/tree/10.0/product_expiry_simple

pedrobaeza:
@lasley is not answering anymore about these things, so we can close this. If he ever wants to retake it, here it is.

PR #402, opened 2018-01-25
[MIG] stock cancel to 10.0

ovnicraft:
migration to v10.0
TODO:
* identify in which cases should only work 
* check previous stock moves and procurements linked.


pedrobaeza:
This module is not valid because it doesn't undo the quant operations. We removed it from previous versions if I don't remember bad.

ovnicraft:
@pedrobaeza please read the code.

ovnicraft:
i read code in stock.move, it handle quant_ids and remove them 'undo quant operations'
Ref: https://github.com/odoo/odoo/blob/10.0/addons/stock/models/stock_move.py#L120
Consider context `force_unlink`
Ref: https://github.com/odoo/odoo/blob/10.0/addons/stock/models/stock_quant.py#L101

pedrobaeza:
haha, but that's incorrect. You can unlink quants except on incoming pickings. The rest proceeds from a possible split with its move history, so it's barely impossible to recreate the previous state before you press validate button on picking.

florian-dacosta:
I think, the possibility to re-open a stock picking would ve a very nice feature to have.
But I also think it is way more complicated the the way it is implemented in this module.

You can't just delete the quants, the stock level would be wrong after the stock picking revert.
Also, if you put a delivery order back to draft, you have to handle the fact that it may be chained to previous stock moves and linked to procurements. So re-validate it could generate all the previous stock moves again...
There are a lot of things that have to be taken care of...

Anyway, if anyne is motivated to do such modules, 2 modules should be made.
1 to make a cancel picking back to draft
And 1 to revert the done pickings

ovnicraft:
OK, so its possible this module only works with incoming pickings ? so i can write a restriction about it.

ovnicraft:
@florian-dacosta thanks for your comment!, yes that what i am trying; with help identify all restrictions.
So taking in care your comment this module try to put picking in draft.

florian-dacosta:
Well it makes the module really limited, but why not, if you put this clearly in the README file.

One case that still is a problem, even with these restrictions : 
If push rules are configured to create a chained picking after the confirmation of the incoming shipment.
Example : 
Validate a purchase order => it creates the incoming shipments => On validation, with push rules, it create a internal picking.
Transfer the incoming shipments, the internal picking is ready.
revert the incoming shipment, when it is confirmed again, it will create another internal picking.

So, after reverting, you will have 1 incoming shipment, and 2 internal pickings.
It is definitily not good.

ovnicraft:
Will analize your case and identify if its possible put restriction in that case, also will wrote a README about it.

AaronHForgeFlow:
I am trying to do some improvements here: https://github.com/Eficent/stock-logistics-workflow/tree/10.0-mig-stock_cancel

* allow to reopen any picking
* reassign location in the quant using the history
* reopen the procurement when the procurement is reopened, cancel it if the order is cancelled
* cancel the account valuation moves
* allow to reopen picking with invoices if those invoices are cancelled

@ovnicraft if you agree I can create a PR to your branch. In that case please rebase your branch to OCA/10.0 first.

cc @jbeficent

ovnicraft:
@aheficent great read your collaboration, so i consider better if you open a new PR here and close this an favor of you  ?

AaronHForgeFlow:
@ovnicraft Sure, I'll create the PR then, including you commits here. Thanks.

ovnicraft:
@aheficent great ! so i close this

PR #411, opened 2018-03-10
[8.0][FIX] not show an error if picking is already assigned, it is a …

sergiocorato:
…subsequent state

### DIFF
--- a/stock_picking_package_preparation/model/stock_picking_package_preparation.py
+++ b/stock_picking_package_preparation/model/stock_picking_package_preparation.py
@@ -161,7 +161,8 @@ class StockPickingPackagePreparation(models.Model):
             raise exceptions.Warning(
                 _('The package has not been generated.')
             )
-        self.picking_ids.do_transfer()
+        self.picking_ids.filtered(lambda x: x.state not in ['done', 'cancel']
+                                  ).do_transfer()
         self.write({'state': 'done', 'date_done': fields.Datetime.now()})
 
     @api.multi
--- a/stock_picking_package_preparation/__openerp__.py
+++ b/stock_picking_package_preparation/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 
 {'name': 'Stock Picking Package Preparation',
- 'version': '8.0.1.0.0',
+ 'version': '8.0.1.0.1',
  'author': 'Camptocamp,Odoo Community Association (OCA)',
  'maintainer': 'Camptocamp',
  'license': 'AGPL-3',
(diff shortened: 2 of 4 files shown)


gurneyalex:
hello @sergiocorato 

can you explain more clearly how to reproduce the issue you are fixing?

sergiocorato:
hello @gurneyalex 
1. Create a sppp
2. Link a picking to sppp
3. Force availability to the picking
4. If you try to put in pack the sppp, it raise the error because not confirmed (but it is already in state assigned)

n.b.: I pushed another fix too, to understand the whole process. This second fix is for a case similar, except:
3a. Transfer the picking
4. If you put in pack the sppp, the picking is duplicated.

rousseldenis (approved):
Code review

bodedra (approved):

rousseldenis:
@gurneyalex

stock_picking_package_preparation/model/stock_picking_package_preparation.py
```diff
@@ -161,7 +161,7 @@ def action_done(self):
             raise exceptions.Warning(
                 _('The package has not been generated.')
             )
-        self.picking_ids.do_transfer()
+        self.picking_ids.filtered(lambda x: x.state != 'done').do_transfer()
```
gurneyalex: Don't you need to also filter out cancelled pickings?

gurneyalex (approved):

stock_picking_package_preparation/model/stock_picking_package_preparation.py
sergiocorato: Theoretically yes, but I didn't found a method to put in pack a cancelled picking
I add the filter anyway

sergiocorato: there are other states too, but duplication is possible only if in `done` state, so it would be unuseful

sergiocorato:
@gurneyalex ready to merge?

stock_picking_package_preparation/model/stock_picking_package_preparation.py
rousseldenis: @sergiocorato I think the @gurneyalex remark makes sense. I remember a fix I did to avoid multiple picking validation (If different users click on button).

And Odoo did quite the same in https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/addons/stock/stock.py#L1457

sergiocorato: done!

rousseldenis:
@sergiocorato Nice, could you just update module version ? Thanks

sergiocorato:
> @sergiocorato Nice, could you just update module version ? Thanks

done!

PR #378, opened 2017-10-09
[10.0] [MIG] product_customer_code_picking

nikul-serpentcs:
Product Customer Code For Stock Picking
=======================================
Based on product_supplierinfo_for_customer, this module loads in every Stock Picking the
customer code defined in the product.

PR #413, opened 2018-03-20
[MIG][10.0] product_customer_code_picking

AaronHForgeFlow:
Supersedes #378 

Depends on https://github.com/OCA/product-attribute/pull/292


AaronHForgeFlow:
As product_customer_code is deprecated. Maybe this module should be renamed to product_supplier_info_for_customer_picking (or something less ugly). @pedrobaeza

product_customer_code_picking/models/stock_move.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @api.multi
+    @api.depends('picking_id.partner_id', 'product_id',
+                 'product_id.customer_ids')
+    def _compute_product_customer_code(self):
+        for move in self.filtered(lambda m: m.picking_id and
+                                  m.picking_id.partner_id and
+                                  m.product_tmpl_id.customer_ids):
+            move.product_customer_code = \
+                move.product_tmpl_id.customers_ids.filtered(
```
rousseldenis: I think the filtered should be on the move recordset (This generates error). But eventually it is no more relevant as the condition is done at the loop level.

Moreover, don't be afraid to depends on 'product_id.customer_ids.product_code'

rousseldenis (changes requested):

pedrobaeza:
Well, for being consistent with the dependencies and other modules, it should be named as `product_supplierinfo_for_customer_picking`, which is not too bad taking into account that is only a technical name.

rousseldenis:
@aheficent Doesn't this need migration script (even if the last installable module is in v7 - for the naming change) ?

AaronHForgeFlow:
@rousseldenis I think it's not needed here, as the field is computed.

We need to pass the product_customer_code to the supplierinfo table. I'm doing that here: https://github.com/OCA/product-attribute/pull/313/

rousseldenis:
@aheficent Ok, that's true for field values, but what about former module data (views, ir_model_data, ...) ?

I'm not really used to that process, but i think this should be done through openupgradelib library.

@pedrobaeza What do you think about ?

pedrobaeza:
That should be done in OpenUpgrade on the branch 10.0, not in the module itself, but the possibility of migrating a v7 DB to v10 with that module installed is very remote.

MiquelRForgeFlow (approved):

JordiBForgeFlow:
@mreficent Can you please add a simple test?

AaronHForgeFlow:
@rousseldenis  is it ok to you to approve this and then add the migration scripts later if needed?

MiquelRForgeFlow:
@jbeficent simple test added.

Travis is  green :green_apple: 

Ready to merge!

MiquelRForgeFlow:
@rousseldenis this is full :green_apple:

JordiBForgeFlow (approved):

PR #414, opened 2018-03-23
[MIG] stock_picking_invoice_link: Migration to 11.0

celm1990:
This PR is continuation of PR #409 
Thank to @Softdil
@pedrobaeza please can you review and merge???

PR #409, opened 2018-03-06
[MIG] stock_picking_invoice_link: Migration to 11.0

aitorbouzas:
Migration to 11.0.
This command failed:
`git format-patch --keep-subject --stdout origin/11.0..origin/10.0 -- $MODULE | git am -3 --keep`
I guess that's why all the files seem to be added instead of modified...

PR #415, opened 2018-03-24
[MIG] stock_picking_invoice_link: Migration to 11.0

celm1990:
This PR is continuation of PR #409 and replace PR #414 
Keep commit history.
Please review and merge if possible


stock_picking_invoice_link/models/stock_move.py
```diff
@@ -10,16 +10,6 @@
 class StockMove(models.Model):
     _inherit = "stock.move"
 
-    # Provide this field for backwards compatibility
-    invoice_line_ids = fields.Many2many(
```
chienandalu: Shouldn't it be a migration script to avoid old data loss?

chienandalu (changes requested):

stock_picking_invoice_link/models/stock_move.py
celm1990: Same data is on field invoice_line_id, is redundant field

pedrobaeza:
Superseeded by #415, preserving all the latest commits from this one and the previous, but refining some of the latest things. Thanks for fixing the tests!

PR #418, opened 2018-04-19
[10.0][MIG] stock_cancel

AaronHForgeFlow:
Supersedes #402 
Tests still pending.
### Known issues:
- The quants are now reassigned to a previous location. It won't return the status to the previous state but the quant at least would be in the same location it was before it was moved:
```
Example: Product A:
                1 quant in stock for 10 units
after delivering 5 units to customer:
                Product A:
                1 quant in stock for 5 units
                1 quant in customers for 5 units
after cancelling: Product A
                2 quants in stock, each one for 5 units.
```
- It is not possible to reopen pickings with multiple pack operation for the same stock move. Impossible to find the original location in that case


stock_cancel/models/stock_quant.py
```diff
@@ -0,0 +1,37 @@
...
+from odoo import api, models
+
+
+class StockQuant(models.Model):
+    _inherit = 'stock.quant'
+
+    @api.multi
+    def _revert(self):
+        for quant in self:
+            previous_move = quant.history_ids[-1]
```
rousseldenis, line 15: @aheficent Are you 100% sure the move -1 in history would be the good one?

AaronHForgeFlow: @rousseldenis No, I'm not. Functional test from my side worked for simple pickings (deliveries, shipments or internal moves when only 2 locations are involved and when the quant have not been used in other place). Pending to check chained stock moves thought. I'll add some tests trying to cover most common cases.

AaronHForgeFlow:
Update: Single moves that are not part of a chain can be reverted with this module. This means to allow users to correct their mistakes when validating manual pickings. Constrains try to don't allow the users to reopen chained pickings. Taken the last location in quant's history will set the proper location for single pickings. This is not tested for chained pickings. It's not meant to be used for orderpoints.

@rousseldenis  I updated the readme so users do not misunderstand the purpose of this module.

misern2 (commented):
@aheficent when a picking is reopened, if the picking is linked to a sales order the delivered quantity on the SO should be updated.

AaronHForgeFlow:
@misern2 you are right, thank you for your review. It's meant to be attended here https://github.com/OCA/stock-logistics-workflow/pull/418/commits/66a5cc98adb7922ff3f67069e4385d27d9f672b7

florian-dacosta:
@aheficent The README says : 'only allow to picking that are not part of a chain', but I don't see any restriction about it, is it normal?

AaronHForgeFlow:
@florian-dacosta yes, that actually means that will not propagate the reopening for to the chained moves. The restriction is still pending, I will add it.

Bilbonet (approved):

misern2 (approved):
👍

HaraldPanten:
Hi all,

I just tried the runbot and I found an issue.

1- I created a SO. I selected a random partner and a random stockable.

2- Normal delivery charges were selected in that SO.

3- I confirmed the SO and I confirmed the delivery order as well.

4- I reopened the picking and I confirmed it again. I did that several times.

5- When I went back to the SO, I found that the delivery charges were duplicated several times.

I think that it shouldn't work like that. When we reopen a picking we are cancelling the delivery to check if there's some mistake. The idea is to confirm it after this.

We are not sending and receiving the items each time that we reopen/validate the stock picking. That's why the delivery charges shouldn't be duplicated in the SO automatically.

Regards,

AaronHForgeFlow:
@haraldqubiq yes you're right. Odoo adds an extra line to the SO when confirming the delivery. I think the best approach is to delete it.

AaronHForgeFlow:
@haraldqubiq The commented feature is related the delivery module and stock_cancel does not depend on it. I'd better create another PR with another module that depends on this one and delivery. This new module should be auto-installed when both are installed in the system.

HaraldPanten:
@aheficent OK! Let me know once you finish or if you want me to give you a hand

AaronHForgeFlow:
@haraldqubiq commit 7e0259e it's supposed to do that. Regards

HaraldPanten:
@aheficent  Now it's OK for me. TY!

AaronHForgeFlow:
@haraldqubiq thanks for reviewing!

stock_cancel/models/stock_picking.py
```diff
@@ -0,0 +1,59 @@
...
+                move.quant_ids._revert()
+
+            picking.group_id.procurement_ids.reset_to_confirmed()
+            sale_lines = picking.group_id.procurement_ids.mapped(
+                'sale_line_id')
+            for line in sale_lines:
+                line.qty_delivered = line._get_delivered_qty()
+            if sale_lines:
+                order = sale_lines[0].order_id
+                delivery_line = order.with_context(
```
fuentes73: Hi,

In a installation without delivery module installed. If your try to re-open a picking that comes from a sale, you'll get this error:

`AttributeError: 'sale.order.line' object has no attribute 'is_delivery'`

You should move this piece of code to stock_cancel_delivery module.

AaronHForgeFlow: I see. I have to move that to the other module proposed, stock_cancel_delivery. Thanks

Jortolsa-S73 (approved):

fuentes73 (approved):
Tested and working as expected

pedrobaeza:
Please squash the commits by author.

AaronHForgeFlow:
Kept the 2 last commits separate because they change different modules

PR #423, opened 2018-05-04
11.0 mig stock picking customer ref

andreasdsp:


stock_picking_customer_ref/models/stock_picking.py
```diff
@@ -0,0 +1,22 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.depends('group_id', 'sale_id.client_order_ref')
+    def _compute_client_order_ref(self):
+        for picking in self:
+            if picking.group_id:
+                cond = [('procurement_group_id', '=', picking.group_id.id)]
+                sale = self.env['sale.order'].search(cond, limit=1)
+                picking.client_order_ref = sale.client_order_ref
```
bimajatiwijaya: I think you need add condition to avoid error AttributeError: 'bool' object has no attribute 'client_order_ref' when variable sale is False.

stock_picking_customer_ref/tests/test_stock_picking_customer_ref.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 AvanzOSC - Alfredo de la Fuente
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import common
+
+
+class TestStockPickingCustomerRef(common.SavepointCase):
```
bimajatiwijaya, line 9: Lets do open discussion. What is advantage using common.Savepoint Case than TransactionCase ?

bimajatiwijaya (changes requested):

stock_picking_customer_ref/tests/test_stock_picking_customer_ref.py
andreasdsp: Thank you for the review. The difference between TransactionCase and SavepointCase is how it run the test. TransactionCase run the test method independently (1 test method, 1 transaction), SavepointCase run all the test methods in one transaction. For example, we create 2 test methods. In TransactionCase, it will create 2 transaction, but in SavepointCase, it will create 1 transaction. So, the advantage is SavepointCase more faster than TransactionCase. CMIIW.

stock_picking_customer_ref/models/stock_picking.py
andreasdsp: Thank you for the review, i'll update the code in the next push.

gurneyalex: requesting this change

gurneyalex (changes requested):

andreasdsp:
done

gurneyalex (approved):

bimajatiwijaya (approved):

PR #425, opened 2018-05-08
[11.0] Add Stock Picking Purchase Propagate module

grindtildeath:
Depends on :
 - [x] https://github.com/OCA/purchase-workflow/pull/550

Stock Picking Purchase Propagate

This module allows to propagate the procurement group and the quantity of the
purchase order at its confirmation to the ensuing picking and stock moves and
their destination moves and pickings.

Usage
=====

This module can help you if your warehouse uses two- or three- steps reception.

In such a case, odoo's scheduler will generate internal transfers pickings with
the procurement group defined on each orderpoint of the products in need.

So, if no procurement group is defined on the orderpoints, Odoo will generate
only one internal transfer picking for all the products having needs, even if
suppliers and delays could be totally different.

If the product is to be purchased, the scheduler will also generate a purchase
order. However when confirming this purchase order, the generated receipt
picking and its move will get the procurement group from the sale order, which
doesn't match the procurement group of the ensuing internal transfers, what
could be baffling for the stock operator who has to find the ensuing internal
transfer.

Moreover, if the quantity is changed before confirming the purchase order, the
receipt picking will be generated with the PO's quantity, whereas the ensuing
moves and picking will still have the original quantity from the orderpoint.
Therefore, if the quantity was reduced on the purchase order, the stock
operator won't be able to close the move line in waiting state, although it's
not expected to receive more quantity until the next purchase order.

Instead of provoking such headaches to stock operators, this module will
propagate the procurement group and the quantity of the purchase order to the
whole chain of moves and reassign them to new pickings.

This allows to have a clear match through the procurement group between
purchase order, receipts and internal transfers, and allows as well stock
operators not to worry about missing quantities which weren't ordered in the
first place.


stock_picking_purchase_propagate/models/stock_move.py
```diff
@@ -0,0 +1,60 @@
...
+        """Propagate group to self and dest moves where propagate is True.
+        Reassign picking to ensure no moves with different group share the
+        same picking.
+
+        Goal is to have all linked destination moves sharing the same
+        procurement group all along the chain.
+        """
+        # As a new picking will be assigned on the moves, save each partner
+        # in a move: partner dict
+        move_partners = {
```
grindtildeath: Actually I'm not sure if that's the best thing to do here, because the issue is here : 
https://github.com/odoo/odoo/blob/dff1e96ed35481c9805ec65b483a5e01e0593d98/addons/stock/models/stock_move.py#L691-L698

Odoo wipes both `partner_id` *and* `origin` if one *or* the other doesn't match.
So I don't know what's better to do, and I'm too tired to figure it out now. 
Any help appreciated.

rousseldenis (approved):
Code review

stock_picking_purchase_propagate/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Picking Purchase Propagate",
+    "summary": "Propagate procurement group and quantity from purchase order",
+    "version": "11.0.1.0.0",
+    "category": "Warehouse",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```
simahawk: remove SA here

stock_picking_purchase_propagate/models/stock_move.py
```diff
@@ -0,0 +1,47 @@
...
+        ).mapped('move_dest_ids').filtered(lambda m: not m.group_id)
+        if to_propagate:
+            # Recursive call on destination moves
+            to_propagate._propagate_procurement_group(group)
+        return res
+
+    @api.multi
+    def _propagate_quantity_to_dest_moves(self):
+        """Propagate the quantity to all dest moves where propagate is True."""
+        for move in self.filtered(lambda m: m.propagate):
```
simahawk: you are already looping: do this check inside the loop

stock_picking_purchase_propagate/tests/test_po_propagate.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class TestPoPropagate(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+
+        def _create_product(name, supplier):
+            product_tmpl_model = cls.env['product.template']
+            return product_tmpl_model.create({
```
simahawk: tracking_disable=True

stock_picking_purchase_propagate/models/purchase.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @api.multi
+    def _create_stock_moves(self, picking):
+        """ When creating the moves from a PO, propagate the procurement group
+            and quantity from the PO lines to the destination moves, and
+            reassign pickings.
+        """
+        moves = super(PurchaseOrderLine, self)._create_stock_moves(picking)
+        destination_moves_to_prop = moves.filtered(
+            lambda m: m.propagate).mapped('move_dest_ids').filtered(
```
simahawk: any chance to get this w/ single search?

simahawk (commented):
LGTM, just some small things

simahawk (changes requested):
LGTM, just some small things

stock_picking_purchase_propagate/models/purchase.py
grindtildeath: The only option I see is bypassing ORM with this :
```
        sql = """
        SELECT dest_mov.id
        FROM stock_move dest_mov
        INNER JOIN stock_move_move_rel rel ON rel.move_dest_id = dest_mov.id
        INNER JOIN stock_move orig_mov ON rel.move_orig_id = orig_mov.id
        WHERE orig_mov.propagate = TRUE 
        AND dest_mov.group_id IS NULL
        AND orig_mov.id IN %s;
        """
        self.env.cr.execute(sql, [tuple(moves.ids)])
        dest_moves_ids = [x[0] for x in self.env.cr.fetchall()]
        destination_moves_to_prop = self.env['stock.move'].browse(dest_moves_ids)
```
Might still be better in terms of performance than using two filtered and a mapped functions. What do you think ?

max3903 (approved):

stock_picking_purchase_propagate/models/purchase.py
simahawk: sounds good. When we get here everything should be already checked regarding permissions, no?

grindtildeath: As the `_create_stock_moves` function already creates records on stock_move (implying user has write access), and the two tables used are `stock_move` and `stock_move_move_rel` (which is only a m2m relation table between two stock_move fields) for a read operation, IMO it should be ok.

grindtildeath:
BTW let's ref this issue here, as it fixes the behaviours described by @max3903  : https://github.com/odoo/odoo/issues/23927#issuecomment-382811329

tschanzt (approved):

jgrandguillaume (approved):
LGMT, thanks !

JordiBForgeFlow:
There's something wrong with this module. The changes to quantities in the PO are propagated incorrectly. I'm going to provide steps to reproduce.

JordiBForgeFlow:
Steps to reproduce:
1. Create a second warehouse, WH2
2. Define a route WH->WH2 as follows:
image

3. Define a product "Widget" with the following routes:
image

4. Create and confirm two sales orders in WH2 for Widget, qty 1
image
This will create 1 internal transfer from WH->WH2 for Widget, qty 2.
This will create 1 PO with destination warehouse WH for Widget, qty 2.

5. Change the PO quantity from 2 to 4.

image

6. Check the status of the delivery orders associated to the sales orders. Will both show 4 as qty to deliver! So two customers order 1 each, and they get 4 each :S
image

JordiBForgeFlow (changes requested):
As described, needs further review.

grindtildeath:
@jbeficent Thanks for your review, I could reproduce the same issue using two orderpoints (one on WH/Stock and another on WH2/Stock) as the module doesn't depends on sale, at least we get the same behaviour. I'm trying now to write a test so we can cover this case.

However, I don't know what's the best solution to this, as we could split the quantity reduction on both stock.move but I don't think it's a good idea, because one stock.move might have a business priority on the other. But in such case there's almost no way to tell on which stock.move the quantity reduction should be done. :disappointed: 

@jgrandguillaume @vrenaville Any idea ?

jgrandguillaume:
Hi,


@jbeficent , @grindtildeath  thank for the deeper review, I only tested "my" case. I was kind of aware of this issue, it would have been worth mentioning it in an earlier stage.. For the record, Odoo didn't find any way to fix this properly. We had a discussion with them on the topic and no good one was found. I'll try to summarized the problem we do face:

 * Manual PO behave differently than order point triggered ones -> that's why we do want to propagate the procurement group
 * Changes in quantities in PO are not propagated neither to moves, which can result (in case of several steps reception) in a headache for the operator. The move from vendor -> input will have X quantity where the following moves (e.g. input -> stock) will have Y quantities. This result in bad behavior of the system, and especially if you lower the quantities -> Your moves will wait for ever.

Propagating the changes as we've done it here will solve those 2 issues BUT if you have a setup mentioned above by Jordi (pulling to an intermediate WH) then the changes is propagated till the SO deliveries as well...

Proposal to solve this:

 * We could propagate the changes up till the stock location of the PO's warehouse and stop there if more moves are chained. This wouldn't break your use case Jordi, but you'll still face the issue of manual fixing (moves from Vendors -> WH). See my examples [1], [2], [3]
 * We should also avoid the PO line to merge in a single line when the procurement group are different -> So when 2 procurement trigger with different procurement group, they'll have 2 PO lines. This way if the user change the qty, we know on which one to propagate or not the changes.

Finally, explaining the limitation imposed by this modules to the user/integrator. Honestly, several DB of our customers has been messed up (even recently) because of this bad behavior in a several step reception. It is very nasty to correct afterwards and if you do not have an intermediate WH, this module work fine as is.

What do you think ? Would you agree with this ?

Regards,

Joël

[1]

In the setup proposed by Jordi (https://github.com/OCA/stock-logistics-workflow/pull/425#issuecomment-389787049), you have the following moves generated:

1. Vendors -> WH
1. WH -> WH2
1. WH2 -> Customer

With my proposal, the propagation will stop at the stock location of the PO's WH, so the move 1. will not propagate the changes to WH2.

- **Pro:** That's probably what the user want. Having more product in WH as he choose to purchase more, but do not ship them all to WH2 if only part of it has been sold to Customer.
- **Cons:** Moves from 2. and 3. are chained moves, so they will always kept the original quantities. In the test case provided by Jordi, you ordered more then pulled by procurement rules so it is fine. But if you ordered less in the PO, then the chained moves 2. and 3. will be kept opened (as you'll received less quantities) fore ever ! But that is already the case today as well so no worst...

[2]

In the standard 2 steps reception, you have the following moves generated:

1. Vendors -> Input
1. Input -> WH

The propagation will go up to 2.

-**Pro:** That's what we want
-**Con:** -

[3]

In the standard 2 step reception with MTO sales, you have the following moves generated:

1. Vendors -> Input
1. Input -> WH
1. WH -> Customer

The propagation will go up to 2 and stop there. So the quantity ship to the customer will be the one he ordered. If you ordered less by lowering the qty in the PO, then your move 3. will never be 100% shipped and you'll have to handle that manually as per the current behavior so it is no worst.

JordiBForgeFlow:
@grindtildeath @jgrandguillaume @max3903 

You have to consider that my testing resulted in 4 units duplicated in SO1 and SO2, where the PO was for 4. So it seems that you are allocating the full quantity of the PO to all the moves, and instead quantities shall be split. So something funny is happening..

On the other hand I think that the key to resolve the problem is to perform the changes to the chained moves IF they are meant to fulfill an orderpoint, but not otherwise.

In case of orderpoints you are in MTS (make to stock) scenario. Inventory accepts variances between what was requested by the orderpoint and what is finally purchased. You are restocking a buffer position, no matter if it's a bit less, a bit more. But in this scenario, as you accept this variance, you want to make sure that the products received will reach the expected final destination, in 2, 3 step receiving.

On the contrary, when you buy for a customer order, or for a manufacturing order in an MTO scenario there's no room to accept more quantity (perhaps you can accept less?) and seems to me that no change should be allowed. 

If you ordered more in the PO in an MTO scenario, the system should keep it in the default receiving location defined by the picking type. Where are you going to send the stock otherwise? Not to the customer or to a specific MO, of course!

Orderpoints/Buffers are flexible. Customer orders or manufacturing orders are not.

grindtildeath:
@jbeficent Sorry I wrote this, before seeing you last comment
I did two things to avoid bad side effects and keep the original purpose of this module :
1. Add a dependency on purchase_line_procurement_group (https://github.com/OCA/purchase-workflow/pull/550). This module splits the po lines for the same product according to their original procurement group. So if the moves following the supplier>input move have different procurement group, we get different lines on the PO, different lines on the supplier>input move, and reducing quantity on the PO line allows the propagation to the correct ensuing moves.
2. ~Add a check before propagating to destination moves, so if the sum of dest moves is different than their original procurement qty, we don't propagate (and we keep the same situation as with standard odoo, the reduction isn't applied, it's not perfect but it's no worse than propagating something false).~

~Now, considering your last comment, don't you think, the check mentioned on point 2 should be enough ?~

grindtildeath:
@jbeficent After discussing with @jgrandguillaume and @max3903, we removed the propagation of quantity to destination moves if there are more than one destination move. As there's no possibility to tell on which destination move the change should be applied, better do nothing than propagate something false.
Also with https://github.com/OCA/purchase-workflow/pull/550 being a dependency, PO lines will not be merged if a procurement group is defined before, so we should be fine with any use case, as any change on the PO will only be propagated in case of 1:1 relation between stock moves.

grindtildeath:
Looks like we have the same issue between OCA/OCB and odoo/odoo (https://github.com/OCA/purchase-workflow/pull/550#issuecomment-390258482)

elicoidal:
Back in v7 we did a similar concept to propagate the split of moves and we opted to add a checkbox at location level to allow to break of the propagation.

grindtildeath:
Tests are now fixed with https://github.com/OCA/stock-logistics-workflow/pull/425/commits/2f7d210fc1e022bd88073da4f27eadf49be9c31f (cf explanation here : https://github.com/OCA/purchase-workflow/pull/550#issuecomment-391384944) 

@jbeficent Can you update your review ? :blush:

jgrandguillaume:
@jbeficent  Gentle reminder on this one, are you ok to merge ?

grindtildeath:
@simahawk @jbeficent Can you guys please update your reviews ?

simahawk (approved):

simahawk:
@grindtildeath base PR merged: can you drop oca_dep?

simahawk:
@jbeficent ping :)

PR #426, opened 2018-05-10
[10.0][MIG] stock_move_backdating

tafaRU:


pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

tafaRU:
Right @pedrobaeza !
Please note I just set WIP label as Travis fails.

tafaRU:
Travis is failing due to an error with `stock_ownership_availability_rules` module. See https://github.com/OCA/stock-logistics-workflow/issues/429

tafaRU:
@OCA/logistics-maintainers any clue about https://github.com/OCA/stock-logistics-workflow/pull/426#issuecomment-388353299?
Thank you in advance.

tafaRU:
@pedrobaeza, could you please take a look at https://github.com/OCA/stock-logistics-workflow/pull/426#issuecomment-388353299 ?
Are still supported options `EXCLUDE` and `INCLUDE` by MQT?

pedrobaeza:
> Are still supported options EXCLUDE and INCLUDE by MQT?

Yes, they are, although you should try to avoid as runbot is only available for the first line

tafaRU:
> Yes, they are, although you should try to avoid as runbot is only available for the first line

Thank you @pedrobaeza for the prompt reply! I'll give it a try then considering that I don't know how to fix https://github.com/OCA/stock-logistics-workflow/issues/429.

tafaRU:
As it turned out, now travis is green!

tafaRU:
Travis is finally green! Please @OCA/logistics-maintainers review it.

stock_move_backdating/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.multi
+    def do_transfer(self):
+        result = super(StockPicking, self).do_transfer()
+        for picking in self:
+            # set date_done as the youngest date among the moves
+            dates = picking.mapped('move_lines.date')
```
rousseldenis: @tafaRU Are move_lines shouldn't be filtered with
```python
picking.move_lines.filtered(lambda m: m.state == 'done').mapped('date')
```
?
You can have cancelled moves in picking

stock_move_backdating/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+            picking.write({'date_done': max(dates)})
+            for move in picking.move_lines:
+                # set date on account.move same as date on stock.move
+                account_moves = self.env['account.move'].search(
```
rousseldenis: @tafaRU Please avoid searching in a loop.

Get all account moves before loop.

stock_move_backdating/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+                    [('ref', '=', picking.name),
+                     ('line_ids.name', '=', move.name)])
+                for account_move in account_moves:
+                    account_move.date = move.date
```
rousseldenis: @tafaRU Please prefer 
```python
account_moves.write({'date': move.date})
```
You are not sure a write will occur and recomputation either.

rousseldenis (changes requested):
Code review

stock_move_backdating/models/stock_picking.py
tafaRU: This module is intended to be used on the `pickings` in state `available`, when the Operations tab is shown, so I don't understand the reason to filter only the moves done.

stock_move_backdating/models/stock_picking.py
tafaRU: could you please show me how to modify it accordingly?

stock_move_backdating/models/stock_picking.py
tafaRU: ok, I agree. Thanks! I'll modify it after your answers (see other open points).

tafaRU:
@rousseldenis could you please check my answers to your requests?
Thank you in advance.

stock_move_backdating/models/stock_picking.py
rousseldenis: @tafaRU What I wanted to say is that on that code line, your moves where already transferred (the super). But I imagine you want to get the dates from dones moves no (or I don't understand the aim)?
Moreover, your comment mention that you want the youngest date, shouldn't it be 'min()'  instead ?

stock_move_backdating/models/stock_picking.py
rousseldenis:
```python
all_moves = self.env['account.move'].search([('ref', '=', picking.name), ('line_ids.name', 'in', picking.mapped('move_lines.name'))])
for move in picking.move_lines:
    for account_move in all_moves.filtered(lambda m, pm=move: pm.name in m.line_ids.mapped('name')):
        account_move.date = move.date
```
Don't forget every search is a read in DB, here you fetch all records once and then, just python code

rousseldenis: @tafaRU The only thing I oppose to this kind of logic is that you rely on name to map two models, is it accurate (can't we have false values) ?

stock_move_backdating/models/stock_picking.py
tafaRU: Let me try to explain what I mean with youngest date with the following example:
date_1 = `2018-06-12` date_2 = `2018-06-13` --> max(date_1, date_2) = date_2 which is the youngest date, isn'it?

stock_move_backdating/models/stock_picking.py
tafaRU: The `account.move.line` values are prepared at https://github.com/odoo/odoo/blob/10.0/addons/stock_account/models/stock.py#L274 so the `name` should not have false values in respect of https://github.com/odoo/odoo/blob/10.0/addons/stock/models/stock_move.py#L26

tafaRU:
@rousseldenis I should have addressed all your comments. Could you please check it?

tafaRU:
@rousseldenis ping!

rousseldenis:
@tafaRU I will come back soon but I have no time for now. Sorry

stock_move_backdating/models/stock_picking.py
rousseldenis: That's it. Understand

stock_move_backdating/models/stock_picking.py
rousseldenis: @tafaRU Yes I understand. What I said is relying on name and not ids to map two models is not accurate. But you don't have other means to do so.

stock_move_backdating/models/stock_pack_operation.py
```diff
@@ -0,0 +1,24 @@
...
+# Copyright 2018 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+def check_date(date):
+    now = fields.Datetime.now()
+    if date > now:
```
rousseldenis: @tafaRU What happen if you void the date_backdating field? Do you have an error?

tafaRU: @rousseldenis, we didn't have error as:

```
  9  	def check_date(date):
 10  	    import pdb;pdb.set_trace()
 11  	    now = fields.Datetime.now()
 12  ->	    if date > now:
 13  	        raise UserError(
 14  	            _("You can not process an actual "
 15  	              "movement date in the future."))
 16  	
 17  	
(Pdb) date
False
(Pdb) date > now
False
```
:wink:

Anyway, it's better to add an explicit check as done at https://github.com/OCA/stock-logistics-workflow/pull/426/commits/b78ca2d009aa14aec19396d3392b1ac1f1345344

tafaRU:
@rousseldenis, is it ok now for you?

tafaRU:
@rousseldenis, ping! Is it ok now for you?

gurneyalex (approved):

rousseldenis (approved):
Code review

stock_move_backdating/wizards/fill_date_backdating.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Alex Comba - Agile Business Group
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
+
+<odoo>
+    <data>
```
SimoRubi: remove `data` tag

stock_move_backdating/wizards/fill_date_backdating.xml
```diff
@@ -0,0 +1,36 @@
...
+        <record id="view_fill_backdating" model="ir.ui.view">
+            <field name="name">Fill Actual Movement Date</field>
+            <field name="model">fill.date.backdating</field>
+            <field name="arch" type="xml">
+              <form string="Fill Actual Movement Date">
+                  <group>
```
SimoRubi: Why nested `group`?

SimoRubi (approved):
Just minor changes, otherwise code review looks good to me

stock_move_backdating/wizards/fill_date_backdating.py
```diff
@@ -0,0 +1,20 @@
...
+    _name = "fill.date.backdating"
+
+    date_backdating = fields.Datetime(string='Actual Movement Date')
+
+    @api.multi
+    def fill_date_backdating(self):
+        """ Fill the Actual Movement Date on all pack operations. """
+        self.ensure_one()
+        picking = self.env['stock.picking'].browse(self._context['active_id'])
+        for pack in picking.pack_operation_product_ids:
```
SimoRubi: Maybe it's better to use `write` for bulk writing here too.
Moreover, don't you have to check here if the date is in the future too?

tafaRU:
@SimoRubi, thank you for your review! Done at https://github.com/OCA/stock-logistics-workflow/pull/426/commits/347ace58df790790d8aaa33ad7d273b0b2c7f6a5

stock_move_backdating/wizards/fill_date_backdating.py
```diff
@@ -3,18 +3,23 @@
...
+        check_date(self.date_backdating)
+
     @api.multi
     def fill_date_backdating(self):
         """ Fill the Actual Movement Date on all pack operations. """
         self.ensure_one()
         picking = self.env['stock.picking'].browse(self._context['active_id'])
         for pack in picking.pack_operation_product_ids:
-            pack.date_backdating = self.date_backdating
+            pack.write({'date_backdating': self.date_backdating})
```
SimoRubi: @tafaRU I meant `picking.pack_operation_product_ids.write({'date_backdating': self.date_backdating})`

tafaRU: done at https://github.com/OCA/stock-logistics-workflow/pull/426/commits/2bf7cd11c0f251812565b6473cc939ebc3a44a2d

tafaRU:
@OCA/logistics-maintainers can this be merged?
Thanks.

gurneyalex (approved):

PR #428, opened 2018-05-11
[FIX] stock_ownership_availability_rules: apply a workaround to get the module compatible with Perpetual inventory valuation

tafaRU:
Whithout that, any stock valuation will be done automatically.
See https://github.com/odoo/odoo/blob/10.0/addons/stock_account/models/stock.py#L92 for further info.

### DIFF
--- a/stock_ownership_availability_rules/README.rst
+++ b/stock_ownership_availability_rules/README.rst
@@ -38,3 +38,4 @@ Contributors
 ------------
 
 * Leonardo Pistone <leonardo.pistone@camptocamp.com>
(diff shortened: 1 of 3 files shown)


gurneyalex (approved):

tafaRU:
@gurneyalex, I close this as Travis clearly fails (https://travis-ci.org/OCA/stock-logistics-workflow/jobs/377716576#L1213). I excluded then the module from tests at https://github.com/OCA/stock-logistics-workflow/pull/426/commits/f0ec30587d322e8a1c77546fc7251083f799f9d1

PR #430, opened 2018-05-18
[MIG] stock_picking_transfer_lot_autoassign: Migration to 11.0

chienandalu:
Auto-assignment of lots on pickings
===================================

When working with lots, it's very uncomfortable to introduce the quantity,
lot by lot, when transferring pickings from your warehouse (outgoing or
internal).

This module automatically assigns the reserved quantity as the done one, so
that you only have to change it in case of divergence, but having the
possibility of transferring directly.

Also this module adds a button in backorder confirmation wizard to auto
complete to do quantities for products without lots.

cc @Tecnativa


pedrobaeza:
@lmgr0312 this PR is for the module you mention. Please review it.

pedrobaeza (approved):

stock_picking_transfer_lot_autoassign/models/stock_picking_type.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = 'stock.picking.type'
+
+    avoid_internal_assignment = fields.Boolean(
```
gurneyalex, line 10: I find this field name confusing, especially since it widely differs from the displayed name. 

Can we take the opportunity of this migration to rename it to "avoid_auto_assignment"?

stock_picking_transfer_lot_autoassign/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+    _inherit = 'stock.move'
+
+    def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
+        """Auto-assign as done the quantity proposed for the lots"""
+        self.ensure_one()
+        res = super(StockMove, self)._prepare_move_line_vals(
+            quantity, reserved_quant,
+        )
+        if (self.picking_id.picking_type_id.avoid_internal_assignment or
+                not res.get('lot_id')):
```
gurneyalex, line 18: this is a strange and asymetric behavior. If I have both tracked and untracked products in a picking, then the update of qty_done will only happen for tracked product, which will result in a backorder if I just click on "Validate" without changing the done quantities, which seems to break the principle of least surprise.

gurneyalex (commented):
I won't block the migration of this module, but I find the provided feature dangerous and likely to cause stock errors. And I also think the specification is really weird, as explained in the review, since tracked and untracked products are not processed in the same way.

stock_picking_transfer_lot_autoassign/models/stock_move.py
sergio-teruel: Perhaps we can add an option in picking type to allow choose if autoassign all (tracked and un tracked products), or directly autoassign all...
@chienandalu @pedrobaeza What is your opinion about this?

stock_picking_transfer_lot_autoassign/models/stock_picking_type.py
sergio-teruel: I agree..

stock_picking_transfer_lot_autoassign/views/stock_picking_type_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record model="ir.ui.view" id="view_picking_type_form_assignment">
+        <field name="name">stock.picking.type.assignment</field>
+        <field name="model">stock.picking.type</field>
+        <field name="inherit_id" ref="stock.view_picking_type_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='use_existing_lots']" position="after">
+                <field name="avoid_internal_assignment" groups="stock.group_production_lot"/>
```
sergio-teruel, line 10: Make the view inheritable only for this group as
```
<field name="inherit_id" ref="stock.view_picking_type_form"/>
<field name="groups_id" eval="[(4,ref('stock.group_production_lot'))]"/>
...
...
```

sergio-teruel (changes requested):

rafaelbn:
@pedrobaeza please change label if this PR still mustn't be tested

chienandalu:
Closing this in favour of: https://github.com/OCA/stock-logistics-workflow/pull/435 wich merges `stock_picking_lot_autoassign` into `stock_pack_operation_autofill`

PR #431, opened 2018-05-31
[10.0] FIX stock_picking_package_preparation_line: do not erase lines fields when picking is created from lines

eLBati:
Use case:
 - Create a preparation, add a line with a product and fill 'note' field
 - Put in pack

'note' field is empty

### DIFF
--- a/stock_picking_package_preparation_line/models/stock_picking_package_preparation.py
+++ b/stock_picking_package_preparation_line/models/stock_picking_package_preparation.py
@@ -180,6 +180,9 @@ class StockPickingPackagePreparation(models.Model):
                         })
                 # Set the picking as "To DO" and try to set it as
                 # assigned
+                # skip_update_line_ids because picking is created based on
+                # preparation lines, updating lines would erase some fields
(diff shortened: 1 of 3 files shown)


stock_picking_package_preparation_line/models/stock_picking_package_preparation.py
```diff
@@ -180,16 +180,20 @@ def action_put_in_pack(self):
                         })
                 # Set the picking as "To DO" and try to set it as
                 # assigned
-                picking.action_confirm()
+                # skip_update_line_ids because picking is created based on
+                # preparation lines, updating lines would erase some fields
+                picking.with_context(
```
tafaRU: wouldn't be better to update the context only one time by `picking = picking.with_context(skip_update_line_ids=True)`?

eLBati:
@tafaRU thanks, done

tafaRU (approved):

SimoRubi (approved):
code review and tested in runbot

LoisRForgeFlow (approved):
Code review

PR #432, opened 2018-06-06
[10.0][MIG] stock_picking_backorder_strategy

rousseldenis:


stock_picking_backorder_strategy/__manifest__.py
```diff
@@ -1,36 +1,20 @@
...
-    "depends": ["stock",
-                ],
-    "data": ["views/stock_views.xml",
-             ],
-    "demo": [],
-    "license": "AGPL-3",
-    'installable': False,
-    "application": False,
+    'name': 'Picking backordering strategies',
+    'version': '10.0.0.1.0',
```
LoisRForgeFlow, line 7: isn't `X.0.1.0.0` the default?

LoisRForgeFlow (approved):
Code review, LGTM :+1:

stock_picking_backorder_strategy/__manifest__.py
rousseldenis: @lreficent Yes of course. I'll change it

Jortolsa-S73:
In 'no_create' mode, the stock move with the remaining quantity remains pending in the same picking. 
In my opinion that move must be cancelled or not created to avoid problems with extra moves.
stock_picking_backorder

- Use no_create option
- Validate picking with remaining quantity
- The picking still pending with a stock move done and the remaining quantity pending in other stock move
- Validate the picking again
- Odoo creates an extra stock move

@rousseldenis do you agree with the solution of not create the stock move with remaining quantity? Could we help you with the fix? We are in the OCA Code Sprint in Sevilla.

gurneyalex (approved):

LoisRForgeFlow:
@rousseldenis could you have a look at @Jortolsa comments?

rousseldenis:
@lreficent I will come back soon (few days). I'm on another subject  now.

rousseldenis:
@Jortolsa @lreficent  I've done modifications (+ tests) in another commit.

Jortolsa-S73 (approved):

pedrobaeza:
Are you going to squash a bit your commits before merging, @rousseldenis ?

rousseldenis:
@pedrobaeza Done. I want to keep these two commits.

PR #438, opened 2018-06-25
[11.0][FIX] Split also chained pickings when splitting a picking

asaunier:
As for now, when a picking is automatically created by a push rule, splitting the initial picking (using ``split_process()``) does not propagate the change to the chained pickings.

### DIFF
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -14,10 +14,9 @@ class StockPicking(models.Model):
 
     @api.multi
     def split_process(self):
-        """Use to trigger the wizard from button with correct context"""
         for picking in self:
 
-            # Check the picking state and condition before split
+            # Check the picking state and condition before splitting
             if picking.state == 'draft':
                 raise UserError(_('Mark as todo this picking please.'))
             if all([x.qty_done == 0.0 for x in picking.move_line_ids]):
@@ -27,6 +26,7 @@ class StockPicking(models.Model):
 
             # Split moves considering the qty_done on moves
             new_moves = self.env['stock.move']
+            product_qty_done = {}
             for move in picking.move_lines:
                 rounding = move.product_uom.rounding
                 qty_done = move.quantity_done
@@ -49,6 +49,8 @@ class StockPicking(models.Model):
                             try:
                                 move_line.write(
                                     {'product_uom_qty': move_line.qty_done})
+                                product_qty_done[move_line.product_id.id] = \
+                                    move_line.qty_done
                             except UserError:
                                 pass
                     new_moves |= self.env['stock.move'].browse(new_move_id)
@@ -78,3 +80,19 @@ class StockPicking(models.Model):
                     'picking_id': backorder_picking.id,
                 })
                 new_moves._action_assign()
+
+                # Propagate split to chained picking/moves
+                chained_moves = picking.move_lines.mapped('move_dest_ids')
+                for chained_move in chained_moves:
+                    # import pdb; pdb.set_trace()
+                    chained_picking = chained_move.picking_id
+                    chained_picking.action_assign()
+                    for move_line in chained_move.move_line_ids:
+                        # FIXME is it the right way to propagate done qty to
+                        # the chained moves?
+                        qty_done = product_qty_done[move_line.product_id.id]
+                        move_line.write({
+                            'qty_done': qty_done,
+                            'product_uom_qty': qty_done,
+                        })
+                    chained_picking.split_process()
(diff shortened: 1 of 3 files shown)


stock_split_picking/models/stock_picking.py
```diff
@@ -78,3 +80,19 @@ def split_process(self):
...
+                    chained_picking.action_assign()
+                    for move_line in chained_move.move_line_ids:
+                        # FIXME is it the right way to propagate done qty to
+                        # the chained moves?
+                        qty_done = product_qty_done[move_line.product_id.id]
+                        move_line.write({
+                            'qty_done': qty_done,
+                            'product_uom_qty': qty_done,
+                        })
+                    chained_picking.split_process()
```
grindtildeath: Before calling this, you probably have to write on chained_move.qty_done, otherwise it will raise the error on line 23.

stock_split_picking/models/stock_picking.py
```diff
@@ -78,3 +80,19 @@ def split_process(self):
                     'picking_id': backorder_picking.id,
                 })
                 new_moves._action_assign()
+
+                # Propagate split to chained picking/moves
+                chained_moves = picking.move_lines.mapped('move_dest_ids')
+                for chained_move in chained_moves:
```
grindtildeath: We probably don't want to propagate the split if `len(chained_moves) > 1`. But I'm not sure yet if we should raise an error or not.

grindtildeath (commented):
Let's have a look together this afternoon.

stock_split_picking/models/stock_picking.py
asaunier: I think we can have more than one chained move here and should handle this case. For instance if the picking contains several products (then several moves) => the split might apply on several moves and cause several chained moves.

stock_split_picking/models/stock_picking.py
```diff
@@ -49,6 +49,8 @@ def split_process(self):
                             try:
                                 move_line.write(
                                     {'product_uom_qty': move_line.qty_done})
+                                product_qty_done[move_line.product_id.id] = \
+                                    move_line.qty_done
```
asaunier: I have tested in the UI and programmatically in unit tests: an error is raised when trying to create several moves with the same product id in the same picking.
As a result I think using the product id as index will not cause conflicts between moves.

stock_split_picking/models/stock_picking.py
grindtildeath: Well, my comment wasn't really clear, but if each move is linked to only one `move_dest_ids` we should propagate. If a move is linked to multiple `move_dest_ids`, there's no possibility to tell which move should be split. :wink: 
Therefore you might have to loop `picking.move_lines` instead of `chained_moves` and proceed from there :
```
for move in picking.move_lines:
    if len(move.move_dest_ids) > 1:
        continue
```

asaunier: I will investigate this hint, thanks!

asaunier:
@grindtildeath I think I have taken your suggestions into account.

I have also created a more sophisticated scenario for the unittests in order to make sure it is correctly managed. And it was useful to spot flaws.

Here is the scenario:
* the initial picking has 3 moves, for 3 distinct products (1 move for each product, I have figured out while writing the tests and playing with the UI that it is not possible to have 2 moves with the same product) with quantities 10, 20 and 30
* because of a push rule, a chained picking is automatically created with moves for 10 product, 20 product2, 30 product3
* after setting ``qty_done`` to 4 product (6 remaining), 20 product2 (0 remaining) and 0 product3 (30 remaining) and triggering a split we get the following pickings:

first step | second step
--------- | ----------
initial picking = 4 product + 20 product2 + 0 product3 | chained picking = 4 product + 20 product2 + 0 product3
backorder = 6 product + 0 product2 + 30 product3 | chained backorder = 6 product + 0 product2 + 30 product3

gurneyalex:
Does not seem to work:

I configure my WH to receive in 2 steps:
image

I create a PO for 10 ipads:

image

I get 2 pickings:

image

I split the incoming picking in two. When I check the internal picking has not been split:

image

@asaunier @grindtildeath Am I missing something obvious?

gurneyalex (changes requested):
interactive test on a clean database with purchase installed seem to indicate that it does not work.

gurneyalex:
cc @leemannd

rousseldenis:
@asaunier

rousseldenis:
@saunier @leemannd Do you plan to finish this ?

rousseldenis:
@asaunier

leemannd:
@rousseldenis I was not working on this at that time. What I've found (to be confirmed) we internally closed this because unit test where not working. Given @gurneyalex 's test. It seems we are wrong.
I sadly have no time to have a look at it, but maybe you want to check with other people from camptocamp present at the code sprint?

rousseldenis:
@leemannd Sorry to hear you will not attend sprint! Ok, I'll look forward people for that.

rousseldenis:
Closing as old. Please feel free to reopen if needed.

PR #439, opened 2018-07-02
[10.0][FIX] stock_picking_invoice_link

MiquelRForgeFlow:
Adapt to https://github.com/odoo/odoo/commit/ef444da57a0e5b3a8426a6fac8cef3f6d856b148.

Also, changed `force_assign()` to `action_assign()` because the needed quantity is in stock, don't need to force.


MiquelRForgeFlow:
@lreficent @pedrobaeza @cubells

pedrobaeza (approved):
We can fast-track this when CIs finish

PR #440, opened 2018-07-10
[10.0][FIX] stock_picking_invoice_link

MiquelRForgeFlow:
Adapt to odoo/odoo@b1a373c.

Sigh... 😒


MiquelRForgeFlow:
@lreficent @pedrobaeza

pedrobaeza:
Maybe we should change the approach for not checking the returning value, but the state of the data?

MiquelRForgeFlow:
Well, but I prefer my lazy approach. I don't think Odoo will change again this 😂😤

MiquelRForgeFlow:
:green_apple: @pedrobaeza ASAP

PR #442, opened 2018-07-11
[FIX] resolve bug for mass transfer

quentinDupont:
PR to fix the bug that happens when you try to mass transfer some stock.picking.

Before that, creation of transfert_wizard returned null because there were several active_ids.


rousseldenis (approved):
Makes sense

gurneyalex (approved):

PR #445, opened 2018-07-26
[FIX] stock_picking_invoice_link: return moves

chienandalu:
cc @Tecnativa


MiquelRForgeFlow (approved):

pedrobaeza (approved):

carlosdauden (approved):

afilomeno:
Hi,
I've installed that module (return move), and I notice that when I going to make a return and then print the invoice, Total Amount  has changed!! Please refer to attached file to view what I mean...

Thanks on advance,




CapturaA

PR #460, opened 2018-10-04
11.0 mig stock picking mass action

mpanarin:
migration of the module to v11.0


stock_picking_mass_action/tests/test_mass_action.py
```diff
@@ -49,38 +49,50 @@ def setUpClass(cls):
         })
 
     def test_mass_action(self):
-        self.assertEqual(self.picking.state, 'draft')
+        self.assertEqual(self.picking.state, 'confirmed')
```
rousseldenis, line 52: Has picking flow changed (default is 'confirmed' ?) ?

stock_picking_mass_action/wizard/mass_action.py
```diff
@@ -23,51 +22,72 @@ def _default_force_availability(self):
     def _default_transfer(self):
         return self.env.context.get('transfer', False)
 
+    def _default_picking_ids(self):
+        self.write({
```
rousseldenis: Quite weird even in a wizard. Maybe update() is more correct

rousseldenis (approved):
Code review. Minor comments

stock_picking_mass_action/__manifest__.py
```diff
@@ -1,10 +1,9 @@
-# -*- coding: utf-8 -*-
-# Copyright 2014 Camptocamp SA - Guewen Baconnier
+# Copyright 2018 Camptocamp SA
```
TDu: Shouldn't the copyright keep the original date like this 2014-2018 ?
And it is the same in other files

stock_picking_mass_action/tests/test_mass_action.py
mpanarin: pickings are now autoconfirmed on creation
https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_picking.py#L508

yvaucher (approved):

PR #463, opened 2018-10-22
[11.0][ADD] stock_picking_auto_create_lot: Create lot for incoming pickings

sergio-teruel:
cc @Tecnativa


stock_picking_auto_create_lot/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    def button_validate(self):
+        if self.picking_type_code == 'incoming':
```
rousseldenis: @sergio-teruel IMHO to be flexible, this should be an option in stock.picking.type (doing this you limit lot auto creation to incoming pickings).

So,

```suggestion
        if self.picking_type_id.auto_create_lot:
```

stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import SavepointCase
+
+
+class TestStockPickingAutoCreateLot(SavepointCase):
+    at_install = False
+    post_install = True
```
rousseldenis: I don't think this module breaks normal behaviour

rousseldenis (changes requested):
Code review. Minor comment

sergio-teruel:
@rousseldenis Changes done!!. About add field to operation types, I was a bit lazy... :smiley:

rousseldenis (approved):
LGTM. Code review

pedrobaeza (approved):

PR #466, opened 2018-10-29
[FIX] stock_pack_operation_auto_fill: Autofill button permissions

chienandalu:
Only stock users should be able to process done quantities.

cc @Tecnativa


pedrobaeza (approved):

rousseldenis (approved):
Fast one

LoisRForgeFlow (approved):

PR #467, opened 2018-10-30
[11.0][MIG] stock_picking_filter_lot

LoisRForgeFlow:
Migration to v11:

* Adapted to new stock.move.line model.


rousseldenis (approved):
Code review

stock_picking_filter_lot/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+To use this module:
+
+#. Go to an open picking with products that are tracked by lots:
```
SimoRubi: Remove semicolon at the end of line

stock_picking_filter_lot/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
...
+#. When you enter detailed operations you will notice that the lots available
+   to select are filtered by the origin of the operation.
```
SimoRubi, line 5: Why filtered by the origin of the operation?
The module in v10 has the field `filter_location_id` that lets the user specify the location to use in the filter

SimoRubi (changes requested):
Thanks for your contribution!
Just code review waiting for runbot's build

stock_picking_filter_lot/readme/USAGE.rst
LoisRForgeFlow: Yes, but why is that needed, if you are moving from a given location you just want to see available lots at that location, why an extra field to filter?

BTW, I was thinking to add pair of improvements:
* Show the the qty available to transfer with a given lot and location filled.
* Add an option on the picking types to remove this filtering, allowing you to force the use of a given lot.

What do you think?

SimoRubi: Ok that makes sense @lreficent.
Regarding the new features (I'd suggest a new dedicated PR):
* The available quantity to transfer could be useful
* That's an interesting feature too, I'd add a check (maybe only visible in debug mode) in the lots popup that inherits from the configuration in the picking type and allows to override that for the current transfer

SimoRubi: Ok that makes sense @lreficent.
Regarding the new features (I'd suggest a new dedicated PR):
* The available quantity to transfer could be useful
* That's an interesting feature too, I'd add a check (maybe only visible in debug mode) in the lots popup that inherits from the configuration in the picking type and allows to override that for the current transfer

SimoRubi (approved):
Just take a look at the semicolon I pointed out in https://github.com/OCA/stock-logistics-workflow/pull/467/commits/c64e3e450de99899570f9beda3352dd4b7a11432#r230679182, obviously not blocking

stock_picking_filter_lot/readme/USAGE.rst
LoisRForgeFlow: Done!

stock_picking_filter_lot/readme/USAGE.rst
LoisRForgeFlow: Ok, let's go for a different PR. I'll merge this when green then.

PR #471, opened 2018-11-20
[IMP] stock_split_picking

gurneyalex:
* allow to split pickings with no available stock
* fix crash when processing a picking which was split twice

### DIFF
--- a/stock_split_picking/models/__init__.py
+++ b/stock_split_picking/models/__init__.py
@@ -3,4 +3,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import stock_picking
-from . import stock.picking.split
+from . import stock_picking_split
--- a/stock_split_picking/models/__init__.py
+++ b/stock_split_picking/models/__init__.py
@@ -3,3 +3,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import stock_picking
+from . import stock.picking.split
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -3,7 +3,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, models
-from odoo.exceptions import UserError
 
 
 class StockPicking(models.Model):
@@ -29,16 +28,3 @@ class StockPicking(models.Model):
                     'res_id': wiz.id,
                     'context': self.env.context,
             }
-
-        for pick in self:
-            if pick.state == 'draft':
-                raise UserError(_('Mark as todo this picking please.'))
-            if all([x.qty_done == 0.0 for x in pick.pack_operation_ids]):
-                raise UserError(
-                    _('You must enter done quantity in order to split your '
-                      'picking in several ones.'))
-            wizard = self.env['stock.backorder.confirmation'].with_context(
-                active_id=pick.id,
-                do_only_split=True,
-            ).create({})
-            wizard.process()
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2015 Camptocamp SA - Nicolas Bessi
+# Copyright 2013-2018 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, models
@@ -14,6 +14,22 @@ class StockPicking(models.Model):
     @api.multi
     def split_process(self):
         """Use to trigger the wizard from button with correct context"""
+        self.ensure_one()
+        wiz = self.env['stock.picking.split'].create({'picking_id': self.id})
+        view = self.env.ref('stock_split_picking.stock_picking_split_form')
+        return {
+                    'name': _('Split Picking'),
+                    'type': 'ir.actions.act_window',
+                    'view_type': 'form',
+                    'view_mode': 'form',
+                    'res_model': 'stock.picking.split',
+                    'views': [(view.id, 'form')],
+                    'view_id': view.id,
+                    'target': 'new',
+                    'res_id': wiz.id,
+                    'context': self.env.context,
+            }
+
         for pick in self:
             if pick.state == 'draft':
                 raise UserError(_('Mark as todo this picking please.'))
--- a/stock_split_picking/models/stock_picking_split.py
+++ b/stock_split_picking/models/stock_picking_split.py
@@ -1,0 +1,131 @@
+# -*- coding: utf-8 -*-
+# Copyright 2013-2018 Camptocamp SA - Alexandre Fayolle
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import _, api, models, fields
+from odoo.exceptions import ValidationError
+from odoo.addons import decimal_precision as dp
+
+
+class StockPickingSplit(models.TransientModel):
+    _name = 'stock.picking.split'
+
+    picking_id = fields.Many2one(
+        'stock.picking',
+        string='Picking',
+        required=True,
+        default=lambda s: s.env.context.get('active_id', False)
+    )
+    line_ids = fields.Many2many(
+        'stock.picking.split.line',
+        'split_id',
+        string='Split Operations',
+    )
+
+    @api.model
+    def create(self, vals):
+        rec = super(StockPickingSplit, self).create(vals)
+        rec.compute_lines()
+        return rec
+
+    @api.multi
+    def process(self):
+        self.ensure_one()
+        picking = self.picking_id
+        if picking.pack_operation_ids:
+            reserve = True
+            picking.do_unreserve()
+        split_picking = picking.copy({
+            'name': '/',
+            'move_lines': [],
+            'pack_operation_ids': [],
+            # 'backorder_id': picking.id
+            }
+        )
+        product_split = {}
+        for line in self.line_ids:
+            product_split[line.product_id] = line.split_qty
+        new_move_ids = []
+        for move in picking.move_lines:
+            if move.state in ('cancel', 'done'):
+                continue
+            to_split = min(product_split[move.product_id],
+                           move.product_uom_qty)
+            if to_split > 0:
+                product_split[move.product_id] -= to_split
+                new_move_ids.append(move.split(to_split))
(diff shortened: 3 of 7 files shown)


stock_split_picking/models/__init__.py
```diff
@@ -3,3 +3,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import stock_picking
+from . import stock.picking.split
```
grindtildeath: Move the import and file to `/wizard`

stock_split_picking/models/stock_picking.py
```diff
@@ -14,6 +14,22 @@ class StockPicking(models.Model):
...
+                    'view_mode': 'form',
+                    'res_model': 'stock.picking.split',
+                    'views': [(view.id, 'form')],
+                    'view_id': view.id,
+                    'target': 'new',
+                    'res_id': wiz.id,
+                    'context': self.env.context,
+            }
+
         for pick in self:
```
grindtildeath: This code (after the return) is probably to remove.

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+    line_ids = fields.Many2many(
+        'stock.picking.split.line',
+        'split_id',
+        string='Split Operations',
+    )
+
+    @api.model
+    def create(self, vals):
+        rec = super(StockPickingSplit, self).create(vals)
+        rec.compute_lines()
```
grindtildeath: IMO calling this `generate_lines` would avoid possible confusion.

stock_split_picking/models/__init__.py
sebalix: Should not this TransientModel be in the `./wizards` subfolder?

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+        product_qty = {}
+        product_availability = {}
+        for move in self.picking_id.move_lines:
+            if move.state in ('cancel', 'done'):
+                continue
+            product = move.product_id
+            if product not in product_qty:
+                product_qty[product] = move.product_uom_qty
+                product_availability[product] = move.reserved_availability
+            product_qty[product] += move.product_uom_qty
```
grindtildeath: Isn't `else` missing here ? Otherwise the qties will be double their value if `product not in product_qty`

stock_split_picking/models/stock_picking.py
```diff
@@ -14,6 +14,22 @@ class StockPicking(models.Model):
...
+                    'name': _('Split Picking'),
+                    'type': 'ir.actions.act_window',
+                    'view_type': 'form',
+                    'view_mode': 'form',
+                    'res_model': 'stock.picking.split',
+                    'views': [(view.id, 'form')],
+                    'view_id': view.id,
+                    'target': 'new',
+                    'res_id': wiz.id,
+                    'context': self.env.context,
```
sebalix, line 29: Does it really work? I encountered a problem yesterday with Odoo 10.0 if we force the user's context: `active_id` can be fooled.

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+            product_availability[product] += move.reserved_availability
+        for product, qty in product_qty.items():
+            vals = {'picking_id': self.picking_id.id,
+                    'product_qty': qty,
+                    'split_qty': product_availability[product],
+                    'product_id': product.id,
+                    'product_uom_id': product.uom_id.id,
```
grindtildeath, line 93: Can't the stock move use another uom than the product defined one ? I'm not sure about this...

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+            (picking.name)
+        )
+        split_picking.action_confirm()
+        if reserve:
+            picking.action_assign()
+            split_picking.action_assign()
+        return True
+
+    def compute_lines(self):
+        line_ids = self.env['stock.picking.split.line']
```
grindtildeath, line 75: Renaming this var `created_lines` or something similar would be cleaner.

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+class StockPickingSplit(models.TransientModel):
+    _name = 'stock.picking.split'
+
+    picking_id = fields.Many2one(
+        'stock.picking',
+        string='Picking',
+        required=True,
+        default=lambda s: s.env.context.get('active_id', False)
+    )
+    line_ids = fields.Many2many(
```
grindtildeath: Shouldn't this be O2m ?

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @api.constrains('product_qty', 'split_qty')
+    def _check_qty(self):
+        for rec in self:
+            if rec.split_qty > rec.product_qty:
+                raise ValidationError(
+                    _('Split quantity for product %s must be less than %.3f') %
+                    (rec.product_id.name, rec.product_qty)
+                )
+            if rec.split_qty < 0:
```
grindtildeath, line 129: Shouldn't 0 be allowed if I don't want to split some products ? Or, if the user is supposed to delete the line, maybe adding a some help text would make sense.

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+        for line in self.line_ids:
+            product_split[line.product_id] = line.split_qty
+        new_move_ids = []
+        for move in picking.move_lines:
+            if move.state in ('cancel', 'done'):
+                continue
+            to_split = min(product_split[move.product_id],
+                           move.product_uom_qty)
+            if to_split > 0:
+                product_split[move.product_id] -= to_split
```
grindtildeath, line 54: I don't really understand what you're doing here. :confused:

grindtildeath (changes requested):
The improvement looks quite good overall, but here are a few things to check.

stock_split_picking/models/stock_picking_split.py
gurneyalex: you are right, this needs a smarter approach.

stock_split_picking/models/stock_picking_split.py
gurneyalex: I missed this in the refactoring (it was originally a m2m on stock.pack.operations) Good catch

stock_split_picking/models/stock_picking_split.py
gurneyalex: 0 is allowed (it's "<" not "<="). 
And removing lines is forbidden in the view definition.

stock_split_picking/models/stock_picking.py
gurneyalex: It does work quite well (I did test before submitting the PR).

stock_split_picking/models/stock_picking_split.py
gurneyalex: This is to handle the case where you have several stock.move with the same product. We summed the qtys for all these moves in the wizard, and now we need to spread back the qtys among the different moves with the same product.

stock_split_picking/models/stock_picking_split.py
grindtildeath: My bad, I have been confused by the error message then.

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,131 @@
...
+    def _check_qty(self):
+        for rec in self:
+            if rec.split_qty > rec.product_qty:
+                raise ValidationError(
+                    _('Split quantity for product %s must be less than %.3f') %
+                    (rec.product_id.name, rec.product_qty)
+                )
+            if rec.split_qty < 0:
+                raise ValidationError(
+                    _('Split quantity for product %s must be greater than 0') %
```
grindtildeath:
```suggestion
                    _('Split quantity for product %s must be greater than or equal to 0') %
```

stock_split_picking/models/stock_picking_split.py
grindtildeath: Would be great if you can state it in a code comment :wink:

grindtildeath (commented):
Seems like this PR broke the tests. :(
Please also move stock_picking_split.py to `/wizard` and we should be good. :)

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,133 @@
...
+        self.ensure_one()
+        picking = self.picking_id
+        if picking.pack_operation_ids:
+            reserve = True
+            picking.do_unreserve()
+        split_picking = picking.copy({
+            'name': '/',
+            'move_lines': [],
+            'pack_operation_ids': [],
+            # 'backorder_id': picking.id
```
rousseldenis, line 41: Please remove comment

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,133 @@
...
+        product_split = {}
+        for line in self.line_ids:
+            product_split[line.product_id] = line.split_qty
+        new_move_ids = []
+        for move in picking.move_lines:
+            if move.state in ('cancel', 'done'):
+                continue
+            to_split = min(product_split[move.product_id],
+                           move.product_uom_qty)
+            if to_split > 0:
```
rousseldenis, line 53: Shouldn't float_compare be used ?

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+    split_qty = fields.Float(
+        'Split', default=0.0,
+        digits=dp.get_precision('Product Unit of Measure'), required=True
+    )
+
+    @api.constrains('product_qty', 'split_qty')
+    def _check_qty(self):
+        for rec in self:
+            if rec.split_qty > rec.product_qty:
```
rousseldenis, line 124: Please use float_compare as they use dp

stock_split_picking/models/stock_picking_split.py
```diff
@@ -0,0 +1,133 @@
...
+                raise ValidationError(
+                    _('Split quantity for product %s must be less than %.3f') %
+                    (rec.product_id.name, rec.product_qty)
+                )
+            if rec.split_qty < 0:
```
rousseldenis, line 129: float_compare

rousseldenis (changes requested):
Code review. Minor changes
Shouldn't be tests to highlight changes ?

PR #472, opened 2018-11-20
[10.0][FIX] stock_picking_package_preparation_line: Do not delete lines when the picking is not removed

SimoRubi:
Before this PR:
If the picking in the current stock_picking_package_preparation is edited (not removed), the `line_ids` of the current stock_picking_package_preparation get deleted

After this PR:
If the picking in the current stock_picking_package_preparation is edited (not removed), the `line_ids` of the current stock_picking_package_preparation _do not_ get deleted

### DIFF
--- a/stock_picking_package_preparation_line/__manifest__.py
+++ b/stock_picking_package_preparation_line/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     'name': 'Stock Picking Package Preparation Line',
-    'version': '10.0.1.0.2',
+    'version': '10.0.1.0.3',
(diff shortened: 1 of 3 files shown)


eLBati (approved):
Thanks.
Reminder: v11

pedrobaeza:
Can you add a test for the new case?

SimoRubi:
@pedrobaeza seems legit, done in https://github.com/OCA/stock-logistics-workflow/pull/472/commits/08705cd819c5ab3330fdf80ea713e240255b0cae

pedrobaeza (approved):

PR #475, opened 2018-11-22
[11.0][IMP] stock_batch_picking: Make move_line_ids writable

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/stock_batch_picking/models/stock_batch_picking.py
+++ b/stock_batch_picking/models/stock_batch_picking.py
@@ -69,28 +69,29 @@ class StockBatchPicking(models.Model):
 
     notes = fields.Text('Notes', help='free form remarks')
 
-    move_lines = fields.One2many(
+    move_lines = fields.Many2many(
         'stock.move',
         readonly=True,
         string='Related stock moves',
         compute='_compute_move_lines'
     )
 
-    move_line_ids = fields.One2many(
+    move_line_ids = fields.Many2many(
         'stock.move.line',
-        readonly=True,
         string='Related pack operations',
-        compute='_compute_move_line_ids'
+        compute='_compute_move_line_ids',
+        # HACK: Allow to write sml fields from this model
+        inverse=lambda self: self,
     )
 
-    entire_package_ids = fields.One2many(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is this working correctly? I see several problems:

* A computed one2many field is doomed to fail in some processes, as `one2many` uses the inverse name for doing certain operations. You should switch to many2many.
* Switching to many2many, when editing, you already have it editable.
* Adding lines is not possible (should be limited), because which picking is the source? This can be bypassed adding the picking field, but it's clumsy.

rousseldenis:
@sergio-teruel Can you attend @pedrobaeza comments ?

pedrobaeza:
@sergio-teruel don't make translations on a PR, but on Weblate. This will conflict for sure, and more as the PR is not going to be merged soon, as the comments are not attended.

rousseldenis:
@sergio-teruel Is it normal that you include several modifications that are not related to this PR? Moreover, they are mixed within same commits. I think about someone who should complain about that :smile:

stock_batch_picking/models/stock_batch_picking.py
```diff
@@ -78,9 +78,9 @@ class StockBatchPicking(models.Model):
 
     move_line_ids = fields.One2many(
         'stock.move.line',
-        readonly=True,
         string='Related pack operations',
-        compute='_compute_move_line_ids'
+        compute='_compute_move_line_ids',
+        inverse=lambda self: self,
```
rousseldenis, line 84: @sergio-teruel Why this ?

pedrobaeza:
@rousseldenis I have been reviewing with @sergio-teruel this PR, and first of all, the rest of unrelated modifications are now on separate PRs. Can you please review them?

About this one, the question is that you can't edit any of the `stock.move.line` fields. For example, "Done" quantity field, so with this little hack, we allow that. We are not able to find a better solution (tried with inverse over the fields theirselves, with regular inverse method, etc...).

What has been done now is that the computed fields are many2many, not one2many.

pedrobaeza (approved):

rousseldenis (approved):
Code review

PR #476, opened 2018-11-24
[MIG]Migrate stock_picking_customer_ref

bizzappdev:
Migrate stock_picking_customer_ref from v11 to v12
#457


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

bizzappdev:
sorry for my lack of knowledge. commit history has been added

stock_picking_customer_ref/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Usage
+=====
+
+Edit the client reference in a sale order with pickings. It will show up in
+every picking associated.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/154/11.0
```
pedrobaeza: Change this or use the new README by fragments system.

stock_picking_customer_ref/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
pedrobaeza: Incorrect URL.

pedrobaeza (commented):
Quick review over the README

bizzappdev:
changes are done

rousseldenis:
@pedrobaeza

stock_picking_customer_ref/tests/test_stock_picking_customer_ref.py
```diff
@@ -0,0 +1,47 @@
...
+        cls.partner = cls.env['res.partner'].create({
+            'name': 'Mr. Odoo',
+        })
+        cls.sale_order = cls.env['sale.order'].create({
+            'partner_id': cls.partner.id,
+            'order_line': [(0, 0, {'product_id': cls.product.id})],
+        })
+        cls.sale_order.with_context(show_sale=True).action_confirm()
+
+    def test_stock_picking_customer_ref(self):
```
rousseldenis, line 30: @bizzappdev Even if it is migration, could you test a picking without sale to cover all figures ?

stock_picking_customer_ref/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.depends('group_id', 'sale_id.client_order_ref')
+    def _compute_client_order_ref(self):
+        stock_pickings = self.filtered(
+            lambda stock_picking: stock_picking.group_id)
+        for picking in stock_pickings:
+            cond = [('procurement_group_id', '=', picking.group_id.id)]
+            sale = self.env['sale.order'].search(cond, limit=1)
```
rousseldenis: @bizzappdev The sale_id is now present in stock picking :

https://github.com/odoo/odoo/blob/12.0/addons/sale_stock/models/stock.py#L66

The traversal is direct through a related field.

rousseldenis (changes requested):
Code review

stock_picking_customer_ref/models/stock_picking.py
rousseldenis: @bizzappdev Your changes are not enough. Please use related field instead.

```suggestion
            client_order_ref = fields.Char(
                string="Sale Reference/Description",
                related="sale_id.client_order_ref",
                store=True)
```

bizzappdev: Yes indeed it should be like this.

stock_picking_customer_ref/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright 2017 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Stock Picking Customer Ref',
+    'version': '12.0.1.0.0',
+    'category': 'Warehouse',
+    'author': 'AvanzOSC,'
+              'Tecnativa,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://odoo-community.org/',
```
nikul-serpentcs: `https://github.com/OCA/stock-logistics-workflow`

stock_picking_customer_ref/tests/test_stock_picking_customer_ref.py
```diff
@@ -0,0 +1,69 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super(TestStockPickingCustomerRef, cls).setUpClass()
+        cls.product = cls.env['product.product'].create({
+            'name': 'Test stuff',
+        })
+        cls.partner = cls.env['res.partner'].create({
+            'name': 'Mr. Odoo',
+        })
+        cls.sale_order = cls.env['sale.order'].create({
```
nikul-serpentcs: @bizzappdev some object `cls.env['sale.order']`  use multiple time, So use as like a global variable.

nikul-serpentcs (commented):
Improve Code

nikul-serpentcs:
@bizzappdev  Use new README by fragment system, copying https://github.com/OCA/maintainer-tools/tree/master/template/module/readme and editing or removing the needed sections. You can then run locally oca-gen-addon-readme if you have OCA/maintainer-tools installed and you want to pre-generate the README.rst file (preferred), or keep the previous README.rst file.

stock_picking_customer_ref/models/stock_picking.py
```diff
@@ -11,13 +10,16 @@ class StockPicking(models.Model):
...
+            if not sale:
+                continue
+            picking.client_order_ref = sale.client_order_ref
 
-    client_order_ref = fields.Char(string="Sale Reference/Description",
-                                   compute="_compute_client_order_ref",
-                                   store=True)
+    client_order_ref = fields.Char(
+        string="Sale Reference/Description",
```
Swapnesh-SerpentCS: Better to use `Customer Reference` as a string instead of `Sale Reference/Description` here (and at all other places too).

Ps: SO number will be already there in origin field
screenshot 11

bizzappdev: > Better to use `Customer Reference` as a string instead of `Sale Reference/Description` here (and at all other places too).
> 
> Ps: SO number will be already there in origin field
> screenshot 11

@Swapnesh-SerpentCS  I doubt about it, because Source Document has already the value of Sale order, and the purpose of the module is to get the Sale order client reference field into Delivery order

bizzappdev: @elicoidal Can you please suggest what should be the String/caption for the field. as we doubt for the same. as per @Swapnesh-SerpentCS it should be Customer Reference (Same as Sale order) and according to my it should have Sale Customer Reference. as it is related field from Sale order.

stock_picking_customer_ref/tests/test_stock_picking_customer_ref.py
bizzappdev: @nikul-serpentcs IMO it is okay as it has been used only once outside the setup method.

bizzappdev:
> @bizzappdev Use new README by fragment system, copying https://github.com/OCA/maintainer-tools/tree/master/template/module/readme and editing or removing the needed sections. You can then run locally oca-gen-addon-readme if you have OCA/maintainer-tools installed and you want to pre-generate the README.rst file (preferred), or keep the previous README.rst file.

Changes are done

stock_picking_customer_ref/models/stock_picking.py
elicoidal: @bizzappdev difficult decision :smile: 
I would tend to simple: "Customer Reference" to follow the usual logic in Odoo ("Source" is called "Source" everywhere whatever the object).

stock_picking_customer_ref/tests/test_stock_picking_customer_ref.py
nikul-serpentcs: @bizzappdev Yes, I know but IMO if you use as a global variable that good for a feature.

stock_picking_customer_ref/models/stock_picking.py
bizzappdev: @elicoidal  Thank you for your feedback. Changes are done as per you feedback.

nikul-serpentcs:
@bizzappdev squash commits

elicoidal (approved):
don't you need to change the PO files too?

bizzappdev:
PO file changed.

stock_picking_customer_ref/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    client_order_ref = fields.Char(
+        string="Customer Reference",
+        related="sale_id.client_order_ref",
+        store=True)
```
rousseldenis, line 14: @bizzappdev I think this should be readonly to avoid security and writes problems

Swapnesh-SerpentCS: @rousseldenis   Related fields are readonly by default. (https://github.com/odoo/odoo/pull/27236)

rousseldenis: @bizzappdev Yes, forgot we were in v12 :smile:

rousseldenis (approved):
LGTM. could you squash your commits ?

stock_picking_customer_ref/views/stock_picking_view.xml
```diff
@@ -0,0 +1,49 @@
...
+
+        <record model="ir.ui.view" id="view_picking_form_inh_customerinfo">
+            <field name="name">view.picking.form.inh.customerinfo</field>
+            <field name="model">stock.picking</field>
+            <field name="inherit_id" ref="stock.view_picking_form" />
+            <field name="arch" type="xml">
+                <field name="group_id" position="after">
+                    <field name="client_order_ref"
+                           attrs="{'invisible':
+                           [('client_order_ref', '=', False)]}" />
```
Swapnesh-SerpentCS: We should use the reference field `sale_id` here instead of `client_order_ref`
IMO it makes more sense to display field even if it is blank.

What are your views on this? @rousseldenis @elicoidal 

Ps: Haven't checked if sale_id is there on view or not.

rousseldenis: @Swapnesh-SerpentCS You're right. It's better to show the field explicitly because if it's hidden, the user does not know the value - even void - of that field. @bizzappdev

bizzappdev:
squash commits is done and also remove bizzappdev as author.

PR #477, opened 2018-11-24
[MIG]Migrate stock_picking_show_return from v11 to v12

bizzappdev:
Migrate from v11 to v12
#457


bodedra (changes requested):
IMHO You have to respect commit history following migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#technical-method-to-migrate-a-module-from-110-to-120-branch

bizzappdev:
Commit history has been added

stock_picking_show_return/README.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Usage
+=====
+
+#. Go to any picking that has a return
+#. See the new tab "Returns" appears.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/154/11.0
```
bodedra, line 44: Should be `12.0`

stock_picking_show_return/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2014-2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Show returns on stock pickings",
+    "version": "12.0.1.0.0",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "category": "Warehouse Management",
+    "website": "https://www.tecnativa.com",
```
bodedra: Should be `https://github.com/OCA/stock-logistics-workflow`

bodedra (changes requested):
@bizzappdev Thanks for migration. Would you please correct minor changes. Other than it's LGTM

bizzappdev:
suggested change are done

bodedra (approved):
Code Review :+1:

stock_picking_show_return/README.rst
pedrobaeza: You can use README by fragments and generate this automatically: https://github.com/OCA/maintainer-tools/#readme-generator

nikul-serpentcs:
@bizzappdev Use new README by fragment system, copying https://github.com/OCA/maintainer-tools/tree/master/template/module/readme and editing or removing the needed sections. You can then run locally oca-gen-addon-readme if you have OCA/maintainer-tools installed and you want to pre-generate the README.rst file (preferred), or keep the previous README.rst file.

stock_picking_show_return/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2014-2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    returned_ids = fields.Many2many(
```
Swapnesh-SerpentCS, line 10: We should use table name (Eg. `relation='returned_picking_relation'`) explicitly to prevent sharing same table name (In case one have m2m with same object)
Related PR on Github: https://github.com/odoo/odoo/pull/29376 and https://github.com/odoo/odoo/pull/29323

stock_picking_show_return/README.rst
bizzappdev: Done

stock_picking_show_return/__manifest__.py
bizzappdev: Done

stock_picking_show_return/models/stock_picking.py
bizzappdev: @Swapnesh-SerpentCS  As it is non-store field it will not create any database table

stock_picking_show_return/README.rst
```diff
@@ -0,0 +1,91 @@
...
+    :target: https://translation.odoo-community.org/projects/stock-logistics-workflow-12-0-mig-stock_picking_show_return/stock-logistics-workflow-12-0-mig-stock_picking_show_return-stock_picking_show_return
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/154/12.0-mig-stock_picking_show_return
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+=================================
+Display returns on stock pickings
```
nikul-serpentcs: @bizzappdev Remove it, Two time added

stock_picking_show_return/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+=================================
+Display returns on stock pickings
```
nikul-serpentcs: same as

nikul-serpentcs (commented):
Improve Code

stock_picking_show_return/readme/DESCRIPTION.rst
bizzappdev: Done

stock_picking_show_return/tests/test_stock_picking_show_return.py
```diff
@@ -0,0 +1,35 @@
...
+        })
+
+    def test_returned_ids_field(self):
+        self.picking.action_assign()
+        self.picking.move_line_ids.qty_done = 1
+        self.picking.action_done()
+        wizard = self.env['stock.return.picking'].with_context(
+            active_ids=self.picking.ids, active_id=self.picking.id).create({})
+        wizard.create_returns()
+        self.assertTrue(self.picking.returned_ids)
```
nikul-serpentcs, line 35: @bizzappdev IMO here use like `self.assertTrue(bool(self.picking.returned_ids))`, It's better for assert True

nikul-serpentcs:
@bizzappdev squash commits

stock_picking_show_return/tests/test_stock_picking_show_return.py
bizzappdev: [](https://docs.python.org/2/library/unittest.html#unittest.TestCase.debug)
it is checking that automatically by itself. no value added from changing code.

nikul-serpentcs: @bizzappdev yes, but if you checked with `bool` (`assertTrue`) then the result must be bool, that why I asking to only change  like this `bool(self.picking.returned_ids)`

bizzappdev: https://docs.python.org/2/library/unittest.html#unittest.TestCase.debug

Method | Checks that | New in
assertTrue(x) | bool(x) is True

the parameter value can be anything. it will automatically convert it to bool after that only it will check.
check the official documentation for the same.

alexey-pelykh: It seems that https://github.com/python/cpython/blob/44a79cc5b3d1fb0c03c99077aa26def85ec26c67/Lib/unittest/case.py#L727 should answer the question completely, in this case casting to `bool` is not needed, IMHO

nikul-serpentcs: @bizzappdev @alexey-pelykh I think that an empty object will return True So that assertTrue seems wrong, it will always be True.

alexey-pelykh: yep, thus I'd compare `len()` or at least `.ids`

EDIT: my bad here, empty recordset would be treated as `False`

alexey-pelykh: due to converting to bool wouldn't change the issue behavior you're describing

nikul-serpentcs: > yep, thus I'd compare `len()` or at least `.ids`

Yes, but better is `bool` :smile:

bizzappdev: > @bizzappdev @alexey-pelykh I think that an empty object will return True So that assertTrue seems wrong, it will always be True.
Naa empty will not return True 
as mentioned by @alexey-pelykh 
https://github.com/python/cpython/blob/44a79cc5b3d1fb0c03c99077aa26def85ec26c67/Lib/unittest/case.py#L729

it has if condition and if the object is empty it will go in the If condition

alexey-pelykh: https://github.com/odoo/odoo/blob/b234af04c1226bd5d5fb28fae7087b7f130e53d4/odoo/models.py#L4907

`if recordset:` is same as `if bool(recordset):` fyi

pedrobaeza: No need for `bool`, `len` or whatever. ORM recordsets are already evaluated as falsy if no records, and trueish otherwise, so `self.assertTrue(self.picking.returned_ids)` is perfect as is.

nikul-serpentcs: @pedrobaeza Thanks for clarifying my confusion!

pedrobaeza: Thanks also to @alexey-pelykh that was saying the same with links to ORM code :wink:

bizzappdev:
> squash commits

sorry but I am not good in Squashing. might be the merger will do.

pedrobaeza:
You can see the guide for that: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

Tell me if you have any doubt. This can't be done at merge time unless the merger downloads locally the branch, do the squashing and force push over your PR. If no luck, I'll do it.

nikul-serpentcs:
> > squash commits
> 
> sorry but I am not good in Squashing. might be the merger will do.

@bizzappdev  follow this command/steps: 
`git rebase -i HEAD~6`
after open file like this `pick c70a0dc ..`, 
you rename `pick`  to `f`, 
Note: not rename first `pick`.( this one [MIG]stock_picking_show_return: migration from 11 to v12)
after save file and  `git push origin branch_name -f`

bizzappdev:
thanks @pedrobaeza  for pointing doc. I had only doubt that last time when I forced push the PR was automatically closed that is why I was worrying. I have force pushed the commit with squash. Please check
cc @nikul-serpentcs

stock_picking_show_return/README.rst
```diff
@@ -0,0 +1,87 @@
...
+Do not contact contributors directly about support or help with technical issues.
+
+Credits
+=======
+
+Authors
+~~~~~~~
+
+* Tecnativa
+* BizzAppDev
```
rousseldenis: @bizzappdev A migration does not make you author.

stock_picking_show_return/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2014-2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Show returns on stock pickings",
+    "version": "12.0.1.0.0",
+    "author": "Tecnativa, "
+              "BizzAppDev, "
```
rousseldenis: Same here

rousseldenis (changes requested):
Code review. Minor changes

stock_picking_show_return/README.rst
bizzappdev: then will it be co-author? and do we also have to remove Contributors also?

rousseldenis: Nope. 'Just' contributors is ok. Usually, author is for the original writer of the module.

bizzappdev: ahh okay

rousseldenis (approved):

PR #478, opened 2018-11-26
[11.0][FIX] stock_picking_package_preparation_line: Do not delete lines when the picking is not removed

SimoRubi:
Forward port of https://github.com/OCA/stock-logistics-workflow/pull/472

### DIFF
--- a/stock_picking_package_preparation_line/__manifest__.py
+++ b/stock_picking_package_preparation_line/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Stock Picking Package Preparation Line',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'author': 'Apulia Software srl, Odoo Italia Network, '
               'Odoo Community Association (OCA)',
     'license': 'AGPL-3',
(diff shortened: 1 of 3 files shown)


SimoRubi:
@eLBati @pedrobaeza forward porting the PR you approved a few days ago, could you please check?
Thanks!

pedrobaeza (approved):

bodedra (approved):

rousseldenis (approved):
Code review

rousseldenis:
@SimoRubi Even if it is not documented like that in Odoo, the use of inverse attribute in standard fields allows not to override write() method. This allows readability.

SimoRubi:
@rousseldenis do you mean that we might move everything that is in the `write` method in the `inverse` attribute of `picking_ids` field?
Thanks for the advice

rousseldenis:
Not for this one, but yes it could be an alternative to override write() method which is not sometimes readable. If you read the code at https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L3107, I think this is safe.

SimoRubi:
ok I better get what you meant now, anyway I don't think using this _side effect_ of `inverse` field attribute could be said to be safe: it may change in future releases because the use you describe is not its real scope.
Anyway thanks for the insight in Odoo core code :)

PR #480, opened 2018-11-30
[FIX] stock_no_negative: Use should_bypass_reservation

etobella:
It was giving errors when using scrap locations.

### DIFF
--- a/stock_no_negative/__manifest__.py
+++ b/stock_no_negative/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Stock Disallow Negative',
-    'version': '11.0.1.0.1',
+    'version': '11.0.1.0.2',
     'category': 'Inventory, Logistic, Storage',
     'license': 'AGPL-3',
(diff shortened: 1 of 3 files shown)


rousseldenis:
@etobella Could you add a test case for this ?

etobella:
I will create the test, but it is an standard feature of odoo:
https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_move_line.py#L223

rousseldenis:
Ok. But I didn't speak about testing the function itself. But adding a test case to highlight that the no negative flow works with scrap locations now

rousseldenis:
@etobella I think by the way this is weird to install that module and authorize negative quants on scrap locations. In which case do you need this ?

etobella:
In my case, I have some locations where stock could be negative for products.
currently, this module is disallowing to have negative stock on all locations internal locations.
Maybe, the right solution should be creating a new flag for allow_negative on locations

PR #485, opened 2018-12-18
8.0 mig stock picking line sequence

mathiasfrancke:
Hello,
I needed a module from Odoo 10 in Odoo 8. Now i would like to share the back port with the OCA.

Best Regards
Mathias


rousseldenis:
@mathiasfrancke Could you please check travis?

stock_picking_line_sequence/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+{
+    'name': 'Stock picking lines with sequence number',
+    'summary': 'Manages the order of stock moves by displaying its sequence',
+    'version': '8.0.1.0.0',
+    'category': 'Warehouse Management',
+    'author': "Camptocamp, "
+              "Eficent, "
+              "Serpent CS, "
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.camptocamp.com',
```
Nikul-Chaudhary: URL should be `https://github.com/OCA/stock-logistics-workflow`

Nikul-Chaudhary (changes requested):
Minor Change

stock_picking_line_sequence/__openerp__.py
mathiasfrancke: corrected

rousseldenis:
@mathiasfrancke This makes stock_picking_reorder_lines module failing. Could you check ?

mathiasfrancke:
> @mathiasfrancke This makes stock_picking_reorder_lines module failing. Could you check ?

@rousseldenis Thank you - I will have a look later.

mathiasfrancke:
So it seems the issue is related to *stock_picking_reorder_lines* which seems to be like the predecessor of *stock_picking_line_sequence* till version 8. Since Odoo 9.0 there is no more *stock_picking_reorder_lines* but instead *stock_picking_line_sequence*.
I guess the CI fails because moth modules introduce the field *sequence* on *stock.move*. So field gets overwritten and behavior changes and tests fail.
So the question is what to do with this backport?

kmatthes:
@mathiasfrancke What options do you see to handle this and which option do you suggest (and why)?

rousseldenis:
@mathiasfrancke What's the status of this ?

mathiasfrancke:
@rousseldenis I think this migration is obsolete. When I did the backport I did not see the module [stock_picking_reorder_lines](https://github.com/OCA/stock-logistics-workflow/tree/8.0/stock_picking_reorder_lines). My backport conflicts with that module and when I have a look on both readme files I think _stock_picking_reorder_lines_ is already the needed one for Odoo 8.
So what to do with this PR? I guess we can close it.

PR #487, opened 2019-01-04
[stock_picking_deactivate_immediate_transfer] missing act.window

JordiBForgeFlow:
Several stock.picking ir.actions.act_window related records were missing. For example, when you wanted to create a picking after accessing from the buttons in the dashboard it would attempt to create an immediate transfer.

image

### DIFF
--- a/stock_picking_deactivate_immediate_transfer/views/stock_picking_views.xml
+++ b/stock_picking_deactivate_immediate_transfer/views/stock_picking_views.xml
@@ -9,7 +9,71 @@
         }
         </field>
     </record>
+    <record id="stock.action_picking_tree_late" model="ir.actions.act_window">
+        <field name="context">{
+                'search_default_picking_type_id': [active_id],
+                'default_picking_type_id': active_id,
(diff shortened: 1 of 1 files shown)


rousseldenis:
@jbeficent Couldn't reproduce it on runbot. The flow seems correct. Seems weird the modification you propose. Is it all about context ?

JordiBForgeFlow:
@rousseldenis Try to create a delivery order after you press button '0 To Do' in the Dashboard. You'll see that it's proposing an immediate transfer.

rousseldenis:
It's creating a stock picking :
image

rousseldenis:
Wtih the good picking type

image

LoisRForgeFlow:
@rousseldenis The problem is that without this fix, you create a picking that is an immediate transfer, you should add operations to see if  you can use planned qty or done qty in order to test this PR.

LoisRForgeFlow (approved):
Tested on runbot

rousseldenis (approved):
Ok, I see.

PR #496, opened 2019-02-06
[10.0][FIX] stock_picking_backorder_strategy: Return value

### LINKED ISSUE #495: stock_picking_backorder_strategy module fails to return backorder record
@api.multi
    def _create_backorder(self, backorder_moves=None):
        if backorder_moves is None:
            backorder_moves = []
        res = False
        # Do nothing with pickings 'no_create'
        pickings = self.filtered(
            lambda p: p.picking_type_id.backorder_strategy != 'no_create')
        pickings_no_create = self - pickings
        pickings_no_create.mapped('move_lines')._cancel_remaining_quantities()
        backorders = super(StockPicking, pickings)._create_backorder(
            backorder_moves=backorder_moves)
        to_cancel = backorders.filtered(
            lambda b: b.backorder_id.picking_type_id.backorder_strategy ==
            'cancel')
        to_cancel.action_cancel()
        return res

The above method should be returning backorders, not 'res'. So, please rename 'res' variable to 'backorders".

rousseldenis:
Fix #495

### DIFF
--- a/stock_picking_backorder_strategy/__manifest__.py
+++ b/stock_picking_backorder_strategy/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Picking backordering strategies',
-    'version': '10.0.0.1.0',
+    'version': '10.0.0.1.1',
     'author': 'ACSONE SA/NV, Odoo Community Association (OCA)',
     'category': 'Warehouse Management',
     'website': 'http://www.acsone.eu',
--- a/stock_picking_backorder_strategy/models/stock_picking.py
+++ b/stock_picking_backorder_strategy/models/stock_picking.py
@@ -21,7 +21,6 @@ class StockPicking(models.Model):
     def _create_backorder(self, backorder_moves=None):
         if backorder_moves is None:
             backorder_moves = []
-        res = False
         # Do nothing with pickings 'no_create'
         pickings = self.filtered(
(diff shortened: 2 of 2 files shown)


max3903 (approved):

stock_picking_backorder_strategy/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'Picking backordering strategies',
-    'version': '10.0.0.1.0',
+    'version': '10.0.0.1.1',
     'author': 'ACSONE SA/NV, Odoo Community Association (OCA)',
     'category': 'Warehouse Management',
     'website': 'http://www.acsone.eu',
```
Nikul-Chaudhary, line 10: URL should be `https://github.com/OCA/stock-logistics-workflow`

rousseldenis: @Nikul-Chaudhary -bot
This is a fix, the module was in 10.0 before. This could be an improvement.

Nikul-Chaudhary: @rousseldenis In v10  [here](https://github.com/OCA/stock-logistics-workflow/blob/10.0/stock_picking_backorder_strategy/__manifest__.py#L10), it's still `http://www.acsone.eu`

So change here, It's not blocking

PR #498, opened 2019-02-06
[8.0] remove update of stock move when invoicing

sergiocorato:


rousseldenis:
@sergiocorato You should do one PR with #499

PR #502, opened 2019-02-19
[10.0][FIX] not show an error if picking is already assigned, it is a…

sergiocorato:
… subsequent state, do not transfer already done pickings, add cancel in state filtered

from #411


tbaden (approved):
code review

rousseldenis (approved):
LGTM

PR #503, opened 2019-02-19
[11.0][FIX] product_expiry_simple: delete duplicate view

enriquemartin:


pedrobaeza:
Can you point where is duplicated?

enriquemartin:
@pedrobaeza sorry! On stock_picking.xml line 13. Same inherited view adding same field but with less attributes.

pedrobaeza (approved):
We can fast-track this as it's pretty straight-forward

PR #509, opened 2019-03-11
[10.0] product_expiry_simple: Replace stored display_name by name_get()

alexis-via:
Replace stored display_name by name_get: make things simpler and avoid troubles when you need to modify the behavior of display_name in another module.

### DIFF
--- a/product_expiry_simple/__manifest__.py
+++ b/product_expiry_simple/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Product Expiry Simple',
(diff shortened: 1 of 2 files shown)


rousseldenis:
@alexis-via Isn't this a too big change on 10.0 branch ? Some people could have overriden 'compute_display_name_field' method.

At least, change module version

rousseldenis (approved):
Code review.

I think this is a fast-tracking one

LoisRForgeFlow (approved):
Yes, much better with this approach

PR #512, opened 2019-03-15
[10.0][FIX][stock_picking_package_preparation] Fix check_tracking() c…

robyf70:
…all to use move instead of moves.

This PR fix a call to `check_tracking()` by using `move` instead `moves`. However we could use straight `moves.check_tracking()` and remove totally the `for`. What do you think?


bodedra (approved):
Code review :+1:

pedrobaeza (approved):

PR #515, opened 2019-04-04
[12.0][MIG] stock_picking_restrict_cancel_with_orig_move

kittiu:
Migrate to v12

PR #518, opened 2019-04-09
[12.0][MIG] mrp_stock_picking_restrict_cancel

kittiu:
WIP

Dependencies
- [x] stock_picking_restrict_cancel_with_orig_move https://github.com/OCA/stock-logistics-workflow/pull/515


kittiu:
Pending to add test script.

rousseldenis:
@kittiu

kittiu:
@rousseldenis OK I will rebase and push again.
For the test, I have been trying hard, but I can't find way to replicate by manual yet (I am not much familiar with WH module).
If you could guide my test case in user step, I will be happy to write the script :)

kittiu:
Note:
So far, I found that the test error on module **stock_picking_package_preparation_line** (when self.preparation.action_put_in_pack() is called)
But only when **mrp_stock_picking_restrict_cancel** is installed along side

PR #523, opened 2019-04-16
[IMP] stock_split_picking: rewrite

sebalix:
Supersedes #471 (adressing review comments)

- allow to split pickings with no available stock
- fix crash when processing a picking which was split twice
- fix mixed UoMs in stock moves
- do not allow to split partially a move if it has an ancestor depending
  on it (the ancestor move is limited to one destination move).
- improve/fix tests


stock_split_picking/models/stock_picking.py
```diff
@@ -1,9 +1,8 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2015 Camptocamp SA - Nicolas Bessi
+# Copyright 2013-2018 Camptocamp SA
```
Tonow-c2c:
```suggestion
# Copyright 2013-2019 Camptocamp SA
```

stock_split_picking/tests/test_stock_split_picking.py
```diff
@@ -3,66 +3,163 @@
...
-from odoo.exceptions import UserError
 
 
 class TestStockSplitPicking(SavepointCase):
 
     @classmethod
     def setUpClass(cls):
         super(TestStockSplitPicking, cls).setUpClass()
-
         cls.src_location = cls.env.ref('stock.stock_location_stock')
```
Tonow-c2c, line 17:
```suggestion
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
         cls.src_location = cls.env.ref('stock.stock_location_stock')
```

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,188 @@
+# -*- coding: utf-8 -*-
+# Copyright 2013-2018 Camptocamp SA - Alexandre Fayolle
```
Tonow-c2c:
```suggestion
# Copyright 2013-2019 Camptocamp SA - Alexandre Fayolle
```

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,188 @@
...
+        self.line_ids = line_ids
+
+
+class StockPickingSplitLine(models.TransientModel):
+    _name = 'stock.picking.split.line'
+
+    split_id = fields.Many2one('stock.picking.split',
+                               ondelete='cascade')
+    product_id = fields.Many2one(
+        'product.product', 'Product', ondelete="cascade", readonly=True
```
Tonow-c2c, line 170: the next time please one variable by line + variable_name=variable,

stock_split_picking/views/stock_partial_picking.xml
```diff
@@ -8,7 +8,7 @@
         <field name="arch" type="xml">
             <field name ="state" position="before">
                 <button name="split_process"
-                        states="draft,assigned,partially_available"
+                        states="draft,assigned,waiting,confirmed,partially_available"
```
i-vyshnevska: no need how in draft, as it trying to do unreserve it

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,188 @@
...
+        # Handle the case where there are several stock.move with the same
+        # product+UoM. We summed the qtys for all these moves in the wizard,
+        # and now we need to spread back the qtys among the different moves
+        # with the same product+UoM.
+        product_split = {}
+        for line in self.line_ids:
+            key = (line.product_id, line.product_uom_id, line.full_split)
+            product_split[key] = line.split_qty
+        new_move_ids = []
+        for move in picking.move_lines:
```
i-vyshnevska, line 90: before wizard creation we union lines per product and uom, and here we also should use same approach 
we need exclude lines with same product and uom if they more that on (otherwise you will have lines with 0 qty)

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,188 @@
...
+                continue
+            reserved_availability = move.product_id.uom_id._compute_quantity(
+                move.reserved_availability, move.product_uom)
+            # key = (Product, UoM, Full split)
+            key = (move.product_id, move.product_uom, bool(move.move_orig_ids))
+            if key not in product_qty:
+                product_qty[key] = move.product_uom_qty
+                product_availability[key] = reserved_availability
+            else:
+                product_qty[key] += move.product_uom_qty
```
i-vyshnevska, line 145: we can't  do it with products which tracked on serial numbers or lots

i-vyshnevska (changes requested):
one missing point when it's using route  there is great feature to send lot and serial numbers in between of pickings, so user have already filed picking with all data and need just check

stock_split_picking/views/stock_partial_picking.xml
sebalix: but we also want to be able to split a draft picking, even if there is nothing to unreserve, or I misunderstood your comment?

i-vyshnevska: ok so do  check before, as when it's in draft and trying to unreserve error appeared

stock_split_picking/wizards/stock_picking_split.py
sebalix: You mean we can get 2 moves corresponding with the same key? I think it is handled in the lines after. In the initial PR there is a comment from @gurneyalex explaining this: https://github.com/OCA/stock-logistics-workflow/pull/471#discussion_r234927811

stock_split_picking/wizards/stock_picking_split.py
sebalix: I updated the PR to add the `stock.move.restrict_lot_id` to the key.

stock_split_picking/wizards/stock_picking_split.py
i-vyshnevska: yes, I saw it and works while you have simple product, but when you track it by lots error raised or from  181 or 193 line
I think maybe we not divide them but split one by one, for instance if you have 2 lines with same product per 10 unit each, and split them in 7 and 13 one pick will have one line with 7 unit and other with 10 and 3, in this case it will be easier to send lots or s/n if we know them
and while written it,  realized we can't split product which tracking by sn on float number, only integers allowed

sebalix: I tested and it is working "well", we are losing the lots/serial number which were already assigned (pack operation removed by the unreserve step), and the reserved ones in the new picking may be different. I don't know if it is an issue.

Good catch about the fact we should not be able to split a move with floating numbers if its product is tracked by serial number, I'll add a check on this.

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        split_picking.message_post(
+            body=_("Split from picking <em>%s</em> <b>created</b>.") %
+            (picking.name)
+        )
+        split_picking.action_confirm()
+        if reserve:
+            picking.action_assign()
+            split_picking.action_assign()
+        return split_picking.id
```
i-vyshnevska:
```suggestion
        return {
            'name': _('Splited Picking'),
            'type': 'ir.actions.act_window',
            'view_type': 'form',
            'view_mode': 'form',
            'res_model': 'stock.picking',
            'target': 'new',
            'res_id': split_picking.id,
            'context': self.env.context,
        }
```

i-vyshnevska (approved):
now works as expected, 
we can return action to user with new picking it's more obviously 
and can mention in  todo to handle lots and sn in case of routes

Tonow-c2c (approved):
LGTM

rousseldenis:
@sebalix Could you check Travis error ?

sebalix:
@rousseldenis Yep I tried, and I don't know why tests are running well on OCB but not on Odoo...
I relaunched the job on both, maybe a new commit on odoo/odoo bringing a regression which were not yet synced in OCA/OCB...? (it seems big, in fact I really don't know, I'm waiting for CI).

nikul-serpentcs:
@sebalix Squash Commit's

sebalix:
@rousseldenis @nikul-serpentcs now tests should be OK (I added flags to run them at post install) + commits squashed.

rousseldenis:
@sebalix So, can close #471 and merge ?

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class StockPickingSplit(models.TransientModel):
+    _name = 'stock.picking.split'
+
+    picking_id = fields.Many2one(
+        'stock.picking',
+        string='Picking',
+        required=True,
+        default=lambda s: s.env.context.get('active_id', False)
```
rousseldenis: Maybe delete='cascade', but non blocking

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,215 @@
...
+        for line in self.line_ids:
+            key = (line.product_id, line.product_uom_id,
+                   line.restrict_lot_id, line.full_split)
+            product_split[key] = line.split_qty
+        new_move_ids = []
+        for move in picking.move_lines:
+            if move.state in ('cancel', 'done'):
+                continue
+            # key = (Product, UoM, Lot, Full split)
+            key = (move.product_id, move.product_uom,
```
rousseldenis: Another improvement should be (Roadmap ?) putting code to get that key in a method to easy inheritance.

stock_split_picking/wizards/stock_picking_split.py
```diff
@@ -0,0 +1,215 @@
...
+        line_ids = self.env['stock.picking.split.line']
+        product_qty = {}
+        product_availability = {}
+        for move in self.picking_id.move_lines:
+            if move.state in ('cancel', 'done'):
+                continue
+            reserved_availability = move.product_id.uom_id._compute_quantity(
+                move.reserved_availability, move.product_uom)
+            # key = (Product, UoM, Lot, Full split)
+            key = (move.product_id, move.product_uom,
```
rousseldenis: Same as above, that could avoid duplicate code

rousseldenis (approved):
Minor comments. Not blocking

rousseldenis:
Other improvement should be increasing coverage testing constrains.

sebalix:
@rousseldenis yes #471 can be closed.

PR #527, opened 2019-05-14
[IMP] stock_picking_mass_action: Fix and improve transfer method

carlosdauden:
@Tecnativa

### DIFF
--- a/stock_picking_mass_action/wizard/mass_action.py
+++ b/stock_picking_mass_action/wizard/mass_action.py
@@ -1,5 +1,6 @@
 # Copyright 2014 Camptocamp SA - Guewen Baconnier
 # Copyright 2018 Tecnativa - Vicent Cubells
+# Copyright 2019 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import fields, api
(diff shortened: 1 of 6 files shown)


stock_picking_mass_action/wizard/mass_action.py
```diff
@@ -56,7 +56,7 @@ def _default_picking_ids(self):
     picking_ids = fields.Many2many(
         string='Pickings',
         comodel_name="stock.picking",
-        default=lambda self: self._default_picking_ids(),
+        default=_default_picking_ids,
```
rousseldenis: @carlosdauden Why changing lambda to non inheritable method ?

rousseldenis:
@carlosdauden Could you add test as you change behaviour of this ?

IMHO these kind of changes should not be done on already merged module as the behaviour is totally different

pedrobaeza:
@rousseldenis in this case it's because the module doesn't work at all.

rousseldenis:
> @rousseldenis in this case it's because the module doesn't work at all.

Indeed, that's a good reason :smile:

stock_picking_mass_action/wizard/mass_action.py
carlosdauden: I thought I had read that it was not necessary anymore, but surely I am wrong because I just tried it and still need.

Changes done.

rousseldenis:
> @carlosdauden Could you add test as you change behaviour of this ?

Don't forget tests, that can help validating the process

carlosdauden:
@rousseldenis I pussed tests just now.
Thanks.

chienandalu (approved):
Works as expected :+1:

rousseldenis (approved):
Code review

PR #529, opened 2019-06-04
[10.0][FIX] stock_no_negative: show product name_get instead of name

Abranes:
Use `product.name_get` instead of `product.name`

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -33,5 +33,5 @@ class StockQuant(models.Model):
                     "stock level of the product '%s'%s would become negative "
                     "(%s) on the stock location '%s' and negative stock is "
                     "not allowed for this product.") % (
-                        quant.product_id.name, msg_add, quant.qty,
+                        quant.product_id.display_name, msg_add, quant.qty,
                         quant.location_id.complete_name))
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

stock_no_negative/models/stock_quant.py
```diff
@@ -33,5 +33,5 @@ def check_negative_qty(self):
                     "stock level of the product '%s'%s would become negative "
                     "(%s) on the stock location '%s' and negative stock is "
                     "not allowed for this product.") % (
-                        quant.product_id.name, msg_add, quant.qty,
+                        quant.product_id.name_get()[0][1], msg_add, quant.qty,
```
rousseldenis:
```suggestion
                        quant.product_id.display_name, msg_add, quant.qty,
```

I think this is more correct

rousseldenis (changes requested):

rousseldenis:
@Abranes Don't forget to upgrade module version

stock_no_negative/models/stock_quant.py
Abranes: Done!

rousseldenis (approved):
Fast-tracking one

PR #533, opened 2019-06-07
[12.0][ADD] stock_picking_product_kit_helper

kittiu:
When sales order contain product kits (product with BOM of type kit),
the delivery order (stock.move) created by it will be exploded to multiple product lines.

Normally, to partially deliver, user will calculate manually the quantity of each product lines to delivery.

This module add new tab "Product Kit Help" to help in calculate quantity in product line with ease.


kittiu:
There is error on addons **stock_picking_package_preparation_line**, and it is error only after module **mrp** is installed (and this module use bom kit).

kittiu:
@eLBati Do you want to take a look first?

stock_picking_product_kit_helper/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Kitti U. - Ecosoft <kittiu@ecosoft.co.th>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Stock Picking Product Kit Helper',
+    'summary': 'Set quanity in picking line based on product kit quantity',
+    'version': '12.0.0.0.0',
```
rousseldenis, line 6:
```suggestion
    'version': '12.0.1.0.0',
```

stock_picking_product_kit_helper/models/stock_picking.py
```diff
@@ -0,0 +1,134 @@
...
+        comodel_name='stock.picking.product.kit.helper',
+        string='Product Kit Helper Lines',
+        inverse_name='picking_id',
+        readonly=False,
+        states={'done': [('readonly', True)], 'cancel': [('readonly', True)]},
+    )
+    has_product_kit = fields.Boolean(
+        string='Has Product Kit',
+        compute='_compute_has_product_kit',
+        default=False,
```
rousseldenis, line 21: No need of this

stock_picking_product_kit_helper/models/stock_picking.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @api.model
+    def _is_product_kit(self, product, company):
+        BOM = self.env['mrp.bom'].sudo()
+        bom = BOM._bom_find(product=product,
+                            company_id=company.id)
+        return bom and bom.type == 'phantom'
+
+    @api.one
+    def _compute_has_product_kit(self):
```
rousseldenis, line 33:
```suggestion
    @api.multi
    def _compute_has_product_kit(self):
    for picking in self:
        if any(self._is_product_kit(line.product_id, line.company_id) for line in picking.move_lines.mapped('sale_line_id')):
            picking.has_product_kit = True
```

kittiu:
@rousseldenis or this module should be in manufacturing repo as it use both mrp and purchase. WDYT?

rousseldenis (commented):
> @rousseldenis or this module should be in manufacturing repo as it use both mrp and purchase. WDYT?

There is no really rule when modules are quite on the border of two repos. Some would have put them in one, some in another.

You put here, we can go on

rousseldenis:
@kittiu After thought about it, I think indeed it should be better in manufacture repo as you will pull mrp dependencies. Are you OK?

kittiu:
> @kittiu After thought about it, I think indeed it should be better in manufacture repo as you will pull mrp dependencies. Are you OK?

I am OK, I will fix per comments then put it there.

eLBati:
> There is error on addons stock_picking_package_preparation_line, and it is error only after module mrp is installed

Can you describe the error by the steps to reproduce it?

kittiu:
If mrp is also installed. When run test of stock_picking_package_preparation_line, it will fail.

PR #536, opened 2019-06-14
[12.0] mig stock_picking_filter_lot

IsabelEsparza:


HaraldPanten (approved):
Functional review on runbot 👍

stock_picking_filter_lot/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2018 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock picking filter lot",
+    "summary": "In picking out lots' selection, "
+               "filter lots based on their location",
+    "version": "12.0.1.0.0",
+    "category": "Warehouse",
+    "website": "https://github.com/OCA/stock-logistics-workflow/tree/"
```
tbaden: website should be: https://github.com/OCA/stock-logistics-workflow

tbaden (changes requested):
thanks for the work 👍 
I did a code review: LGTM, just a minor change.

stock_picking_filter_lot/__manifest__.py
IsabelEsparza: I just changed it.

alan196 (approved):
Functional Review! 👍

stock_picking_filter_lot/views/stock_picking_view.xml
```diff
@@ -9,7 +9,7 @@
         <field name="model">stock.picking</field>
         <field name="inherit_id" ref="stock.view_picking_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='move_line_ids']/tree/field[@name='lot_id']" position="attributes">
+            <xpath expr="//field[@name='move_line_ids_without_package']/tree/field[@name='lot_id']" position="attributes">
```
LoisRForgeFlow, line 12: why this change?

LoisRForgeFlow (commented):
Functioanl test :+1: 

Comment inline:

stock_picking_filter_lot/views/stock_picking_view.xml
tbaden: @lreficent move_line_ids no longer exists this way in the record. It got changed to move_line_ids_without_package

tbaden (approved):

stock_picking_filter_lot/views/stock_picking_view.xml
LoisRForgeFlow: Oh, I see, you are right. Thanks.

damendieta:
Hi, I found a (kind of) bug in this PR.

As this module computes the filter based on quant location, there is a problem when you have a quant that has been transfered and it's quantity is 0.

I have tried a modification with this changes:

```python
    @api.depends('quant_ids', 'quant_ids.location_id', 'quant_ids.quantity')
    def _compute_location_ids(self):
        for lot in self:
            lot.location_ids = lot.mapped('quant_ids') \
                .filtered(lambda q: q.quantity > 0).mapped('location_id') \
                .filtered(lambda l: l.usage == 'internal')
```

It gives a more real filter, even excludes not internal locations (to avoid vendor and customer locations).

I don't know if a new PR is worth it, but I leave this suggestion here if someone needs it.

LoisRForgeFlow:
@damendieta Thanks for noticing, I think it is worth, so please open a PR with the proposed fix.

PR #542, opened 2019-06-26
[FIX] stock_batch_picking: Error when default code is False

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):

rousseldenis (approved):
Code review. Makes sense

PR #546, opened 2019-07-01
[FIX] stock_pack_operation_auto_fill: Avoid auto fill qty_done when an extra_move has been created

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/stock_pack_operation_auto_fill/models/stock_move.py
+++ b/stock_pack_operation_auto_fill/models/stock_move.py
@@ -8,13 +8,27 @@ from odoo import models
 class StockMove(models.Model):
     _inherit = 'stock.move'
 
+    def _create_extra_move(self):
+        """
+        When user set on stock move line done a quantity greater than initial
+        demand Odoo creates an extra stock move with the difference and it is
+        posible to create an extra stock move line with qty_done = 0 which will
+        be deleted in _action_done method.
+        This method set a context variable to prevent set qty_done for these
+        cases.
+        """
+        my_self = self
+        if self.picking_id.auto_fill_operation:
+            my_self = self.with_context(skip_auto_fill_operation=True)
+        return super(StockMove, my_self)._create_extra_move()
+
     def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
         """Auto-assign as done the quantity proposed for the lots"""
-        self.ensure_one()
         res = super(StockMove, self)._prepare_move_line_vals(
             quantity, reserved_quant,
         )
-        if not self.picking_id.auto_fill_operation:
+        if (self.env.context.get('skip_auto_fill_operation') or
+                not self.picking_id.auto_fill_operation):
             return res
         elif (self.picking_id.picking_type_id.avoid_lot_assignment and
               res.get('lot_id')):
--- a/stock_pack_operation_auto_fill/tests/test_stock_picking_auto_fill.py
+++ b/stock_pack_operation_auto_fill/tests/test_stock_picking_auto_fill.py
@@ -322,3 +322,38 @@ class TestStockPicking(TransactionCase):
         # The expected result is only opertions with product_id set and
         self.assertFalse(product_8_op.qty_done)
         self.assertTrue(product_9_op.qty_done)
+
+    def test_receipt_more_qty_than_demand(self):
+        # Active auto fill operations although they have assigned lots
+        self.picking_type_in.write({
+            'auto_fill_operation': True,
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Please fix tests according this change

stock_pack_operation_auto_fill/models/stock_move.py
```diff
@@ -8,12 +8,25 @@
...
+        be deleted in _action_done method.
+        This method set a context variable to prevent set qty_done for these
+        cases.
+        """
+        my_self = self.with_context(skip_auto_fill_operation=True)
+        return super(StockMove, my_self)._create_extra_move()
+
     def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
         """Auto-assign as done the quantity proposed for the lots"""
-        self.ensure_one()
```
pedrobaeza, line 13: No need to remove this

stock_pack_operation_auto_fill/models/stock_move.py
```diff
@@ -8,12 +8,25 @@
...
         res = super(StockMove, self)._prepare_move_line_vals(
             quantity, reserved_quant,
         )
+        if self.env.context.get('skip_auto_fill_operation', False):
```
pedrobaeza: No need of the `, False` part

stock_pack_operation_auto_fill/models/stock_move.py
```diff
@@ -8,12 +8,25 @@
...
+            return res
         if not self.picking_id.auto_fill_operation:
```
pedrobaeza: Join both conditions in a single `if`

pedrobaeza (changes requested):

stock_pack_operation_auto_fill/models/stock_move.py
```diff
@@ -8,12 +8,25 @@
...
+    def _create_extra_move(self):
+        """
+        When user set on stock move line done a quantity greater than initial
+        demand Odoo creates an extra stock move with the difference and it is
+        posible to create an extra stock move line with qty_done = 0 which will
+        be deleted in _action_done method.
+        This method set a context variable to prevent set qty_done for these
+        cases.
+        """
+        my_self = self.with_context(skip_auto_fill_operation=True)
```
rousseldenis: You should change context if move.picking_id.auto_fill_operation is True. Without that, you will create new environnment at each extra move creation(even for 'normal' flow).

stock_pack_operation_auto_fill/models/stock_move.py
sergio-teruel: Yes, I know, but the method already has this check.
https://github.com/odoo/odoo/blob/963475bcfd75fe0a35401c7343d14937aa7f0d74/addons/stock/models/stock_move.py#L810

sergio-teruel:
@rousseldenis @pedrobaeza Changes done

stock_pack_operation_auto_fill/models/stock_move.py
pedrobaeza: The call is not for doing the check, but for raising the error on a meaningful line, at the beginning on the method, instead of a weird line where it's more difficult to know which one should be a singleton. Removing this line here, the error will raise in the super, and thus maybe you are not aware that the problem is in the previous call. Just this case it's not critical, but for you to know why the call.

stock_pack_operation_auto_fill/models/stock_move.py
```diff
@@ -8,13 +8,27 @@
...
+        return super(StockMove, my_self)._create_extra_move()
+
     def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
         """Auto-assign as done the quantity proposed for the lots"""
-        self.ensure_one()
         res = super(StockMove, self)._prepare_move_line_vals(
             quantity, reserved_quant,
         )
-        if not self.picking_id.auto_fill_operation:
+        if (self.env.context.get('skip_auto_fill_operation') and
```
pedrobaeza: s/and/or

pedrobaeza (changes requested):

rafaelbn:
@pedrobaeza , @sergio-teruel made changes to :+1:

pedrobaeza (approved):

pedrobaeza:
But he never notified us...

Let's merge this for having the patch.

/ocabot merge patch

rafaelbn (approved):

pedrobaeza:
Is this needed in 12.0, @sergio-teruel ?

PR #555, opened 2019-08-09
[MIG][WIP] stock_batch_picking: Migration to 12.0 

carlosdauden:


pedrobaeza:
@carlosdauden both CIs are red.

i-vyshnevska:
@pedrobaeza do you know what base advantage in front https://github.com/odoo/odoo/tree/12.0/addons/stock_picking_batch seems the base work relies on same wizards?

pedrobaeza:
Until v11, the advantages of this module were:

* Unified report for showing you all the product you have to pick together, not picking by picking. Now it seems v12 includes a new report, but not sure if it's enough.
* Possibility of doing operations in batch. Again it seems Odoo has now changed something about this in core.

So maybe we can avoid this migration, but further analysis is needed. There's another question around: people have based custom layers over this module, and adapt them to new module can be more work than simply migrating it.

gurneyalex:
@pedrobaeza  @i-vyshnevska  the OCA module has numerous advantages over the standard Odoo module in 12.0:

* the Odoo module works by changing the state of all the pickings in the batch when the batch state changes (and not handing important cases such as production lots, back orders), whereas the OCA module computes the state of the batch from the state of the pickings in the batch
* the reporting in the OCA module is more flexible (esp. grouping wise) than the one in Odoo 


The data models are very close but the behavior is different enough to motivate having both versions available. 

What could be nice is having a convergence of the data models: 
* make the OCA module depend on the Odoo module (probably renaming the module to something like stock_picking_batch_extended)
* rename the model and fields in the OCA module
* add a field on the modet stock picking batch model which acts as a switch between the Odoo behavior and the OCA behavior. This could be set as a default through a settings screen in the Inventory application
* not quite sure how to handle the data migration, but maybe @pedrobaeza  can suggest something

This would enable having e.g. the reporting from the OCA module on top of the Odoo module, but keeping the canonical behavior for people who like it better.

pedrobaeza:
Yes, we can handle data migration, but we can't rename tables, as you may have Odoo module installed on previous version, so a copy is needed instead. Are you going to handle then the conversion of the module?

carlosdauden:
This PR is included in #570

PR #562, opened 2019-08-29
[12.0][MIG] stock_picking_purchase_order_link

jaumeplanas:


AdriaGForgeFlow (approved):
Code review :+1:

HaraldPanten (approved):
Functional review. Works as expected 👍

pedrobaeza:
Please squash migration commits together

stock_picking_purchase_order_link/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+        Link between picking and purchase order""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Eficent Business and IT Consulting Services S.L.,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/stock-logistics-workflow',
+    "application": False,
+    "installable": True,
+    'depends': [
+        'purchase'
```
rousseldenis: You have to add purchase_stock as that module was added in v12.

Your tests are passing because of https://github.com/OCA/stock-logistics-workflow/pull/562/files#diff-dd273efb1c4dfddc20425a7b87ce6478R9

That's why I don't like tests on post_install...

stock_picking_purchase_order_link/tests/test_stock_picking_purchase_order_link.py
```diff
@@ -0,0 +1,54 @@
...
+                    'name': self.product.name,
+                    'product_id': self.product.id,
+                    'product_qty': 1.0,
+                    'product_uom': self.product.uom_po_id.id,
+                    'price_unit': 10.0,
+                    'date_planned': date.today(),
+                })],
+        })
+        purchase_order.button_confirm()
+        self.picking = self.stock_picking_obj.search(
```
rousseldenis: What is the aime of this test function as nothing is tested ???

rousseldenis (changes requested):
Code review

stock_picking_purchase_order_link/__manifest__.py
pedrobaeza: @rousseldenis that doesn't work anymore on v12, so that's not the reason why they are passing.

pedrobaeza: You can check in https://travis-ci.org/OCA/stock-logistics-workflow/jobs/578365850#L1575 that they are been run immediately after the module loading. There's a warning about an invalid field (vendor), but the test logic works with already loaded modules (maybe by luck).

rousseldenis: What do you mean ???

Tests are ok : https://travis-ci.org/OCA/stock-logistics-workflow/jobs/578365850#L1585

rousseldenis: > the test logic works with already loaded modules (maybe by luck).

That's the point ! That's why I requested dependency on purchase_stock

pedrobaeza: I mean that they are not executed at the end, as that variables are no longer valid (you can do it, but through tags).

pedrobaeza: > That's the point ! That's why I requested dependency on purchase_stock

Yes, I'm not debating on that. Simply pointing that the reason is not by the `post_install = True` variable (it's useless on v12).

rousseldenis: > (it's useless on v12).

Ok, was not aware. But it could be the problem.

jaumeplanas: So I undestand that requested chages are remove `post_install = True` in tests file (it's useless), add a dependency on `purchase_stock` and squah all v12 migration commits together.

pedrobaeza: Yes, I think so

rousseldenis (approved):
Code review

PR #563, opened 2019-08-30
[FIX] stock_lock_lot: from picking product_id from production default_product_id

oihane:
when a lot is created from the production wizard the context has 'default_product_id' but when it is created from the transferring wizard of pickings the context has 'product_id' so to have both options, I've done this change. Not sure if it is correct.


oihane:
cc @disna-sistemas @daniel-CA @rousseldenis

Daniel-CA (approved):

disna-sistemas (approved):

PR #564, opened 2019-09-05
[12.0][ADD] stock_dropshipping_whole_supplier

bealdav:
When at least one product is handled with dropshipping in a sale order,
all products, of the order, provided with the same vendor are also automatically dropshipped.

It allows to reduce delivery operations and cost.

This behavior is configured by supplier, with a boolean field.

Screenshot from 2019-09-06 13-19-40

Please @kittiu @rousseldenis could you have a look

## TODO

- [x] unit test


bealdav:
@benoitguillot @florian-dacosta DONE, review welcome

kittiu:
@bealdav I do manual test and found 2 points.
1) User need to be in group multi currency to see the "Overall Dropshipping"?
image

2) It not works for me,
I have installed the module, then, I try with 2 existing product "FURN_6666", "E-COM11" both has supplier "Wood Corner". Only "FURN_6666" check "Dropship", not "E-COM11".
But still when confirmed SO, it still create DO for E-COM11. Only if E-COM11 check "Dropship", then no DO is created from SO.

Anything can be possible wrong for me?

stock_dropshipping_whole_supplier/models/sale.py
```diff
@@ -0,0 +1,67 @@
...
+    def _set_dropship_route_on_adhoc_cases(self):
+        """ Assign a dropship route to sale lines which have
+            the same vendor than dropship product
+        """
+        self.ensure_one()
+        vendor_dropsh, vendor_line = self._get_dropship_info()
+        for vendor, line in vendor_dropsh.items():
+            lines_to_dropshipping = vendor_line.get(vendor)
+            if lines_to_dropshipping:
+                lines_to_dropshipping.write(
```
rousseldenis: I'd prefer here building a recordset and then write the value(as it is the same). It will reduces sql UPDATEs

rousseldenis (changes requested):
Code review. Minor comment

bealdav:
Thanks @kittiu I'll fix it.
Any suggestion for a better module name ? cc @rafaelbn

stock_dropshipping_whole_supplier/models/sale.py
```diff
@@ -0,0 +1,67 @@
...
+                {partner_obj.id: sale_line_obj}
+        """
+        self.ensure_one()
+        dropsh_route = self.env.ref(
+            "stock_dropshipping.route_drop_shipping")
+        line_obj = self.env["sale.order.line"]
+        vendor_dropsh = {}
+        vendor_line = {}
+        for line in self.order_line:
+            vendor = line.product_id._select_seller().name
```
bealdav: @kittiu the bug comes from I omit to specify `quantity=line.product_uom_qty` 
in  `_select_seller` method

bealdav: Done

stock_dropshipping_whole_supplier/models/sale.py
bealdav: Done

bealdav:
@kittiu I implemented your case as unit test

HaraldPanten (approved):
Functional review 👍 seems to be OK

kittiu:
Functional Tested :+1:

kittiu (approved):

bealdav:
@rousseldenis requested changes are done.
Thanks

bealdav:
I don't know why CI is triggered with this kind of commit ? 
How to save the planet ?

rousseldenis (approved):
Code review

rousseldenis:
@bealdav Could you squash your commits (even if that will trigger CI job - but will make oca bot happy :smile: ) ?

bealdav:
AFAIK fixup! squash automatically, so i don't know which commit merge ?

rousseldenis:
> AFAIK fixup! squash automatically, so i don't know which commit merge ?

I think this is not fixed yet in bot merge, see https://github.com/OCA/oca-github-bot/issues/57

pedrobaeza:
And now that the strategy is to do a merge operation since https://github.com/OCA/oca-github-bot/pull/70, I would say that even less possible to be done.

rousseldenis:
> And now that the strategy is to do a merge operation since [OCA/oca-github-bot#70](https://github.com/OCA/oca-github-bot/pull/70), I would say that even less possible to be done.

So, is the autosquash strategy the default one ?

pedrobaeza:
> So, is the autosquash strategy the default one ?

No, mostly the contrary: a merge is now done over main branch. You would need a rebase with auto squash for doing that. I think https://github.com/OCA/oca-github-bot/pull/71 is trying to address this.

rousseldenis:
Ok, so @bealdav you need to squash you fixup commits.

bealdav:
please @rousseldenis maybe you can merge ? Thanks

PR #535, opened 2019-06-12
[12.0][MIG] stock_batch_ picking

AdriaGForgeFlow:
Standard migration of module stock_batch_picking to v12

cc ~ @Eficent

PR #570, opened 2019-09-23
[12.0][MIG]stock_batch_picking: Migration to 12.0

i-vyshnevska:
migrate addon plus refactoring to become compatible with odoo as advised in https://github.com/OCA/stock-logistics-workflow/pull/555#issuecomment-524357888


stock_picking_batch_oca/data/batch_picking_actions_server.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Tecnativa - Carlos Dauden
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="0">
```
nikul-serpentcs: @i-vyshnevska IMO Here no need to `noupdate="0"`

carlosdauden:
Sorry, but all commits after 11/19/2018 are missing https://github.com/OCA/stock-logistics-workflow/pull/555/commits.
I based my PR #555 on the PR #535 that was outdated.
At the beginning of the month I realized that and that there were things in the PR that I didn't quite like so I did the complete migration since v11

carlosdauden:
Great job, by the way :1st_place_medal:

stock_picking_batch_oca/data/batch_picking_actions_server.xml
i-vyshnevska: I didn't add it, I think it more acceptable to left such record with noupdate

i-vyshnevska:
@carlosdauden  I rebased on your latest work, and tests finally pass

rousseldenis:
@i-vyshnevska Could you see runbot warnings (it's about _description field in a model and duplicate string value in fields)

stock_picking_batch_oca/views/stock_batch_picking.xml
```diff
@@ -0,0 +1,193 @@
...
+              parent='stock.menu_stock_root'
+              id='menu_action_stoch_batch_picking'
+              sequence='3'/>
+
+
+    <!-- Remove odoo menuitem -->
+    <function model="ir.ui.menu" name="write">
+        <value model="ir.ui.menu" eval="ref('stock_picking_batch.stock_picking_batch_menu')"/>
+        <value eval="{
+            'parent_id': False
```
carlosdauden: This change don't hide menu, if menu option hasn't parent It's showed in root menu:
Selección_341

carlosdauden (changes requested):

stock_picking_batch_oca/views/stock_batch_picking.xml
```diff
@@ -0,0 +1,193 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- stock.picking.batch form view -->
+    <record model="ir.ui.view" id="stock_batch_picking_form">
+        <field name="name">stock.picking.batch.form</field>
+        <field name="model">stock.picking.batch</field>
```
carlosdauden:
```suggestion
        <field name="model">stock.picking.batch</field>
        <field name="priority" eval="10"/>
```
With default priority the below method returns stock_picking_batch.stock_picking_batch_form:
https://github.com/i-vyshnevska/stock-logistics-workflow/blob/fdd8ba5cbdaa33cc587317d027830105128b7eae/stock_picking_batch_oca/wizard/batch_picking_creator.py#L158

carlosdauden (changes requested):

stock_picking_batch_oca/views/stock_batch_picking.xml
i-vyshnevska: yes you are correct, I updated this

carlosdauden (approved):

carlosdauden:
Good job @i-vyshnevska

pedrobaeza (changes requested):
Doing a quick review, there are several comments:

1. Module name. It can't be called `stock_picking_batch_oca`. We put this kind of name when we are providing an alternative implementation of other module that has the same name. Examples: `account_bank_statement_import_camt_oca`, `account_budget_oca`. But not for extending an existing community one. `stock_picking_batch_extension` is not a good name either, as it doesn't provide information on what is extending, but if no other option, it's better than the other. Please indicate the features that this finally includes for trying to suggest a name.
2. The README must change to reflect the new conditions.
3. There aren't migration scripts to adapt data from v11.
4. Maintainers must be GitHub login names, that's to say, persons, not companies.

i-vyshnevska:
@pedrobaeza  I am not familiar with migration staff, the rest is ok to change

pedrobaeza:
OK, but can you answer about the features now present in the module?

Also for me doing the migration scripts, I need to know a bit the changes you have made in the data layout. Can you summarize then for not having to analyze deeply the code?

stock_picking_batch_oca/models/stock_batch_picking.py
```diff
@@ -0,0 +1,243 @@
...
+
+    # added state to be compatible with picking_ids
+    state = fields.Selection(
+        selection_add=[
+            ('assigned', 'Available'),
+        ],
+        string='State',
+        readonly=True, index=True,
+        help='the state of the batch picking. '
+             'Workflow is draft -> in_progress -> done or cancel',
```
sebalix: what about 'assigned/Available'?

stock_picking_batch_oca/models/stock_batch_picking.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class StockBatchPicking(models.Model):
+    """ This object allow to manage multiple stock.picking at the same time.
+    """
+    # renamed stock.batch.picking -> stock.picking.batch
+    _inherit = 'stock.picking.batch'
+
+    name = fields.Char(
+        string='Name',
```
sebalix: Should we overwrite the standard user's label? Original is 'Batch Picking Name' https://github.com/odoo/odoo/blob/12.0/addons/stock_picking_batch/models/stock_picking_batch.py#L15

stock_picking_batch_oca/models/stock_batch_picking.py
```diff
@@ -0,0 +1,243 @@
...
+    entire_package_detail_ids = fields.Many2many(
+        comodel_name='stock.quant.package',
+        compute='_compute_entire_package_ids',
+        help='Those are the entire packages of a picking shown in the view of '
+             'detailed operations',
+    )
+
+    @api.depends('picking_ids')
+    def _compute_move_lines(self):
+        for batch in self:
```
sebalix: Should we prevent to compute fields if `use_oca_batch_validation` is not set?

stock_picking_batch_oca/models/stock_batch_picking.py
```diff
@@ -0,0 +1,243 @@
...
+            if len(self) == 1:
+                message = _('This Batch has no pickings')
+            else:
+                message = _('These Batches have no pickings')
+
+            raise UserError(message)
+
+        return self.filtered(lambda b: len(b.picking_ids) != 0)
+
+    def verify_state(self, expected_state=None):
```
sebalix: As this method is called from standard picking methods (action_cancel, action_done...) it should not update related batches if `use_oca_batch_validation` is not set.

i-vyshnevska:
> OK, but can you answer about the features now present in the module?
> 
> Also for me doing the migration scripts, I need to know a bit the changes you have made in the data layout. Can you summarize then for not having to analyze deeply the code?

now it's possible to choose between odoo and oca batch validation  they are not conflicting (to be added more detailed in README)

for data changes 
table stock.batch.picking doesn't exist anymore, all data should be moved to stock.picking.batch
for model stock.batch.picking:
   - picker_id changed to user_id
for model stock.picking:
  - batch_picking_id -> batch_id

if any missed, changed fields have old_name attribute

stock_picking_batch_oca/views/stock_batch_picking.xml
```diff
@@ -0,0 +1,185 @@
...
+        </field>
+    </record>
+
+    <record id="action_stock_batch_picking_tree"
+            model="ir.actions.act_window">
+        <field name="name">Batch Picking</field>
+        <field name="res_model">stock.picking.batch</field>
+        <field name="view_mode">tree,form</field>
+         <field name="view_id" ref="stock_batch_picking_tree"/>
+            <field name="view_ids" eval="[(5, 0, 0),
```
sebalix: Do we really need to remove existing related views here?
Also, not mandatory, but I think creating `act_window.view` separately is better to overload them easily (example: https://github.com/odoo/odoo/blob/8f4b98e67fe49516c0c830524de351db67c58357/addons/hr_timesheet/report/hr_timesheet_report_view.xml#L21) to change the sequence for instance.

sebalix (changes requested):

stock_picking_batch_oca/models/stock_batch_picking.py
i-vyshnevska: verify_state called from action_cancel and odoo cancel method called cancel_picking, methods from odoo and this addon just different they are not inherited, they accessible only via form view which hide buttons in dependency of use_oca_batch_validation

stock_picking_batch_oca/models/stock_batch_picking.py
i-vyshnevska: in progress = assigned 
one used for odoo, other for oca, doc updated

stock_picking_batch_oca/models/stock_batch_picking.py
i-vyshnevska: no harm from it, but updated

stock_picking_batch_oca/views/stock_batch_picking.xml
i-vyshnevska: not removing now

tschanzt (approved):
looks good to me

pedrobaeza:
Please squash migration commits together and it's good to go.

stock_picking_batch_extended/models/stock_picking.py
```diff
@@ -0,0 +1,75 @@
...
+            if pick.state != 'assigned':
+                pick.action_assign()
+                if pick.state != 'assigned':
+                    continue
+
+            if force_qty:
+                for pack in pick.move_line_ids:
+                    pack.qty_done = pack.product_uom_qty
+            else:
+                if all(pack.qty_done == 0 for pack in pick.move_line_ids):
```
rousseldenis: Please use float_is_zero here.

stock_picking_batch_extended/__manifest__.py
```diff
@@ -0,0 +1,37 @@
...
+    'version': '12.0.1.0.0',
+    'author': "Camptocamp, "
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    'development_status': 'Mature',
+    'maintainers': [
+        'Camptocamp',
+        'Tecnativa',
+    ],
+    'category': 'Inventory',
```
rousseldenis: Set to 'Warehouse Management' as it won't appear in good category on OCA shop

stock_picking_batch_extended/report/batch_report.py
```diff
@@ -0,0 +1,89 @@
...
+    def key_level_0(self, operation):
+        return operation.location_id.id, operation.location_dest_id.id
+
+    @api.model
+    def key_level_1(self, operation):
+        return operation.product_id.id
+
+    @api.model
+    def new_level_0(self, operation):
+        level_0_name = u'{} \u21E8 {}'.format(
```
rousseldenis: 'u' is not needed anymore in python3

stock_picking_batch_extended/report/batch_report.py
```diff
@@ -0,0 +1,89 @@
...
+            'location': operation.location_id,
+            'location_dest': operation.location_dest_id,
+            'l1_items': {},
+        }
+
+    @api.model
+    def new_level_1(self, operation):
+        return {
+            'product': operation.product_id,
+            'product_qty': operation.product_qty or operation.qty_done,
```
rousseldenis: I would have written :

```suggestion
            'product_qty': not float_is_zero(operation.product_qty) or operation.qty_done,
```

stock_picking_batch_extended/report/batch_report.py
```diff
@@ -0,0 +1,89 @@
...
+            'operations': operation,
+        }
+
+    @api.model
+    def update_level_1(self, group_dict, operation):
+        group_dict['product_qty'] += (
+            operation.product_qty or operation.qty_done)
```
rousseldenis: Same as above

stock_picking_batch_extended/wizard/batch_picking_creator.py
```diff
@@ -0,0 +1,182 @@
...
+        help='Date on which the batch picking is to be processed'
+    )
+
+    user_id = fields.Many2one(
+        'res.users', string='Picker',
+        oldname='picker_id',
+        default=lambda self: self._default_user_id(),
+        help='The user to which the pickings are assigned',
+    )
+    notes = fields.Text('Notes', help='free form remarks')
```
rousseldenis: s/free/Free

stock_picking_batch_extended/wizard/batch_picking_creator.py
```diff
@@ -0,0 +1,182 @@
...
+    _name = 'stock.picking.batch.creator.group.field'
+    _description = 'Batch Picking Creator Group Field'
+    _order = 'sequence, id'
+
+    batch_picking_creator_id = fields.Many2one(
+        comodel_name='stock.picking.batch.creator',
+        ondelete="cascade",
+        required=True,
+    )
+    sequence = fields.Integer(help='Group by picking field')
```
rousseldenis: Maybe a default (nitpicking)

rousseldenis (changes requested):
Code review. LGTM. Some required changes (on float values). 

Is it normal that 'stock_picking_mass_action' module change is included here?

i-vyshnevska:
@rousseldenis  your comments taken into account, 
fix for mass_action here because this addon had an appropriate test to catch it, AFAIN we can have commits for different addons in one pr while they are separate

rousseldenis:
> @rousseldenis your comments taken into account,
> fix for mass_action here because this addon had an appropriate test to catch it, AFAIN we can have commits for different addons in one pr while they are separate

For a better module version management, could you update it on mass_action in this? So, we can merge it.

i-vyshnevska:
done, @rousseldenis can you check before rebasing?

rousseldenis (approved):
LGTM. @i-vyshnevska As the modification on mass module is more a patch, maybe you could change just the last version part, but nitpicking.

pedrobaeza (approved):

pedrobaeza:
Let's move with this: /ocabot merge

pedrobaeza:
I will create migration scripts in a later step.

stock_picking_batch_extended/report/batch_report.py
carlosdauden: I don't understand this...

pedrobaeza: Do you mean @rousseldenis's comment or the original code?

i-vyshnevska: the change supposes to say that 
if product_qty = 0.00000001 and precision is set 0.001, product_qty is equal to 0, 
if use just (not product_qty) than 0.00000001 will be considered as not 0

PR #576, opened 2019-10-14
[IMP] stock_picking_invoice_link: Allow invoice lines linked with more than one stock move

sergio-teruel:
cc @Tecnativa TT20017

### DIFF
--- a/stock_picking_invoice_link/i18n/es.po
+++ b/stock_picking_invoice_link/i18n/es.po
@@ -9,27 +9,38 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-12-18 03:48+0000\n"
-"PO-Revision-Date: 2017-12-18 03:48+0000\n"
+"POT-Creation-Date: 2019-10-15 08:30+0000\n"
+"PO-Revision-Date: 2019-10-15 10:31+0200\n"
 "Last-Translator: Pedro M. Baeza <pedro.baeza@gmail.com>, 2017\n"
 "Language-Team: Spanish (https://www.transifex.com/oca/teams/23907/es/)\n"
 "Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
-"Content-Transfer-Encoding: \n"
+"Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Generator: Poedit 2.0.6\n"
+
+#. module: stock_picking_invoice_link
+#: model:ir.ui.view,arch_db:stock_picking_invoice_link.view_picking_form
+msgid "<span class=\"o_stat_text\">Invoices</span>"
+msgstr "<span class=\"o_stat_text\">Facturas</span>"
 
 #. module: stock_picking_invoice_link
 #: model:ir.model,name:stock_picking_invoice_link.model_account_invoice
-#: model:ir.ui.view,arch_db:stock_picking_invoice_link.view_picking_form
 msgid "Invoice"
 msgstr "Factura"
 
 #. module: stock_picking_invoice_link
 #: model:ir.model,name:stock_picking_invoice_link.model_account_invoice_line
-#: model:ir.model.fields,field_description:stock_picking_invoice_link.field_stock_move_invoice_line_id
+#: model:ir.model.fields,field_description:stock_picking_invoice_link.field_stock_move_invoice_line_ids
 msgid "Invoice Line"
 msgstr "Línea de factura"
+
+#. module: stock_picking_invoice_link
+#: model:ir.ui.view,arch_db:stock_picking_invoice_link.view_move_form
+#: model:ir.ui.view,arch_db:stock_picking_invoice_link.view_move_picking_form
+msgid "Invoice Lines"
+msgstr "Líneas de factura"
 
 #. module: stock_picking_invoice_link
 #: model:ir.model.fields,field_description:stock_picking_invoice_link.field_stock_picking_invoice_ids
@@ -85,9 +96,10 @@ msgid "Transfer"
 msgstr "Transferir"
 
 #. module: stock_picking_invoice_link
-#: model:ir.ui.view,arch_db:stock_picking_invoice_link.view_picking_form
-msgid "View Invoice"
-msgstr "Ver factura"
+#: code:addons/stock_picking_invoice_link/models/stock_move.py:34
+#, python-format
+msgid "You can not modify an invoiced stock move"
+msgstr "No puede modificar un movimiento de almacén facturado"
 
-#~ msgid "Invoice Lines"
-#~ msgstr "Líneas de factura"
+#~ msgid "View Invoice"
+#~ msgstr "Ver factura"
--- a/stock_picking_invoice_link/migrations/11.0.2.0.0/post-migrate.py
+++ b/stock_picking_invoice_link/migrations/11.0.2.0.0/post-migrate.py
@@ -1,0 +1,9 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.m2o_to_x2m(env.cr, env['stock.move'], 'stock_move',
+                           'invoice_line_ids', 'invoice_line_id')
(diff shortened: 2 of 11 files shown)


stock_picking_invoice_link/i18n/es.po
```diff
@@ -89,5 +90,11 @@ msgstr "Transferir"
 msgid "View Invoice"
 msgstr "Ver factura"
 
+#. module: stock_picking_invoice_link
+#: code:addons/stock_picking_invoice_link/models/stock_move.py:28
+#, python-format
+msgid "You can not modify an invoiced stock move"
+msgstr "No puedes modificar un movimiento de almacén facturado"
```
pedrobaeza: Always "trato de usted"

```suggestion
msgstr "No puede modificar un movimiento de almacén facturado"
```

stock_picking_invoice_link/models/account_invoice.py
```diff
@@ -19,16 +19,28 @@ class AccountInvoice(models.Model):
              "(only when the invoice has been generated from a sale order).",
     )
 
+    @api.model
+    def _refund_cleanup_lines(self, lines):
+        result = super(AccountInvoice, self)._refund_cleanup_lines(lines)
+        if self.env.context.get('mode') == 'modify':
+            for i, line in enumerate(lines):
```
pedrobaeza, line 26: Isn't there any `_prepare_refund_line` or similar for not using this weak link through line index?

stock_picking_invoice_link/models/stock_move.py
```diff
@@ -3,15 +3,27 @@
...
         string='Invoice Line',
         copy=False,
         readonly=True,
     )
+
+    def write(self, vals):
+        if 'product_uom_qty' not in vals:
+            return super().write(vals)
+        for move in self:
+            if move.state == 'done' and move.invoice_line_ids:
```
pedrobaeza: This is difficult to be read, having also 2 calls to super. Please use 3 levels of indentation this time.

stock_picking_invoice_link/models/stock_move.py
```diff
@@ -3,15 +3,27 @@
...
         comodel_name='account.invoice.line',
+        relation='stock_move_invoice_line_rel',
+        column1='move_id',
+        column2='invoice_line_id',
         string='Invoice Line',
         copy=False,
         readonly=True,
     )
+
+    def write(self, vals):
```
pedrobaeza, line 23: Please a docstring explaining why you are preventing this.

stock_picking_invoice_link/views/stock_view.xml
```diff
@@ -13,12 +13,13 @@
...
-                    groups="account.group_account_invoice"
-                    attrs="{'invisible':[('invoice_ids', '=', False)]}"
-                    widget="many2many_tags"
-                    context="{'tree_view_ref': 'account.invoice_tree', 'form_view_ref': 'account.invoice_form'}"/>
+            <xpath expr="//notebook" position="inside">
+                <page name="invoice" string="Invoices" attrs="{'invisible':[('invoice_ids', '=', [])]}">
+                    <field name="invoice_ids" string="Invoice"
+                        groups="account.group_account_invoice"
+                        widget="one2many_list"
+                        context="{'tree_view_ref': 'account.invoice_tree', 'form_view_ref': 'account.invoice_form'}"/>
```
pedrobaeza: This is not valid when `stock_picking_link_purchase` finally arrives.

pedrobaeza: And why putting this on an extra tab and invoice lines in the same form?

pedrobaeza (changes requested):

stock_picking_invoice_link/models/account_invoice.py
sergio-teruel: I think that this ok

stock_picking_invoice_link/views/stock_view.xml
sergio-teruel: > This is not valid when stock_picking_link_purchase finally arrives.

Ohh yea...
I thin that I can remove this and make a new smart button... but when we will have two invoice types... What view display??

> And why putting this on an extra tab and invoice lines in the same form?

I do not understand

sergio-teruel:
@pedrobaeza some changes done!

stock_picking_invoice_link/views/stock_view.xml
pedrobaeza: Yeah, a smart-button with a method that checks linked invoice type and return the corresponding form/tree is a good option.

stock_picking_invoice_link/migrations/11.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    sql = """
```
pedrobaeza: There's an openupgradelib method for avoiding you to write this SQL: https://github.com/OCA/openupgradelib/blob/884331200591e098ed1a0963934660bcef8075c0/openupgradelib/openupgrade.py#L1235

stock_picking_invoice_link/models/stock_move.py
pedrobaeza: Reading the current docstring, shouldn't it block `qty_done` instead of `product_uom_qty`?

sergio-teruel: for a picking in done state when user modifies the units... the field product_uom_qty is writed too.

sergio-teruel:
Changes done

pedrobaeza (approved):

carlosdauden (approved):

gustavovalverde:
This broke everything on production. This PR is using `openupgradelib` which is not in this repo requirements.

cc: @pedrobaeza

pedrobaeza:
@gustavovalverde it shouldn't be in requirements. It's only for upgrading, not for using the module in a fresh installation. And it doesn't break anything: just revert to previous commit, or have openupgradelib in your system.

gustavovalverde:
Pulling this changes in **running environment** does not allow the users to use pickings or invoices because of the missing fields `invoice_line_id` and `move_line_ids` which were changed from One2many to Many2one.

The views will complain as this fields are missing. And trying to upgrade the module will crash because this migration depends on `openupgradelib` which is not a requirement on this repo.

This library should be added in the requirements, as anyone who updates this source code will probably go through the same.

pedrobaeza:
I don't agree and this has been discussed other times. Any way, just add the library and update again. I don't know why you don't have it already: it's something basic in OCA.

gustavovalverde:
As a good practice, we're not supposed to add libraries "just in case" in production. And I'm not aware of any OCA documentation that implies this library must be in all installations which have an OCA repo in it.

pedrobaeza:
You can contribute with this reference in the place you consider. As said, it's already discussed in other places why it's not put in external dependencies (Odoo doesn't init at all, etc), so I don't want to keep this discussion again. Take note from now on to put openupgradelib everywhere and you won't have this problem again.

gustavovalverde:
I'll add that note here: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

pedrobaeza:
I don't think that place is the right one, as they are guidelines for development, not any kind of deployment guide.

andres-pcg:
Alright @pedrobaeza, I understood the whole thing here. You have made your point and it's clear; but, where should we have that note added to? Do you have any suggestions? I know and we will for sure make a note of this. But i think there should be something shared with whole community.

pedrobaeza:
For me, we should add openupgradelib to https://github.com/OCA/OCB/blob/12.0/requirements.txt, so that we stated that in "OCA world", that library is used with no explicit dependency declaration, the same as the rest of the libraries here, and we should add it in the source form for getting latest version:
```
git+https://github.com/OCA/openupgradelib.git@master
```

rousseldenis:
> @gustavovalverde it shouldn't be in requirements. It's only for upgrading, not for using the module in a fresh installation. And it doesn't break anything: just revert to previous commit, or have openupgradelib in your system.

Hi @pedrobaeza. The good practice is to put it in external dependencies in manifest. If you manage your project with setuptools-odoo, it will integrate it.

But as you said, the good practice is to include it in deployments. @gustavovalverde You should have discovered that error during your integration tests.

pedrobaeza:
@rousseldenis we had this discussion already in the past. The question is that the module doesn't need that library for a fresh install, only for migrating.

rousseldenis:
> @rousseldenis we had this discussion already in the past. The question is that the module doesn't need that library for a fresh install, only for migrating.

Ok, but I like explicit declarations. The migration code is part of the module, so...

pedrobaeza:
And then I talked to you about the module avoiding to init the Odoo service when introducing the external dependency without no clear clue in the logging, which it's worst, and that's why we decided this is the less bad.

PR #582, opened 2019-10-29
[ADD] product_cost_price_avco_sync: New module to sync product cost with stock move price unit

sergio-teruel:
cc @Tecnativa TT19351


sergio-teruel:
@rafaelbn @chienandalu  Can you review?

product_cost_price_avco_sync/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class PurchaseOrderLine(models.Model):
```
pedrobaeza: This part shouldn't be in this module, but on a module in OCA/purchase-workflow called `purchase_stock_price_unit_sync`

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+        return res
+
+    def price_unit_update(self):
+        procesed_moves = []
+        for move in self.sorted('date'):
+            if (move.id in procesed_moves or move.state != 'done' or
+                    not move._is_in() or
+                    move.product_id.cost_method != 'average'):
+                continue
+            # not take in account move's quantities
```
pedrobaeza: don't take into account move quantities

product_cost_price_avco_sync/models/product.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def get_cost_history_date(self, date):
+        return self.env['product.price.history'].search([
```
pedrobaeza: `self.ensure_one()`

product_cost_price_avco_sync/models/product.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def get_cost_history_date(self, date):
```
pedrobaeza: The table contains datetime, so you should pass a datetime argument and call the variable according it. Maybe `dt` for avoiding collapsing with library names.

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+            previous_qty = move.product_id.with_context(
+                to_date=move.date
```
pedrobaeza: It's worth to mention here that this field contains the date where the move is done for avoiding future questions

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+                ('date', '>', move.date),
+            ], order='date')
+            for af_move in affected_moves:
+                if af_move._is_in():
+                    if af_move.inventory_id:
+                        af_move.with_context(
+                            skip_cost_update=True).price_unit = price
+                    # Avoid reprocess move if included in self.write
+                    elif af_move.id in procesed_moves:
+                        procesed_moves.append(af_move.id)
```
pedrobaeza: This is never going to be reached. If you want to not add several times the same move, use a set instead.

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+
+from odoo import models
+from odoo.tools.float_utils import float_compare
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    def write(self, vals):
+        res = super(StockMove, self).write(vals)
```
pedrobaeza, line 15: Please get used to add docstrings to the methods you touch, specially overwritings. In this case, I would highlight that you can call for the whole recordset the method without the need to call the update sequentially.

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+                    previous_qty -= af_move.quantity_done
+                    if previous_qty < 0.0:
+                        previous_qty = 0.0
+            if float_compare(move.product_id.standard_price, price,
+                             precision_rounding=move.product_uom.rounding):
+                # Write the standard price, as SUPERUSER_ID because a warehouse
+                # manager may not have the right to write on products
+                move.product_id.with_context(
+                    force_company=move.company_id.id,
+                ).sudo().write({'standard_price': price})
```
pedrobaeza, line 103: You have to overwrite the cost price history, not to write this last value, or successive changes won't be correct. Imagine this scenario of cost history:

-  Move 1 - AVCO: 5 €
-  Move 2 - AVCO: 4 €
-  Move 3 - AVCO: 5 €

You modify move 2 price and the new AVCO is 4.5 € and the final one 4.7 €. According this code, the cost price history will be:

- Move 1 date - AVCO: 5 €
- Move 2 date - AVCO: 4 €
- Move 3 date - AVCO: 5 €
- Modification date - AVCO: 4.7 €

Then if you modify move 3 price unit, this code will take 4 € as starting point for redoing all the computations instead of 4.5 €!

And maybe you need to write final cost price via SQL for not logging a new cost price history (it depends if there's code for detecting same amount).

product_cost_price_avco_sync/models/stock_picking.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = 'stock.picking'
+
+    def action_done(self):
+        return super(StockPicking, self.with_context(
```
pedrobaeza, line 12: Docstring saying: "Avoid AVCO cost price recomputation when validating picking"

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+                if af_move._is_in():
+                    if af_move.inventory_id:
+                        af_move.with_context(
+                            skip_cost_update=True).price_unit = price
+                    # Avoid reprocess move if included in self.write
+                    elif af_move.id in procesed_moves:
+                        procesed_moves.append(af_move.id)
+                    price = af_move.get_average_price(previous_qty, price)
+                    previous_qty += af_move.quantity_done
+                elif af_move._is_out():
```
pedrobaeza, line 91: Maybe we have to do this generic for covering the case of manufacturing moves.

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,64 @@
...
+                    af_move.with_context(
+                        skip_cost_update=True).price_unit = -price
+                    previous_qty -= af_move.quantity_done
+                    if previous_qty < 0.0:
```
pedrobaeza: Why this adjustment? Standard is not considering this AFAIK and the idea is to recreate the same computation.

product_cost_price_avco_sync/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+                'skip_cost_update'):
+            self.price_unit_update()
+        return res
+
+    def price_unit_update(self):
+        for line in self:
+            if line.state not in ['purchase', 'done']:
+                continue
+            line.move_ids.write({
+                'price_unit': line.price_unit,
```
pedrobaeza: You must call here `_get_stock_move_price_unit` instead for taking into account several things like taxes price included, purchase discount module, etc.

product_cost_price_avco_sync/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import models
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = 'purchase.order.line'
+
+    def write(self, vals):
+        res = super().write(vals)
+        if 'price_unit' in vals and not self.env.context.get(
```
pedrobaeza: You should take into account other values that will vary the final price unit, like taxes. Purchase discount modification will also modify the move price unit. Although not performant, the only way I see for not needing a lot of glue modules is that you call `_get_stock_move_price_unit` before calling write records per record, and again after, and propagate the price unit if changed.

pedrobaeza (changes requested):

rafaelbn (approved):
Functionally tested :+1: 

 @chienandalu  missing you functional review

chienandalu (changes requested):
@pedrobaeza comments should be addressed as the module is not compatible with other modules affecting the price unit like purchase_discount.

pedrobaeza:
Is it ready to be reviewed?

carlosdauden:
Changes done

product_cost_price_avco_sync/models/stock_move.py
pedrobaeza: Hehe, docstring seems to be not enough yet

product_cost_price_avco_sync/models/stock_move.py
pedrobaeza: Answer not given

pedrobaeza: Answer not given

product_cost_price_avco_sync/models/stock_move.py
pedrobaeza: What about this?

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,101 @@
...
+                continue
+            previous_instant = move._previous_instant_date()
+            previous_qty = move.product_id.with_context(
+                to_date=previous_instant
+            ).qty_available
+            previous_price = move.product_id.get_history_price(
+                move.company_id.id, previous_instant) if previous_qty else 0.0
+            price = move.get_average_price(previous_qty, previous_price)
+            move._remove_after_history_price()
+            move._create_price_history(price)
```
pedrobaeza, line 73: There's duplicated code that can be avoided reordering a bit the code and adding an extra `if`, but it's worth for avoiding duplicating logic. We can see it together if you want.

product_cost_price_avco_sync/models/stock_move.py
```diff
@@ -0,0 +1,101 @@
...
+                    # Avoid worst price because core don't works negatives fine
+                    if previous_qty < 0.0:
+                        previous_qty = 0.0
+            if float_compare(move.product_id.standard_price, price,
+                             precision_rounding=move.product_uom.rounding):
+                # Write the standard price, as SUPERUSER_ID because a warehouse
+                # manager may not have the right to write on products
+                move.product_id.with_context(
+                    force_company=move.company_id.id,
+                ).sudo().write({'standard_price': price})
```
pedrobaeza, line 103: Is this duplicating latest cost price in history?

pedrobaeza (changes requested):
Still several doubts, but this sounds better

chienandalu (approved):
Looks good to me :+1:

pedrobaeza:
My comments still remains, so this can't be merged.

product_cost_price_avco_sync/models/stock_move.py
carlosdauden: I do not know if I understand well the alternative that you propose, I have considered several options but the code is not so understandable.
Feel free to make a suggestion or a commit.

product_cost_price_avco_sync/models/stock_move.py
carlosdauden: It is possible, but it is something that usually occurs in odoo and that in v13 has changed

product_cost_price_avco_sync/models/stock_move.py
carlosdauden: Outdated

product_cost_price_avco_sync/models/stock_move.py
carlosdauden: I think that manufacture moves are covered with the affected moves search

rousseldenis:
@sergio-teruel

sergio-teruel:
Comments attended!! I agree with @carlosdauden commits.
For me ready to merge

product_cost_price_avco_sync/models/stock_move.py
pedrobaeza: Not sure at all, but we don't have this case for now, so let's move on.

pedrobaeza (approved):
Let's move on as I'm not able to review deeply again due to time constraints, but it's working on production in 2 customers since a month or so, and in v13 everything will change

PR #583, opened 2019-10-29
[11.0][ADD] sale_margin_sync: Recompute sale margin when related stock…

carlosdauden:
… move cost price is changed

@Tecnativa


pedrobaeza:
This one should go to https://github.com/OCA/margin-analysis/

PR #585, opened 2019-11-11
[12.0] Add module stock_quant_package_product_packaging

grindtildeath:


stock_quant_package_product_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,54 @@
...
+        for pack in self:
+            if (
+                not pack.product_packaging_id
+                and pack.single_product_id
+                and pack.single_product_qty
+            ):
+                packaging = self.env['product.packaging'].search(
+                    [
+                        ('product_id', '=', pack.single_product_id.id),
+                        ('qty', '=', pack.single_product_qty)
```
jbaudoux, line 49: Any multiple of the packaging qty should also match. If box qty=10 and you receive 2 boxes (qty=20), you should still match that you have boxes. If pallet qty=50, when you receive 50, then you should match pallet (and not anymore box)

jbaudoux (changes requested):
Missing packaging multiples in matching algorithm

stock_quant_package_product_packaging/models/stock_quant_package.py
guewen: That's an issue we'll have to figure out more globally. If you have 3 boxes of 20 in a move line (so qty of 60) and a bin has room for 2 boxes, you should be able to put 2 boxes in a bin and one in another. But since the put-aways are evaluated for a move line, if they are not split by a quantity that match the packaging, they would never be put in such bins. For now, the best course of action is to split the move lines by their packaging quantity.

rousseldenis:
@grindtildeath @guewen What's the status of this ?

ferran-S73:
LGTM

jbaudoux (approved):
Same code was ported & merged in v13 and v14
So make sense to also merge it in v12

I'll move my comment to an issue as this is still valid

ferran-S73 (approved):

rousseldenis:
@guewen Could you attend your comment ?

guewen:
Wow, I no longer work at Camptocamp, so am no involved longer in this project. I don't remember what this comment is about 😄

rousseldenis:
> Wow, I no longer work at Camptocamp, so am no involved longer in this project. I don't remember what this comment is about smile

Yes I know but don't wanted to frustrate if we merge :smiley:

guewen:
Please go on, no frustration on my side. Thanks 😉

rousseldenis:
@grindtildeath It seems that this is breaking other modules.

rousseldenis:
@grindtildeath Do you plan to finish this ?

grindtildeath:
We can close this as we don't have any customer using it and we only did a POC for them in 12.0

PR #586, opened 2019-11-21
[ADD][11.0] stock_return_request: New module

chienandalu:
This module allows to return stock from a location to either suppliers,
customers or internal locations reverting the corresponding pickings.

cc @Tecnativa TT20569


pedrobaeza:
Please put Travis in green.

stock_return_request/models/stock_return_request.py
```diff
@@ -0,0 +1,517 @@
...
+        copy=False,
+    )
+    to_refund = fields.Boolean(
+        string="To refund",
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+    )
+    show_to_refund = fields.Boolean(
+        compute='_compute_show_to_refund',
+        default=lambda self: 'to_refund' in self.env['stock.move']._fields,
```
rousseldenis, line 106: Why this as there is a compute function ?

stock_return_request/models/stock_return_request.py
```diff
@@ -0,0 +1,517 @@
...
+
+    def action_confirm(self):
+        """Wrapper for multi. Avoid recompute as it delays the pickings
+           creation"""
+        with self.env.norecompute():
+            for one in self:
+                one._action_confirm()
+        self.recompute()
+
+    def _action_confirm(self):
```
rousseldenis, line 234: Maybe split this into smaller functions.

rousseldenis (commented):
Code review. Minor remarks. LGTM

stock_return_request/models/stock_return_request.py
chienandalu: If I remember well, it was to assure we had it visible with a record in a new state (not yet recorded)

chienandalu:
Travis is happy now

stock_return_request/models/stock_move.py
```diff
@@ -0,0 +1,32 @@
...
+        readonly=True,
+        store=True,
+    )
+
+    @api.depends('state', 'returned_move_ids', 'quantity_done',
+                 'reserved_availability', 'returned_move_ids.qty_returnable')
+    def _compute_qty_returnable(self):
+        """Looks for chained returned moves to compute how much quantity
+           from the original can be returned """
+        for one in self.filtered(lambda x: x.state not in ['draft', 'cancel']):
```
carlosdauden: if one is move better it's named as move?

stock_return_request/models/stock_return_request.py
```diff
@@ -0,0 +1,517 @@
...
+        ], limit=1)
+        return warehouse
+
+    def _compute_show_to_refund(self):
+        for one in self:
+            if 'to_refund' not in self.env['stock.move']._fields:
+                continue
+            one.show_to_refund = True
+
+    def _create_picking(self, picking, picking_moves):
```
carlosdauden: Can you do this to pass extra vals avoiding extra write calls:
```suggestion
    def _create_picking(self, picking, picking_moves, extra_vals):
```

carlosdauden (changes requested):

stock_return_request/models/stock_return_request.py
carlosdauden: Better option is create _prepare_picking and use copy_data function:
https://github.com/odoo/odoo/blob/dd620c788fe46131db161449f1f988083d700aac/odoo/models.py#L3848

chienandalu: Sure! I went for the second choice. Thanks!

rafaelbn (approved):
:+1:  thanks

pedrobaeza:
@carlosdauden @rousseldenis can you update your review, please?

carlosdauden (approved):
@rousseldenis comments

stock_return_request/models/stock_return_request.py
chienandalu: Some parts already are. I'll keep it in my for future revisions.

rousseldenis (approved):

PR #590, opened 2019-12-04
[12.0] Add stock_quant_package_dimension

grindtildeath:
Depends on: 
 - [ ] https://github.com/OCA/stock-logistics-workflow/pull/585
 - [ ] <s>https://github.com/OCA/product-attribute/pull/531</s> https://github.com/OCA/product-attribute/pull/593

### DIFF
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -49,7 +49,7 @@ class StockQuantPackage(models.Model):
         #  is happening: https://github.com/odoo/odoo/issues/41353
         if self.product_packaging_id:
             vals = self.product_packaging_id.read(
-                fields=['length', 'width', 'height', 'weight']
+                fields=['length', 'width', 'height', 'max_weight']
             )[0]
             vals['weight'] = vals['max_weight']
             vals.pop('id')
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -1,0 +1,55 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class StockQuantPackage(models.Model):
+    _inherit = 'stock.quant.package'
+
+    weight = fields.Float('Weight (kg)')
+    length = fields.Integer('Length (mm)', help='length in millimeters')
+    width = fields.Integer('Width (mm)', help='width in millimeters')
+    height = fields.Integer('Height (mm)', help='height in millimeters')
+    volume = fields.Float(
+        'Volume (m³)',
+        digits=(8, 4),
+        compute='_compute_volume',
+        readonly=True,
+        store=False,
+        help='volume in cubic meters',
+    )
+
+    @api.depends('length', 'width', 'height')
+    def _compute_volume(self):
+        for pack in self:
+            pack.volume = (
+                pack.length * pack.width * pack.height
+            ) / 1000.0 ** 3
+
+    def auto_assign_packaging(self):
+        self = self.with_context(_auto_assign_packaging=True)
+        res = super().auto_assign_packaging()
+        return res
+
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        if (
+            self.env.context.get('_auto_assign_packaging')
+            and vals.get('product_packaging_id')
+        ):
+            for package in self:
+                package.onchange_product_packaging_id()
+        return res
+
+    @api.onchange('product_packaging_id')
+    def onchange_product_packaging_id(self):
+        # FIXME somehow the values returned by this onchange are not handled
+        #  correctly in basic_model.js on _.each at line 1594 where an overflow
+        #  is happening
+        if self.product_packaging_id:
+            vals = self.product_packaging_id.read(
+                fields=['length', 'width', 'height', 'weight']
+            )[0]
+            vals.pop('id')
+            self.update(vals)
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -47,7 +47,7 @@ class StockQuantPackage(models.Model):
     def onchange_product_packaging_id(self):
         if self.product_packaging_id:
             vals = self.product_packaging_id.read(
-                fields=['length_alt', 'width', 'height', 'max_weight']
+                fields=['length', 'width', 'height', 'max_weight']
             )[0]
             vals['weight'] = vals['max_weight']
             vals.pop('id')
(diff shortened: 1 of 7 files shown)


jgrandguillaume (approved):
LGTM

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class StockQuantPackage(models.Model):
+    _inherit = 'stock.quant.package'
+
+    weight = fields.Float('Weight (kg)')
+    length = fields.Integer('Length (mm)', help='length in millimeters')
```
grindtildeath: :warning: https://github.com/odoo/odoo/issues/41353

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class StockQuantPackage(models.Model):
+    _inherit = 'stock.quant.package'
+
+    weight = fields.Float('Weight (kg)')
```
grindtildeath: :warning: `max_weight` is defined in `delivery`: https://github.com/odoo/odoo/blob/12.0/addons/delivery/models/product_packaging.py#L13 cf https://github.com/OCA/product-attribute/pull/531#pullrequestreview-326963615

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -0,0 +1,57 @@
...
+        return res
+
+    @api.onchange('product_packaging_id')
+    def onchange_product_packaging_id(self):
+        # FIXME somehow the values returned by this onchange are not handled
+        #  correctly in basic_model.js on _.each at line 1594 where an overflow
+        #  is happening: https://github.com/odoo/odoo/issues/41353
+        if self.product_packaging_id:
+            vals = self.product_packaging_id.read(
+                fields=['length', 'width', 'height', 'weight']
```
gurneyalex:
```suggestion
                fields=['length', 'width', 'height', 'weight', 'max_weight']
```

stock_quant_package_dimension/models/stock_quant_package.py
gurneyalex: you can redefine it here, it's not really a problem.

stock_quant_package_dimension/models/stock_quant_package.py
gurneyalex: or maybe not... But you are not reading that field but using it on line 54. And then you are reading weight, but overwriting the value with max_weight, so it is strange.

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -0,0 +1,57 @@
...
+        store=False,
+        help='volume in cubic meters',
+    )
+
+    @api.depends('length', 'width', 'height')
+    def _compute_volume(self):
+        for pack in self:
+            pack.volume = (
+                pack.length * pack.width * pack.height
+            ) / 1000.0 ** 3
```
Tonow-c2c, line 28: like in https://github.com/OCA/product-attribute/pull/531/files#diff-7204ef8da825976a99b3672ada1c75c8R26 you can add a warning comment in the code to say that will work only if the input stay in mm

stock_quant_package_dimension/models/stock_quant_package.py
grindtildeath: Actually I got messed up between the fields. Here we shouldn't use `weight` because it's a computed field: https://github.com/odoo/odoo/blob/12.0/addons/delivery/models/stock_picking.py#L28

benbrich:
Hi @grindtildeath, I am searching for a while already for a module to define package types incl. details (dims, weight, material, DG classes). When encoutering your modules I was quite happy to find an aproach that seems to go in this direction. Thanks for your work! The only thing which is keeping me from implementing you modules is that in our usecase different products are added to packages and the product packaging should work more as a kind of template for the Packages where dims and total weight should be computed for eg shipping purposes. As far as I understand things correctly, in the current implementation a packaging always needs to be linked to a single product, so I cannot assign a package containing two or more different products in several quantities to a predifined product packaging. Is this correct? If yes, do you see a chance to include this case designwise? If I am wrong, could you be of some help, about the how? :-) Thanks from Hamburg, Ben

grindtildeath:
Hi @benbrich 
Thanks for your interest in these modules.
Our use case for this module is strongly linked with what we did in OCA/wms repository. We use this in the reception flow, in order to compute putaway locations according to packaging dimensions and what we called storage types (that we chose to define on the `product.packaging` model): https://github.com/OCA/wms/pull/12 and https://github.com/OCA/wms/pull/13
However, v12.0 was more a prototype of what's possible to do, and we brought some improvements in the v13.0 PRs.
Now, considering your use case, it might make sense to split this module into two different modules, one defining the fields on `stock.quant.package` and another to auto-assign the packaging according for packages including only one product.

@jgrandguillaume what do you think ?

rousseldenis:
> Hi @grindtildeath, I am searching for a while already for a module to define package types incl. details (dims, weight, material, DG classes). When encoutering your modules I was quite happy to find an aproach that seems to go in this direction. Thanks for your work! The only thing which is keeping me from implementing you modules is that in our usecase different products are added to packages and the product packaging should work more as a kind of template for the Packages where dims and total weight should be computed for eg shipping purposes. As far as I understand things correctly, in the current implementation a packaging always needs to be linked to a single product, so I cannot assign a package containing two or more different products in several quantities to a predifined product packaging. Is this correct? If yes, do you see a chance to include this case designwise? If I am wrong, could you be of some help, about the how? :-) Thanks from Hamburg, Ben

Hi, you have to distinguish two concepts inside Odoo which are sometimes misunderstood even by some Odoo people.

product.packaging model is used to define 'how' the product can be served (by you or your supplier). e.g.: a bottle, a pack of 6 bottles, etc... That's why this one is tied to one product in particular.

stock.quant.package is the 'container' used for 'packaging' one or several products (same or different), usually for shipping (but can be used internally too).


I think you should take a look at this PR as it reintroduce some concepts Odoo deleted:

https://github.com/OCA/stock-logistics-tracking/pull/23/files

I hope it helped you

jgrandguillaume:
Hi @benbrich ,

Packaging in Odoo is really related to product it self: how do you sell your product. PACK (stock.quant.package) is used to pack them together. The module here propose to attribute the proper packaging to a PACK if it only host one single product with quantity equal a known packaging.

e.g. Bier is sold by packaging of "6 box2 or "pallet of 6002, so if you create a PACK of 6 bier, it'll give it the packaging "6box". Therefor, it seems you're trying to use packaging where you should may be use PACK ?

Hope this helps.

Joël

simahawk:
shall we close this?

PR #596, opened 2019-12-18
[13.0][MIG] stock_lock_lot: Migration to 13.0

ernestotejeda:
Cc @Tecnativa TT20491


stock_lock_lot/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2015 Serv. Tec. Avanzados - Pedro M. Baeza (http://www.serviciosbaeza.com)
+# Copyright 2015 AvanzOsc (http://www.avanzosc.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Stock Lock Lot",
+    "Summary": "Restrict blocked lots in Stock Moves and reservations",
+    "version": "13.0.1.0.0",
+    "author": "OdooMRP team, "
```
pedrobaeza: Remove this author

stock_lock_lot/models/stock_quant.py
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2015 AvanzOsc (http://www.avanzosc.es)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    locked = fields.Boolean(
```
pedrobaeza: I think this is not adding any value and takes time to computed (imagine on installing the module...). Or is it used in any place?

pedrobaeza (commented):
It looks good in a first sight

stock_lock_lot/models/stock_quant.py
ernestotejeda: It is only used to show in the tree an form view if its product is locked, so i can remove it.

ernestotejeda:
changes done

pedrobaeza (approved):
I have added some tests, although there's some part still untested, but I think it's good enough for now.

chienandalu (approved):
Tested on runbot :+1:

dreispt:
Hello @ernestotejeda @pedrobaeza, this migration introduced behaviour changes to the original design, that are incompatible with some use cases.

I'm concerned with [this](https://github.com/OCA/stock-logistics-workflow/pull/1686/commits/2c1ecf483864aad5fe10503de56a1c65565f859e#diff-bbd385fc21b8901e79614f2d708717b7bd9bcc4b8bf417958a05925478fbc137R51-R57) particular change:
- Previously locking a Lot would undo any reservations on it. This sounds reasonable and useful, but was removed.
- After unlocking a lot that has reservations is not allowed. This is not reasonable to me, as it prevents unlocking a lot that is reserved in a chained move and waiting for the lot to be "released"/unlocked.

Right now, this prevent using this module to implement quality verification workflows in a multi-step receiving scenario.
I wonder about the design considerations behind the behavior change, to decide if I should go ahead with a different implementation fitting my need, or propose a change to this module.

Thank you

pedrobaeza:
I sincerely don't remember about this work done 6 years ago. We are currently not using this module, so proceed as you prefer.

PR #603, opened 2019-12-27
[12.0][ADD]stock_picking_auto_revert

AaronHForgeFlow:
This module means to substitute `stock_cancel`. 

**stock_cancel** cannot be migrated without changing all Odoo logic. It is impossible to handle quant history with the current design. If a user cancels and old picking, all the history should be recreated from the time the wrong picking was completed. To achieve this we need to reinvent the wheel.

I created this as separate module of https://github.com/ForgeFlow/stock-logistics-workflow/tree/12.0/stock_picking_back2draft because they cover similar but not same cases:

- `stock_picking_back2draft` handles confirmed pickings but not done pickings.
- `stock_picking_auto_revert` handles done pickings, but not confirmed or draft pickings. 

I feel like some companies may want  the ability to set to draft pickings that are ready but not to do quick returns of the ones that are done. I would let the two modules separated. 

Waiting for some feedback here.


stock_picking_auto_revert/README.rst
```diff
@@ -0,0 +1,81 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/stock-logistics-workflow-10-0/stock-logistics-workflow-10-0-stock_picking_auto_revert
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/154/10.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+Corrrect mistakes in pickings. Instead of cancel this module will create a
```
bealdav:
```suggestion
Correct mistakes in pickings. Instead of cancel this module will create a
```

stock_picking_auto_revert/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2019 ForgeFlow
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Stock Picking Auto Revert",
+    "version": "12.0.1.0.0",
+    "category": "Stock",
```
rousseldenis: Change this to "Warehouse"

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,73 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models, exceptions, _
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _check_restrictions(self):
+        # to be implemented by other modules to add restrictions to returns
```
rousseldenis: What the aim of this ?

Shouldn't it be located in another module ?

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,73 @@
...
+        pass
+
+    @api.multi
+    def action_revert_recreate(self):
+        self.ensure_one()
+        pick = self
+        pick._check_restrictions()
+        msg = _(
+            "Too bad. This picking cannot be returned because the products "
```
rousseldenis: :smile: Even if I enjoy the beginning of the sentence, I would prefer something more consensual.

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,73 @@
...
+            "are not available in the destination location"
+        )
+        # Create return picking
+        StockReturnPicking = self.env["stock.return.picking"]
+        default_data = StockReturnPicking.with_context(
+            active_ids=pick.ids, active_id=pick.ids[0]
```
rousseldenis: Why not 
```suggestion
            active_ids=pick.ids, active_id=pick.id
```
as pick is protected by ensure_one()

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,73 @@
...
+                "parent_location_id",
+                "location_id",
+            ]
+        )
+        for rm in default_data["product_return_moves"]:
+            if len(rm) > 2 and isinstance(rm[2], dict):
+                sm = pick.move_lines.filtered(
+                    lambda x: x.product_id.id == rm[2]["product_id"]
+                )
+                if rm[2]["quantity"] < sm.product_uom_qty:
```
rousseldenis: I think you should use float_compare

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,73 @@
...
+                    raise exceptions.UserError(msg)
+            else:
+                raise exceptions.UserError(msg)
+        return_wiz = StockReturnPicking.with_context(
+            active_ids=pick.ids, active_id=pick.ids[0]
```
rousseldenis: Same as above

stock_picking_auto_revert/view/stock_view.xml
```diff
@@ -0,0 +1,17 @@
...
+
+    <record id="stock_picking_cancel_form" model="ir.ui.view">
+        <field name="name">stock.picking.cancel.form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form"/>
+        <field name="arch" type="xml">
+            <field name="state" position="before">
+                <button name="action_revert_recreate" string="Return and recreate"
+                    states="done" type="object" icon="fa-ban"
+                    confirm="Are you sure you want to return a recreate this picking?"/>
```
rousseldenis:
```suggestion
                    confirm="Are you sure you want to return and recreate this picking?"/>
```

rousseldenis (changes requested):
Code review. Some changes.

@AaronHForgeFlow A security group should be great to access that feature or not

stock_picking_auto_revert/models/stock_picking.py
AaronHForgeFlow: It was just a hook for other modules to add restrictions on the use of this button. But it was not a good idea. I removed it.

stock_picking_auto_revert/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+{
+    "name": "Stock Picking Auto Revert",
+    "version": "12.0.1.0.0",
+    "category": "Warehouse",
+    "summary": "Returns and recreate the picking",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "httpa://forgeflow.com",
+    "license": "AGPL-3",
+    "depends": ["sale_stock"],
+    "data": ["security/stock_security.xml", "view/stock_view.xml",],
```
ps-tubtim: I can't see this file `security/stock_security.xml`.

ps-tubtim (commented):
Please fix Runbot and Travis.

stock_picking_auto_revert/__manifest__.py
AaronHForgeFlow: Edited: actually it was a requirement, so will put back

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+                "product_return_moves",
+                "parent_location_id",
+                "location_id",
+            ]
+        )
+        for rm in default_data["product_return_moves"]:
+            if len(rm) > 2 and isinstance(rm[2], dict):
+                sm = pick.move_lines.filtered(
+                    lambda x: x.product_id.id == rm[2]["product_id"]
+                )
```
nguyenminhchien, line 36: Hi, if "pick" has 2 lines having the same product, "sm" may not ensure having 1 record. so, "sm.product_uom_qty" (at the line 43) will return an error in that case.

AaronHForgeFlow: Good point. Corrected.

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,77 @@
...
+        ).default_get(
+            [
+                "move_dest_exists",
+                "original_location_id",
+                "product_return_moves",
+                "parent_location_id",
+                "location_id",
+            ]
+        )
+        for rm in default_data["product_return_moves"]:
```
rousseldenis, line 32: In order to get better visibility and to allow overriding to maybe complete conditions for a picking to be returned, I would have done this computation in another function. And better a computed field 'can_be_returned' or 'products_available_for_return'. What do you think ?

rousseldenis:
@AaronHForgeFlow What's the status of this ?

AaronHForgeFlow:
I am not working on this anymore. But I will probably try to add the same module in a newer version (14.0 or 15.0) I will create a PR for that in any case.

PR #604, opened 2020-01-02
[FIX] stock_picking_batch_extended drop previous created keys on update

i-vyshnevska:
need to drop associated view keys

### DIFF
--- a/stock_picking_batch_extended/readme/DESCRIPTION.rst
+++ b/stock_picking_batch_extended/readme/DESCRIPTION.rst
@@ -3,6 +3,8 @@ but reworked for matching Odoo new picking system (with pack operations).
 
 Since v12 it is compatible with odoo addon https://github.com/odoo/odoo/blob/12.0/addons/stock_picking_batch/models/stock_picking_batch.py
 It allows to define which approach for batch handling will be used per company.
+By default after installation this option will be activated for all companies, you cal disable
+it in settings.
 
 Batch picking allows you to manage several pickings at the same time.
 You can create a batch with all pickings selected or multiple batch depends on
--- a/stock_picking_batch_extended/readme/DESCRIPTION.rst
+++ b/stock_picking_batch_extended/readme/DESCRIPTION.rst
@@ -3,7 +3,7 @@ but reworked for matching Odoo new picking system (with pack operations).
 
 Since v12 it is compatible with odoo addon https://github.com/odoo/odoo/blob/12.0/addons/stock_picking_batch/models/stock_picking_batch.py
 It allows to define which approach for batch handling will be used per company.
-By default after installation this option will be activated for all companies, you cal disable
+By default after installation this option will be activated for all companies, you can disable
 it in settings.
 
 Batch picking allows you to manage several pickings at the same time.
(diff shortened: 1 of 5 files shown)


rousseldenis:
@i-vyshnevska Could you add the module you're fixing in PR's subject ?

ajaniszewska-dev (approved):

stock_picking_batch_extended/readme/DESCRIPTION.rst
```diff
@@ -3,6 +3,8 @@ but reworked for matching Odoo new picking system (with pack operations).
 
 Since v12 it is compatible with odoo addon https://github.com/odoo/odoo/blob/12.0/addons/stock_picking_batch/models/stock_picking_batch.py
 It allows to define which approach for batch handling will be used per company.
+By default after installation this option will be activated for all companies, you cal disable
```
yvaucher:
```suggestion
By default after installation this option will be activated for all companies, you can disable
```

yvaucher:
@i-vyshnevska shouldn't the default value set to `True` here?

https://github.com/OCA/stock-logistics-workflow/blob/13.0/stock_picking_batch_extended/models/res_config_settings.py#L20

stock_picking_batch_extended/views/stock_batch_picking.xml
```diff
@@ -170,8 +170,12 @@
         <field name="name">Batch Picking</field>
         <field name="res_model">stock.picking.batch</field>
         <field name="view_mode">tree,form</field>
-         <field name="view_id" ref="stock_batch_picking_tree"/>
-            <field name="search_view_id" ref="stock_batch_picking_search"/>
+        <field name="view_id" ref="stock_batch_picking_tree"/>
+        <field name="view_ids" eval="[
+            (5, 0, 0),
+            (0, 0, {'view_mode': 'tree', 'view_id': ref('stock_batch_picking_tree')}),
+            (0, 0, {'view_mode': 'form', 'view_id': ref('stock_batch_picking_form')})]"/>
```
hailangvn, line 177: Good day @i-vyshnevska, we are porting this PR to 14.0. We are wondering what is the purpose of these three lines and are they needed for 14.0. Could you help to advise?

i-vyshnevska: @hailangvn yes this PR you also can include to migration
the first line is to clear previous relations if exist - so this will ensure there are no other views linked to this action
than we setting correct views we want to use

hailangvn: Thanks @i-vyshnevska for the advice. We did tests with and without this code but could not find any difference. Could you please provide more background information that lead to this implementation?

i-vyshnevska: @hailangvn will fail if you try to update on installation, check for constrains "act_window_view_unique_mode_per_action" BTW could be different in v15

rousseldenis:
@i-vyshnevska What's the status of this ?

i-vyshnevska:
@rousseldenis the code is valid and should be forward ported to the latest branches, regards Yannick comment it is not crucial we can introduce it on the latest branch

PR #607, opened 2020-01-08
[13.0][MIG] stock_quant_package_product_packaging

grindtildeath:
Forward port of #585


stock_quant_package_product_packaging/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2019 Camptocamp SA
```
Tonow-c2c, line 1:
```suggestion
# Copyright 2020 Camptocamp SA
```

and idem for others

Tonow-c2c (approved):

rousseldenis:
@grindtildeath isn't it time to add tests ?

grindtildeath:
@rousseldenis I'll try to add a few here if I still have time after migrating the "top modules" from wms repo :wink:

stock_quant_package_product_packaging/__manifest__.py
grindtildeath: Again, not needed as there were no changes this year. :wink:

stock_quant_package_product_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    single_product_qty = fields.Float(compute="_compute_single_product")
+
+    @api.depends("quant_ids", "quant_ids.product_id")
+    def _compute_single_product(self):
+        for pack in self:
+            pack_products = pack.quant_ids.mapped("product_id")
+            if len(pack_products) == 1:
+                pack.single_product_id = pack_products.id
+                # TODO handle uom
```
simahawk, line 34: is this TODO a blocker?

simahawk (commented):
@grindtildeath still have to work on this?

stock_quant_package_product_packaging/models/stock_quant_package.py
grindtildeath: Looking at it now, I think the comment can be dropped because:
 - UOM on `stock.quant` is related to the product: https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_quant.py#L57
 - UOM on `product.packaging` is also related to the product: https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L700

simahawk: makes sense. Remove it then :)

simahawk (approved):

simahawk:
@grindtildeath check commits, this is likely an error https://github.com/OCA/stock-logistics-workflow/pull/607/commits/e2c68b7bc9ad16f378166e6831489399168f2e00

PR #620, opened 2020-02-06
[IMP] stock_return_request: lot suggestions

chienandalu:
Lot suggestion wizard so the user can easily guess the lot that is going to be returned.

image

cc @Tecnativa TT20569

### DIFF
--- a/stock_return_request/wizard/suggest_return_request_lot.py
+++ b/stock_return_request/wizard/suggest_return_request_lot.py
@@ -1,0 +1,99 @@
+# Copyright 2020 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class SuggestReturnRequestLot(models.TransientModel):
+    _name = "suggest.return.request.lot"
+    _description = "Suggest lots for the return request line"
+
+    request_line_id = fields.Many2one(
+        comodel_name="stock.return.request.line",
+        default=lambda self: self._default_request_line_id(),
+        required=True,
+        readonly=True,
+        ondelete="cascade",
+    )
+    lot_suggestion_mode = fields.Selection(
+        selection=[
+            ('sum', 'Total by lot'),
+            ('detail', 'Total by move'),
+        ],
+        default='sum',
+    )
+    suggested_lot = fields.Selection(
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Travis is failing

stock_return_request/wizard/suggest_return_request_lot.py
```diff
@@ -0,0 +1,97 @@
...
+        if (
+            self.env.context.get("active_model", False)
+            != "stock.return.request.line"
+        ):
+            return (False, False)
+        request_line = (
+            self.request_line_id or
+            self.request_line_id.browse(self.env.context.get('active_id')))
+        if not request_line:
+            (False, False)
```
carlosdauden:
```suggestion
            return (False, False)
```

carlosdauden (changes requested):

stock_return_request/wizard/suggest_return_request_lot.py
chienandalu: Sure :) thanks

carlosdauden (approved):

stock_return_request/wizard/suggest_return_request_lot.py
```diff
@@ -0,0 +1,97 @@
...
+            return
+        return [(
+            ml.lot_id.id, '{} - {} - {} - {}'.format(
+                ml.date, ml.lot_id.name,
+                ml.reference, suggested_lots_moves[ml])
+        ) for ml in suggested_lots_moves.keys()]
+
+    def action_confirm(self):
+        if self.lot_suggestion_mode == 'sum' and self.suggested_lot:
+            self.request_line_id.lot_id = self.suggested_lot
```
pedrobaeza, line 96: I don't get how this is working assigning that selection value. This seems very hackish...

chienandalu: Well, the selection value is the id of the lot. Not black magic was made :)

pedrobaeza: Ah, OK, I see. Then add a mention that on next version, you can't put an integer as selection key (only strings), so you have to convert from/to in this code.

chienandalu: Added comment

PR #623, opened 2020-02-07
[12.0][IMP] stock_picking_invoice_link: Allow invoice lines linked with more than one stock move

sergio-teruel:
This commit is forwardport from a v11.0 commit (https://github.com/OCA/stock-logistics-workflow/commit/c087ce6c20cade3823f4b8ae985b0fab51bfe519)

Mybe I need a migration script for new v12..0 installations, but if the instance come from a v11.0 the field is many2many yet... 
openupgrade.m2o_to_x2m(env.cr, env['stock.move'], 'stock_move', 'invoice_line_ids', 'invoice_line_id')

cc @Tecnativa

### DIFF
--- a/stock_picking_invoice_link/models/sale_order.py
+++ b/stock_picking_invoice_link/models/sale_order.py
@@ -2,37 +2,39 @@
 # Copyright 2017 Jacques-Etienne Baudoux <je@bcim.be>
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import api, models
+from odoo import models
+from odoo.tools import float_compare
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    @api.multi
-    def invoice_line_create_vals(self, invoice_id, qty):
-        self.mapped(
-            'move_ids'
-        ).filtered(
-            lambda x: x.state == 'done' and
-            not x.invoice_line_id and
-            not x.location_dest_id.scrap_location and
-            x.location_dest_id.usage == 'customer'
-        ).mapped(
-            'picking_id'
-        ).write({'invoice_ids': [(4, invoice_id)]})
-        return super(SaleOrderLine, self).invoice_line_create_vals(invoice_id,
-                                                                   qty)
+    def get_stock_moves_link_invoice(self):
+        return self.mapped('move_ids').filtered(
+            lambda x: (
+                x.state == 'done' and not (any(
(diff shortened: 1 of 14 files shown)


sergio-teruel:
@carlosdauden Can you review please??

manuelcalerosolis (approved):
I have tested the functionality of the module, and everything works correctly.

pedrobaeza (changes requested):
Indeed we need here a migration script for converting one2many to many2many, but we have to do it conditionally (you can detect if the m2m table is empty as condition), as if not, when migrating from v11 with the upgrade module 11.0.2.0.0 to v12, there will be duplicated data.

stock_picking_invoice_link/models/sale_order.py
```diff
@@ -2,37 +2,37 @@
...
-            not x.scrapped and (
-                x.location_dest_id.usage == 'customer' or
-                (x.location_id.usage == 'customer' and
-                 x.to_refund)
-            )).ids
-        vals['move_line_ids'] = [(6, 0, move_ids)]
+        vals = super()._prepare_invoice_line(qty)
+        stock_moves = self.get_stock_moves_link_invoice()
+        # Invoice returned moves marked as to_refund
+        if qty < 0.0:
```
rousseldenis: Please use float_compare()

rousseldenis (changes requested):
Code review. Indeed as @pedrobaeza this needs a migration script.

carlosdauden:
I don't understand...

Are you see this?:
https://github.com/OCA/stock-logistics-workflow/blob/11.0/stock_picking_invoice_link/migrations/11.0.2.0.0/post-migrate.py#L6-L9

This field is changed in v11.
The instance must be updated before major version migration

pedrobaeza:
@carlosdauden the migration script is needed for those with version 12.0.1.0.0, but the migration should check for avoiding applying twice for those migrating from 11.0.2.0.0. Let's have a meeting if you still don't understand.

carlosdauden:
Ok, sorry.
I thought this was the PR migration module to v12...

sergio-teruel:
@pedrobaeza Added migration script
@rousseldenis Changes done!

carlosdauden (approved):

carlosdauden:
@pedrobaeza @rousseldenis Can you check?

rousseldenis (approved):
Code review

pedrobaeza (approved):

PR #597, opened 2019-12-19
[13.0][MIG] stock_no_negative

qdelcourte:

PR #628, opened 2020-02-21
13.0 mig stock no negative

jimhoefnagels:


ps-tubtim:
@jimhoefnagels This module had migration pull request on https://github.com/OCA/stock-logistics-workflow/pull/597

PR #630, opened 2020-02-25
[12.0] [FIX] stock_picking_line_sequence install with unuseful and very slow update of stock moves

sergiocorato:
When install module stock_picking_line_sequence on an existing db with data, it force recomputation of all stock.move with deletion and re-creation of all stock.picking.package.preparation.line


AaronHForgeFlow (approved):
:+1:

LoisRForgeFlow (approved):

PR #632, opened 2020-02-26
[12.0][IMP] stock_picking_batch_extended: Add followers, attachments, chatt…

carlosdauden:
…er and activities

@Tecnativa

### DIFF
--- a/stock_picking_batch_extended/__manifest__.py
+++ b/stock_picking_batch_extended/__manifest__.py
@@ -1,5 +1,5 @@
 # Copyright 2012-2014 Alexandre Fayolle, Camptocamp SA
-# Copyright 2018 Tecnativa - Carlos Dauden
+# Copyright 2018-2020 Tecnativa - Carlos Dauden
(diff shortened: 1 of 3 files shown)


pedrobaeza:
@carlosdauden Travis is red. Can you please take also the opportunity to change the maintainers for being real GitHub users?

pedrobaeza (approved):

manuelcalerosolis (commented):
I think we should include the file USAGE.rst to explain that we should use this module, it must contain something like this.

To configure this module, you need to:

* Goto (Inventory > Configuration > and select "Use OCA approach to validate Picking Batch")

manuelcalerosolis (approved):
Only include USAGE.rst

pedrobaeza:
That would be CONFIGURE.rst, not USAGE.rst. Any way, let's merge it and let that as an improvement not directly related with this PR.

/ocabot merge minor

Please forward port it to v13 as well.

PR #633, opened 2020-03-04
[FIX] product_cost_price_avco_sync: Aleatory error because compare value

carlosdauden:
before BD write.
Invalidate cache don't works to avoid the problem.

@Tecnativa


pedrobaeza (approved):
While not having a better solution, it's OK for me.

sergio-teruel (approved):

AaronHForgeFlow (approved):
:+1:

PR #638, opened 2020-03-11
[12.0][FIX]always reset sequence on creation of stock moves

### LINKED ISSUE #631: stock moves sequence-error created from other modules
module: stock_picking_line_sequence
version: 12.0
**Current behavior**
This module does not integrate with other modules such as (sales, purchase )

for example the stock move sequence created from a" sale order" is not correct, instead of (1,2,3,4....etc) you will find (9,999)for all stock move lines?.
also the stock move sequence created from a" purchase order" also is not correct, you will find lines with a sequence step equal to 5 such as (5,10,15,20,25,......etc.)?

for your information I tested the module with a fresh database but the same error.
could you please check that!

AaronHForgeFlow:
Fixes https://github.com/OCA/stock-logistics-workflow/issues/631

### DIFF
--- a/stock_picking_line_sequence/views/stock_view.xml
+++ b/stock_picking_line_sequence/views/stock_view.xml
@@ -15,13 +15,6 @@
                     <field name="sequence" widget="handle"/>
                     <field name="sequence2"/>
                 </xpath>
-                <xpath expr="//notebook/page/field[@name='move_ids_without_package']" position="attributes">
-                    <attribute name="context">{'picking_type_code': picking_type_code, 'default_picking_id': id,
-                    'form_view_ref':'stock.view_move_picking_form', 'address_in_id': partner_id,
-                    'default_picking_type_id': picking_type_id, 'default_location_id': location_id,
-                    'default_location_dest_id': location_dest_id, 'default_sequence': max_line_sequence}
-                    </attribute>
-                </xpath>
             </field>
         </record>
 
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
Basically, the issue comes when stock moves are created individually and the picking is created later. For example, the standard procedure when confirming sales orders.

LoisRForgeFlow:
cc @mahmoudmk1 can you review that your problem is fixed?

AaronHForgeFlow:
Tests are failing, will fix that.

AaronHForgeFlow:
Travis error unrelated: fixed here: https://github.com/OCA/stock-logistics-workflow/pull/633

AaronHForgeFlow:
It seems the error in a test of module `product_cost_price_avco_sync` sometimes arises and sometimes not, but seems unrelated to this module.

AaronHForgeFlow:
cc @mahmoudmk1 can you review that your problem is fixed?

stock_picking_line_sequence/views/stock_view.xml
```diff
@@ -15,13 +15,6 @@
                     <field name="sequence" widget="handle"/>
                     <field name="sequence2"/>
                 </xpath>
-                <xpath expr="//notebook/page/field[@name='move_ids_without_package']" position="attributes">
-                    <attribute name="context">{'picking_type_code': picking_type_code, 'default_picking_id': id,
```
rousseldenis, line 19: Why removing this ?

AaronHForgeFlow: It is defaulting the sequence on creation, and it seems not needed anymore.

rousseldenis:
@AaronHForgeFlow Could you rebase ?

rousseldenis (approved):

rousseldenis:
@mahmoudmk1

PR #641, opened 2020-03-23
[ADD] stock_picking_group_by_partner_by_carrier

gurneyalex:
This module can be used if your customers expect that several different orders
they passed will be shipped in a single delivery order.

With this module installed, when a sale order is confirmed, the stock moves for
the lines of the sale order can be placed in an existing delivery order that
shares the same delivery address and carrier (or lack thereof).

Sale orders with a Shipping Policy set to 'When all products are ready' always
get their own shipping.


stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -0,0 +1,78 @@
...
+        if self.sale_line_id.order_id.picking_policy == 'one':
+            # use the normal domain from the stock module
+            domain = [
+                ('group_id', '=', self.group_id.id),
+                ('location_id', '=', self.location_id.id),
+                ('location_dest_id', '=', self.location_dest_id.id),
+                ('picking_type_id', '=', self.picking_type_id.id),
+                ('printed', '=', False),
+                ('immediate_transfer', '=', False),
+                ('state', 'in', states)]
```
guewen: Shouldn't you add the filter on `move_type` as well here? Or is the filter on `group_id` enough?

gurneyalex: it is. 
and I have an automated test for it :)

jgrandguillaume (approved):
Works perfectly !

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,99 @@
...
+    @api.depends("move_lines.sale_line_id.order_id")
+    def _compute_sale_ids(self):
+        for rec in self:
+            rec.sale_ids = rec.mapped("move_lines.sale_line_id.order_id")
+
+    def write(self, values):
+        if self.env.context.get("picking_no_overwrite_partner_origin"):
+            written_fields = set(values.keys())
+            if written_fields == {"partner_id", "origin"}:
+                values = {}
```
simahawk, line 24: any chance to land here w/ other values you'd better preserve?

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,99 @@
...
+            # we are creating a backorder from
+            picking = self.move_lines[0]._search_picking_for_assignation()
+            if picking:
+                return picking
+        return super(
+            StockPicking, self.with_context(picking_no_copy_if_can_group=0)
+        ).copy(defaults)
+
+    def do_something(self):
+        return "bla bla"
```
simahawk, line 57: I hope it does :smile:

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,99 @@
...
+                    for move in sale_moves:
+                        sales_and_moves.append(move)
+                return sales_and_moves
+            else:
+                return moves
+        else:
+            moves = self.move_lines.sorted(lambda m: m.sale_line_id.order_id)
+            if len(moves.mapped("sale_line_id.order_id")) > 1:
+                sales_and_moves = []
+                for sale, sale_moves in groupby(
```
simahawk, line 88: can we split this part as it's the same as at line 55?

florian-dacosta:
Thanks for the module
Just one functional remark, I think that not allowing to group "When all products are ready" pickings is too restrictive.
A lot of business configure it this way to avoid partial delivery, which maybe expensive. But it does not mean they would not want to group this kind of picking, too reduce the shipping cost.

I think we should at least have the possibility to group the pickings having this shipping policy if needed, maybe making it configurable or something.
What do you think @gurneyalex ?

gurneyalex:
@florian-dacosta Yes we can make this configurable. I'll add the code for this. Thanks for your feedback.

guewen:
@gurneyalex needs to merge this first https://github.com/gurneyalex/stock-logistics-workflow/pull/4

pedrobaeza:
ocabot stuck. Please relaunch

PR #642, opened 2020-03-31
[13.0][ADD] stock_partner_delivery_window

grindtildeath:
Depends on:
 - [x] https://github.com/OCA/server-tools/pull/1798
 - [x] https://github.com/OCA/partner-contact/pull/876

This module allows to define time scheduling preference for delivery orders on
partners, in order to raise a warning when changing a scheduled date to a time
window that is not preferred by this customer.

### DIFF
--- a/stock_partner_delivery_window/models/res_partner.py
+++ b/stock_partner_delivery_window/models/res_partner.py
@@ -1,6 +1,7 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ResPartner(models.Model):
@@ -19,8 +20,22 @@ class ResPartner(models.Model):
     )
 
     delivery_time_window_ids = fields.One2many(
-        'partner.delivery.time.window', 'partner_id'
+        'partner.delivery.time.window',
+        'partner_id',
+        string="Delivery time windows",
     )
+
+    @api.constrains('delivery_time_preference', 'delivery_time_window_ids')
+    def _check_delivery_time_preference(self):
+        for partner in self:
+            if (
+                partner.delivery_time_preference == "time_windows" and
+                not partner.delivery_time_window_ids
+            ):
+                raise ValidationError(_(
+                    "Please define at least one delivery time window or change"
+                    " preference to Any time"
+                ))
 
     def get_delivery_windows(self, day_name=None):
         """
--- a/stock_partner_delivery_window/models/res_partner.py
+++ b/stock_partner_delivery_window/models/res_partner.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.addons.partner_tz.tools import tz_utils
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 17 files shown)


stock_partner_delivery_window/models/res_partner.py
```diff
@@ -0,0 +1,70 @@
...
+        windows = self.env["partner.delivery.time.window"].search(
+            domain
+        )
+        for window in windows:
+            if not res.get(window.partner_id.id):
+                res[window.partner_id.id] = self.env["partner.delivery.time.window"].browse()
+            res[window.partner_id.id] |= window
+        return res
+
+    def is_in_delivery_window(self, date_time):
```
jbaudoux, line 66: Add a comment that this method is expecting a datetime in partner tz and not in UTC

grindtildeath: Actually the method must expect a tz naive datetime in UTC (as is the case in the rest of Odoo), so I will fix it. But good catch :+1:

grindtildeath: FTR: The issue was that the conversion done in `float_to_time` did use `datetime.now` instead of the picking's date.

Corrected in ae5ab0f with tests fixed.

grindtildeath:
Reverted `oca_dependencies.txt` to avoid conflicts on branch aggregation

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,52 @@
...
+    def _onchange_scheduled_date(self):
+        self.ensure_one()
+        if (
+            not self.partner_id
+            or self.partner_id.delivery_time_preference != "time_windows"
+            or self.picking_type_id.code != "outgoing"
+        ):
+            return
+        p = self.partner_id
+        sec_lead_time = self.company_id.security_lead
```
LoisRForgeFlow: this is added by "sale_stock" you need to change the dependencies of this module then or adapt it.

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,52 @@
...
+
+from odoo import _, api, models
+from odoo.tools.misc import format_datetime
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    @api.onchange("scheduled_date")
+    def _onchange_scheduled_date(self):
```
LoisRForgeFlow, line 14: Why only check in the onchange? this seems to me a little bit weak, you can write any scheduled_date with any other action. Why don't you extend '_compute_scheduled_date' instead or use a api.constraint?

LoisRForgeFlow (changes requested):
Some comments below

LoisRForgeFlow:
Also, I would explicit timezone in the error messages:
image

As you can see in the above image, only regarding the message you could think the date should be inside the window (but partner is at MT and I'm at CEST - that's why I get the error, which is correct)

stock_partner_delivery_window/models/stock_picking.py
grindtildeath: Good catch :+1:

stock_partner_delivery_window/models/stock_picking.py
grindtildeath: IIRC the idea was to still allow overriding and get the warning in such case. 

@jgrandguillaume can you confirm we don't want to restrict this?

grindtildeath:
@LoisRForgeFlow Thx for the review, comments attended.

simahawk:
replaced by https://github.com/OCA/stock-logistics-workflow/pull/755

PR #643, opened 2020-04-03
[12.0] NEW stock_picking_filter_product: 

alan196:
A new module that allows filtering products shown in pickings and stock scrap to select only available products in origin location.

This module is based on the logic of stock_picking_filter_lot module.


lmgr0312 (approved):
LGTM!

Tino480:
Looks and works as described

Tino480 (approved):

rousseldenis:
@alan196 Thanks for your contribution. I don't really like the module name as it does not reflect what it does on a functional point of view. Could you enhance a little bit ?

alan196:
🤔 What do you think of 

- stock_filter_product_by_location
- stock_filter_product_by_source_location
- stock_show_products_source_location

@rousseldenis

PR #648, opened 2020-04-15
[13.0][MIG] stock_picking_filter_lot

newtratip:
I changed the location for make sure quantity > 0
Selection_333

Is there any concern for change this ?


rousseldenis:
> I changed the location for make sure quantity > 0
> Selection_333
> 
> Is there any concern for change this ?

I think this is ok.

rousseldenis (approved):
Code review

AdriaGForgeFlow (approved):
Code review :+1:

lk-eska:
> > I changed the location for make sure quantity > 0
> > Selection_333
> > Is there any concern for change this ?
> 
> I think this is ok.

No it isn't ok.

Reason 1: It will not compute when quantity is changed positive to negative (when it becomes zero it is ok since it will delete the quant and trigger compute) 

Reason 2: This formula will eliminate Vendor location to be in locations (because the quantity is negative), thus it will not possible to buy the same product with the same lot number.

PR #652, opened 2020-04-20
[10.0][ADD]stock_picking_mass_return

MPoch-PlanetaTIC:
A wizard to select a supplier, a warehouse and a list of products with their quantities.
When clicking "Return Product" button a new stock picking out will be created to return the indicated products and quantites getting the stock from the last received purchases.

cc @PlanetaTIC


rousseldenis:
@MPoch-PlanetaTIC Could you add tests ?

rousseldenis:
@MPoch-PlanetaTIC What's the status of this ?

MPoch-PlanetaTIC:
This module worked, but now we don't use it because we no longer use Odoo10.
And we haven't migrated it to new Odoo versions.

Feel free to do whatever you need to ;)

rousseldenis:
Ok, so, closing as no interest from others

PR #654, opened 2020-04-20
[10.0] [ADD] stock picking purchase on no backorder

MPoch-PlanetaTIC:
Purchase the missing stock on a reception when selecting "No Backorder"

cc @PlanetaTIC


rousseldenis:
@MPoch-PlanetaTIC What's the status of this ?

PR #657, opened 2020-05-18
[12.0][FIX] sale_order_global_stock_route: Solved problems related with the method product_id_change()

CarlosRoca13:
When we add order lines and the data was updated using that method, the field of the lines was not updated. With this fix, we solve this and the module is easier to understand

cc @Tecnativa TT23962

### DIFF
--- a/sale_order_global_stock_route/models/sale_order.py
+++ b/sale_order_global_stock_route/models/sale_order.py
@@ -2,7 +2,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import fields, models, api
-from lxml import etree
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please check Travis

pedrobaeza (changes requested):
Well, this salomonical solution as product is mandatory on sales order lines works, although not very academic. The problem is if you create orders from code/XML-RPC. I think you need to override also sale.order.line create / write method for handling this. On v13 it will be easier thanks to computed writeable fields.

pedrobaeza:
@sergio-teruel your opinion here is welcome.

sergio-teruel:
Yes!! @CarlosRoca13 @pedrobaeza is right. By this way, any external application like as a mobile app  would have to call all onchange methods. But if we open this door... Do we have to apply this to all the modules that already exist? Anyway... You can extend write and create method instead of onchange method and you write a comment for remove this methods and use a computed writable fields....

pedrobaeza:
Yeah, in v13 we fortunately can rid off this problem. @sergio-teruel if you don't consider that as needed for current use cases, then let's go as is.

sale_order_global_stock_route/models/sale_order.py
```diff
@@ -21,27 +20,27 @@ def _onchange_route_id(self):
...
+    _inherit = 'sale.order.line'
+
+    @api.onchange('product_id')
+    def product_id_change(self):
+        res = super().product_id_change()
+        self.route_id = self.order_id.route_id
+        return res
+
+    @api.model
+    def create(self, vals_list):
```
pedrobaeza, line 42: Make use of create multi here

pedrobaeza (approved):

sergio-teruel (approved):

PR #660, opened 2020-05-20
[12.0][FIX] stock_no_negative: Use product display_name instead of name

Abranes:
Use `product.display_name` instead of `product.name`

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -42,5 +42,5 @@ class StockQuant(models.Model):
                     "stock level of the product '%s'%s would become negative "
                     "(%s) on the stock location '%s' and negative stock is "
                     "not allowed for this product and/or location.") % (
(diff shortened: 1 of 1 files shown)


stock_no_negative/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     'name': 'Stock Disallow Negative',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
```
pedrobaeza: Don't change the module version. Now this is done by the bot.

pedrobaeza (approved):
The rest seems OK, thanks.

stock_no_negative/__manifest__.py
Abranes: @pedrobaeza Ok!

fuentes73 (approved):

HaraldPanten (approved):
LGTM 👍

PR #661, opened 2020-05-20
Create requirements.txt

ovnicraft:
migrations use library

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+openupgradelib


gurneyalex (approved):
(Red Travis is unrelated)

pedrobaeza:
This is not needed at all, as in Travis openupgradelib is not needed, and for your system, you need to have it, but not for this specific repo.

ovnicraft:
@pedrobaeza so as external library used on modules, must not declared on requirements ?

pedrobaeza:
It's only used for migrations, not for regular operation, so it's not needed to be put on requirements.txt.

PR #662, opened 2020-05-27
[12.0][FIX] stock_request: CI random fails

chienandalu:
When the CI is very fast, the moves could have the same datetime, so the
date asc order would not work as expected

cc @Tecnativa TT24083

### DIFF
--- a/stock_return_request/tests/test_stock_return_request_common.py
+++ b/stock_return_request/tests/test_stock_return_request_common.py
@@ -1,6 +1,7 @@
 # Copyright 2019 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
+from dateutil.relativedelta import relativedelta
+from odoo import fields
 from odoo.tests.common import SavepointCase
 
 
@@ -181,6 +182,12 @@ class StockReturnRequestCase(SavepointCase):
         cls.picking_supplier_2.action_confirm()
         cls.picking_supplier_2.action_assign()
         cls.picking_supplier_2.action_done()
+        # Test could run so fast that the move lines date would be in the same
+        # second. We need to sort them by date, so we'll be faking the line
+        # dates after the picking is done.
+        cls.picking_supplier_2.move_lines.write({
+            'date': fields.Datetime.now() + relativedelta(seconds=1)
+        })
         # We'll have 100 units of each product
         # No we deliver some products
         cls.picking_customer_1 = cls.picking_obj.create({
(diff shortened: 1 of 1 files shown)


HaraldPanten (approved):
small [FIX]. Seems to be OK

pedrobaeza (changes requested):
Can't you write only once in `setUp`? You should also a comment before that lines saying why you do that, or migrating the module in the future these lines will be removed as useless.

chienandalu:
> Can't you write only once in setUp? 

What do you mean?

pedrobaeza:
You are setting several times the date in 4 places. Can you reduce this to once (maybe having good data in `setUp`/`setUpClass`)?

pedrobaeza:
@chienandalu any news about this?

chienandalu:
Sorry. The patch is ugly, but the thing here is that the `date` of the `stock.move.line` is forced to `now()`: https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_move_line.py#L454 so when the test go very fast they get done in the same second, so the expected sort in the test won't behave well. In real life, this is not issue an issue, as it's quite improbable an if it happend it wouldn't be a problem at all.

rousseldenis:
> Sorry. The patch is ugly, but the thing here is that the `date` of the `stock.move.line` is forced to `now()`: https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_move_line.py#L454 so when the test go very fast they get done in the same second, so the expected sort in the test won't behave well. In real life, this is not issue an issue, as it's quite improbable an if it happend it wouldn't be a problem at all.

You can maybe init the date in a variable at the beginning of setupClass and reuse it.

Usually, when dates are important in tests, the good way is mocking fields.Datetime.now()

rousseldenis:
There is also the excellent library to do that : https://pypi.org/project/freezegun/

chienandalu:
> You can maybe init the date in a variable at the beginning of setupClass and reuse it.
> Usually, when dates are important in tests, the good way is mocking fields.Datetime.now()
> There is also the excellent library to do that : https://pypi.org/project/freezegun/

@rousseldenis Good to know, but aren't we overcomplicating a simple solution? A code comment is needed indeed to clarify anyway :sweat_smile:

rousseldenis:
> > You can maybe init the date in a variable at the beginning of setupClass and reuse it.
> > Usually, when dates are important in tests, the good way is mocking fields.Datetime.now()
> > There is also the excellent library to do that : https://pypi.org/project/freezegun/
> 
> @rousseldenis Good to know, but aren't we overcomplicating a simple solution? A code comment is needed indeed to clarify anyway

It was an informative comment :smiley: here, just put the date at the beginning and reuse it.

chienandalu:
Thanks. I think is neater now :)

pedrobaeza:
Travis is red

yajo (approved):

JordiBForgeFlow (approved):

pedrobaeza (approved):

marcelsavegnago (approved):

PR #663, opened 2020-06-02
[13.0] Fix "record does not exist" when validating move line

guewen:
When _action_done is called on a transfer, it may delete a part of
the move lines. The extension of `_action_done()` that assigns
a packaging fails with "Record does not exist or has been deleted".

Check if the if lines still exist before writing on them.


grindtildeath (approved):
:+1:

HaraldPanten (approved):
Seems to be OK 👍

rousseldenis (approved):

jgrandguillaume (approved):
LGTM

PR #664, opened 2020-06-04
[11.0][FIX] product_expiry_simple: fix bad decorator on _action_done()

alexis-via:


alexis-via:
Here is the original _action_done() method https://github.com/odoo/odoo/blob/11.0/addons/stock/models/stock_move_line.py#L369
As you can see, there is no decorator on it... so we should not have @api.model on that method in this module !

HaraldPanten (approved):
👍  Small fix

apollet (approved):

rousseldenis (approved):
Fast one

PR #667, opened 2020-06-11
[10.0][ADD] stock_account_returns_price_difference

pepetreshere:
Hello, we would like to have this module considered into the OCA repositories, thank you very much. 

It creates a journal entry for the price difference when returning goods in average costing method (equivalente function is available in anglo-saxon accounting but not all companies uses the anglo-saxon option)

It is based on methods existing in stock_account module


pepetreshere:
Hello @pedrobaeza, would you please consider the inclusion of this feature in OCA repositories?, or perhaps provide me some guidance on who should I contact instead?. Of course if you also have some technical suggestions I would like to know.

pedrobaeza:
You need reviewers interested in your PR, and I'm afraid I'm not one of them. You can review other PRs in the repo of the same magnitude and ask in exchange that they review yours.

Technically, the module name shouldn't contain plurals.

rousseldenis:
@pepetreshere What's the status of this ?

PR #675, opened 2020-06-26
[12.0][FIX] stock_no_negative: When a transfer is packaged and unpacked, it creates a separate quant

deivislaya:
This causes that when making a transfer, take only one quant and leave
it negative instead of taking it from the other quants of the same
product

Issue related https://github.com/odoo/odoo/issues/53535

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -22,7 +22,9 @@ class StockQuant(models.Model):
         )
         if not check_negative_qty:
             return
-
+        # Execute generic merge method before doing comparisons for avoiding
+        # existing negative quants not merged with their counterparts
+        self._merge_quants()
         for quant in self:
             disallowed_by_product = \
                 not quant.product_id.allow_negative_stock \


pedrobaeza:
I don't see this as good soluiton. This seems because you had negative quants before installing the module, and it's provoking that problem for you. Then the solution would be to not allow the installation of the module in that case.

deivislaya:
It is correct, I have negative quants, before installing the module, but this behavior also occurs once the module is installed and it raises the warning, even when the product is available at the location.

moylop260:
Odoo runs this method using the product button `On Hand`:
  - Screen Shot 2020-06-30 at 15 12 56

The code is here:
 - https://github.com/odoo/odoo/blob/bbdf7488e331533c2dd807ec6a87b14bd9d411b9/addons/stock/models/product.py#L382

If you had a negative stock it process doesn't change that negative value just merge quants that should be merged by Odoo but there are cases where it is not correctly merged.
Even if you didn't have negative quants, see the following example:
 - https://github.com/odoo/odoo/issues/53535

I mean,
If you have 2 quants -10 and +15 it will be merged to +5 using just one quant.
It is not a negative stock and a merge is required before check it in order to avoid raising a false negative.

If you have 2 quants -10 and +6 it will be merged to -4 and the constraint will be raised correctly (even if you run the merge process)

What do you think @JordiBForgeFlow @alexis-via @JayVora-SerpentCS @pedrobaeza ?

JayVora-SerpentCS:
> Odoo runs this method using the product button `On Hand`:
> 
>     * Screen Shot 2020-06-30 at 15 12 56
> 
> 
> The code is here:
> 
>     * https://github.com/odoo/odoo/blob/bbdf7488e331533c2dd807ec6a87b14bd9d411b9/addons/stock/models/product.py#L382
> 
> 
> If you had a negative stock it process doesn't change that negative value just merge quants that should be merged by Odoo but there are cases where it is not correctly merged.
> Even if you didn't have negative quants, see the following example:
> 
>     * [odoo/odoo#53535](https://github.com/odoo/odoo/issues/53535)
> 
> 
> I mean,
> If you have 2 quants -10 and +15 it will be merged to +5 using just one quant.
> It is not a negative stock and a merge is required before check it in order to avoid raising a false negative.
> 
> If you have 2 quants -10 and +6 it will be merged to -4 and the constraint will be raised correctly (even if you run the merge process)
> 
> What do you think @JordiBForgeFlow @alexis-via @JayVora-SerpentCS @pedrobaeza ?

Good catch! Should not the code be added upon unpack() process itself? ( override it and check) ?

stock_no_negative/models/stock_quant.py
```diff
@@ -22,7 +22,7 @@ def check_negative_qty(self):
         )
         if not check_negative_qty:
             return
-
+        self._merge_quants()
```
pedrobaeza, line 27: Add comment for knowing why is here in later references
```suggestion
        # Execute generic merge method before doing comparisons for avoiding
        # existing negative quants not merged with their counterparts
        self._merge_quants()
```

pedrobaeza (commented):
Thanks for the clarifications. I didn't know the internals of that method.

rousseldenis (approved):
LGTM

pedrobaeza (approved):

moylop260 (approved):

PR #679, opened 2020-07-10
[ADD] shipping_expedition: New module to manage shipments made throug by external delivery carrier

victoralmau:
This module allows you to create a new model called: shipping.expedition that is integrated with the Inventory application in order to manage external shipments that are made by external transport companies.
The creation of shipments is allowed from the model: stock.picking according to the carrier defined in it.

It is a base functionality that through other addons will allow integration with different transport companies: Nacex (http://www.nacex.es/), CBL (https://www.cbl-logistica.com/) y TXT (https://txt.es/)
A common model is generated and its data will be used to connect to the corresponding webservice and / or generate the .txt file that the programs need.
Additionally (with the additional addons) the status of the expeditions will be updated.


rousseldenis:
@victoralmau Maybe this should go to https://github.com/OCA/delivery-carrier

pedrobaeza:
And looks similar to https://github.com/OCA/delivery-carrier/pull/266

rousseldenis:
> And looks similar to [OCA/delivery-carrier#266](https://github.com/OCA/delivery-carrier/pull/266)

Yes, and it seems this contain a lot of things that could be split in different modules or outed.

victoralmau:
I have reviewed it and it looks similar in some things but I think that they really are different things: The labels that we generate from the picking to print them from Odoo (own labels with the design that you want regarding a custom addon to print them on our own printer), what is proposed here would be something different; If our company has the shipment of the merchandise externally (through a transport agency such as: ups, mrw, etc.), it needs to generate an expedition (shipment of the merchandise) regarding the picking of which we want to do it in the corresponding transport company. 

In some cases, the web service of the transport company returns information about the expedition that has been created (for example, the delegation of the transport company, the telephone number of the transport company) as well as obtaining the image of the label that We will have to print with the specific printer, in other cases it will be necessary to generate a customized .txt file for each transport company to send that file to the computer with the specific program of the transport company so that it is, generates the expedition. 

All this helps to avoid having to manually create the data of each picking on each website of each transport company and we can know what state they are in. It is proposed to create another model and have it separate (although linked to picking) since it is intended that they be linked to the user_id of the corresponding sale order, so that users can check the status of shipments, if there are incidents, auto- send email of the status of the expedition (it is developed in another addon linked to this that would have to be uploaded), create an activity to review the incidence, define the cost of each expedition in order to be able to get a report of them or, for example, be able to define ex-post (when we know the cost of the house shipped according to the supplier's invoice issued by the carrier) the purchase price (and therefore the margin) and recalculate the margin of the transport line of the linked order. 

All these functionalities that I indicate are those that are developed (as well as each addon for each transport company) in other addons (pending upload) linked against the one I propose.

pedrobaeza:
That's the same that is done in subsequent modules that depends on the other mentioned module. Examples: https://github.com/OCA/l10n-spain/pull/1382 and https://github.com/OCA/l10n-spain/pull/1392.

And also the modularity is a thing, as this include too much stuff (and thus, dependencies), in a "all-in-one" bundle, so you take all or nothing.

victoralmau:
Ok, perfect, with the seur example I have seen that it is exactly the same that I propose, sorry, so then it will be much better to adapt the specifics of: TXT, Nacex and Cbl so that they adapt to delivery_state and work on what already exists. 

I did not know that this addon existed (perhaps because we started creating it for V10).

pedrobaeza:
Yes, indeed it has been a pity that this has been around outside OCA, but we did a code sprint in May and the result was the normalization of this and the proposal of all the existing addons. For CBL and Nacex, if they are specific for Spain, `delivery_cbl` and `delivery_nacex` will be interesting to be extracted and depending on the other module.

PR #682, opened 2020-07-14
[12.0][ADD] stock_picking_sale_order: New module to add some fields in pickings from sale orders

victoralmau:
Add the order_id and confirmation_date_order fields in the picking that are defined when validate sale order.


pedrobaeza:
This is similar to https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_picking_sale_order_link

victoralmau:
Although in this case the fields were added in the model regarding the linked sale_order and not the button to access it, I have seen that the relationship exists:

**V12**
stock.picking> group_id (procurement.group)> sale_id (sale.order)

**V10**
stock.picking> group_id (procurement.group)> procurement_ids (procurement.order)> sale_line_id (sale.order_line)

As this relationship already exists natively in Odoo this addon should not exist and those that depend on it should adapt to get in sale_id of each stock_picking according to the correct way.

Thank you.

pedrobaeza:
But there's no need for extra fields. `sale_stock` already includes the needed field (`sale_id`).

PR #687, opened 2020-07-20
[13.0] Add stock_putaway_by_route

guewen:
# Stock Putaway Hook

Add hooks allowing modules to add more putaway strategies.
It adds hooks to the core putaway method
``StockLocation._get_putaway_strategy()`` allowing to plug other
strategies and makes the selector fields in the tree views dynamic
(required/readonly).

# Stock Putaway By Route

Based on stock_putaway_hook.

Add the possibility to apply a putaway rule based on a route.

For instance, a move generated for a replenishment (e.g. using the module
``Stock Orderpoint Route`` or ``Stock Buffer Route``) will be related to the
rule (so the route) that generated it. When the putaway rules are applied, if a
rule is defined on the replenishment route, the destination move will be changed
to the one of the rule.

### DIFF
--- a/setup/stock_putaway_by_route/odoo/addons/stock_putaway_by_route
+++ b/setup/stock_putaway_by_route/odoo/addons/stock_putaway_by_route
@@ -1,0 +1,1 @@
+../../../../stock_putaway_by_route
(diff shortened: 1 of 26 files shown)


sebalix (approved):
Code review only

guewen:
Sorry, I didn't mean to merge. I can revert if anyone wants

grindtildeath:
@guewen Although late, I still did the review, and have nothing to add. Good job :+1:

PR #689, opened 2020-07-27
[13.0] Create module delivery_package_default_shipping_weight

grindtildeath:
This module allows to define a Default shipping weight on Delivery packages
(`product.packaging`) in order to set automatically the Shipping weight on the
packages (`stock.quant.package`) using this Package type.


delivery_package_default_shipping_weight/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Delivery Package Default Shipping Weight",
+    "summary": "Set default package shipping weight according to packaging",
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
```
rousseldenis: Hi @grindtildeath, thansk for this.

Is it the norm at your side to make all modules 'Alpha' ?

I don't see the needs here as this is a tiny module with few implications.

grindtildeath: Hell @rousseldenis ,

I guess it depends on the contributor, and personally I tend to default on using "Alpha".

Thing is, specs on the WMS can change, and we're not sure if this module will be still be used, changed or dropped in the next couple of weeks.

But I have no issue to use a more mature level for this module anyway :wink:

rousseldenis: > Hell @rousseldenis ,
> 
> I guess it depends on the contributor, and personally I tend to default on using "Alpha".
> 
> Thing is, specs on the WMS can change, and we're not sure if this module will be still be used, changed or dropped in the next couple of weeks.
> 
> But I have no issue to use a more mature level for this module anyway

Ok, but independently of your specs, I tend to think 'Alpha' should not prevail as a 'default'. This IMHO generates some noise through OCA repos. Even if maturity level has been asked by lot of people, I hope this would not lead to have a lot of 'dead' modules through major branches and no one going back to fix, update and get them to 'Production/Stable' stage.

TDu (approved):

guewen (approved):

guewen:
Calm down ocabot. Looks like it's stuck in a loop.

guewen:
Congrats @OCA-git-bot, you did it!

PR #690, opened 2020-08-07
[FIX] stock_putaway_by_route: AttributeError: 'stock.move.line' object has no attribute 'rule_id'

sebalix:

PR #692, opened 2020-08-12
[13.0] Add match by product's route in putaway by routes

guewen:
The module matching a putaway rule by a route can now:

* find a putaway rule by the route related to a stock move
* or a route related to the product on the move

The changes required to enable this feature:

* In stock_putaway_hook: in the part that match the value, support "in"
  when the value is a container (recordset, list, tuple), so we can pass
  product.route_ids (and match any route set on a put-away rule)
* In stock_putaway_by_route, in the various parts where
  `StockLocation._get_putaway_strategy()` is called, propagate the
  product's routes as well
* In StockMoveLine.onchange_product_id, remove the computation based on
  self.move_id: it cannot change.

### DIFF
--- a/stock_putaway_by_route/models/stock_move.py
+++ b/stock_putaway_by_route/models/stock_move.py
@@ -10,6 +10,8 @@ class StockMove(models.Model):
     def _generate_serial_move_line_commands(self, lot_names, origin_move_line=None):
         if self.rule_id.route_id:
             self = self.with_context(_putaway_route_id=self.rule_id.route_id)
+        elif self.product_id.route_ids:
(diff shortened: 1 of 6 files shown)


guewen:
@rousseldenis regarding your comment on https://github.com/OCA/stock-logistics-workflow/pull/690#pullrequestreview-465837415, thinking about how to write a test revealed that it was in fact something that can't happen, so I removed this part. In this new PR, I had to do a similar thing in the onchange based on the product, which can change on new records edited manually, added a test this time

jgrandguillaume (approved):
LGTM, functional test works fine

rousseldenis (approved):

PR #693, opened 2020-08-13
[13.0] Create module sale_stock_mto_as_mts_orderpoint - alpha

grindtildeath:
This module aims to materialize the triggering of another stock rule through
an orderpoint in the procurement of MTO products from a sale order.

When a sales order is confirmed, if the delivery order move is generated by
the stock rule linked to the standard MTO route, and the MTO route is marked
on the product, an orderpoint will be created on the Stock location of the
warehouse, with min/max quantities of zero.

This allows to regenerate procurement according to the procure_recommended_qty
using module `stock_orderpoint_manual_procurement`.

Finally, orderpoints with min/max quantities of zero will be archived if the
MTO route is removed on the product.


jgrandguillaume (approved):
LGTM, functional test worked fine

guewen (approved):

guewen:
Rebased and squashed the fixup.

/ocabot merge patch

PR #700, opened 2020-09-02
[FIX] stock_picking_auto_create_lot: Avoid unreserved move lines to fix error when quantity done is mora then initial demand

sergio-teruel:
cc @Tecnativa TT25365

### DIFF
--- a/stock_picking_auto_create_lot/models/stock_picking.py
+++ b/stock_picking_auto_create_lot/models/stock_picking.py
@@ -25,4 +25,8 @@ class StockPicking(models.Model):
                 line.lot_id = self.env["stock.production.lot"].create(
                     {"product_id": line.product_id.id, "company_id": line.company_id.id}
                 )
+            if self.picking_type_id.code == "incoming":
+                return super(
+                    StockPicking, self.with_context(do_not_unreserve=True)
+                ).button_validate()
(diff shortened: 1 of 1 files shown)


marcelsavegnago (approved):

pedrobaeza (changes requested):
Why do you need such context? Include in the commit/PR message steps to reproduce the problem, as I don't get why it's needed and why it's only applied on incoming pickings.

victoralmau (commented):
I don't understand this code add.
I found this in Odoo code https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_move.py#L452 but i don't understand the problem that exist to skip this to lot creation.

sergio-teruel:

- Create a storable product and enable "Auto create lot".
- Enable "Auto create lot" in incoming picking operation type.
- Create an incoming picking with 10 units as initial demand
- Mark as todo
- Set quantity done to more quantities than initial demand (15 units) 
- Validate the picking

Error : "You need to supply a Lot/Serial number for product ..."

**Sorry!! Sorry!!**

... After a long time to research the problem..
This occours if stock_move_line_auto_fill module is installed.
Fixed in this PR: https://github.com/OCA/stock-logistics-workflow/pull/704

PR #701, opened 2020-09-15
[13.0] Fix groups of transfers on sales

guewen:
When grouping transfers of different sales orders in the same
procurement.group, we have an issue with only the first sales order
showing the "Transfers" button on the SO view, because
"sale.picking_ids" returns an empty recordset on the other sales orders.

This is because the link between sales and pickings is done through the
"group_id.sale_id" field of the stock moves. When we group transfers of
different sales, we end up with several group_id in the moves of the
same picking, which is a bit unexpected, and when new transfers are
created (dynamic routing, backorders, ...), they'll only get the first
group_id and be attached only to the first sale order.

The solution implemented here is the following:

* Add a M2m "sale_ids" field on the procurement group
* When several moves with different groups are merged into the same
  group, clone the group, update its "sale_ids" field to contain all the
  sales (sale_id will keep the first one though) and set in on the moves
* Update the computed fields for the relation between the sales and
  pickings to use this M2M field
* Keep the original group in the stock move as "original_group_id": used
  when a sales order is cancelled

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_move.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_move.py
@@ -1,33 +1,72 @@
 import re
 from collections import namedtuple
 
-from odoo import models
+from odoo import fields, models
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    # store the first group the move was in when created, used for cancellation
+    # from a sales order
+    original_group_id = fields.Many2one(
+        comodel_name="procurement.group", string="Procurement Group",
+    )
+
     def _assign_picking(self):
-        return super(
+        result = super(
             StockMove, self.with_context(picking_no_overwrite_partner_origin=1)
         )._assign_picking()
+        self._on_assign_picking_merge_group()
+        return result
 
     def _assign_picking_post_process(self, new=False):
         res = super()._assign_picking_post_process(new=new)
         if not new:
-            picking = self.mapped("picking_id")
-            picking.ensure_one()
-            sales = self.mapped("sale_line_id.order_id")
-            for sale in sales:
-                pattern = r"\b%s\b" % sale.name
-                if not re.search(pattern, picking.origin):
-                    picking.origin += " " + sale.name
-                    picking.message_post_with_view(
-                        "mail.message_origin_link",
-                        values={"self": picking, "origin": sale},
-                        subtype_id=self.env.ref("mail.mt_note").id,
-                    )
+            self._on_assign_picking_message_link()
         return res
+
+    def _prepare_merge_group_values(self, sales):
+        return {
+            "sale_ids": [(6, 0, sales.ids)],
+            "name": ", ".join(sales.mapped("name")),
+        }
+
+    def _on_assign_picking_merge_group(self):
+        moves = {}
+        for move in self:
+            moves.setdefault(move.picking_id, self.browse())
+            moves[move.picking_id] |= move
+        for picking, moves in moves.items():
+            if not picking.picking_type_id.group_pickings:
+                return
+            base_group = picking.group_id
+            base_sales = base_group.sale_ids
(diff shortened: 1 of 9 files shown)


simahawk (approved):

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -1,34 +1,76 @@
...
+from odoo import fields, models
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    # store the first group the move was in when created, used for cancellation
+    # from a sales order
+    original_group_id = fields.Many2one(
+        comodel_name="procurement.group", string="Procurement Group",
```
jbaudoux:
```suggestion
        comodel_name="procurement.group", string="Procurement Group Original",
```

jbaudoux: To prevent having 2 fields on stock.move with the same name

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -1,34 +1,76 @@
...
+    def _on_assign_picking_merge_group(self):
+        moves = {}
+        for move in self:
+            moves.setdefault(move.picking_id, self.browse())
+            moves[move.picking_id] |= move
+        for picking, moves in moves.items():
+            if not picking.picking_type_id.group_pickings:
+                return
+            base_group = picking.group_id
+            base_sales = base_group.sale_ids
```
jbaudoux: This can contain sales that are already done. For example procurement group is linked to SO1 & SO2. SO1 is delivered. Back order contains only SO2 but procurement group is still linked to SO1 & SO2. When SO3 occurs, you have in the backorder SO2 & SO3 but the base group contains SO1 & SO2. The new group should only contain SO2 & SO3 without SO1. Otherwise, on the link from SO1 you see too much pickings (and all the one coming in the future) and the proc group link and name will grow too much.

jbaudoux (changes requested):

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -26,34 +28,40 @@ def _assign_picking_post_process(self, new=False):
             self._on_assign_picking_message_link()
         return res
 
-    def _prepare_merge_group_values(self, sales):
+    def _prepare_merge_group_values(self, move_groups):
+        sales = move_groups.sale_id
```
jbaudoux: mapped?

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -26,34 +28,40 @@ def _assign_picking_post_process(self, new=False):
...
+                continue
             base_group = picking.group_id
-            base_sales = base_group.sale_ids
             # If we have different sales in the line's group, it means moves
             # have been merged in the same picking/group but they come from a
             # different sale.
-            line_sales = moves.group_id.sale_id
-            all_sales = base_sales | line_sales
+            moves = picking.move_lines
+            moves_groups = moves.original_group_id
```
jbaudoux: mapped?

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -26,34 +28,40 @@ def _assign_picking_post_process(self, new=False):
...
+            moves_sales = moves_groups.sale_id
```
jbaudoux: mapped?

stock_picking_group_by_partner_by_carrier/models/stock_move.py
guewen: no longer necessary on relations in 13.0 :)

sebalix (approved):

yvaucher (approved):

PR #703, opened 2020-09-27
[12.0][ADD] stock_landed_costs: Add support to avco costs method.

marcelsavegnago:


pedrobaeza:
I suppose this is your alternative for not using `purchase_landed_cost`, isn't it? Is there any feature missing in the other for using the core one?

marcelsavegnago:
@pedrobaeza I really liked the purchase_landed_cost and what I missed was the generation of the accounting entry in the stock journal. Stock_landed_cost does the journal entry but does not update the average price. Talking with @rvalyi , i commented that a good option is to implement yet another new method of updating the price (today we only have the price update directly). What do you think ?

cc @renatonlima

pedrobaeza:
Yes, indeed it can be another option. Till now, we hadn't got that need, as stock accounting is not required here in our country. In v13, the support for AVCO has been added to core `stock_landed_costs`, so I will study as well if we can switch to that one and deprecated `purchase_landed_cost`, so for now I don't have clear the long term vision for this set of modules.

marcelsavegnago:
In Brazil it is necessary. I had not seen that in V13 there was already support for AVCO and I am happy to know that 😄 .

I'm going to take a look at V13 and study purchase_landed_cost a little more.

Anyway, what do you think about following this PR to include support for the average price in stock_landed_cost?

pedrobaeza:
Yeah, totally if you think so. I try to focus in the minimum set of modules, and that's why I say about analyzing situation:

- Study v13 `stock_landed_costs`.
- If not enough, expand it through a `stock_landed_costs_expanded` or similar or migrate `purchase_landed_cost`?
- If enough, use for new v12 customers that need landed cost the module in this PR and ease migration path? For that, it's interesting that you harmonize this code with v13 one.
- If enough, prepare migration scripts for current v12 customers with `purchase_landed_cost` for transitioning to v13 core module.

marcelsavegnago:
@pedrobaeza Thank you for your help, information and guidance.

PR #705, opened 2020-09-30
[11.0][ADD]stock_push_delay v11

AaronHForgeFlow:
Some times we done want to apply push rules right away. For example, when receiving partial shipments in several steps, we want to trigger manually the picking form the packing zone to Input and to Stock. That helps to track better quantities on chained moves, because in many use cases, the pickings are not completed entirely in one step, and specific actions are required depending on the receipt location, among other reasons.

@ForgeFlow


stock_push_delay/models/stock.py
```diff
@@ -0,0 +1,25 @@
...
+
+class StockPicking(models.Model):
+
+    _inherit = 'stock.picking'
+
+    @api.multi
+    def action_done(self):
+        res = super(StockPicking, self).action_done()
+        for picking in self:
+            picking.move_lines.with_context(manual_push=True)._push_apply()
```
rousseldenis: Don't do that with context, so bad for performances.

Either you create another method (e.g.: _push_apply_manual()) , either you put context on self.

rousseldenis (changes requested):

rousseldenis (approved):
Can you squash ?

stock_push_delay/models/stock.py
AaronHForgeFlow: Thanks for the comment, did not realize about that performance issue.

JordiBForgeFlow (approved):

AdriaGForgeFlow (approved):

PR #709, opened 2020-10-06
[13.0] Fix package dimensions overriden on write

guewen:
When a packaging is auto-assigned on a package, it should not
override any previously set dimensions.

### DIFF
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -35,17 +35,28 @@ class StockQuantPackage(models.Model):
         if self.env.context.get("_auto_assign_packaging") and vals.get(
             "product_packaging_id"
         ):
-            for package in self:
-                package.onchange_product_packaging_id()
+            self._update_dimensions_from_packaging(override=False)
         return res
+
+    def _update_dimensions_fields(self):
+        # source: destination
+        return {
+            "lngth": "lngth",
+            "width": "width",
+            "height": "height",
+            "max_weight": "pack_weight",
+        }
+
+    def _update_dimensions_from_packaging(self, override=False):
+        for package in self:
+            if not package.product_packaging_id:
+                continue
+            dimension_fields = self._update_dimensions_fields()
+            for source, dest in dimension_fields.items():
+                if not override and package[dest]:
+                    continue
+                package[dest] = package.product_packaging_id[source]
 
     @api.onchange("product_packaging_id")
     def onchange_product_packaging_id(self):
-        if self.product_packaging_id:
-            vals = self.product_packaging_id.read(
-                fields=["lngth", "width", "height", "max_weight"]
-            )[0]
-            vals["pack_weight"] = vals["max_weight"]
-            vals.pop("id")
-            vals.pop("max_weight")
-            self.update(vals)
+        self._update_dimensions_from_packaging(override=True)
--- a/stock_quant_package_dimension/tests/__init__.py
+++ b/stock_quant_package_dimension/tests/__init__.py
@@ -1,0 +1,1 @@
+from . import test_package_dimension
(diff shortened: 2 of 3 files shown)


stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -35,17 +35,28 @@ def write(self, vals):
...
+            "height": "height",
+            "max_weight": "pack_weight",
+        }
+
+    def _update_dimensions_from_packaging(self, override=False):
+        for package in self:
+            if not package.product_packaging_id:
+                continue
+            dimension_fields = self._update_dimensions_fields()
+            for source, dest in dimension_fields.items():
```
lmignon, line 55: @guewen dimensions fields are not required on the packaging, that means that if your previous package has for exemple no `max_height` defined but all the other dimensions fields you'll update the  `pack_weight` field on the `stock.quant.package`with the new value but not the others fields....

guewen: @lmignon are you sure? fields are updated independently

guewen: note: I'm tired so I may miss even obvious things :sleepy:

lmignon: :smirk:

guewen: reading again: 

> you'll update the pack_weight field on the stock.quant.package with the new value but not the others fields

Not sure what you mean in fact.
The goal here is:
* if you change the packaging from the UI, all the dimensions are overridden from the packaging (even if they are zero) which match the general behavior of onchanges
* if the packaging is changed through the "auto assign" mechanism (`_auto_assign_packaging` in context), only the package fields != 0 are set from the packaging, as we don't want to silently erase previously set dimensions

lmignon: > if the packaging is changed through the "auto assign" mechanism (_auto_assign_packaging in context), only the package fields != 0 are set from the packaging, as we don't want to silently erase previously set dimensions

In this case. If the  `pack_weight` field on the `stock.quant.package` is 0 and the value of `max_height` of the new package provided by the "auto assign" mechanism is 234, you'll update the `pack_weight` field. BUT if the others dimensions fields on the `stock.quant.package` are > 0, these are not updated. IOW you'll mix the value from the original package and the new one for all dimensions fields with value 0 into the original package that are not 0 into the "auto assigned" package. 

IMO, you should prevent the update of the dimension fields as of one of the dimensions fields has a value <> 0.

```python
if not override and any(package[f] for f in dimension_fields):
    continue
for source, dest in dimension_fields.items():
    package[dest] = package.product_packaging_id[source]
```

guewen: > IOW you'll mix the value from the original package and the new one for all dimensions fields with value 0 into the original package that are not 0 into the "auto assigned" package.

Ok, I see your point now. I don't know what is more correct though. If for instance you use a (hardware) scale that sets the weight but not the other dimensions, maybe what you would expect is indeed to mix the values to complete what the scale didn't put on the package? Like, the dimensions on the packaging acting as sort of default values?

@jgrandguillaume your thoughts?

jgrandguillaume: @guewen @lmignon Dimension on the packaging absolutely act as a default values IMO for the very exact example you provided (using a weight machine for example).

guewen: I amended the commit to add more details in the body.

lmignon: Thank you for your thoughts @jgrandguillaume I have a better understanding of the functional context behind this change now.

lmignon (approved):

jgrandguillaume (approved):
LGTM like this from m perspective

PR #711, opened 2020-10-07
[13.0][FIX] stock_picking_sale_order_link: New sale line added to done picking

carlosdauden:
if access from picking to sale with smartbutton

@Tecnativa TT26009

### DIFF
--- a/stock_picking_sale_order_link/models/stock_picking.py
+++ b/stock_picking_sale_order_link/models/stock_picking.py
@@ -12,4 +12,9 @@ class StockPicking(models.Model):
         of given picking.
         """
         self.ensure_one()
-        return self.sale_id.get_formview_action()
+        # Remove default_picking_id to avoid defaults get
+        # https://github.com/odoo/odoo/blob/e4d22d390c8aa8edf757e36704a9e04b2b89f115/
+        # addons/stock/models/stock_move.py#L410
+        ctx = self.env.context.copy()
+        ctx.pop("default_picking_id", False)
+        return self.with_context(ctx).sale_id.get_formview_action()


stock_picking_sale_order_link/models/stock_picking.py
```diff
@@ -12,4 +12,7 @@ def action_view_sale_order(self):
         of given picking.
         """
         self.ensure_one()
-        return self.sale_id.get_formview_action()
+        # Remove default_picking_id to avoid defaults get
+        # https://github.com/odoo/odoo/blob/e4d22d390c8aa8edf757e36704a9e04b2b89f115/
+        # addons/stock/models/stock_move.py#L410
+        return self.with_context(default_picking_id=False).sale_id.get_formview_action()
```
pedrobaeza: Better to simply remove the key as others do, as sometimes the inexistence of the key is not the same than having it with False:
```suggestion
        ctx = self.env.context.copy()
        ctx = ctx.pop(default_picking_id, False)
        return self.with_context(ctx).sale_id.get_formview_action()
```

victoralmau (changes requested):
Exist travis error related to this change (https://travis-ci.org/github/OCA/stock-logistics-workflow/jobs/733630849#L1307), maybe because `sale_id` field may no't be defined (`False` value)
What dou you think about it?

carlosdauden:
@victoralmau travis error was solved here:
https://github.com/OCA/stock-logistics-workflow/compare/9126e11534374d0f37529004c304def59812890f..1b3c0cb452532e92073dffe03ec8c8470c6b343d

pedrobaeza:
Sorry for my typo in the suggestion... :cry:

carlosdauden:
Travis green

PR #714, opened 2020-10-15
[ADD] stock_split_picking: Mode selection to split in different ways

hbrunn:
this adds two additional modes of splitting, either every move in its own picking, or some selected moves into a new picking.

As the original is the default choice, we end up with the same amount of clicks for splitting by done quantity as we had before.

### DIFF
--- a/stock_split_picking/__init__.py
+++ b/stock_split_picking/__init__.py
@@ -2,3 +2,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import models
+from . import wizards
--- a/stock_split_picking/__manifest__.py
+++ b/stock_split_picking/__manifest__.py
@@ -17,6 +17,7 @@
         'stock',
     ],
     'data': [
+        'wizards/stock_split_picking.xml',
         'views/stock_partial_picking.xml',
     ],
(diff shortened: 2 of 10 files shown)


StefanRijnhart (approved):
Of course, it's making me unreasonably mad that TransientModels are not put in <module>/models, and especially that their views are not put in <module>/views but after some investigation I have to conclude that the practices in core Odoo as well as the OCA are split in half concerning the issue so that gives me no reason to disapprove this PR. Maybe people who are old enough to remember the old OpenERP 5 wizard classes, and don't want to be reminded of them ever again, need to go do something else. This looks very good otherwise.

hbrunn:
I'm also not super happy about this separation, but I'm happy with conventions, and they [say so](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#14directories).

rousseldenis (approved):
Code review

rousseldenis:
@hbrunn Ready ?

hbrunn:
@rousseldenis ready for what?

rousseldenis:
Merge ;-)

hbrunn:
ah, yes

PR #716, opened 2020-10-19
[12.0][ADD] stock_picking_tier_validation

TimLai125:
This module extends the functionality of Pickings to support a tier validation process.
The original author the module is https://github.com/sistemasdedatos/odoo-addons, yet the module is generic similar to [purchase_request_tier_validation](https://github.com/OCA/purchase-workflow/tree/12.0/purchase_request_tier_validation).


stock_picking_tier_validation/models/stock_picking.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class Picking(models.Model):
+    _name = "stock.picking"
+    _inherit = ["stock.picking", "tier.validation"]
+    _state_from = ["draft"]
+    _state_to = ["waiting", "confirmed", "assigned", "done"]
```
yostashiro:
```suggestion
    _state_from = ["draft", "waiting", "confirmed", "assigned"]
    _state_to = ["done"]
```
@TimLai125 Shouldn't this part be like this?

TimLai125: Yes, we should include other states as well since picking created from orders (PO/SO etc.) may already be proceed to assigned state in the beginning. Thanks for the comment.

TimLai125:
base_tier_validation does not support state field as a compute field. Closing the PR now, please feel free to re-open it once there is a workaround.

TimLai125:
> base_tier_validation does not support state field as a compute field. Closing the PR now, please feel free to re-open it once there is a workaround.

I am adding an extra constrains method to the stock move model as workaround. Picking that tries to move to the `_state_to` without reviewers will be blocked by the constrains.

stock_picking_tier_validation/views/stock_picking_views.xml
```diff
@@ -0,0 +1,109 @@
...
+        <field name="name">stock.picking.internal.search.tier</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_internal_search" />
+        <field name="arch" type="xml">
+            <filter name="activities_my" position="before">
+                <filter
+                    name="needs_review"
+                    string="Needs my Review"
+                    domain="[('reviewer_ids','in',uid), ('state', 'not in', ['done'])]"
+                    help="My Picking to review"
```
yostashiro, line 98:
```suggestion
                    help="Pickings to review"
```

stock_picking_tier_validation/views/stock_picking_views.xml
```diff
@@ -0,0 +1,109 @@
...
+    </record>
+    <record id="tier_view_picking_internal_search" model="ir.ui.view">
+        <field name="name">stock.picking.internal.search.tier</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_internal_search" />
+        <field name="arch" type="xml">
+            <filter name="activities_my" position="before">
+                <filter
+                    name="needs_review"
+                    string="Needs my Review"
```
yostashiro, line 96:
```suggestion
                    string="Need My Review"
```

stock_picking_tier_validation/models/stock_picking.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2020 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class Picking(models.Model):
+    _name = "stock.picking"
+    _inherit = ["stock.picking", "tier.validation"]
+    _state_from = ["draft", "waiting", "confirmed", "assigned"]
```
yostashiro, line 10:
```suggestion
    _state_from = ["waiting", "confirmed", "assigned"]
```
Validate button doesn't show when the state is draft.

yostashiro (changes requested):
@TimLai125 I have created this PR https://github.com/OCA/server-ux/pull/230 and once it's merged, the adjustments can be made simpler like below, instead of having the constraint on stock.move.

```.py
from odoo import models


class Picking(models.Model):
    _name = "stock.picking"
    _inherit = ["stock.picking", "tier.validation"]
    _state_from = ["waiting", "confirmed", "assigned"]

    def _check_state_conditions(self, vals):
        self.ensure_one()
        if self.move_line_ids.filtered(lambda x: x.state == "done"):
            return True
        return False
```

Ryoko04 (approved):
This module makes it possible to approve delivery order, receipt order and transfer order before validating the order.
I ran this module on runbot and worked fine!

rousseldenis:
@TimLai125 Please check runbot warnings :

image

PR #718, opened 2020-10-21
[14.0] stock_picking_invoice_link migration

mileo:


rousseldenis:
@mileo Can you rebase ?

joao-p-marques:
As this seems to be stopped, I will be superseeding this PR with some changes that are missing, and also including the latest commits as requested. ping @pedrobaeza

PR #726, opened 2020-11-05
[REF] stock_no_negative: Make inheritable no-negative check

luisg123v:
This puts the code to check if negative inventory is allowed for a
given location into a separate method, so it may be reused or inherited.

### DIFF
--- a/stock_no_negative/models/stock_location.py
+++ b/stock_no_negative/models/stock_location.py
@@ -2,7 +2,7 @@
 # @author Jordi Ballester <jordi.ballester@eficent.com.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models, fields
+from odoo import api, models, fields
 
 
 class StockLocation(models.Model):
@@ -12,3 +12,8 @@ class StockLocation(models.Model):
         string='Allow Negative Stock',
         help="Allow negative stock levels for the stockable products "
         "attached to this location.")
+
+    @api.multi
+    def _is_negative_allowed(self):
+        self.ensure_one()
+        return self.allow_negative_stock or self.usage not in ['internal', 'transit']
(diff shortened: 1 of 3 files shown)


luisg123v:
@moylop260 @Yajo,

Could you review this, please?

luisg123v:
CC @hugho-ad @dsabrinarg

stock_no_negative/models/stock_location.py
```diff
@@ -12,3 +12,7 @@ class StockLocation(models.Model):
         string='Allow Negative Stock',
         help="Allow negative stock levels for the stockable products "
         "attached to this location.")
+
+    @api.multi
+    def _is_negative_allowed(self):
+        return self.allow_negative_stock or self.usage not in ['internal', 'transit']
```
rousseldenis, line 19: I would have added self.ensure_one() here or would have implemented a filtered() on a real multi recordset

rousseldenis (changes requested):
LGTM. Just a little change.

stock_no_negative/models/stock_location.py
hugho-ad, line 19: Why or?

it should be AND because the old code 
`quant.location_id.usage in ['internal', 'transit'] and disallowed_by_product and disallowed_by_location`

luisg123v: @hugho-ad it's the opposite code. The one you reference is to check if it's disallowed, and here is to determine if it's allowed.

luisg123v: Done, added `ensure_one` to be [consistent with Odoo](https://github.com/odoo/odoo/blob/f845124535e2c882656bf0ff212dcd539d004ae3/addons/stock_account/models/stock.py#L227)

hugho-ad (approved):

hugho-ad:
LGTM :+1:

rousseldenis (approved):

luisg123v:
Ouch, it seems I don't have merge permissions.

@pedrobaeza or @moylop260, could you help me, please?

stock_no_negative/models/stock_location.py
moylop260: FYI it doesn't have ensure_one

Is it fine?

luisg123v: @moylop260 You are absolutely right! It seems I forgot to push :confused:

PR #728, opened 2020-11-05
[13.0][IMP] stock_picking_auto_create_lot: Improve performances and UX

rousseldenis:
* Create lots using create_multi to reduce queries in case of large pickings.
* Hide the serial assignement for products that are not enabled for it.
* Improve tests

image

PR #729, opened 2020-11-06
[13.0][ADD] stock_picking_auto_create_lot_serial_only

rousseldenis:
Includes:

* [ ] #728


stock_picking_auto_create_lot_serial_only/views/product_template.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field
+            name="inherit_id"
+            ref="stock_picking_auto_create_lot.view_template_property_form"
+        />
+        <field name="arch" type="xml">
+            <field name="tracking" position="attributes">
+                <attribute
+                    name="attrs"
+                >{'invisible':[('tracking', '!=', 'serial')]}</attribute>
+            </field>
```
sergio-teruel, line 19: By default tracking field is none... so the field in view is invisible... How do I can change this value if the field is invisible.
Selección_029

ThomasBinsfeld (approved):

ThomasBinsfeld:
Rebased on 13.0

PR #740, opened 2020-11-26
Add stock_picking_group_by_partner_by_carrier_by_date

simahawk:


stock_picking_group_by_partner_by_carrier_by_date/models/stock_move.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def _domain_search_picking_for_assignation(self):
+        domain = super()._domain_search_picking_for_assignation()
+        cutoff_domain = self._domain_search_picking_for_assignation_cutoff()
+        domain.extend(cutoff_domain)
+        return domain
+
+    def _skip_search_picking_for_assignation_cutoff(self):
+        return (
+            self.picking_type_id.code != "outgoing"
```
simahawk: maybe to specify that it's only for delivery? Or make it configurable at picking type level?

jgrandguillaume (approved):
LGTM thanks

stock_picking_group_by_partner_by_carrier_by_date/models/stock_move.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def _domain_search_picking_for_assignation_cutoff(self):
+        domain = []
+        if self._skip_search_picking_for_assignation_cutoff():
+            return domain
+        date_planned = self.date_expected
+        date_planned_end = date_planned.replace(hour=23, minute=59, second=59)
+        domain = [
+            ("scheduled_date", "<=", date_planned_end),
+            ("scheduled_date", ">=", date_planned),
```
jbaudoux, line 31: @jgrandguillaume Are you sure about this?

jbaudoux: Seems good

jbaudoux (approved):

stock_picking_group_by_partner_by_carrier_by_date/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+{
+    "name": "Stock Picking: group by partner and carrier and scheduled date",
+    "Summary": """Delivery orders will be matched by date too.
+    """,
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "category": "Warehouse Management",
+    "depends": ["stock_picking_group_by_partner_by_carrier"],
```
guewen, line 16: A remark which may or may not apply here. `stock_picking_group_by_partner_by_carrier` adds a hook to customize the domain used in `_assign_picking`, which forces us to depend on `stock_picking_group_by_partner_by_carrier` if we want to customize the domain. Which was maybe necessary in your case? Anyway, not in my case :p, so at some point, it would be good to extract the hook in a base module, which is extended in `stock_picking_group_by_partner_by_carrier`, maybe in `stock_picking_group_by_scheduled_date` (?) and so on.

simahawk: I'll add a TODO.

stock_picking_group_by_partner_by_carrier_by_date/models/stock_move.py
simahawk: I add a TODO for this as well

stock_picking_group_by_partner_by_carrier_by_date/__manifest__.py
simahawk: done

stock_picking_group_by_partner_by_carrier_by_date/models/stock_move.py
simahawk: done

stock_picking_group_by_partner_by_carrier_by_date/models/stock_move.py
```diff
@@ -0,0 +1,33 @@
...
+            or not self.picking_type_id.group_pickings
+            or self.group_id.sale_id.picking_policy == "one"
+        )
+
+    def _domain_search_picking_for_assignation_by_date(self):
+        domain = []
+        if self._skip_search_picking_for_assignation_by_date():
+            return domain
+        date_planned = self.date_expected
+        date_planned_end = date_planned.replace(hour=23, minute=59, second=59)
```
jbaudoux, line 28: This is not timezone safe. date_planned is in UTC. It should be end of day in the timezone of the warehouse.
Something like:
tz_date_planned = date_planned.astimezone(tz_wh)
tz_date_planned_end = tz_date_planned.replace(hour=23, minute=59, second=59)
date_planned_end = tz_utils.tz_to_utc_naive_datetime(tz_wh, tz_date_planned_end)

jbaudoux (changes requested):

stock_picking_group_by_partner_by_carrier_by_date/models/stock_move.py
simahawk: @jbaudoux when the cutoff module is installed all expected dates will be set considering the wh tz already.
Here I'm using a date that is already matching its own WH TZ. Am I wrong?

jbaudoux: As discussed, I'll approve this PR as you prepare the fix in another coming PR

jbaudoux (approved):

PR #741, opened 2020-11-26
[13.0] stock_picking_group_by_partner_by_carrier_by_date cutoff glue module

simahawk:
Includes https://github.com/OCA/stock-logistics-workflow/pull/740
Depends on https://github.com/OCA/sale-workflow/pull/1104


jgrandguillaume (approved):
LGTM thx

guewen:
> 1. stock_picking_group_by_partner_by_carrier
> 2. sale_cutoff_time_delivery
> 3. stock_picking_group_by_partner_by_carrier_by_date
>
> which makes sure module 3 is installed when 1 and 2 are.

How does it work? `auto_install` won't trigger unless the 3 addons are installed right? and if you install the 3 addons, you don't need this one?

simahawk:
> > 1. stock_picking_group_by_partner_by_carrier
> > 2. sale_cutoff_time_delivery
> > 3. stock_picking_group_by_partner_by_carrier_by_date
> > 
> > which makes sure module 3 is installed when 1 and 2 are.
> 
> How does it work? `auto_install` won't trigger unless the 3 addons are installed right? and if you install the 3 addons, you don't need this one?

@guewen well, you are right. I was only thinking in the scope of our project whereas the other 2 modules are installed already.
I was just trying to force its usage somehow.

PR #743, opened 2020-11-27
[10.0] backport stock_quant_package_product_packaging from 13.0

lmignon:

PR #744, opened 2020-11-27
[10.0] backport stock_quant_package_dimension

lmignon:
* [ ] includes #743 
* [ ] depends https://github.com/OCA/product-attribute/pull/695


stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -0,0 +1,63 @@
...
+        help="volume in cubic meters",
+    )
+
+    @api.depends("lngth", "width", "height")
+    def _compute_volume(self):
+        for pack in self:
+            pack.volume = (pack.lngth * pack.width * pack.height) / 1000.0 ** 3
+
+    def auto_assign_packaging(self):
+        self = self.with_context(_auto_assign_packaging=True)
```
lmignon, line 153: I am not a fan of passing arguments to methods via the context. In versions <13.0, this type of approach has a significant impact on performance since it invalidates the cache. However, the way the code is extended does not leave us any other solution.  For me the current approach is not good in the sense that the context of an environment relates to the data and should be dissociated from the context necessary for method execution. I wonder if it would be interesting to develop a decorator that would allow us to specify elements useful for an execution context...
ex:
```python
from odoo.addons.call_context import call_context

    ...
    
    def auto_assign_packaging(self):
        with call_context(_auto_assign_packaging=True):
            return super(StockQuantPackage, self).auto_assign_packaging()

    def write(self, vals):
        res = super(StockQuantPackage, self).write(vals)
        if call_context.get("_auto_assign_packaging") and vals.get(
            "product_packaging_id"
        ):
            self._update_dimensions_from_packaging(override=False)
       return res

```

PR #748, opened 2020-12-03
[14] Add module stock_restrict_lot

florian-dacosta:
The purpose of this module is to be able to restrict the lot reservation of a stock move.
The restrict lot id did exist in previous Odoo version (v10 and less).
This module allows to build other module on top with different logic where we would like to restrict a stock move to a specific lot.
One example is in module sale_order_lot_selection, which was buggy but should be more reliable depending on this module.

@sebastienbeau @bguillot

### DIFF
--- a/stock_restrict_lot/__init__.py
+++ b/stock_restrict_lot/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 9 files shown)


bealdav (approved):

rousseldenis (approved):
Code review

bodedra (approved):

bguillot (approved):
Code review

PR #751, opened 2020-12-09
[13.0][FIX] sale_stock_mto_as_mts_orderpoint

TDu:
For user without proper rights the confirmation of a sale.order can lead
to an access error on the orderpoint creation.
The sudo fixes that.


rousseldenis:
@TDu Could you add a test that highlight this ?

simahawk:
@TDu ping

TDu:
:ping_pong: @simahawk 
@rousseldenis I just added a test which fails (I tested that locally) without the sudo implementation, is it what you add in mind ?
Oh and thanks for your reviews

simahawk (approved):

rousseldenis (approved):

PR #750, opened 2020-12-09
[13.0][IMP] stock_picking_invoice_link: Better link invoices to stock_moves process

carlosdauden:
Steps to reproduce the issue before this PR:
 - Invoice sale order
 - Create Credit note whith third option
 - Remove new draft invoice
 - Invoice sale order
 - Print invoice

@Tecnativa TT27156

PR #752, opened 2020-12-09
[12.0][IMP] stock_picking_invoice_link: Better link invoices to stock_moves process

sergio-teruel:
cc @Tecnativa TT27184
ping @carlosdauden @chienandalu

### DIFF
--- a/stock_picking_invoice_link/models/sale_order.py
+++ b/stock_picking_invoice_link/models/sale_order.py
@@ -3,23 +3,32 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from odoo import models
-from odoo.tools import float_compare
+from odoo.tools import float_compare, float_is_zero
 
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
Same comment about tests than in #750

stock_picking_invoice_link/models/sale_order.py
```diff
@@ -3,23 +3,32 @@
...
-                    inv.state != 'cancel' for inv in x.invoice_line_ids.mapped(
-                        'invoice_id'))) and not x.scrapped and (
-                    x.location_dest_id.usage == 'customer' or
-                    (x.location_id.usage == 'customer' and
-                     x.to_refund))
-            )
-        )
+        moves_linked = self.env["stock.move"]
+        for stock_move in self.move_ids:
+            if stock_move.state != 'done' or stock_move.scrapped or (
```
rousseldenis, line 15: @sergio-teruel IMHO a helper function or field to filter would be welcome in order to better readability and comprehension (apart of the fact it could be overidden easily too)

PR #753, opened 2020-12-10
s_picking_group_by_partner_by_carrier: split hook + add merge wizard

simahawk:
The hook for taking control of stock.move picking assign has been moved to `stock_move_assign_picking_hook`.

A "Merge transfers" wizard is now available to merge pickings manually.


stock_move_assign_picking_hook/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Stock Move picking assignment hook",
+    "Summary": "Base module that adds an hook to override picking assignment on moves.",
```
TDu, line 6:
```suggestion
    "Summary": "Base module that adds a hook to override picking assignment on moves.",
```

stock_move_assign_picking_hook/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "version": "13.0.1.0.0",
+    "development_status": "Alpha",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "category": "Warehouse Management",
+    "depends": ["stock"],
+    "data": [],
```
TDu, line 13:
```suggestion
```

TDu (approved):
Only a quick review, but LG

stock_picking_group_by_partner_by_carrier/tests/test_merge_wizard.py
```diff
@@ -0,0 +1,290 @@
...
+                        "partner": self.partner1,
+                        "pickings": tuple([x for x in valid_expected]),
+                        "has_todo": False,
+                    }
+                ],
+                "discarded_pickings": wiz.discarded_picking_ids,
+                "something_todo": False,
+            },
+        )
+        message = "No picking can be merged!"
```
sebalix, line 174:
```suggestion
        message = "No picking can be merged."
```

stock_picking_group_by_partner_by_carrier/wizard/stock_picking_merge_wiz.py
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+    def _get_info_template(self):
+        mod = "stock_picking_group_by_partner_by_carrier"
+        return self.env.ref(mod + ".stock_picking_merge_wiz_info")
+
+    def action_merge(self):
+        self.ensure_one()
+        if self.nothing_todo:
+            raise exceptions.UserError(_("No picking can be merged!"))
```
sebalix, line 117:
```suggestion
            raise exceptions.UserError(_("No picking can be merged."))
```

sebalix (approved):
Minor comments, LG

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -53,59 +53,52 @@ def _on_assign_picking_message_link(self):
...
-        states = ("draft", "confirmed", "waiting", "partially_available", "assigned")
+        domain = super()._domain_search_picking_for_assignation()
         if (
             not self.picking_type_id.group_pickings
             or self.group_id.sale_id.picking_policy == "one"
         ):
-            # use the normal domain from the stock module
-            domain = [
-                ("group_id", "=", self.group_id.id),
+            return domain
```
sebalix, line 62: @simahawk Should we also call `grouping_domain = self._assign_picking_group_domain()` and add it to `domain` before returning?

simahawk: Hmmm this is actually returning the default domain. It was already like this before the split.

sebalix: Before we were not returning `domain`, and we were reaching the following block code:
```python
        if self.env.context.get("picking_no_copy_if_can_group"):
            # we are in the context of the creation of a backorder:
            # don't consider the current move's picking
            domain.append(("id", "!=", self.picking_id.id))
```
And this filter on `id` was enough to not get Shopfloor test failing. So the behavior has changed.

simahawk: FTR we said that the fix should be in `_create_backorder` which should not set the flag if picking type is not grouping enabled

sebalix: Fixed here: https://github.com/OCA/stock-logistics-workflow/pull/757

PR #754, opened 2020-12-11
[14.0][ADD] create stock_lending

i-vyshnevska:


stock_lending/models/stock_picking.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    has_return_date = fields.Boolean(related="picking_type_id.has_return_date")
```
rousseldenis: Do these fields are typical to 'stock_lending' ???

rousseldenis:
@i-vyshnevska @TDu I think this module name is not correct. At least, the code does not reflect it.

i-vyshnevska:
@rousseldenis  can you suggest a more suitable name?

rousseldenis:
@i-vyshnevska I don't know exactly what is your business case ? Maybe answering that will generate it. :smiley:

i-vyshnevska:
@rousseldenis updated

PR #757, opened 2020-12-17
[13.0][FIX] stock_picking_group_by_partner_by_carrier: regression when creating backorders

sebalix:
If the group pickings option is disabled, the creation of a backorder when validating a partial qty of a picking should work as usual.


simahawk (approved):

simahawk:
this unlocks many tests failing on satellite PRs (eg: wms repo)

PR #762, opened 2020-12-28
[13.0][add] stock_valuation_layer_usage, stock_account_product_run_fifo_hook

JordiBForgeFlow:
This PR adds two modules

stock_valuation_layer_usage
======================
This module allows to trace the usage of stock valuation layer. That is,
makes it possible to identify in what stock moves was a given valuation layer
used, and how much quantity was taken by the particular stock move.

This kind of traceability is important in case that at some point you want
to conduct an revaluation (for example, in case that the purchase order price
changes after the products have been received into stock).

stock_account_product_run_fifo_hook
=============================
Technical module that adds the necessary required hooks in odoo standard methods where the necessary logic cannot be extended by inheritance mechanism.

TODO:

* [x] Add tests
* [x] Complete README

cc @LoisRForgeFlow @JordiMForgeFlow @AaronHForgeFlow


AaronHForgeFlow:
The usage tab does not show much information right now:
image

JordiBForgeFlow:
@AaronHForgeFlow fixed. Thanks for spotting it.
image

AaronHForgeFlow (approved):
Functional test :+1:

Receive at different prices, and play with deliveries. Works fine.

JordiMForgeFlow (approved):
Functional Review 👍🏼

PR #769, opened 2021-01-10
[13.0][ADD] stock_move_line_qty_done_constraint

rousseldenis:


stock_move_line_qty_done_constraint/models/stock_move_line.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo.tools import float_compare
+
+
+class StockMoveLine(models.Model):
+
+    _inherit = "stock.move.line"
+
+    @api.constrains("product_qty", "qty_done")
+    def _check_qty_done(self):
+        for rec in self.filtered("picking_id.picking_type_id.qty_done_constraint"):
```
yostashiro, line 14:
```suggestion
        for rec in self.filtered("move_id.picking_type_id.qty_done_constraint"):
```
Could be a bit more generic to make it like this if there is a chance of this to be applied on move lines from manufacturing orders?

stock_move_line_qty_done_constraint/models/stock_picking_type.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+
+    _inherit = "stock.picking.type"
+
+    qty_done_constraint = fields.Boolean(
```
lmignon, line 11: @rousseldenis The name is meaningless... Why not `disallow_reserved_qty_overflows`?

lmignon (changes requested):
@rousseldenis Thank you. LGTM (Code review). A little comment on the name of the field on the picking_type

lmignon (approved):
@rousseldenis Thank you. LGTM (Code review). A little comment on the name of the field on the picking_type

PR #773, opened 2021-01-12
[13.0] [MIG] stock_lot_scrap: Migration to v13

CarlosRoca13:
cc @Tecnativa TT27135

please, @carlosdauden @joao-p-marques can you review this?


rousseldenis (approved):
Code review

stock_lot_scrap/models/stock_production_lot.py
```diff
@@ -72,7 +71,18 @@ def action_scrap_lot(self):
...
+        scrap_location_id = (
+            self.env["stock.location"]
+            .search(
+                [
+                    ("company_id", "=", self.company_id.id),
+                    ("scrap_location", "=", True,),
+                ]
+            )
+            .id
+        )
```
joao-p-marques: I think this search may cause problems in the future, you can make it more resilient by limiting the search results to 1:
```suggestion
        # The data model has disappeared and right now is a function who creates the
        # scrap location. That's why we have to search by company_id and scrap_location.
        scrap_location_id = (
            self.env["stock.location"]
            .search(
                [
                    ("company_id", "=", self.company_id.id),
                    ("scrap_location", "=", True,),
                ],
                limit=1,
            )
            .id
        )

File: stock_lot_scrap/models/stock_production_lot.py [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+    _inherit = "stock.production.lot"
+
+    @api.model
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):  # pragma: no cover
+        """Inject the button here to avoid conflicts with other modules
+         that add a header element in the main view.
+        """
+        res = super(StockProductionLot, self).fields_view_get(
```
joao-p-marques:
```suggestion
        res = super().fields_view_get(
```

stock_lot_scrap/models/stock_scrap.py
```diff
@@ -0,0 +1,17 @@
...
+
+class StockScrap(models.Model):
+    _inherit = "stock.scrap"
+
+    def action_validate(self):
+        self.ensure_one()
+        self.lot_id.message_post(
+            body=_("Lot was scrapped by <b>%s</b>.") % self.env.user.name
+        )
+        return super(StockScrap, self).action_validate()
```
joao-p-marques:
```suggestion
        return super().action_validate()
```

CarlosRoca13:
@joao-p-marques Changes done!

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "restructuredtext.confPath": ""
+}
```
joao-p-marques: It seems like this was added here by mistake. It shouldn't go here.

CarlosRoca13: Oh true, sorry. Removed

joao-p-marques (approved):
Code review :+1:

.vscode/settings.json
joao-p-marques: No problem. I always got the same issue :sweat_smile:

pedrobaeza (approved):

PR #776, opened 2021-01-13
[13.0][ADD] delivery_total_weight_from_packaging

sebalix:
Change the way the weight is computed on move, picking and package to include packaging weight instead of the weight of the product only.

Ref. 1900


guewen (approved):

jbaudoux (approved):

TDu (commented):
Looks like you did not include any readme files...

sebalix:
@TDu good point

delivery_total_weight_from_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,32 @@
...
+    @api.depends("quant_ids")
+    @api.depends_context("picking_id")
+    def _compute_weight(self):
+        # Override method from `delivery` module to compute a more accurate
+        # weight by including the weight of the packaging
+        # NOTE: code copied/pasted and adapter from `delivery`
+        for package in self:
+            weight = 0.0
+            if self.env.context.get("picking_id"):
+                current_picking_move_line_ids = self.env["stock.move.line"].search(
```
simahawk, line 19: could we avoid a search for each item?

sebalix: Don't know which is the best: querying all move lines for all packages at once, then group them by package, or searching them as it is done right now.

sebalix: Before I update this one, your review is welcome on https://github.com/OCA/stock-logistics-workflow/pull/777

PR #777, opened 2021-01-13
[13.0][ADD] stock_quant_package_dimension_total_weight_from_packaging

sebalix:
First, we add a new `estimated_pack_weight` on the package in `stock_quant_package_dimension` module which is similar to the `weight` field implemented in the standard `delivery` module: it is done on purpose to not depend on `delivery`. This weight is thus computed from product weight.

Second, we re-implement this field in `stock_quant_package_dimension_total_weight_from_packaging` module (`auto_install=True`) which make use of the weight computation mechanism from `product_total_weight_from_packaging`. The weight here is computed from the packaging weight.

Ref. 1900

### DIFF
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -7,6 +7,12 @@ class StockQuantPackage(models.Model):
     _inherit = "stock.quant.package"
 
     pack_weight = fields.Float("Weight (kg)")
+    estimated_pack_weight = fields.Float(
+        "Estimated weight (kg)",
+        digits="Product Unit of Measure",
+        compute="_compute_estimated_pack_weight",
+        help="Based on the weight of the product.",
+    )
     # lngth IS NOT A TYPO: https://github.com/odoo/odoo/issues/41353
     lngth = fields.Integer("Length (mm)", help="length in millimeters")
     width = fields.Integer("Width (mm)", help="width in millimeters")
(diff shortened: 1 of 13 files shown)


guewen (approved):

TDu (approved):

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -20,6 +26,32 @@ class StockQuantPackage(models.Model):
...
 
+    @api.depends("quant_ids")
+    @api.depends_context("picking_id")
+    def _compute_estimated_pack_weight(self):
+        # NOTE: copy-pasted and adapted from `delivery` module
+        # because we do not want to add the dependency against 'delivery' here.
+        for package in self:
+            weight = 0.0
+            if self.env.context.get("picking_id"):
+                current_picking_move_line_ids = self.env["stock.move.line"].search(
```
simahawk: same comment as #776

stock_quant_package_dimension_total_weight_from_packaging/models/stock_quant_package.py
```diff
@@ -0,0 +1,36 @@
...
+        # 'product_total_weight_from_packaging' module
+        for package in self:
+            weight = 0.0
+            if self.env.context.get("picking_id"):
+                current_picking_move_line_ids = self.env["stock.move.line"].search(
+                    [
+                        ("result_package_id", "=", package.id),
+                        ("picking_id", "=", self.env.context["picking_id"]),
+                    ]
+                )
```
simahawk: can't you split the code so that you don't have to override the whole method again?
Something like `_get_weight_for_picking` and `_get_weight_without_picking` or similar...

sebalix: I pushed a fixup which split things in several methods.

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -20,6 +26,49 @@ class StockQuantPackage(models.Model):
         help="volume in cubic meters",
     )
 
+    def _get_picking_move_line_ids_per_package(self, picking_id):
+        if not picking_id:
+            return {}
+        move_lines = self.env["stock.move.line"].search(
+            [
+                ("result_package_id", "in", self.ids),
+                ("picking_id", "=", self.env.context["picking_id"]),
```
simahawk: you should use the picking_id passed as arg no?

stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -20,6 +26,49 @@ class StockQuantPackage(models.Model):
...
+        # NOTE: copy-pasted and adapted from `delivery` module
+        # because we do not want to add the dependency against 'delivery' here.
+        move_line_ids_per_package = self._get_picking_move_line_ids_per_package(
+            self.env.context.get("picking_id")
+        )
+        for package in self:
+            weight = 0.0
+            if self.env.context.get("picking_id"):
+                move_line_ids = move_line_ids_per_package[package]
+                move_lines = self.env["stock.move.line"].browse(move_line_ids)
```
simahawk:
```suggestion
        picking_id = self.env.context.get("picking_id")
        move_line_ids_per_package = self._get_picking_move_line_ids_per_package(
            picking_id
        )
        for package in self:
            weight = 0.0
            move_line_ids = move_line_ids_per_package.get(package)
            if move_line_ids:
                move_lines = self.env["stock.move.line"].browse(move_line_ids)
```

sebalix: I'm not sure about this one: with your implementation, if there is no move lines we will get the weight computed from quants, even if a picking_id was passed in the context.
But at the same time if we pass the picking_id there for sure move lines related (at least in Shopfloor, don't know the use case in `delivery` with the `weight` field).
So I'll replace your `if move_line_ids` with a `if picking_id`

sebalix:
Test is failing on `stock_putaway_by_route` since https://github.com/odoo/odoo/commit/db2172c9a74efbf8160aafdabe9c680d3bf4a8a5 has been merged. I try to fix it.

EDIT: fixed

stock_quant_package_dimension/models/stock_quant_package.py
simahawk: I thought it would be a good idea anyway to have the std weight when no move lines are found. Fine with your explanation :)

simahawk (approved):

PR #779, opened 2021-01-22
[13.0][FIX] stock_picking_group_by_partner_by_carrier

TDu:
When a sale order is cancelled then set to quotation again, all
related picking are cancelled also.

If before reconfirming the sale order for a 2nd time the delivery
carrier is changed, the procurement group linked on the sale order
will be used and related to the wrong carrier.

Removing the procurement group on the cancelation of the sale order
fixes the issue.

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/sale_order.py
+++ b/stock_picking_group_by_partner_by_carrier/models/sale_order.py
@@ -13,6 +13,11 @@ class SaleOrder(models.Model):
                 SaleOrder, sale_order.with_context(cancel_sale_id=sale_order.id)
             ).action_cancel()
 
+    def action_draft(self):
+        res = super().action_draft()
+        self.procurement_group_id = False
+        return res
+
     def get_name_for_delivery_line(self):
         """Get the name for the sale order displayed on the delivery note"""
         self.ensure_one()


simahawk (commented):
seems legit but wouldn't be better to do it when the sale order is rolled back to draft instead? :thinking:

TDu:
@simahawk Agreed, I did the change you proposed.

TDu:
The failing test seems to be unrelated to this change !?

simahawk:
seems so https://travis-ci.com/github/OCA/stock-logistics-workflow/jobs/475603854#L1309
As it fails on one job only it might be a flappy test.

sebalix:
@TDu I fixed the flappy test here: https://github.com/OCA/stock-logistics-workflow/pull/777/commits/a976f375a66444f7cdb9c537be7937807246fee6 (PR https://github.com/OCA/stock-logistics-workflow/pull/777)
Build will be green once it's merged.

sebalix (approved):

TDu:
@sebalix Thanks

stock_picking_group_by_partner_by_carrier/models/sale_order.py
```diff
@@ -13,6 +13,10 @@ def action_cancel(self):
                 SaleOrder, sale_order.with_context(cancel_sale_id=sale_order.id)
             ).action_cancel()
 
+    def action_draft(self):
+        super().action_draft()
```
rousseldenis: As Odoo standard function returns, you should also

rousseldenis (changes requested):
LGTM.

stock_picking_group_by_partner_by_carrier/models/sale_order.py
simahawk: @TDu ping

TDu: @rousseldenis Thanks for the review, fixed.

rousseldenis (commented):
LGTM.

But one question remains: is this part of code has its place here ? Wouldn't this behaviour be generalized (or located in a more agnostic module) ?

TDu:
It has been added in this module, because the problem was discovered when this module didn't behave properly as explained in the commit message.
Could the problem occur in other situation without this module installed, I don't know ?

simahawk:
@TDu can you rebase and squash? @rousseldenis worth to keep it blocked?

rousseldenis (approved):

rousseldenis:
@TDu Could you check Travis ? Or rebase.

TDu:
Rebased and Travis is :green_heart:

PR #786, opened 2021-02-16
[13.0][FIX] stock_picking_group_by_partner_by_carrier_by_date: Add freezegun requirement

victoralmau:
Add freezegun requirement (used in tests https://github.com/OCA/stock-logistics-workflow/blob/13.0/stock_picking_group_by_partner_by_carrier_by_date/tests/test_grouping_by_date.py#L3)

Error in https://travis-ci.com/github/OCA/stock-logistics-workflow/jobs/483493713#L1957

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+freezegun


victoralmau:
Ping @simahawk

simahawk:
wondering: how is possible that the build on 13.0 is green if we miss this?
BTW `stock_partner_delivery_window` requires it as well.

pedrobaeza:
Shouldn't this be put on `requirements_test.txt`? cc @sbidoul

simahawk:
`test-requirements.txt` should be the right place but this works only w/ MQT=PIP.
See https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_install_nightly#L136
We should probably use it in all cases.

victoralmau:
Actually 13.0 is red https://travis-ci.com/github/OCA/stock-logistics-workflow/jobs/483493713#L1957

pedrobaeza:
OK, merging for now as is for solving the red, as switching to PIP MQT may provoke other problems.

PR #800, opened 2021-03-03
[14.0][FIX] stock_restrict_lot

newtratip:
No validate `Inconsistencies between reserved quant and lot restriction on stock move` if vals['lot_id'] is False

I'm migrating on [`sale_tier_validation`](https://github.com/OCA/sale-workflow/pull/1455) module and found some error on `sale_order_lot_generator` module
Selection_198

**Step to reproduce**
1. Create Quotation with product as tracking = lot
2. Confirm Quotation
3. Add sale_order_line with product as tracking = lot
4. Save Quotation and found message `Inconsistencies between reserved quant and lot restriction on stock move`

This is occur from have lot_id on sale_order_line but no have on stock_move_line

Peek 2021-03-03 14-55

I fix no check this if vals['lot_id'] is False. Is there any concern ? @florian-dacosta

### DIFF
--- a/stock_restrict_lot/models/stock_move.py
+++ b/stock_restrict_lot/models/stock_move.py
@@ -23,7 +23,11 @@ class StockMove(models.Model):
             quantity=quantity, reserved_quant=reserved_quant
         )
         if self.restrict_lot_id:
-            if "lot_id" in vals and vals["lot_id"] != self.restrict_lot_id.id:
+            if (
+                "lot_id" in vals
+                and vals["lot_id"] is not False
+                and vals["lot_id"] != self.restrict_lot_id.id
+            ):
                 raise exceptions.UserError(
                     _(
                         "Inconsistencies between reserved quant and lot restriction on "


newtratip:
CI Red
https://github.com/OCA/sale-workflow/pull/1455
https://github.com/OCA/sale-workflow/pull/1476
https://github.com/OCA/sale-workflow/pull/1477

newtratip:
@kittiu, you can review this ?

florian-dacosta:
Hello @newtratip 

I did not manage to reproduce the issue, following the same step as you, having configured both product with lot management and lot auto generation.

So I am not sure why this happend, I guess it is related to the fact that you must have some stock with no lot for a product configured with lot management, which is kind of inconsistent, although not impossible in Odoo.

If you could provide a test with the case failing, it would help a lot to reproduce the issue to know what is the source problem.

Indeed, if we restrict a lot on a stock move, we do not want this move to be able to reserve a product with no lot...
And AFAIK, that is what is happening in your case

florian-dacosta (approved):
Well, I looked a bit more into it and it seems that Odoo will always allow to reserve a product with no lot
So I guess we have no way to avoid it, even if the situation where the stock of a product manager with lot tracking which has no lot is not normal.

So the fix is ok for me

marcelsavegnago (approved):

britoederr (approved):

marcelsavegnago:
Up

LoisRForgeFlow (approved):
thank you

PR #802, opened 2021-03-04
[13.0][MIG][WIP] product_cost_price_avco_sync: Migration to 13.0

carlosdauden:
This module is used to maintain the stock valuation with correct data retroactively.
For example:
On 2021-06-28 merchandise is received, but someone wrongly entered the price or quantity in the purchase order, putting 1000 instead of 10.00.
On 2021-07-01 the error is detected and solved.
When taking a valuation report on 2021-06-30 the information will appear incorrect (although it will be correct on 2021-07-02)

If, in addition, we have deliveries made on 2021-06-29, the svl related with this delivery will be wrong forever.

What this module does is synchronize the corrections so that all the records that are affected by that modification are modified sequentially.

In combination with other modules, sales order margins can be correctly calculated.

The only "drawback" of using this module is that if a report is obtained by selecting the same date at two different times, the information obtained may be different.

@Tecnativa


pedrobaeza:
Please check Travis

chienandalu:
> Please check Travis

I'd say it just fails in pre-commit as tests are commented :sweat_smile:

pedrobaeza:
@carlosdauden can you please finish this and re-activate tests?

AaronHForgeFlow:
Hi @carlosdauden @sergio-teruel  Can you please let me know why is this module needed. Is it to cover the case where the product changes its cost method? or is there an inconsistency in standard Odoo in the process of computing the price? Thanks in advance!

carlosdauden:
Hi @AaronHForgeFlow 

This module is used to maintain the stock valuation with correct data retroactively.
For example:
  On 2021-06-28 merchandise is received, but someone wrongly entered the price or quantity in the purchase order, putting 1000 instead of 10.00.
  On 2021-07-01 the error is detected and solved.
  When taking a valuation report on 2021-06-30 the information will appear incorrect (although it will be correct on 2021-07-02)

If, in addition, we have deliveries made on 2021-06-29, the svl related with this delivery will be wrong forever.

What this module does is synchronize the corrections so that all the records that are affected by that modification are modified sequentially.

In combination with other modules, sales order margins can be correctly calculated.

The only "drawback" of using this module is that if a report is obtained by selecting the same date at two different times, the information obtained may be different.

AaronHForgeFlow:
Hi @carlosdauden , Thank you for your answer, I will try to reproduce that example. Sometimes I get weird average standard cost in the products for some reason and it is hard to tell why that happens. 

In this case, the correction consist of fixing the price in the stock moves, those prices are propagated to the layers, and finally this module updates the standard price correctly, is that correct?

pedrobaeza:
@carlosdauden please rebase and check Travis for merging this

carlosdauden:
Ping @LoisRForgeFlow

product_cost_price_avco_sync/models/stock_valuation_layer.py
```diff
@@ -214,10 +215,12 @@ def process_avco_svl_manual_adjustements(self, svls_dic):
                 and not svl_dic["unit_cost"]
                 and not svl.stock_move_id
             ):
-                standard_price = float(svl.description.split(" ")[-1][:-1])
-                svl_dic["value"] = (
-                    standard_price * accumulated_qty
-                ) - accumulated_value
+                match_price = re.findall(r"[+-]?[0-9]+\.[0-9]+\)$", svl.description)
```
chienandalu, line 214: Can you add a rationale about this change?

rafaelbn (approved):
Thank you for this migration! I cannot tested but I guess this has been in production for a long time! 👍

pedrobaeza (approved):
Let's merge this one as this has been battle-tested. Further improvements or hooks can be done on successive PRs.

/ocabot merge nobump

PR #809, opened 2021-03-10
12.0 add stock picking unreserve line

MPoch-PlanetaTIC:
Permits to unreserve stock.move

### DIFF
--- a/stock_picking_unreserve_line/__manifest__.py
+++ b/stock_picking_unreserve_line/__manifest__.py
@@ -1,0 +1,22 @@
+# Copyright 2020 PlanetaTIC - Marc Poch <mpoch@planetatic.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+{
+    'name': 'Stock Picking Unreserve Line',
+    'summary': 'Permits to unreserve lines of stock picking',
+    'version': '12.0.1.0.0',
+    'category': 'Stock',
+    'author': 'PlanetaTIC'
+              ' Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/stock-logistics-workflow',
+    'license': 'AGPL-3',
+    'depends': [
+        'stock',
+    ],
+    'data': [
+        'views/stock_move_views.xml',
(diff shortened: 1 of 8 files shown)


ypapouin (changes requested):
Don't you think that a specific security group could be added for this ?

stock_picking_unreserve_line/readme/CONSTRIBUTRORS.rst
```diff
@@ -0,0 +1 @@
+* Marc Poch <mpoch@planetatic.com>
```
StefanRijnhart, line 1: Typo in file name `CONSTRIBUTRORS.rst `

stock_picking_unreserve_line/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Permits to unreserve stock.move
```
StefanRijnhart, line 1: Can you be a bit more verbose? What can the user expect after the installation of the module?

StefanRijnhart:
@ypapouin in core Odoo, there is no security group on the picking's Unreserve button, so adding one here would not be necessary I think https://github.com/odoo/odoo/blob/12.0/addons/stock/views/stock_picking_views.xml#L222

stock_picking_unreserve_line/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 PlanetaTIC - Marc Poch <mpoch@planetatic.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+{
+    'name': 'Stock Picking Unreserve Line',
+    'summary': 'Permits to unreserve lines of stock picking',
+    'version': '12.0.1.0.0',
+    'category': 'Stock',
+    'author': 'PlanetaTIC'
```
StefanRijnhart, line 10: Missing comma after PlanetaTIC breaking the linting

StefanRijnhart (changes requested):
Please add a simple test

PR #810, opened 2021-03-10
12.0 add stock unassign wizard

MPoch-PlanetaTIC:
Permits to unassign reserved stock from other stock.move

### DIFF
--- a/stock_unassign_wizard/models/stock_move.py
+++ b/stock_unassign_wizard/models/stock_move.py
@@ -20,7 +20,7 @@ class StockMove(models.Model):
             '&',
             ('picking_id', '!=', False),
             ('picking_id', '!=', self.picking_id.id),
-            ('picking_type_id.code', '=', 'outgoing'),
+            ('picking_type_id.code', 'in', ('outgoing', 'internal')),
             '&',
             ('raw_material_production_id', '!=', False),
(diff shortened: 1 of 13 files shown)


stock_unassign_wizard/models/stock_move.py
```diff
@@ -0,0 +1,52 @@
...
+            ('product_id', '=', self.product_id.id),
+            ('state', 'in', ['assigned', 'partially_available']),
+            '|',
+            '&',
+            '&',
+            ('picking_id', '!=', False),
+            ('picking_id', '!=', self.picking_id.id),
+            ('picking_type_id.code', 'in', ('outgoing', 'internal')),
+            '&',
+            ('raw_material_production_id', '!=', False),
```
rousseldenis, line 25: @MPoch-PlanetaTIC IMHO you should put the domain retrieval in a function in order to split this into two modules as pulling mrp constrains a little bit inheritance. What do you think ?

MPoch-PlanetaTIC: You are right.
I'll split it in two functions:
* the main function: open_unassign_wizard
* and add the new function get_unassign_wizard_domain
Currently I have no time to do this fix, but I will do it.

Thanks

PR #811, opened 2021-03-16
[13.0][OU-MIG] stock_picking_invoice_link: Migration scripts

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/stock_picking_invoice_link/migrations/13.0.1.0.0/post-migration.py
+++ b/stock_picking_invoice_link/migrations/13.0.1.0.0/post-migration.py
@@ -1,0 +1,32 @@
+# Copyright 2021 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+from psycopg2.extensions import AsIs
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    old_table = AsIs(openupgrade.get_legacy_name("stock_move_invoice_line_rel"))
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO stock_move_invoice_line_rel
+        (invoice_line_id, move_id)
+        SELECT aml.id, rel.move_id
+        FROM %s rel
+        JOIN account_move_line aml ON aml.old_invoice_line_id = rel.invoice_line_id
+        """,
+        (old_table,),
+    )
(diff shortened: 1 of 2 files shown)


carlosdauden (approved):

stock_picking_invoice_link/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.remove_tables_fks(env.cr, ["stock_move_invoice_line_rel"])
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE stock_move_invoice_line_rel rel
```
pedrobaeza: I'm afraid you can't do it this way due to 2 main reasons:

- It's destructive: you are removing the traceability in case something is needed, so we always try to preserve previous values when feasible, so a re-query can be done in case of need.
- The worst one: m2m tables are very special in the ORM, and the FK constraints and the rest of the table specification is only done at table creation. With this, you are having a dangling table with no consistency (as you have removed the FK).

What you need to do is:

- In pre-migration, rename the table `stock_move_invoice_line_rel`, and drop FK the same as here.
- In post, do an INSERT INTO in the new created table by the ORM process with the proper values.

stock_picking_invoice_link/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+        SET invoice_line_id = aml.id
+        FROM account_move_line aml
+        WHERE aml.old_invoice_line_id = rel.invoice_line_id
+        """,
+    )
+    openupgrade.remove_tables_fks(env.cr, ["account_invoice_stock_picking_rel"])
```
pedrobaeza, line 22: The same here.

pedrobaeza (changes requested):

stock_picking_invoice_link/migrations/13.0.1.0.0/post-migration.py
sergio-teruel: Here the relation is new

sergio-teruel:
@pedrobaeza changes done!! Can you review?

jjscarafia:
I believe this pr can be closed, isn't it @sergio-teruel ?

pedrobaeza:
Yes, done on e748e3f502356a55de1ddf6913a92191b5731d14

PR #816, opened 2021-03-22
[12.0][IMP] stock_picking_invoice_link: Support products with ordered qties

grindtildeath:


grindtildeath:
@pedrobaeza @carlosdauden @sergio-teruel Your feedback would be welcome here :pray:

pedrobaeza:
Let's make a deal: if you review other 2/3 PRs, I review this one :wink: 

Anyway, please start putting CI green.

grindtildeath:
@pedrobaeza I guess I did my part with https://github.com/OCA/stock-logistics-workflow/pull/817 :wink:

pedrobaeza:
Well, that's not exactly making some reviews, hehe, but maintaining own modules, but OK for this time.

Doing a check on your code, I see chances of some move lines to be linked incorrectly, as there can be lots of cases (returns, partial invoicing, etc), so I wouldn't risk this version that is end of life (according Odoo policy), for changing the behavior.

What about adding a module on top for adding such behavior and do in the module this one the needed hooks?

gurneyalex:
Hello @pedrobaeza 

Our current customer needing this (to be able to print the serial numbers of the invoices) runs 12.0 so for us it makes sense to have this on this enhancement at least on this version. And we would like to port this to later versions. 

This change should not break existing installations (they should be configured to use invoice after delivery). But if you think it is better for the migration path we can add a setting to enable linking moves of product which are not invoiced on delivery to the invoices and ensure BW compatibility, without artificially adding another module to compensate for limitations of this one. 

I still think this is a good enhancement, and we can add unit tests for covering cases which could be problematic. We welcome suggestions for such tests.

pedrobaeza:
@gurneyalex, I see 2 potential problems merging this as is:

* The change of `picking_ids` in `account.invoice` to computed field. This can change the content of the current fields when a change is performed in the existing documents, and also store other values different from the expected ones, and also breaking possible adaptations that stores there picking references without having the lines linked. This change is also non related to the main one you want to put. Don't get me wrong, this change is good for keeping consistency, but the design principles were different (you can store header information non related to lines), and it's too late in this version for changing that design IMO.
* Current users of the module that use lines with ordered quantities, will see when this gets merged, that appears new links/lines from something that previously don't, so that change of behavior is not OK for such a stable version. Providing a configuration option for opt-in will make the module more complicated and less readable. That's why I propose to create an extension module `stock_picking_invoice_link_ordered_qty` that provides the option. If you install it, you have it. Not installing it, you have the same as previously. If any hook is needed in the base module, go ahead with it.

i-vyshnevska (approved):
do we need to increase the version of addon?
LG

pedrobaeza (changes requested):
This is still conflicting. See my last comment.

grindtildeath:
@pedrobaeza Should we try to merge this in the base module for v15.0?

Then we could add this in a separated module with the proper hooks in the base for v12.0 to v14.0 as you proposed.

pedrobaeza:
OK, it can be an option. I'm still afraid of the change to compute, where no manual changes will be allowed anymore, or the immutability of previous links, but we can discuss about it.

PR #818, opened 2021-03-24
[13.0] delivery_total_weight_from_packaging improvements

simahawk:
See atomic commits.

If https://github.com/odoo/odoo/pull/68273 gets merged we can trash the override here.

ref: 2258

### DIFF
--- a/delivery_total_weight_from_packaging/models/stock_quant_package.py
+++ b/delivery_total_weight_from_packaging/models/stock_quant_package.py
@@ -1,32 +1,45 @@
 # Copyright 2021 Camptocamp SA
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class StockQuantPackage(models.Model):
     _inherit = "stock.quant.package"
+
+    # This fiels already exists and it's stored.
+    # Here is made computed but kept editable using readonly=False
+    shipping_weight = fields.Float(compute="_compute_shipping_weight", readonly=False)
 
     @api.depends("quant_ids")
     @api.depends_context("picking_id")
     def _compute_weight(self):
         # Override method from `delivery` module to compute a more accurate
         # weight by including the weight of the packaging
-        # NOTE: code copied/pasted and adapter from `delivery`
         for package in self:
-            weight = 0.0
-            if self.env.context.get("picking_id"):
-                current_picking_move_line_ids = self.env["stock.move.line"].search(
-                    [
-                        ("result_package_id", "=", package.id),
-                        ("picking_id", "=", self.env.context["picking_id"]),
-                    ]
+            package.weight = package._get_weight_from_packaging()
+
+    @api.depends("quant_ids")
+    @api.depends_context("picking_id")
+    def _compute_shipping_weight(self):
+        for package in self:
+            package.shipping_weight = package._get_weight_from_packaging()
+
+    def _get_weight_from_packaging(self):
+        # NOTE: code copied/pasted and adapter from `delivery`
+        weight = 0.0
+        if self.env.context.get("picking_id"):
+            current_picking_move_line_ids = self.env["stock.move.line"].search(
+                [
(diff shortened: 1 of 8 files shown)


delivery_total_weight_from_packaging/models/stock_quant_package.py
```diff
@@ -1,32 +1,43 @@
...
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
 
-from odoo import api, models
+from odoo import api, fields, models
 
 
 class StockQuantPackage(models.Model):
     _inherit = "stock.quant.package"
 
+    shipping_weight = fields.Float(compute="_compute_shipping_weight", readonly=False)
```
sebalix, line 12: Add a comment saying it's an override? Could you explain again what is the purpose of `readonly=False` here? To not break tests which set a value on it?

delivery_total_weight_from_packaging/models/stock_quant_package.py
```diff
@@ -1,32 +1,43 @@
...
-                        ("result_package_id", "=", package.id),
-                        ("picking_id", "=", self.env.context["picking_id"]),
-                    ]
+            package.weight = package._get_weight_from_packaging()
+
+    @api.depends("quant_ids")
+    @api.depends_context("picking_id")
+    def _compute_shipping_weight(self):
+        for package in self:
+            package.shipping_weight = package._get_weight_from_packaging()
```
sebalix, line 26: Should we not compute the shipping weight based on the delivery packaging (`package.packaging_id`)? Here I suppose that `_get_weight_from_packaging()` and `product.get_total_weight_from_packaging()` compute the weight from storage packaging, not delivery ones.

sebalix (commented):
Some questions, Travis is also failing

delivery_total_weight_from_packaging/models/stock_quant_package.py
simahawk: the purpose is pretty simple: I  don't want to change its default behavior. People expect to be able to customize it.
The wizard itself is going to update that value. It just makes sense IMO :)
I'll add a comment.

delivery_total_weight_from_packaging/models/stock_quant_package.py
simahawk: hmm good point! I'll check!

simahawk: @sebalix answer: no. The weight on the delivery package is just the max weight allowed. Also, in theory, the weight of the pallet is already considered in the product packaged on its pallet packaging type. We might have issues w/ parcels but we'll tackle it in another round.

jbaudoux: There is no weight on a delivery package. And the weight from packaging can already include the weight of the packaging. It all depend if you repack in a bigger packaging. That is a very complex other problem that should not be addressed here.

delivery_total_weight_from_packaging/models/stock_quant_package.py
jbaudoux: @simahawk How do you set the measured value if it is a computed non stored field?
_compute_shipping_weight returns the theoretical weight. When you use a scale, you need to be able to enter the value
That field should store the value coming from the wizard

simahawk: this is not a new field, it's already stored. Here I make it computed and keep stored by setting readonly=false

jbaudoux: Ah. I didn't know that this implied store=True automatically

jbaudoux (approved):

sebalix (approved):

delivery_total_weight_from_packaging/models/stock_quant_package.py
sebalix: I didn't know neither, thanks for the info

TDu (approved):

PR #821, opened 2021-03-30
[13.0][ADD] Send delivery notification from any operation

sebalix:
Requires https://github.com/OCA/stock-logistics-warehouse/pull/1115

New module `delivery_send_to_shipper_at_operation` allowing to send the delivery notification from any operation (among chained operations) before the delivery.
With a pick/pack/ship setup this allows to send the delivery notification to the carrier when the pack operation is marked as done for instance (by default Odoo sends it on the ship/delivery).

Once the notification has been sent from a non-delivery operation, it won't be sent again when the delivery operation is validated.

**Screenshots:**

delivery_carrier1
delivery_carrier2

**Drawback:**

This way of sending delivery notifications is not compatible if we have several pack operations to process before the delivery (backorders?). In such a case, the notification will be generated from only part of the goods to deliver, which is not expected.

Ref. 2286


delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+        # FIXME: almost the same code than 'get_carrier_from_chained_picking'
+        # of 'stock_helper_delivery' module, maybe to refactor/merge
+        # to make 'get_carrier_from_chained_picking' returning both carrier
+        # and its related picking?
+        self.ensure_one()
+        if self.carrier_id:
+            return self
+
+        def get_moves_dest(moves):
+            return moves.move_dest_ids.filtered(
```
jbaudoux: Maybe no need to traverse all the moves of the picking, just the first done should provide a valid move_dest_id.

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+                # configured to match the current operation type: force the
+                # sending of the delivery notice to the carrier
+                carrier = picking.get_carrier_from_chained_picking()
+                related_ship = picking._get_carrier_operation_from_chained_picking()
+                if (
+                    carrier.integration_level == "rate_and_ship"
+                    and carrier.send_delivery_notice_on == "custom"
+                    and picking.picking_type_id
+                    in carrier.send_delivery_notice_picking_type_ids
+                    and not related_ship.delivery_notification_sent
```
jbaudoux:
```suggestion
```
When the first packing is done, it will make a notification to the carrier and then set 
`related_ship.delivery_notification_sent = True`
Then, all the other packing won't be declared to the carrier

jbaudoux (changes requested):

pedrobaeza:
Why this is not in OCA/delivery-carrier?

sebalix:
@pedrobaeza it could be moved there yes. I'll do that later

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: fixed

delivery_send_to_shipper_at_operation/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2020 Camptocamp SA
```
TDu:
```suggestion
# Copyright 2021 Camptocamp SA
```

TDu (approved):

delivery_send_to_shipper_at_operation/models/delivery_carrier.py
```diff
@@ -0,0 +1,42 @@
...
+    'Defined operations': send the carrier notification after the operation
+    of the chosen type bellow has been marked as done. In that case, the
+    carrier will be taken looking at the next operations until we found the
+    carrier (e.g. in a pick + pack + ship setup, if you set the notification
+    to be sent on the pack operation type, it'll look at the carrier defined
+    in the next operation, being here the ship).
+
+    If no notification have been sent when reaching the ship step
+    (e.g. re-route goods from carrier A to B), then a fallback ensure the
+    ship operation will trigger the call anyway.
```
simahawk:
```suggestion
    'Ship operation' (default): send the carrier notification after the
    operation having a delivery method defined has been marked as done.

    'Defined operations': send the carrier notification after the operation
    of the chosen types has been marked as done. In this case, the
    carrier will be taken looking at the next operations until a
    carrier is found. For instance, in a pick + pack + ship setup,
    if you set the notification to be sent on the pack operation type,
    it will look at the carrier defined in the next operation (the ship in this case).

    If no notification have been sent when reaching the ship step
    (e.g. re-route goods from carrier A to B), then a fallback ensure the
    ship operation will trigger the call anyway.
```

delivery_send_to_shipper_at_operation/models/delivery_carrier.py
```diff
@@ -0,0 +1,42 @@
...
+
+    send_delivery_notice_on = fields.Selection(
+        selection=[("ship", "Ship operation"), ("custom", "Defined operations")],
+        string="Send delivery notice on",
+        default="ship",
+        help=SEND_DELIVERY_NOTICE_ON_HELP,
+    )
+    send_delivery_notice_picking_type_ids = fields.Many2many(
+        comodel_name="stock.picking.type",
+        string="Operation types",
```
simahawk:
```suggestion
        string="Send delivery notice operation types",
```

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,85 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    delivery_notification_sent = fields.Boolean(default=False)
+
+    def _send_confirmation_email(self):
+        for picking in self:
+            if not picking.carrier_id:
+                # If the current operation has no carrier defined, but a carrier
```
simahawk: can you move this to `_handle_delivery_notification()`?

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,85 @@
...
+                related_ship = picking._get_carrier_operation_from_chained_picking()
+                if (
+                    carrier.integration_level == "rate_and_ship"
+                    and carrier.send_delivery_notice_on == "custom"
+                    and picking.picking_type_id
+                    in carrier.send_delivery_notice_picking_type_ids
+                ):
+                    picking.carrier_id = carrier
+                    picking.send_to_shipper()
+                    picking.carrier_id = False
```
simahawk: wouldn't make sense to keep the carrier here?

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,85 @@
...
+        return super().send_to_shipper()
+
+    def _get_carrier_operation_from_chained_picking(self):
+        """Returns the first operation having a carrier defined in the next
+        chained operations.
+        """
+        # FIXME: almost the same code than 'get_carrier_from_chained_picking'
+        # of 'stock_helper_delivery' module, maybe to refactor/merge
+        # to make 'get_carrier_from_chained_picking' returning both carrier
+        # and its related picking?
```
simahawk: yeah, or simply make it return the ship picking only?

delivery_send_to_shipper_at_operation/models/stock_picking.py
simahawk: yes, just a loop and pick the 1st one matching

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,85 @@
...
+            for move_dest in moves_dest:
+                if move_dest.picking_id.carrier_id:
+                    return move_dest.picking_id
+            moves_dest = get_moves_dest(moves_dest)
+        return self.browse()
+
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        # Override to hide the "Send to shipper" button if the delivery
```
simahawk, line 59: why do you need this? Just place `delivery_notification_sent` hidden in the view and hide the btn, no?

delivery_send_to_shipper_at_operation/tests/test_send_to_shipper.py
```diff
@@ -0,0 +1,129 @@
...
+
+import mock
+
+from odoo.tests.common import SavepointCase
+
+
+class TestDeliverySendToShipper(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestDeliverySendToShipper, cls).setUpClass()
```
simahawk:
```suggestion
        super().setUpClass()
```

delivery_send_to_shipper_at_operation/tests/test_send_to_shipper.py
```diff
@@ -0,0 +1,129 @@
...
+            mocked.assert_not_called()
+            self.assertFalse(self.shipping.delivery_notification_sent)
+        # Validate the ship: delivery notification is sent
+        with mock.patch.object(
+            type(self.shipping.carrier_id), "send_shipping"
+        ) as mocked:
+            self._validate_picking(self.shipping)
+            self.assertEqual(self.shipping.state, "done")
+            mocked.assert_called()
+            self.assertFalse(self.shipping.delivery_notification_sent)
```
simahawk, line 97: I would expect the flag to be true here. Or, do we expect to be able to send it multiple times on the ship?

simahawk (changes requested):

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: I preferred to unset it but good question, now the pack is done, we could keep it for traceability, especially if for whatever reason we change the carrier later on the ship (maybe something to consider too: unset the flag on the ship if we change its carrier).

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: There is already a lot of `attrs` on this button in the `delivery` module:
```xml
<button name="send_to_shipper" string="Send to Shipper" type="object" attrs="{'invisible':['|','|','|','|',('carrier_tracking_ref','!=',False),('delivery_type','in', ['fixed', 'base_on_rule']),('delivery_type','=',False),('state','not in',('done')),('picking_type_code', '=', 'incoming')]}"/>
```
I wanted to do like @guewen did in another module to tune dynamically the attrs.

simahawk: https://github.com/OCA/server-tools/tree/13.0/base_view_inheritance_extension like this?

delivery_send_to_shipper_at_operation/tests/test_send_to_shipper.py
sebalix: I was wondering the same, due to a lack of time I didn't do it: the thing is, right now with the default config we do not change at all the std behavior, but yes we could set this field in all cases. This would allows to filter operations from which we sent a delivery notification.

sebalix: We can not do that easily: if we set the `delivery_notification_sent` on a ship (with default Odoo behavior), the check done in `send_to_shipper` on this field will remove the possibility to call again this method to re-send the notification to the carrier, we should avoid that IMO (even if the button is removed from the form as soon as a tracking ref has been retrieved from the carrier, technically we should still be able to call this method).

delivery_send_to_shipper_at_operation/models/delivery_carrier.py
```diff
@@ -0,0 +1,41 @@
...
+        selection=[("ship", "Ship operation"), ("custom", "Defined operations")],
+        string="Send delivery notice on",
+        default="ship",
+        help=SEND_DELIVERY_NOTICE_ON_HELP,
+    )
+    send_delivery_notice_picking_type_ids = fields.Many2many(
+        comodel_name="stock.picking.type",
+        string="Send delivery notice operation types",
+        help="""
+            When an operation of the listed type will be mark as done, the
```
mmequignon, line 36:
```suggestion
            When an operation of the listed type will be marked as done, the
```

delivery_send_to_shipper_at_operation/models/stock_picking.py
mmequignon: I don't have any idea about good practices to do this, but this seems kind of tricky to me.

mmequignon (approved):
LGTM, even if the attrs update looks weird.

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: Yes, it is tricky, and for that need there is the module @simahawk put in link but I don't like adding a dependency on that in OCA modules... it's something which should be supported by Odoo.

sebalix: It has already been done here: https://github.com/OCA/delivery-carrier/blob/13.0/delivery_carrier_pricelist/models/stock_picking.py

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: done

delivery_send_to_shipper_at_operation/models/stock_picking.py
sebalix: I removed the line, now the carrier is kept on the operation.

delivery_send_to_shipper_at_operation/models/stock_picking.py
```diff
@@ -0,0 +1,90 @@
...
+    transfer_node_to_modifiers,
+)
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    delivery_notification_sent = fields.Boolean(default=False)
+
+    def _send_confirmation_email(self):
```
simahawk, line 21: Seems we have an issue when hooking here: the delivery fee lines are invoiced already.
To be checked.

sebalix:
Moved to https://github.com/OCA/delivery-carrier/pull/355, closing this one.

PR #823, opened 2021-03-31
FIX stock_shifting_: prevent to use a bad domain when multiple lots

bealdav:
Please could you have a look ?
cc @Kev-Roche @apollet


Kev-Roche (approved):

bealdav:
cc @hparfr

hparfr (approved):

PR #825, opened 2021-04-06
[13.0][ADD] New module 'stock_picking_delivery_link'

sebalix:
Adds link to the delivery on all intermediate operations (pick and pack with a pick/pack/ship setup).

It aims to replace https://github.com/OCA/stock-logistics-warehouse/pull/1115 to get a useful module on its own instead of a "helper" one.

image

Ref. 2356


stock_picking_delivery_link/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    ship_carrier_id = fields.Many2one(
+        comodel_name="delivery.carrier",
+        compute="_compute_ship_picking_data",
+        string="Related carrier",
+    )
+
+    def _compute_ship_picking_data(self):
+        for picking in self:
+            ship = self._get_ship_from_chain()
```
simahawk: :fire: 
```suggestion
            ship = picking._get_ship_from_chain()
```

simahawk (changes requested):
shouldn't stay in delivery-carrier?

stock_picking_delivery_link/models/stock_picking.py
sebalix: fixed

sebalix:
Not sure, to me it's more related to the workflow of operations than the configuration of delivery carriers (https://github.com/OCA/delivery-carrier/#readme )

pedrobaeza:
Shouldn't this go in OCA/delivery-carrier?

sebalix:
@pedrobaeza As said I don't think but I could move it there if you prefer?

pedrobaeza:
For me all modules depending on `delivery` has that special repo, as this one is very populated, but I don't want to impose it.

simahawk:
@sebalix to be honest it makes sense. If I'm looking for something related to "delivery" that's the 1st place where I'd go normally.

sebalix:
Moved to https://github.com/OCA/delivery-carrier/pull/354, closing this one.

PR #833, opened 2021-04-16
[ADD] stock_move_manual_lot

hbrunn:
This module forces users to manually select lots, undoing Odoo's automatic assignment of lots.

Usage
=====

To use this module, you need to check the box `Use manual lot selection` on some
picking type.

For pickings using this type, users will be forced to select lot numbers manually.


stock_move_manual_lot/tests/test_stock_move_manual_lot.py
```diff
@@ -0,0 +1,76 @@
...
+        self.assertEqual(self.env['stock.quant'].search([
+            ('location_id', '=', self.picking.location_dest_id.id),
+            ('product_id', '=', self.product.id),
+        ]).mapped('lot_id'), self.lot1)
+
+    def test_reassign_reservation(self):
+        picking2 = self.picking.copy()
+        picking2.action_assign()
+        self.assertEqual(picking2.move_line_ids.lot_id, self.lot2)
+        picking2.move_line_ids.manual_lot_id = self.lot1
```
StefanRijnhart, line 93: You can set the lot here but because there is no unreserved quantity for this lot Odoo will reset the move line's product quantity.
```
-> self.assertEqual(picking2.move_line_ids.lot_id, self.lot2)
(Pdb) picking2.move_line_ids.read(["lot_id", "product_qty"])
[{'id': 53, 'lot_id': (41, 'lot2'), 'product_qty': 1.0}]
(Pdb) n
> /home/bruno/buildout/ou120/parts/stock-logistics-workflow/stock_move_manual_lot/tests/test_stock_move_manual_lot.py(68)test_reassign_reservation()
-> picking2.move_line_ids.manual_lot_id = self.lot1
(Pdb) 
> /home/bruno/buildout/ou120/parts/stock-logistics-workflow/stock_move_manual_lot/tests/test_stock_move_manual_lot.py(71)test_reassign_reservation()
-> with self.assertRaises(UserError), self.env.clear_upon_failure():
(Pdb) picking2.move_line_ids.read(["lot_id", "product_qty"])
[{'id': 53, 'lot_id': (40, 'lot1'), 'product_qty': 0.0}]
```
So I maintain you need to unreserve explicitely first.

hbrunn: hmm, in my first naive version I called [_update_available_quantity(-this.product_qty)](https://github.com/OCA/OCB/blob/12.0/addons/stock/models/stock_quant.py#L189), but because [stock.move.line#write](https://github.com/OCA/OCB/blob/12.0/addons/stock/models/stock_move_line.py#L268) does so too, this breaks when we've only 2 on hand. Will investigate next week

yostashiro:
@hbrunn Is this module trying to handle different business needs than what https://github.com/OCA/stock-logistics-warehouse/tree/12.0/stock_quant_manual_assign does?

hbrunn:
@yostashiro thanks for the hint, but it's indeed another business requirement - here we want to make sure that a user manually chose a serial number, the addon you link allows us to do so, but doesn't enforce it.

stock_move_manual_lot/tests/test_stock_move_manual_lot.py
hbrunn: yes, we indeed need to [free the reservation](https://github.com/OCA/stock-logistics-workflow/pull/833/files#diff-8e597cc2aff6a170901caba6b907f585528d98ac5913bc38aa002775aa97a2dcR32) - the codepath in write [doing so](https://github.com/OCA/OCB/blob/12.0/addons/stock/models/stock_move_line.py#L308) only does it after it calculates the available quantity

stock_move_manual_lot/models/stock_move_line.py
```diff
@@ -46,5 +52,15 @@ def _reserve_manual_lot(self, vals):
...
+                )
+                updated_picking_ids = list(map(
+                    operator.itemgetter(0), self.env.cr.fetchall(),
+                ))
 
             this.lot_id = this.manual_lot_id
+
+            if updated_picking_ids:
+                # see if we can assign pickings that were unassigned above
+                self.env['stock.picking'].browse(updated_picking_ids).action_assign()
```
StefanRijnhart: I think `action_assign` might raise, so you might want to wrap in try + saveoint.

stock_move_manual_lot/models/stock_move_line.py
```diff
@@ -36,7 +37,12 @@ def _reserve_manual_lot(self, vals):
                 package_id=this.package_id,
                 owner_id=this.owner_id,
             )
+            updated_picking_ids = []
             if available_qty < this.product_qty:
+                # free_reservation might unassign pickings, so we record which
+                # ones have been changed
+                self.env.cr.execute('select now()')
```
StefanRijnhart: `now() at time zone 'UTC'`

stock_move_manual_lot/models/stock_move_line.py
hbrunn: thanks, done

stock_move_manual_lot/tests/test_stock_move_manual_lot.py
```diff
@@ -0,0 +1,117 @@
...
+        picking2.action_assign()
+        self.assertEqual(picking2.move_line_ids.lot_id, self.lot2)
+        self.assertTrue(self.picking.move_line_ids)
+        picking2.move_line_ids.manual_lot_id = self.lot1
+        self.assertFalse(self.picking.move_line_ids)
+        with self.assertRaises(
+            UserError
+        ), self.env.clear_upon_failure(), self.env.cr.savepoint():
+            self.picking.button_validate()
+        self.picking.action_assign()
```
StefanRijnhart: Can we do without this now?

stock_move_manual_lot/models/stock_move_line.py
```diff
@@ -0,0 +1,73 @@
...
+                updated_picking_ids = list(map(
+                    operator.itemgetter(0), self.env.cr.fetchall(),
+                ))
+
+            this.lot_id = this.manual_lot_id
+
+            if updated_picking_ids:
+                # see if we can assign pickings that were unassigned above
+                with self.env.cr.savepoint():
+                    try:
```
StefanRijnhart: `savepoint` under the scope of `try`, otherwise the rollback won't happen when an exception occurs.

hbrunn:
@StefanRijnhart fixed

StefanRijnhart:
Thanks for the updates! Would it be a good idea to set manual_lot_id to lot_id in a post init hook for move lines in state `done`?

hbrunn:
I [think so](https://github.com/OCA/stock-logistics-workflow/pull/833/commits/9f01bf18a36b586c1ce6ab5f5ecb3337e8b7fcc3#diff-b5c90afa440ac94561dabefc34a0c62373b79981cc33dd436972356cef732709R5)

stock_move_manual_lot/models/stock_picking.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    use_manual_lot_selection = fields.Boolean(
+        related="picking_type_id.use_manual_lot_selection",
+    )
+
+    def button_validate(self):
```
StefanRijnhart, line 13: Seem to have missed this earlier, but I would appreciate a docstring clarifying what this does.

hbrunn: [done](https://github.com/OCA/stock-logistics-workflow/pull/833/commits/94059211c9a410ad3fd7a3ac9a4e6fd89d71da53#diff-e6d83dd27f05ea77e0d86d96158521c823c7e9c6f662f885effedd1e8a9bc7c9R14)

StefanRijnhart (approved):
Perfect, thanks!

StefanRijnhart:
OK nearly perfect
`stock_move_manual_lot/hooks.py:3:1: E302 expected 2 blank lines, found 0`

hbrunn:
done

hbrunn:
@StefanRijnhart please re-review my fixup commit: 6a17e8099168a4086267cdc5e1a050c8d5161e63

rousseldenis:
@chrisdec I think this one should be your interest

stock_move_manual_lot/views/stock_move_line.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field name="arch" type="xml">
+            <form position="inside">
+                <field name="use_manual_lot_selection" invisible="True" />
+            </form>
+            <field name="lot_id" position="before">
+                <field
+                    name="manual_lot_id"
+                    groups="stock.group_production_lot"
+                    context="{'default_product_id': product_id, 'active_picking_id': picking_id}"
+                    attrs="{'invisible': ['|', '|', ('use_manual_lot_selection', '=', False), ('lot_id', '!=', False), ('lot_name', '!=', False)]}"
```
StefanRijnhart: This way, `manual_lot_id` is invisible when either
* the` functionality is disabled for the picking type, or
* lot_id is set, or
* a name for a new lot is set

Isn't it the case that lot_id is already set when the move line is reserved with just any lot (which is our basic use case)? In that case, we will want to see this field I think.

hbrunn: yes, I think now it's more sensible

stock_move_manual_lot/models/stock_move_line.py
StefanRijnhart: Thanks! You may squash now.

stock_move_manual_lot/views/stock_move_line.xml
StefanRijnhart: Thanks! You may squash now.

astirpe (approved):

PR #834, opened 2021-04-21
[13.0][IMP] s_p_g_by_partner_by_carrier: disable grouping on partner

sebalix:
Allow to disable the grouping feature on partner and fallback on the standard Odoo behavior.

Ref. 2468

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_move.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_move.py
@@ -57,6 +57,7 @@ class StockMove(models.Model):
         domain = super()._domain_search_picking_for_assignation()
         if (
             not self.picking_type_id.group_pickings
+            or self.partner_id.disable_picking_grouping
             or self.group_id.sale_id.picking_policy == "one"
         ):
             return domain
--- a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
@@ -58,11 +58,14 @@ class StockPicking(models.Model):
             return super().action_cancel()
 
     def _create_backorder(self):
-        if self.picking_type_id.group_pickings:
-            self = self.with_context(picking_no_copy_if_can_group=1)
-        backorders = super()._create_backorder()
-        if self.picking_type_id.group_pickings:
-            backorders._merge_procurement_groups()
+        backorders = self.browse()
+        for picking in self:
+            if not picking._is_grouping_disabled():
+                picking = picking.with_context(picking_no_copy_if_can_group=1)
+            backorder = super(StockPicking, picking)._create_backorder()
+            if not picking._is_grouping_disabled():
+                backorder._merge_procurement_groups()
+            backorders |= backorder
         return backorders
 
     def _prepare_merge_procurement_group_values(self, move_groups):
@@ -74,7 +77,7 @@ class StockPicking(models.Model):
 
     def _merge_procurement_groups(self):
         for picking in self:
-            if not picking.picking_type_id.group_pickings:
+            if picking._is_grouping_disabled():
                 continue
             if picking.picking_type_id.code != "outgoing":
                 continue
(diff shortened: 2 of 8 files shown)


stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -100,6 +101,11 @@ def _domain_search_picking_handle_move_type(self):
         return [("move_type", "=", self.group_id.move_type)]
 
     def _key_assign_picking(self):
+        if (
+            not self.picking_id.picking_type_id.group_pickings
+            and self.picking_id.partner_id.disable_picking_grouping
```
gurneyalex: the previous diff chunk as an "or" and I think it is correct. Can you check?
```suggestion
            or self.picking_id.partner_id.disable_picking_grouping
```

gurneyalex: suggestion: have a helper method _is_grouping_disabled returning 

```
    not self.picking_type_id.group_pickings
    or self.partner_id.disable_picking_grouping
```

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -77,6 +79,8 @@ def _merge_procurement_groups(self):
         for picking in self:
             if not picking.picking_type_id.group_pickings:
                 continue
+            if picking.partner_id.disable_picking_grouping:
+                continue
```
gurneyalex, line 81: use if picking._is_grouping_disabled() to group the first two if

stock_picking_group_by_partner_by_carrier/models/stock_move.py
jgrandguillaume: Agreed

jgrandguillaume (approved):
All good but the small suggestion by Alex, think it worth it. Thanks for adding tests :)

stock_picking_group_by_partner_by_carrier/models/stock_move.py
jbaudoux: :+1:

simahawk (approved):
LG apart from what @gurneyalex pointed out

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -59,6 +59,8 @@ def action_cancel(self):
             return super().action_cancel()
 
     def _create_backorder(self):
+        if self.partner_id.disable_picking_grouping:
```
simahawk: :fire: this override is wrong. You should loop on self.
Found out because our CI is failing :wink:

simahawk (changes requested):

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
simahawk: also, the check below for setting the ctx key should be probably done by grouping per picking type? :thinking:

sebalix: What is strange is this issue wasn't triggered before on the existing check, we were lucky.
I reimplemented the override to loop on each picking, your review is welcome.

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -57,6 +57,7 @@ def _domain_search_picking_for_assignation(self):
         domain = super()._domain_search_picking_for_assignation()
         if (
             not self.picking_type_id.group_pickings
+            or self.partner_id.disable_picking_grouping
```
gurneyalex, line 60: _is_grouping_disabled() ?

simahawk (approved):
preapproving

stock_picking_group_by_partner_by_carrier/models/stock_move.py
sebalix: In this method (stock.move level) we have no picking yet (it's the purpose of the method to find one), so we can't use the `_is_grouping_disabled` method of the picking here. We could implement one for the move maybe but it's only used here.

PR #841, opened 2021-05-04
[13.0][IMP] stock_valuation_layer_usage: output layer selection for MTO case

AaronHForgeFlow:
This new module should be used together with stock_valuation_layer_usage.

This module means to fix this siutation:

When creating the out svl, Odoo takes the first layer available with value. The module stock_valuation_layer_usage arises an issue with that Odoo process. Odoo will take always the oldest svl. However, that is not the case for MTO. When there is a fixed link between the incoming move and the outgoing move, at the time of creating the outgoing svl the system should take the specific incoming svl not just the oldest one. Also, when the case is not MTO, the system should avoid using svl that are "reserved".

@ForgeFlow @MiquelRForgeFlow @JordiBForgeFlow

### DIFF
--- a/stock_account_product_run_fifo_hook/hooks.py
+++ b/stock_account_product_run_fifo_hook/hooks.py
@@ -16,18 +16,15 @@ def post_load_hook():
         # Find back incoming stock valuation layers (called candidates here)
         # to value `quantity`.
         qty_to_take_on_candidates = quantity
+        # START HOOK Search Candidates
+        candidates_domain = self._get_candidates_domain(company)
         candidates = (
             self.env["stock.valuation.layer"]
             .sudo()
             .with_context(active_test=False)
-            .search(
-                [
-                    ("product_id", "=", self.id),
-                    ("remaining_qty", ">", 0),
-                    ("company_id", "=", company.id),
-                ]
-            )
+            .search(candidates_domain)
         )
+        # END HOOK Search Candidates
         new_standard_price = 0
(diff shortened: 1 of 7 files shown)


stock_valuation_layer_usage_mto/readme/HISTORY.rst
```diff
@@ -0,0 +1,5 @@
+13.0.1.0.0 (2020-01-03)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* [13.0][MIG] stock_valuation_layer_usage
```
MiquelRForgeFlow:
```suggestion
* [13.0][MIG] stock_valuation_layer_usage_mto
```

MiquelRForgeFlow (commented):
But I think both modules should be fused. I don't get why make it separated.

AaronHForgeFlow:
> But I think both modules should be fused. I don't get why make it separated.

I almost agree but I am a bit unsure because you may want to see the layer usages as a report, but maybe you don't want to change the standard layer selection (even if it is wrong).

Now I joined the two modules, not 100% sure. What do you think @JordiBForgeFlow ?

JordiBForgeFlow:
@AaronHForgeFlow ok, works for me.

MiquelRForgeFlow (approved):

AaronHForgeFlow:
Changed the logic to a more efficient one now. The hook for the svl selection is just for the domain

JordiMForgeFlow (approved):
LGTM 👍🏼

PR #846, opened 2021-05-21
[MIG] stock_split_picking: Migration to 14.0

RujiaL:
@gdgellatly


Reyes4711-S73:
Functional review

sbejaoui (approved):

rousseldenis (approved):

Abranes (approved):
LGTM

dominiquekonsuntack:
There is case where this module does not behave intuitively.
if the stock picking has e.g. two stock moves - (product 1 = 1000), (product 2 = 1000)
let's say that we want to split partially the first line. So the user will enter product 1 qty_done = 200, and product 2 qty_done = 0

the expected behaviour is:
picking 1: (product 1, qty=200)
picking 2: (product 1, qty = 800,)  and (product 2, qty = 1000)

but this module will give instead:
picking 1: (product 1, qty=200) and (product 2, qty=1000)
picking 2: (product 1, qty=800)

in other word, the handling of qty_done = 0 is not handled in a manner consistent with when qty_done > 0

gdgellatly:
> There is case where this module does not behave intuitively.
> if the stock picking has e.g. two stock moves - (product 1 = 1000), (product 2 = 1000)
> let's say that we want to split partially the first line. So the user will enter product 1 qty_done = 200, and product 2 qty_done = 0
> 
> the expected behaviour is:
> picking 1: (product 1, qty=200)
> picking 2: (product 1, qty = 800,) and (product 2, qty = 1000)
> 
> but this module will give instead:
> picking 1: (product 1, qty=200) and (product 2, qty=1000)
> picking 2: (product 1, qty=800)
> 
> in other word, the handling of qty_done = 0 is not handled in a manner consistent with when qty_done > 0

@dominiquekonsuntack This isn't to do with this PR. It is just a straight migration, there were no introduced behaviour changes in this PR. Really you should raise an issue for that or raise a separate PR. That way it can be back/forwardported easier.

PR #847, opened 2021-05-21
[MIG] stock_picking_backorder_strategy: Migration to 14.0

RujiaL:
@gdgellatly


andreampiovesana:
if you use batch picking you have an error -> suggested change are:
    def _check_backorder(self):
        #self.ensure_one()
        # If strategy == 'manual', let the normal process going on
        for picking in self:
            if picking.picking_type_id.backorder_strategy == "manual":
                return super(StockPicking, self)._check_backorder()
        return self.browse()

stock_picking_backorder_strategy/models/stock_picking.py
```diff
@@ -8,10 +8,10 @@ class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     def _check_backorder(self):
-        self.ensure_one()
-        # If strategy == 'manual', let the normal process going on
-        if self.picking_type_id.backorder_strategy == "manual":
-            return super(StockPicking, self)._check_backorder()
+        for picking in self:
```
gdgellatly: This is wrong. The guts of it is as I can tell is we are supposed to return a list of pickings if the backorder_strategy is manual and nothing if not. In theory multiple pickings could have multiple types and multiple strategies. Therefore what needs to happen is pretty much something like in pseudo code.

```
self = self.filtered(<backorder_strategy> == 'manual')
return super()
```

andreampiovesana (approved):
go

stock_picking_backorder_strategy/models/stock_picking.py
```diff
@@ -12,7 +12,7 @@ def _check_backorder(self):
         # If strategy == 'manual', let the normal process going on
         if self.picking_type_id.backorder_strategy == "manual":
             return super(StockPicking, self)._check_backorder()
-        return False
+        return self.browse()
```
sebalix: This method should be re-implemented as the standard one is able to check multiple pickings ([link](https://github.com/odoo/odoo/blob/3711c299523c8da4abb6066e505ec05090d1567d/addons/stock/models/stock_picking.py#L1017-L1041)), here the `self.ensure_one()` is breaking this.

gdgellatly: @sebalix This should have been fixed, I did the review myself and saw the code. It must never have been pushed. Anyway RujiaL won't be working on Odoo anymore as she left so make a PR and I can review.

gdgellatly: @sebalix I don't know where you got your snippet, it is not the current code

andreampiovesana (approved):
test ok

sebalix:
Hi @gdgellatly @andreampiovesana , I move this PR to https://github.com/OCA/stock-logistics-workflow/pull/895 , your reviews are welcome there :)
I'm closing this one.

stock_picking_backorder_strategy/models/stock_picking.py
sebalix: I don't know myself! Sorry for the noise.

PR #853, opened 2021-06-10
[WIP][13.0][ADD] stock_quant_constraint New module to guarantee quant information

NL66278:
Prevent problems with quants that are inconsistent with stock moves and/or stock move lines.


stock_quant_constraint/models/stock_quant.py
```diff
@@ -0,0 +1,206 @@
...
+        if not quants_unique:
+            raise UserError(_("Cron to check quants aborted because of duplicat keys"))
+        # Continue checking quants for mismatches.
+        cursor.execute(STOCK_STATEMENT)
+        for row in cursor.fetchall():
+            _logger.error(_("Mismatch in stock %s"), str(row))
+            self._do_repair(row)
+
+    def _do_repair(self, row):
+        """Repair (or create) stock_quant, based on stock moves."""
```
gdgellatly, line 206: @NL66278 Well the docstring looks promising!

I have quite some experience in this problem. The biggest issue is of course failed to reserve/unreserve. I have never quite traced it but I feel it has to do with stock.inventory validation. In general the total qty is correct, but the movelines are not. The other one is within a failed transaction, again never quite tracked it, but the quant gets committed and the move rolled back, which of course breaks 2 quants.

I am very interested in this problem and have some database full of inconsistencies to test. V12 and V14 only mind you but looks pretty easy backport for me once you have something. Just ping when ready.

NL66278: @gdgellatly I did not have time to work further on this, but I have now asked one of our programmers to get involved.

stock_quant_constraint/models/stock_quant.py
```diff
@@ -0,0 +1,206 @@
...
+        readonly=True,
+    )
+    quantity_computed = fields.Float(
+        "Quantity in Stock Move Lines",
+        help="Difference between computed and actual quantity",
+        readonly=True,
+    )
+
+    @api.constrains("product_id", "reserved_quantity")
+    def check_reserved_quantity(self):
```
ntsirintanis, line 128: The problem here is that this constraint method is called before `stock.move.line` is updated, therefore it is risen any time we perform` action_assign()` in  a stock picking. More specifically, the aforementioned action calls

https://github.com/OCA/OCB/blob/13.0/addons/stock/models/stock_move.py#L1123

and a call to `self.env['stock.quant']._update_reserved_quantity()` therein.

NL66278: Good catch. Please feel free to change anything that is needed. You can fork my branch and then put a PR on my repo....

ntsirintanis: I am trying to think of a use case where having this constraint will be useful. I am not sure, can you please elaborate on that decision? In my opinion, it is not needed. What we might need is an action that finds faulty quants/stock move lines, and a cron job that corrects them, with proper logging etc. The action that finds faulty stuff is already there, it is the constraint method. I can implement a correction method, that would basically equate `quant.reserved_quantity` and `sum(move_lines.product_qty`

NL66278: The thing is that with a repair action, without a constraint, you can never be sure where the inconsistencies come from. When an exception is thrown the moment the database threatens to be made inconsistent, you know from the stack trace exactly where the error is coming from.

stock_quant_constraint/README.rst
```diff
@@ -0,0 +1,95 @@
...
+bad performance.
+
+With stock quants performance improved immensely, but is also lead to redundancy in
+the database, and therefore inconsistency.
+
+This module will provide methods to report on inconsistencies, and where possible to
+repair them.
+
+The information from the stock moves (and lines) will be considered 'The Truth'.
+Quants will have to confirm to these.
```
thomaspaulb, line 41: conform : P

PR #854, opened 2021-06-14
[13.0] Add stock_dangerous_goods

mmequignon:
This glue module between `stock` and `l10n_eu_product_adr` adds a few utility fields


stock_dangerous_goods/models/product_product.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    is_lq_product = fields.Boolean(compute="_compute_is_lq_product")
```
simahawk: add `help`

stock_dangerous_goods/models/stock_quant_package.py
```diff
@@ -0,0 +1,16 @@
...
+class StockQuantPackage(models.Model):
+    _inherit = "stock.quant.package"
+
+    has_lq_products = fields.Boolean(compute="_compute_has_lq_products")
+
+    def _compute_has_lq_products(self):
+        for record in self:
+            record.has_lq_products = any(
+                record.mapped("quant_ids.product_id.is_lq_product")
+            )
```
simahawk:
```suggestion
            record.has_lq_products = record._has_lq_product()

    def _has_lq_product(self):
        for quant in self.quant_ids:
            if quant.product_id.is_lq_product:
                return True
```

simahawk (changes requested):

stock_dangerous_goods/models/product_product.py
```diff
@@ -0,0 +1,20 @@
...
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    is_lq_product = fields.Boolean(
+        compute="_compute_is_lq_product",
+        help="Whether this product is a Limited Quantity product or not",
+    )
+
+    def _compute_is_lq_product(self):
```
TDu, line 16: The function needs an `api.depends`

stock_dangerous_goods/models/stock_quant_package.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import fields, models
+
+
+class StockQuantPackage(models.Model):
+    _inherit = "stock.quant.package"
+
+    has_lq_products = fields.Boolean(compute="_compute_has_lq_products")
+
+    def _compute_has_lq_products(self):
```
TDu, line 13: Also needs an `api.depends`

stock_dangerous_goods/models/stock_quant_package.py
```diff
@@ -0,0 +1,20 @@
...
+        for record in self:
+            record.has_lq_products = record._has_lq_products()
+
+    def _has_lq_products(self):
+        self.ensure_one()
+        for quant in self.quant_ids:
+            if quant.product_id.is_lq_product:
+                return True
```
TDu, line 21: Would be nicer to return `False` if there is no lq products, instead of relying to the default return value

stock_dangerous_goods/models/stock_quant_package.py
mmequignon: No, this fields isn't stored and will be computed every time

stock_dangerous_goods/models/product_product.py
mmequignon: Not stored, no need to set an api.depends

stock_dangerous_goods/models/stock_quant_package.py
simahawk: Actually depends is not just for stored field. It makes sure cache is invalidated where already computed.
It also documents on which fields you depend on functionally.
So, not mandatory, good to have.

TDu: I would think the `compute` method is also useful for non stored fields. So its value is not recomputed every time it is used, it is still temporarily stored in memory.

simahawk (approved):
LG, non blocking comments on depends. Please squash on your next push.

stock_dangerous_goods/models/product_product.py
sebalix: `api.depends` is not used only when `store=True`, computed fields also act as onchange, here if I tick the `is_dangerous` field on the form it's better to see the `is_lq_product` field updated on the fly, for that you need to define an `api.depends`.

mmequignon:
@TDu @sebalix I updated my PR.

sebalix (approved):

rousseldenis:
Just one remark, I would have named it l10n_eu_stock_dangerous_good as it is not a 'l10n_eu' agnostic module.

dreispt:
Travis is red.

simahawk:
```
----------

FAIL: TestStockValuationLayerUsage.test_02_drop_ship

  File "/home/travis/build/OCA/stock-logistics-workflow/stock_valuation_layer_usage/tests/test_stock_valuation_layer_usage.py", line 294, in test_02_drop_ship

    self.assertEquals(balance_grni, -10.0)

AssertionError: 0 != -10.0
```
https://app.travis-ci.com/github/OCA/stock-logistics-workflow/jobs/532510229#L1971

no time to solve it now :cry:

simahawk:
NOTE: I made some unrelated changes to boost some test classes

simahawk:
@JordiBForgeFlow hola, could you check this failing test? https://app.travis-ci.com/github/OCA/stock-logistics-workflow/jobs/535992180#L1874 There's no change in this PR that affects it.

I've tried w/ and without this module or its dependencies. Nothing. The test was failing even before I refactored its setup to use setUpClass of course. 

The issue seems caused by the lack of AML found for the product.

mmequignon:
> @JordiBForgeFlow hola, could you check this failing test? https://app.travis-ci.com/github/OCA/stock-logistics-workflow/jobs/535992180#L1874 There's no change in this PR that affects it.
> 
> I've tried w/ and without this module or its dependencies. Nothing. The test was failing even before I refactored its setup to use setUpClass of course.
> 
> The issue seems caused by the lack of AML found for the product.

I don't know how this module tests can run correctly, since it uses `product.template.invoice_policy` which is introduced by the `sale` module, on which it does not depends.
I'm unable to run `stock_valuation_layer_usage` when installed alone.

simahawk:
> > @JordiBForgeFlow hola, could you check this failing test? https://app.travis-ci.com/github/OCA/stock-logistics-workflow/jobs/535992180#L1874 There's no change in this PR that affects it.
> > I've tried w/ and without this module or its dependencies. Nothing. The test was failing even before I refactored its setup to use setUpClass of course.
> > The issue seems caused by the lack of AML found for the product.
> 
> I don't know how this module tests can run correctly, since it uses `product.template.invoice_policy` which is introduced by the `sale` module, on which it does not depends.
> I'm unable to run `stock_valuation_layer_usage` when installed alone.

I fixed it here https://github.com/OCA/stock-logistics-workflow/pull/854/commits/9cd0fe5fb3d24b45b1d6882423ad4d63ede10767

You can run tests if you reset hard on this branch version ;)

simahawk:
FTR a test w/out all the tests improvements has been made here https://github.com/OCA/stock-logistics-workflow/pull/887

TDu (approved):

rousseldenis:
@simahawk @mmequignon What's the status of this concerning tests ?

simahawk:
> @simahawk @mmequignon What's the status of this concerning tests ?

status is: no time to get back to it so far :sob: 

We'll have to do it soon though. But we are facing issue on v14 too because of the refactoring of the base module.... we'll see :man_shrugging:

simahawk:
@mmequignon pls try to get back to this when you have time :pray:

rousseldenis:
@mmequignon

rousseldenis:
@mmequignon Could you rebase ?

bosd:
ping @mmequignon

rousseldenis:
@mmequignon

mmequignon:
sorry, totally forgot about that...

bosd:
> @mmequignon Could you rebase ?

:pray:

rousseldenis:
@mmequignon What's the status of this ?

PR #857, opened 2021-06-21
[10.0] stock_cost_method_last: also update price when move is from production

alexis-via:


alexis-via:
Link to the native code product_price_update_before_done(): https://github.com/odoo/odoo/blob/10.0/addons/stock_account/models/stock.py#L200
In the native code, we have "move.location_id.usage in ('supplier', 'production')", so we must do the same here.

RaulOForgeFlow (approved):

PR #858, opened 2021-06-22
[13.0][IMP] stock_split_picking: allow splitting pickings waiting for other operations

JordiMForgeFlow:
Make the split option available when the picking is in the "Waiting Another Operation" status. If, for example, you want to split two MTO moves into different pickings, you should be able to do so.

### DIFF
--- a/stock_split_picking/views/stock_partial_picking.xml
+++ b/stock_split_picking/views/stock_partial_picking.xml
@@ -8,7 +8,7 @@
             <field name="state" position="before">
                 <button
                     name="%(stock_split_picking.action_stock_split_picking)s"
-                    states="draft,confirmed,assigned"
+                    states="draft,confirmed,waiting,assigned"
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (commented):
If the picking is split both pickings would be waiting, right?

JordiMForgeFlow:
> If the picking is split both pickings would be waiting, right?

Yes, the pickings will be in the state computed based on their stock moves.

AaronHForgeFlow (approved):

LoisRForgeFlow (approved):
I don't see any drawback :+1:

JordiMForgeFlow:
Hi @HaraldPanten, could you review this change? :)

HaraldPanten (approved):
@JordiMForgeFlow just changes considering the picking status. LGTM 👍

PR #859, opened 2021-06-22
[13.0] s_p_group_by_partner_by_carrier: Fix translations in report

mmequignon:
On the deliveryslip report, in the remaining products to deliver part, the lang have to be passed in order to make everything correctly translated.

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.py
+++ b/stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.py
@@ -65,6 +65,10 @@ class DeliverySlipReport(models.AbstractModel):
 
     @api.model
     def get_remaining_to_deliver(self, picking):
+        # As data fetched here is meant to be displayed on a report,
+        # set the lang
+        self = self.with_context(lang=picking.partner_id.lang)
+        picking = picking.with_context(lang=picking.partner_id.lang)
         sales_data = self._get_remaining_to_deliver(picking)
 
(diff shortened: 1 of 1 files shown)


stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.py
```diff
@@ -65,6 +65,10 @@ def _get_remaining_to_deliver(self, picking):
 
     @api.model
     def get_remaining_to_deliver(self, picking):
+        # As data fetched here is meant to be displayed on a report,
+        # set the lang
+        self = self.with_context(lang=picking.partner_id.lang)
+        picking = picking.with_context(lang=picking.partner_id.lang)
```
sebalix, line 71: We should expect that `picking` has already the right context as the lang is passed here https://github.com/odoo/odoo/blob/9ad46806804862d2bc4cf331c00530763a055e67/addons/stock/report/report_deliveryslip.xml#L102 .
Unless the issue is coming from here https://github.com/OCA/stock-logistics-workflow/blob/b6950de2adbb4c4ae973a0f7a3836da1b7577ce2/stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.py#L92-L104 the lang could be set on `docs` directly? Weird that the translation issue occurs only in the "remaining to deliver" block.

simahawk: I'd set it on docs indeed.

PR #860, opened 2021-06-23
[14.0][mig] stock_picking_sale_order_link

dreispt:
- [9.0][ADD] stock_picking_sale_order_link: Allow navigate from picking to sale order
- OCA Transbot updated translations from Transifex
- [UPD] Update stock_picking_sale_order_link.pot
- [11.0][MIG] stock_picking_sale_order_link: Migration to v11.0
- [UPD] Update stock_picking_sale_order_link.pot
- [UPD] README.rst
- [MIG] stock_picking_sale_order_link: Migration to 12.0
- [UPD] README.rst
- [UPD] Update stock_picking_sale_order_link.pot
- Update translation files
- [UPD] README.rst
- Translated using Weblate (Portuguese (Brazil))
- [IMP] stock_picking_sale_order_link: black, isort
- [MIG] stock_picking_sale_order_link: Migration to v13.0
- [UPD] Update stock_picking_sale_order_link.pot
- pre-commit update
- Added translation using Weblate (Chinese (Simplified))
- Translated using Weblate (Chinese (Simplified))
- [FIX] stock_picking_sale_order_link: New sale line added to done picking if access form picking to sale with smartbutton TT26009
- stock_picking_sale_order_link 13.0.1.0.1
- [IMP] stock_picking_sale_order_link: black, isort, prettier
- [MIG] stock_picking_sale_order_link: Migration to 14.0

PR #861, opened 2021-06-30
14.0 mig stock picking sale order link

Zar-SW:


andreampiovesana:
you shoud close for https://github.com/OCA/stock-logistics-workflow/pull/860

StefanRijnhart:
Indeed. This is a duplicate. But also, it seems like you did not apply the `git format-patch` method as decribed on https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0 (under `Technical method`). As you can see on the other PR, you should end up with a branch that contains all the previous commits from the 13.0 branch that altered the source code of this particular module. Please keep that in mind for your next port!

PR #862, opened 2021-06-30
[13.0] stock_partner_delivery_window:: Fix default value on delivery_time_preference

grindtildeath:
The default value introduced by workdays feature breaks the installation
of Demo databases as the change is not reflected in sale_partner_delivery_window.

cc @simahawk @pedrobaeza @mmequignon


grindtildeath:
Closes: https://github.com/OCA/sale-workflow/issues/1621

pedrobaeza (approved):

PR #867, opened 2021-07-16
[ADD] stock_picking_force_assign

hbrunn:
This module adds the possibility to automatically unreserve currently reserved
pickings in order to assign some currently unreserved picking.


StefanRijnhart:
Looking great! Could we consider posting a comment on the unreserved pickings? And 'auto unreserve' is more of an implementation detail, and maybe less suitable for the module and button name. How about stock_picking_force_reserve (or stock_picking_force_assign for old times' sake)?

hbrunn:
I was actually tempted to call this exactly `stock_picking_force_assign` but then feared this would be misleading because it doesn't force in the sense of negative quants. But I think I agree, what we actually do is an implementation detail that might change or be amended.

So we have

- [x] post message on unreserved pickings why they were unreserved (can we by now link in a convenient way in mail.messages?)
- [x] rename module to stock_picking_force_assign
- [x] rename button to 'Force reservation'

?

StefanRijnhart:
Perfect!

antoniocanovas (approved):
Functional review ok

hbrunn:
@StefanRijnhart followup done

stock_picking_force_assign/security/stock_picking_force_assign_security.xml
```diff
@@ -0,0 +1,11 @@
+<odoo>
+    <record id="group_stock_force_assign" model="res.groups">
+        <field name="name">Auto unreserve pickings</field>
+        <field name="category_id" ref="base.module_category_warehouse_management"/>
```
StefanRijnhart: Does this not break the warehouse manager -> warehouse user hierarchy, preventing those groups to show up as a selection field on the user tab?

hbrunn: yes, moved this to extra rights then

stock_picking_force_assign/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+        """ Return moves to unreserve in order to reserve pickings in self """
+        location = self.mapped('location_id')
+        result = self.env['stock.move']
+        assert len(location) == 1, 'Pickings need to be from the same location'
+        float_compare = functools.partial(
+            tools.float_compare,
+            precision_digits=result._fields['product_qty'].digits,
+        )
+
+        for move_line in self.mapped('move_lines'):
```
StefanRijnhart: You could consider changing variable name `move_line` to `stock_move` or `move`, as an act of resistance against the terminology fu in core Odoo, but it's just a nice to have.

StefanRijnhart (approved):

stock_picking_force_assign/models/stock_picking.py
hbrunn: I'm all for the resistance!

StefanRijnhart: Alright comrade!

PR #868, opened 2021-07-23
[13.0][ADD] stock_move_change_source_location

AdriaGForgeFlow:
This PR adds a new module that allows you to change the source location of one or several moves from a stock picking.

It's quite similar to this one https://github.com/OCA/stock-logistics-workflow/tree/13.0/stock_picking_operation_quick_change but that one changes the destination location of the stock move lines.

To use this module:
To change the source location of one or several stock moves of a picking, go to a specific picking and click ont the button Change Source Location.
You will have three options:

* Change All moves (the source location of all stock moves will be changed)
* Change only moves with matched OLD location (you need to provide a source location and only those moves with the same source location will be updated)
* Select moves to change (you will need to select the moves to update)

Once the option has been selected, choose the new source location and click Apply.

cc ~ @ForgeFlow


stock_move_change_source_location/wizards/stock_move_change_source_location_wizard.py
```diff
@@ -0,0 +1,119 @@
...
+            raise UserError(
+                _(
+                    "You can not change operations destination location if "
+                    "picking state is not in %s"
+                )
+                % ",".join(self._get_allowed_picking_states())
+            )
+
+    def _check_allowed_moves(self, moves):
+        if any([move.move_orig_ids for move in moves]):
```
JordiBForgeFlow, line 104: Perhaps you should also check for allowed states in moves (as you also do in pickings). Sometimes you could have a cancelled move in an non cancelled picking

AdriaGForgeFlow: You are right, added the extra check now :+1:

stock_move_change_source_location/wizards/stock_move_change_source_location_wizard.xml
```diff
@@ -0,0 +1,61 @@
...
+                    <field name="warehouse_view_location_id" invisible="1" />
+                    <field name="moves_to_change" />
+                    <field
+                        name="old_location_id"
+                        attrs="{'invisible': [('moves_to_change', '!=', 'matched')]}"
+                        options="{'no_create': True, 'no_open': True}"
+                    />
+                    <field
+                        name="new_location_id"
+                        domain="[('id', 'child_of', warehouse_view_location_id)]"
```
JordiMForgeFlow: I think that due to this domain the view location is being added as an available location:

image

AdriaGForgeFlow: I've added `("usage", "=", "internal")` in the domain and it should not appear anymore.

Thanks!

stock_move_change_source_location/views/stock_picking_view.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <header position="inside">
+                <button
+                    name="%(action_stock_move_change_source_location)d"
+                    type="action"
+                    string="Change Source Location"
+                    groups="stock.group_stock_user"
+                >
+                </button>
```
JordiMForgeFlow:
```suggestion
                <button
                    name="%(action_stock_move_change_source_location)d"
                    type="action"
                    string="Change Source Location"
                    groups="stock.group_stock_user"
                    attrs="{'invisible': [('picking_type_id.code', 'not in', ('outgoing', 'internal'))]}"
                >
                </button>
```

AdriaGForgeFlow: Done :+1:

JordiMForgeFlow (approved):
Code + functional review 👍🏼

rousseldenis (approved):
Seems great. Code review

JordiBForgeFlow (approved):
:+1: Functional and technical review.

PR #878, opened 2021-08-13
[12.0][IMP] stock_picking_force_assign: allow partial

StefanRijnhart:


hbrunn (approved):

stock_picking_force_assign/models/stock_picking.py
```diff
@@ -24,6 +24,12 @@ def action_force_assign_pickings(self):
         )))
         return self.action_assign()
 
+    def _force_assign_allow_partial(self):
+        """Indicate if partial forced assignments should be allowed"""
+        return self.env["ir.config_parameter"].sudo().get_param(
+            "stock_picking_force_assign.allow_partial"
+        ).strip().lower() in ("1", "true", "yes")
```
sebalix: You could use [`str2bool`](https://github.com/odoo/odoo/blob/6f8b65c5a0c65d2c1a49f152423e69bfa6b3896d/odoo/tools/misc.py#L454)?

sebalix (commented):
One remark, LG

stock_picking_force_assign/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo noupdate="1">
+    <record model="ir.config_parameter" id="parameter_force_assign_allow_partial">
+        <field name="key">stock_picking_force_assign.allow_partial</field>
```
rousseldenis, line 4: @StefanRijnhart Would it be more flexible to configure it on stock.picking.type level ?

StefanRijnhart: Yes, there can be a lot of ways to implement the configuration of this behaviour: on the picking type, or on the company or depending on the picking policy of a related sale order etc. Our requirement, in a complex multicompany, multiwarehouse setup is just a change in the global behaviour, I would want to avoid a configuration by picking type because we have a lot of picking types, and there is already a constant state of confusion with regards to the existing settings on this level up to the point where the behaviour of Odoo seems unpredicatable and unreliable to its warehouse users.

stock_picking_force_assign/models/stock_picking.py
StefanRijnhart: Yes, I can, thanks! Refactored accordingly. I was a bit surprised that using `tools.str2bool` gave me the one from `tools/convert.py` instead of this one, so I selected it explicitely.

PR #881, opened 2021-08-21
Requires "Setting" access rights to access documents

Espzai:
Fix the issue to access pickings from invoices and vice versa.
This will solve this problem.

### DIFF
--- a/stock_picking_invoice_link/models/account_move.py
+++ b/stock_picking_invoice_link/models/account_move.py
@@ -35,8 +35,7 @@ class AccountMove(models.Model):
         """
         self.ensure_one()
         form_view_name = "stock.view_picking_form"
-        action = self.env.ref("stock.action_picking_tree_all")
-        result = action.read()[0]
+        result = self.env["ir.actions.actions"]._for_xml_id("stock.action_picking_tree_all")
         if len(self.picking_ids) > 1:
             result["domain"] = "[('id', 'in', %s)]" % self.picking_ids.ids
         else:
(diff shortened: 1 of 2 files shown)


rousseldenis:
@Espzai Thanks for this. Could you provide more information about which access error you have (screenshot) ?

Espzai:
> @Espzai Thanks for this. Could you provide more information about which access error you have (screenshot) ?

@rousseldenis 
```
action = self.env.ref("stock.action_picking_tree_all")
result = action.read()[0]
```
here read requires "Administration/setttings" rights. Either we do it as result = action.sudo().read()[0]
or we use the proposed method in the PR directly where Odoo is using sudo() to read the action.

Screenshot from 2021-08-23 16-46-14

Regards.

rousseldenis:
> > @Espzai Thanks for this. Could you provide more information about which access error you have (screenshot) ?
> 
> @rousseldenis
> 
> ```
> action = self.env.ref("stock.action_picking_tree_all")
> result = action.read()[0]
> ```
> 
> here read requires "Administration/setttings" rights. Either we do it as result = action.sudo().read()[0]
> or we use the proposed method in the PR directly where Odoo is using sudo() to read the action.
> 
> Screenshot from 2021-08-23 16-46-14
> 
> Regards.

Indeed, that's the good way to do it.

rousseldenis (approved):

pedrobaeza:
Please squash commits into one and pass pre-commit

rousseldenis:
@Espzai Could you squash your commits as @pedrobaeza said ?

Moreover with a more significant message

rousseldenis:
Closing this as it has been done in another PR.

PR #885, opened 2021-09-03
[13.0][FIX] stock_picking_show_return: Use Form in tests to avoid crash tests

sergio-teruel:
Crash test due to this Odoo commit
OCA/OCB@8ca10a8
The wizard lines have the field uom_id related to move.product_uom and readonly=False, if you call directly to wiz.onchage_picking_id a write in stock move is executed.

cc @Tecnativa
ping @carlosdauden @CarlosRoca13


pedrobaeza:
Can you please check Travis?

sergio-teruel:
https://github.com/OCA/stock-logistics-workflow/pull/886

PR #887, opened 2021-09-09
[13.0] Test build

simahawk:
testing #854 without changes 1st and after adding stock_dangerous_goods. Fails the same way.


sebalix:
I rebased on last 13.0 to check...

PR #889, opened 2021-09-14
[13.0][ADD] stock_landed_costs_purchase_auto: Auto-create stock landed cost when the order is confirmed.

victoralmau:
Auto-create stock landed cost when the order is confirmed.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30782


CarlosRoca13 (approved):
LGTM Code and Functional review :+1:

stock_landed_costs_auto_create/models/purchase_order.py
```diff
@@ -0,0 +1,30 @@
...
+    _inherit = "purchase.order"
+
+    landed_costs_ids = fields.One2many(
+        comodel_name="stock.landed.cost",
+        inverse_name="purchase_id",
+        string="Landed Costs",
+    )
+
+    def _create_stock_landed_cost_extra(self):
+        lc_form = Form(
```
joao-p-marques: I don’t think you should be using a `Form` here, as that is a test resource… Maybe just use the `create()` method.

joao-p-marques (commented):
Code review. Just a minor issue:

stock_landed_costs_auto_create/models/purchase_order.py
victoralmau: I have used `Form` here with the objetivo that the behavior is the same every time. A use case of this would be this: https://github.com/OCA/stock-logistics-workflow/pull/890

joao-p-marques: I understand but, as I mentioned, it is a test resource used to replicate the "real" behavior, but it is not meant for this type of situations... Check the documentation in https://github.com/odoo/odoo/blob/13.0/odoo/tests/common.py#L1461 for more information...

victoralmau: Ok, it's already changed. Thanks.

pedrobaeza (changes requested):
The landed cost should be created and linked no matter what is the content of the PO, as a placeholder for possible extra costs (which will be the products which such mark)

stock_landed_costs_auto_create/models/purchase_order.py
```diff
@@ -0,0 +1,26 @@
...
+    def _prepare_slc_values(self):
+        return {
+            "purchase_id": self.id,
+        }
+
+    def _create_picking(self):
+        res = super()._create_picking()
+        for order in self:
+            self.env["stock.landed.cost"].with_context(
+                company_id=order.company_id.id
```
rousseldenis: Why in context and not with_company() ?

victoralmau: Sorry but `.with_company()` is available in v14.

pedrobaeza: Shouldn't it be `force_company` context anyways?

stock_landed_costs_auto_create/models/stock_picking.py
```diff
@@ -0,0 +1,19 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def action_done(self):
+        res = super().action_done()
+        for item in self.filtered(
+            lambda x: x.state == "done"
+            and x.picking_type_code == "incoming"
+            and x.purchase_id
+            and x.purchase_id.landed_costs_ids
```
joao-p-marques: I think this is equivalent:
```suggestion
            and x.purchase_id.landed_costs_ids
```

joao-p-marques (commented):
Apart from what was already discussed, just a minor comment

joao-p-marques (approved):
Code and functional review :+1: 

Apart from the [latest comment](https://github.com/OCA/stock-logistics-workflow/pull/889/files#r718587173) from @pedrobaeza, looks good

victoralmau:
It's ready to merge, right?

stock_landed_costs_auto_create/models/purchase_order.py
joao-p-marques: @victoralmau you left this unattended.

victoralmau: Ok, sorry, I forgot about this change.

stock_landed_costs_auto_create/models/purchase_order.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    landed_costs_ids = fields.One2many(
```
rousseldenis: @victoralmau Shouldn't both behaviour separated in two modules ? Storing landed costs on purchase order and auto creation ? @pedrobaeza Could you update your review?

victoralmau: In my opinion, although this field is not necessary, I think it can be useful to keep it and not create another addon just for this. what do you think @pedrobaeza ?

victoralmau: ping @pedrobaeza

stock_landed_costs_auto_create/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock landed costs auto create",
+    "version": "13.0.1.0.0",
+    "category": "Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["stock_landed_costs"],
```
pedrobaeza:
```suggestion
    "depends": ["stock_landed_costs", "purchase_stock"],
```

stock_landed_costs_auto_create/models/purchase_order.py
```diff
@@ -0,0 +1,26 @@
...
+        comodel_name="stock.landed.cost",
+        inverse_name="purchase_id",
+        string="Landed Costs",
+    )
+
+    def _prepare_slc_values(self):
```
pedrobaeza: More meaningful name:

```suggestion
    def _prepare_landed_cost_values(self):
```

stock_landed_costs_auto_create/models/purchase_order.py
```diff
@@ -0,0 +1,26 @@
...
+        return {
+            "purchase_id": self.id,
+        }
+
+    def _create_picking(self):
+        res = super()._create_picking()
+        for order in self:
+            self.env["stock.landed.cost"].with_context(
+                force_company=order.company_id.id
+            ).create(order._prepare_slc_values())
```
pedrobaeza: You have to link the recently created picking already, and remove the picking code.

pedrobaeza (changes requested):
Rename the module to `stock_landed_costs_purchase_auto`, for reflecting such part. @rousseldenis no need to split, as all of this comes from the purchase part. Having only the link SLC <> purchase without the auto-creation doesn't add too much value IMO.

@victoralmau please add also an smart-button with the number of linked SLC and for accessing them.

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    landed_costs_ids = fields.One2many(
```
pedrobaeza:
```suggestion
    landed_cost_ids = fields.One2many(
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+        comodel_name="stock.landed.cost",
+        inverse_name="purchase_id",
+        string="Landed Costs",
+    )
+    landed_costs_number = fields.Integer(compute="_compute_landed_costs_number")
```
pedrobaeza:
```suggestion
    landed_cost_number = fields.Integer(compute="_compute_landed_cost_number")
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def _compute_landed_costs_number(self):
```
pedrobaeza:
```suggestion
    def _compute_landed_cost_number(self):
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+        for item in self:
+            item.landed_costs_number = len(item.landed_costs_ids)
+
+    def _prepare_landed_costs_values(self):
```
pedrobaeza: Sorry, I slipped out an `s` in between:

```suggestion
    def _prepare_landed_cost_values(self):
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+        return {
+            "purchase_id": self.id,
+        }
+
+    def button_approve(self, force=False):
```
pedrobaeza: Why changing the approach to only on approve? It was correct on the picking creation, as there can be several pickings created across the lifetime of the order (for example, adding a new line or increasing quantity when the initial picking is already done).

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,39 @@
...
+                self.env["stock.landed.cost"]
+                .with_context(force_company=order.company_id.id)
+                .create(order._prepare_landed_costs_values())
+            )
+            order.write({"landed_costs_ids": [(6, 0, [landed_costs_id.id])]})
+        return res
+
+    def action_view_stock_landed_cost(self):
+        action = self.env.ref("stock_landed_costs.action_stock_landed_cost").read()[0]
+        action["domain"] = [("purchase_id", "=", self.id)]
```
pedrobaeza: Better to put a filter than a domain, isn't it?

stock_landed_costs_purchase_auto/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def action_done(self):
+        res = super().action_done()
+        for item in self.filtered(
+            lambda x: x.state == "done"
+            and x.picking_type_code == "incoming"
+            and x.purchase_id.landed_costs_ids
+        ):
+            for lc in item.purchase_id.landed_costs_ids:
+                lc.write({"picking_ids": [(6, 0, [item.id])]})
```
pedrobaeza: This shouldn't be done here. Please link the picking on the picking creation.

pedrobaeza (changes requested):

stock_landed_costs_purchase_auto/models/stock_picking.py
victoralmau: Ok, I have used the `_prepare_landed_cost_values()` function to set the linked `picking_ids`.

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,37 @@
...
+        string="Landed Costs",
+    )
+    landed_cost_number = fields.Integer(compute="_compute_landed_cost_number")
+
+    def _compute_landed_cost_number(self):
+        for item in self:
+            item.landed_cost_number = len(item.landed_cost_ids)
+
+    def _prepare_landed_cost_values(self):
+        return {"purchase_id": self.id, "picking_ids": self.picking_ids}
```
pedrobaeza: Don't add all existing pickings, but only the one that is being created, so you need to pass such picking as argument, and you have to indicate with the proper command the ID.

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def _create_picking(self):
+        res = super()._create_picking()
+        for order in self:
+            landed_cost_ids = (
+                self.env["stock.landed.cost"]
+                .with_context(force_company=order.company_id.id)
+                .create(order._prepare_landed_cost_values())
+            )
+            order.write({"landed_cost_ids": [(6, 0, [landed_cost_ids.id])]})
```
pedrobaeza: Don't overwrite existing LCs:

```suggestion
            order.write({"landed_cost_ids": [(4, landed_cost.id)]})
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,37 @@
...
+        for item in self:
+            item.landed_cost_number = len(item.landed_cost_ids)
+
+    def _prepare_landed_cost_values(self):
+        return {"purchase_id": self.id, "picking_ids": self.picking_ids}
+
+    def _create_picking(self):
+        res = super()._create_picking()
+        for order in self:
+            landed_cost_ids = (
```
pedrobaeza: Please name variables properly according their content:
```suggestion
            landed_cost = (
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,37 @@
...
+    landed_cost_ids = fields.One2many(
+        comodel_name="stock.landed.cost",
+        inverse_name="purchase_id",
+        string="Landed Costs",
+    )
+    landed_cost_number = fields.Integer(compute="_compute_landed_cost_number")
+
+    def _compute_landed_cost_number(self):
+        for item in self:
+            item.landed_cost_number = len(item.landed_cost_ids)
```
pedrobaeza: You can do this with `read_group` for performance reason.

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,46 @@
...
+    def _compute_landed_cost_number(self):
+        domain = [("purchase_id", "in", self.ids)]
+        res = self.env["stock.landed.cost"].read_group(
+            domain=domain, fields=["purchase_id"], groupby=["purchase_id"]
+        )
+        landed_cost_dict = {x["purchase_id"][0]: x["purchase_id_count"] for x in res}
+        for item in self:
+            item.landed_cost_number = landed_cost_dict.get(item.id, 0)
+
+    def _prepare_landed_cost_values(self, params):
```
pedrobaeza: Pass directly `picking` as argument, not a generic `params` one, as it doesn't favor inheritance. Take into account that you don't control the initial call to this method, so changing the params in other part of the stack is the same as other technique (like injecting a context).

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,46 @@
...
+        return {"purchase_id": self.id, "picking_ids": params.get("picking_ids", False)}
+
+    def _create_picking(self):
+        res = super()._create_picking()
+        for order in self:
+            pickings = order.picking_ids.filtered(
```
pedrobaeza: Instead of using this technique, be practical and also avoid problems with orders with no stockable products: if no picking is created, no need to create extra landed cost, so:

- Get all the pickings linked to the purchase before calling super.
- Call super.
- Get again all the linked pickings, substract from the previous one, and if there's any, then get the first one for creating the landed cost.

And another thing: you have forgotten `_create_or_update_picking`, that should use the same technique.

pedrobaeza (changes requested):

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,56 @@
...
+            domain=domain, fields=["purchase_id"], groupby=["purchase_id"]
+        )
+        landed_cost_dict = {x["purchase_id"][0]: x["purchase_id_count"] for x in res}
+        for item in self:
+            item.landed_cost_number = landed_cost_dict.get(item.id, 0)
+
+    def _prepare_landed_cost_values(self, picking):
+        return {
+            "purchase_id": self.id,
+            "picking_ids": picking.ids if picking else False,
```
pedrobaeza: Remember that this is not valid when preparing a dictionary vals for a m2m, and at this level, `picking` **mustn't** be False. If False, no call to this method should be done:
```suggestion
            "picking_ids": [(4, picking.id)],
```

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+    def _create_picking_with_stock_landed_cost(self, all_pickings):
```
pedrobaeza: Please don't transfer as argument something that is pending of an operation of such low level. Only pass the proper picking to create the LC. It's also legit to put `self.ensure_one()` for this method, as there can't be a "shared" picking between several purchase orders.

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,64 @@
...
+        res = super()._create_picking()
+        for order in self:
+            order_pickings = order.picking_ids - all_pickings
+            if order_pickings:
+                order._create_picking_with_stock_landed_cost(
+                    fields.first(order_pickings)
+                )
+        return res
+
+    def _create_or_update_picking(self):
```
pedrobaeza, line 60: This one should be in `purchase.order.line`, not here.

stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -0,0 +1,68 @@
...
+        action = self.env.ref("stock_landed_costs.action_stock_landed_cost").read()[0]
+        action["context"] = {"search_default_purchase_id": self.id}
+        return action
+
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    def _create_or_update_picking(self):
+        all_pickings = self.order_id.mapped("picking_ids")
```
pedrobaeza:
```suggestion
        all_pickings = self.mapped("order_id.picking_ids")
```

pedrobaeza (approved):

PR #890, opened 2021-09-14
[13.0][ADD] stock_landed_costs_delivery: Auto create cost line at destination with the shipping information of the purchase order.

victoralmau:
Auto create cost line at destination with the shipping information of the purchase order.

Locked by: 
- [x] https://github.com/OCA/stock-logistics-workflow/pull/889

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30782


stock_landed_costs_delivery/README.rst
```diff
@@ -0,0 +1,96 @@
...
+=====
+
+#. Go to Inventory > Configuration > Shipping methods
+#. Create a record with LC Product.
+#. Go to Purchase > Orders > Requests for Quotation.
+#. Create an order with 2 lines: storable product + service product (landed cost).
+#. Set Custom carrier.
+#. Confirm Order.
+#. Go to Inventory > Operations > Landed Costs.
+#. Record related to the purchase order and and a cost line created with the shipping price.
```
joao-p-marques:
```suggestion
#. A record related to the purchase order and a cost line was created with the shipping price.
```

joao-p-marques (approved):
LGTM, just a minor comment about the docs :+1

stock_landed_costs_delivery/models/delivery_carrier.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    lc_product_id = fields.Many2one(
```
pedrobaeza: Is it required to use a product with such flag? As we are programmatically inserting this line, can we use the same product that the delivery carrier already has?

stock_landed_costs_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    def _prepare_slc_line_values(self):
```
pedrobaeza: Use meaningful names:

```suggestion
    def _prepare_landed_cost_line_delivery_values(self):
```

stock_landed_costs_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,23 @@
...
+            "name": self.carrier_id.name,
+            "account_id": tmpl.get_product_accounts()["stock_input"].id,
+            "price_unit": self.delivery_price,
+            "split_method": "equal",
+        }
+
+    def _prepare_slc_values(self):
+        res = super()._prepare_slc_values()
+        if self.carrier_id.lc_product_id:
+            res["cost_lines"] = [(0, 0, self._prepare_slc_line_values())]
```
pedrobaeza: Be tolerant with other possible modules adding lines:

```suggestion
            if "cost_lines" not in res:
                res["cost_line"] = []
            res["cost_lines"].append((0, 0, self._prepare_slc_line_values()))
```

pedrobaeza (changes requested):

stock_landed_costs_delivery/models/delivery_carrier.py
victoralmau: I think this was necessary to "separate" if you only wanted to create new lines for only some carriers.

pedrobaeza: Well, I don't see why are you going to put the delivery carrier in the purchase if it's not for this, but if so, better to put a flag for excluding/including them instead of another product.

victoralmau:
The dependency is already merged and rebase, ready to review it again.

stock_landed_costs_delivery/models/delivery_carrier.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    allow_create_landed_cost_lines = fields.Boolean(
```
pedrobaeza: Simplify name

```suggestion
    create_landed_cost_line = fields.Boolean(
```

stock_landed_costs_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    def _prepare_landed_cost_line_delivery_values(self):
+        tmpl = self.carrier_id.product_id.product_tmpl_id
+        return {
+            "product_id": self.carrier_id.product_id.id,
+            "name": self.carrier_id.name,
+            "account_id": tmpl.get_product_accounts()["stock_input"].id,
```
pedrobaeza: Account is not required, so better to not populate it.

stock_landed_costs_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+            "price_unit": self.delivery_price,
+            "split_method": "equal",
```
pedrobaeza: I think we should incorporate this selection to the delivery carrier, as it may depend on the carrier (some charges by weight, other by volume, etc). And the default one should be by quantity. Get the selection dynamically as in https://github.com/OCA/sale-workflow/blob/8ab5521eb2aa968d83a19ce79674043c4e8f63a8/sale_order_type/models/sale_order_type.py#L16

stock_landed_costs_delivery/models/purchase_order.py
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+    def _prepare_landed_cost_values(self, picking):
+        res = super()._prepare_landed_cost_values(picking)
+        if self.carrier_id.allow_create_landed_cost_lines:
+            if "cost_lines" not in res:
+                res["cost_lines"] = []
```
pedrobaeza: See this abbreviated form for the same:

```suggestion
            res.setdefault("cost_lines", [])
```

pedrobaeza (changes requested):

pedrobaeza (approved):

pedrobaeza:
Please check Travis

victoralmau:
The error has already been added in tests of the `stock_picking_batch_extended` addon. Ping @joao-p-marques

pedrobaeza:
Great, working.

/ocabot merge nobump

PR #893, opened 2021-09-21
[13.0][ADD] stock_picking_carrier_from_rule

TDu:
This module sets a shipping method on stock picking based on the partner address
set on the stock rule.
If the value for `carrier_id` is not yet set on the picking. And the partner as
a delivery method set.


TDu:
Moving it to the delivery-carrier repository
https://github.com/OCA/delivery-carrier/pull/417

Issue #865, opened 2021-07-12
[14.0] stock_picking_filter_lot applying on all picking types not only outgoing


remi-filament:
Although stock_picking_filter_lot states the following in README : in a picking **out** or a scrap, the domains applied seem not limited to outgoing picking but applies on any picking.

The following use case is then not working for me :
I receive for the second time product from supplier (from Vendor location) with already existing lot number. In that case, the lot_id field does not propose any SN (since there are none in Vendor location), but I also cannot create a new lot number since it already exists...

I am not sure what should be the best way-forward ?
I was thinking at first to enforce domain only for outgoing picking, but it could definitely make sense for internal (at least some of those) and even incoming when you receive back from customer, you could filter to get only delivered lots...
Maybe it should be an option on stock picking type whether to apply filter or not ?

As far as I could tell, this is the same since version 10.0 of this module

gdgellatly:
by default in odoo, incoming operations do not allow existing lots. you must explicitly allow it. your report does not state whether you have or not.

remi-filament:
Hi @gdgellatly thank you for looking into it.
Yes I have allowed both existing lots and creating new lots.

The issue is related to the domain condition added  by stock_picking_filter : ('location_ids', 'child_of', location_id)
When receiving lots already created in system, they for sure do not have a location_id set and therefore would not be shown in the list.

Since it is not found, you need to create a new lot, which is failing because lot already exists.


I am not sure how to overcome this issue, maybe we need to add an extra boolean on stock_location_type (for instance "Use available lots only") and apply domain only if this is True ?
If this makes sense for community, I could create a PR but I actually do not know technically how to apply domain only if a variable is True. Any help welcome here !

remi-filament:
Found a solution by using web_domain_field see #894

PR #894, opened 2021-09-22
[14.0][FIX] stock_picking_filter_lot configurable per operation_type

remi-filament:
I have finally found a solution for #865 

This PR proposes to :
- add boolean use_filter_lots on stock_picking_type
- apply filter on lots only for picking where picking_type has use_filter enabled

### DIFF
--- a/stock_picking_filter_lot/models/stock_move_line.py
+++ b/stock_picking_filter_lot/models/stock_move_line.py
@@ -1,0 +1,13 @@
+# Copyright 2021 Le Filament (https://le-filament.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import fields, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    picking_type_use_filter_lots = fields.Boolean(
+        related="picking_id.picking_type_id.use_filter_lots"
+    )
--- a/stock_picking_filter_lot/__manifest__.py
+++ b/stock_picking_filter_lot/__manifest__.py
@@ -6,10 +6,14 @@
     "version": "14.0.1.0.1",
     "category": "Warehouse",
     "website": "https://github.com/OCA/stock-logistics-workflow",
-    "author": "Agile Business Group, Odoo Community Association (OCA)",
+    "author": "Le Filament, Agile Business Group, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
     "depends": ["stock"],
-    "data": ["views/stock_move_line_view.xml", "views/stock_scrap_view.xml"],
+    "data": [
+        "views/stock_move_line_view.xml",
+        "views/stock_picking_type_view.xml",
+        "views/stock_scrap_view.xml",
+    ],
 }
--- a/stock_picking_filter_lot/models/__init__.py
+++ b/stock_picking_filter_lot/models/__init__.py
@@ -1,1 +1,3 @@
+from . import stock_move_line
+from . import stock_picking_type
 from . import stock_production_lot
--- a/stock_picking_filter_lot/models/stock_picking_type.py
+++ b/stock_picking_filter_lot/models/stock_picking_type.py
@@ -1,0 +1,15 @@
+# Copyright 2021 Le Filament (https://le-filament.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    use_filter_lots = fields.Boolean(
+        "Use only available lots",
+        default=True,
+        help="If this is checked, only lots available in source location \
+            would be displayed in drop down list for selecting lot.",
+    )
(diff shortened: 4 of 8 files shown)


remi-filament:
Just added updates on README I forgot in initial commit

stock_picking_filter_lot/models/stock_move_line.py
```diff
@@ -0,0 +1,38 @@
...
+                        ("location_ids", "child_of", rec.location_id.id),
+                    ]
+                )
+            else:
+                rec.lot_id_domain = json.dumps(
+                    [
+                        ("product_id", "=", rec.product_id.id),
+                        ("company_id", "=", rec.company_id.id),
+                    ]
+                )
```
bealdav: Ok I understand the topic here.
Please you may avoid to have these duplicates lines by adding conditionnally ("location_ids", "child_of", rec.location_id.id) to first 2 tuples and send to json.dumps

gdgellatly:
Hi

Are we sure this is the way to go? It seems really complicated way to solve a simple issue. And usability wise it isn't nice. Not to mention merging will break existing installations. Then even when they do get the dependencies, they need to update all existing picking types, surely the default must be true here. 

The problem only presents in a non standard configuration, mostly for receiving existing lots. Personally I feel this change needs to be its own module. But I also feel we are missing a much easier solution. I just can't think of it but it revolves around location_id.usage.

Also computed and related fields are readonly by default. There is no need to set that.

remi-filament:
@gdgellatly I have updated code with your comment to get default to true in order not to disturb existing installations and removed redundant readonly and store = False.
I tend to think also that there should be a better way to implement this but I could not get to it (but I am no Python expert so I am most probably missing something).

Fine with me if you prefer this to go in another module I can do that also.
Let me know if you can think of a better/easier way to implement it ?

gdgellatly:
I had a think about it. For me I'd prefer this picking type stuff is in another module and we do something like this to avoid the dependency. But done without the dependency, and a default of true on the picking types I wouldn't block it for in the main module.

In pseudocode

```python
class MoveLine(...)
    filter_lots = fields.Boolean(related=<picking_type_field>)

class PickingType(..)
  filter_lots = fields.Boolean(default=True)
```

We can just use the boolean value in the view domain.

Then on the views we just add
`<field name='filter_lots', invisible='1' />`

And for the field domain we just have
`[(product),(company)] + (filter_lots and [(location)] or [])`

remi-filament:
Thanks for giving this point some thoughts !
I am wondering whether it is possible to set domain as you propose directly in view ?

Otherwise I guess we still need web_domain_field dependency, no ?

Also, unless I miss something (like you can set domain with variable directly in view), I feel like what you propose is quite similar to what I did (except that you propose to make domain with and / or instead of my if), right ?

gdgellatly:
https://github.com/odoo/odoo/blob/97e3115bce518cd9d98fe2454e341077077a79a8/addons/purchase/views/account_move_views.xml#L17

Yes you can define in view for simple boolean test and changes.

The difference is this won't break any installations, either functionally or with new dependencies. It will add a field but such is life. That happens in OCA sometimes.

remi-filament:
Great I did not know that, thanks for the tip, it looks much better now.
Only one point that still could be improved : I did not manage to add tuples in domain so I had to repeat the product and company part on both sides of the or.

I tried with a + like in python code but got the following error :  unsupported operand type(s) for +: 'tuple' and 'tuple'

Any idea ?

gdgellatly:
@remi-filament you don't need to add tuples, you need to add lists of tuples. e.g. [(tuple),(tuple)] + ([(tuple)] or [])

remi-filament:
This is what I tried :

- [('product_id','=', product_id), ('company_id', '=', company_id)] + (picking_type_use_filter_lots and [('location_ids', 'child_of', location_id), ] or [])
- [('product_id','=', product_id), ('company_id', '=', company_id)] + (picking_type_use_filter_lots and [('location_ids', 'child_of', location_id)] or [])

Both give me the same error when trying to set lot number : unsupported operand type(s) for +: 'tuple' and 'tuple'

gdgellatly:
@remi-filament idk what issue is. Maybe you have a comma wrong somewhere, maybe ORM interprets it differently and converts list to a tuple before evaluating, as first one works fine. Second one I can't tell difference, but it doesn't matter if you go long way. See below for test.

```python
Python 3.8.10 (default, Jun  2 2021, 10:49:15) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> product_id = 4
>>> company_id = 2
>>> picking_type_use_filter_lots = True
>>> location_id = 5
>>> [('product_id','=', product_id), ('company_id', '=', company_id)] + (picking_type_use_filter_lots and [('location_ids', 'child_of', location_id), ] or [])
[('product_id', '=', 4), ('company_id', '=', 2), ('location_ids', 'child_of', 5)]
>>> picking_type_use_filter_lots = False
>>> [('product_id','=', product_id), ('company_id', '=', company_id)] + (picking_type_use_filter_lots and [('location_ids', 'child_of', location_id), ] or [])
[('product_id', '=', 4), ('company_id', '=', 2)]
```

remi-filament:
Thanks for looking into it, using plus on list of tuples is properly handled in Python code but not un XML view...
I guess we can go for it as is, it is already much better than my original idea thanks to your help !

jordancoest (approved):

remi-filament:
Just rebased on 14.0

rousseldenis (approved):
Code review

rousseldenis:
@remi-filament Do you plan to forwardport this ?

remi-filament:
Hi @rousseldenis I never used v15 so far, so I am not feeling comfortable to forward-port to v15

rousseldenis:
> Hi @rousseldenis I never used v15 so far, so I am not feeling comfortable to forward-port to v15

You can basically use oca-port to do it and make a PR. If it works as is, fine !

PR #896, opened 2021-09-22
[14.0][ADD] stock_picking_filter_lot_configurable

remi-filament:
Following discussions on #894 in order to solve #865 :
I have created another module for making lot filter configurable per stock picking operation type.

@gdgellatly Should I let the 2 PR open and let PSC choose whether to implement it directly in stock_picking_filter_lot or use separate module ?


remi-filament:
With the updated domains from #894 I guess this one makes no sense and would prefer to get #894 in so I close this one

PR #897, opened 2021-09-22
[13.0][FIX] stock_picking_group_by_partner_by_carrier_by_date: group moves same day different time

sebalix:
<s>The first commit add a test to reproduce the issue while the second one fix it.</s> squashed

### DIFF
--- a/stock_picking_group_by_partner_by_carrier_by_date/tests/test_grouping_by_date.py
+++ b/stock_picking_group_by_partner_by_carrier_by_date/tests/test_grouping_by_date.py
@@ -56,17 +56,17 @@ class TestGroupByDate(TestGroupByDateBase):
             p1, p2 = self._create_orders_and_pickings()
             self.assertEqual(p1, p2)
 
-        # Same day, inside time window
+        # Same day
         with freeze_time("2020-11-27 11:00:00"):
             p3, p4 = self._create_orders_and_pickings()
             self.assertEqual(p3, p4)
             self.assertEqual(p3, p1)
 
-        # Same day, out of time window
-        with freeze_time("2020-11-27 18:00:00"):
+        # Same day, one minute later
+        with freeze_time("2020-11-27 11:01:00"):
             p5, p6 = self._create_orders_and_pickings()
             self.assertEqual(p5, p6)
-            self.assertNotEqual(p5, p1)
+            self.assertEqual(p5, p1)
 
         # Delay 1st picking date
         p1.scheduled_date = "2020-11-27 18:30:00"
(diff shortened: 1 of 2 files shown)


simahawk (approved):
you can squash commits together IMO

sebalix:
@simahawk the test `# Same day, out of time window` is failing, I'm not sure to understand it, what did you mean by "Out of time window"? I see nothing regarding such time windows, not even through dependencies.

simahawk:
> @simahawk the test `# Same day, out of time window` is failing, I'm not sure to understand it, what did you mean by "Out of time window"? I see nothing regarding such time windows, not even through dependencies.

"out of the time window of the other picking". Indeed, this change is breaking the old assumption about time matching...

simahawk (changes requested):
putting on hold as we are not sure if this is always wanted

mmequignon (approved):

mmequignon:
This field is based on the `stock.move.date_expected` field (the min() or max() one, depending on the `move_type`).
The thing is that we have a cutoff time applied to those dates in the sale flow in our project.
For the resupply flow, we do not have such thing, and therefore we have to find another way to merge them.
I'm a bit lost about what you're saying. I can't find a case where it should be an issue.
Do you have examples?

TDu (commented):
Makes sense. I don't know about that failing test.

TDu (approved):

jbaudoux:
> > @simahawk the test `# Same day, out of time window` is failing, I'm not sure to understand it, what did you mean by "Out of time window"? I see nothing regarding such time windows, not even through dependencies.
> 
> "out of the time window of the other picking". Indeed, this change is breaking the old assumption about time matching...

The time window is the time when you can go to the customer. This has nothing to do with the WH planning. This test is not relevant and should be removed.

jbaudoux (approved):

sebalix:
Test updated accordingly.

simahawk (approved):

stock_picking_group_by_partner_by_carrier_by_date/tests/test_grouping_by_date.py
```diff
@@ -56,17 +56,17 @@ def test_group_by_date(self):
...
-        # Same day, inside time window
+        # Same day
         with freeze_time("2020-11-27 11:00:00"):
             p3, p4 = self._create_orders_and_pickings()
             self.assertEqual(p3, p4)
             self.assertEqual(p3, p1)
 
-        # Same day, out of time window
-        with freeze_time("2020-11-27 18:00:00"):
+        # Same day, one second later
```
mmequignon:
```suggestion
        # Same day, one minute later
```
Or maybe change the frozen time to `2020-11-27 11:00:01` instead

mmequignon (approved):
Small remark. Looks good.

stock_picking_group_by_partner_by_carrier_by_date/tests/test_grouping_by_date.py
sebalix: Fixed.

PR #900, opened 2021-09-27
[14.0][MIG] delivery_total_weight_from_packaging: Migration to 14.0

hiepnh97:


delivery_total_weight_from_packaging/__manifest__.py
```diff
@@ -1,14 +1,14 @@
 # Copyright 2020 Camptocamp SA
-# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
```
sebalix: Why this license change?

hiepnh97: Hi @sebalix, I changed the license because I think it is the same thing with this [PR](https://github.com/OCA/stock-logistics-warehouse/pull/1256).

sebalix: Ok, I checked internally and this seems OK. @simahawk are you agree with this change?

simahawk: I don't get the reference to https://github.com/OCA/stock-logistics-warehouse/pull/1256 :thinking: 
There's no license mix here AFAIS (on v13 at least).
Also, "hiding" the license change in the same commit IMO it's not that good.
If there's no special / good reason to make it AGPL I'd roll back this change

hailangvn: Thanks @sebalix, @simahawk for the comments. We will roll back the change.

hailangvn: This PR has been rebased and updated. Thanks.

sebalix: Runbot was red, I restarted it to check the logs

mmequignon (approved):

simahawk:
there's a problem w/ runbot not related to this PR. Maybe related to https://github.com/OCA/maintainer-quality-tools/pull/697

/ocabot merge nobump

PR #910, opened 2021-10-14
[13.0][FIX] stock_picking_group_by_partner_by_carrier - Properly merge the procurement groups

jbaudoux:
### Fixing the procurement group of pulled internal moves related to SO
There is an inconsistency depending on the usage of this module. The grouping configuration can only be enabled on a delivery order (SHIP). This updates the procurement group of that SHIP.

* If this module is used with the release of operation module, then this grouping module set the same procurement to all the moves. Then, at release of operations, the pulled moves that already all shares the same procurement group will be regrouped in the internal operations.
* Elif this module is used with the dynamic routing to reclassify internal moves inside the right operation, then as this grouping module set the same procurement to all the moves, the pulled moves that already all shares the same procurement group will be regrouped in the internal operations.
* Else, only the ship moves belong to the same procurement group but not the pulled internal moves as there were assigned to picking before the merge of the procurement group. Conceptually, the purpose of grouping the ship moves is because we want to make a single pick and pack for multiple SO. I have added a note in the readme about this current limitation.

Note: the fix here consist of ensuring that all the moves of all the pickings related to a procurement group have the same procurement group. This wasn't the case for added SO line after confirmation.

### Fixing the grouping of pulled internal moves related to inter-WH resupply
In case of a resupply delivery order (i.e. not linked to any SO), the grouping was effective but the procurement group was not merged properly preventing any pulled internal move to be regrouped

### Merge procurement group by name
The new regrouped procurement group will contain the name of all regrouped procurement group sorted by name

### Reduce the amount of created procurement group
Do not create a new procurement group when not necessary and a rename can be applied.
In case of backorder, the backorder will keep the same procurement group as the original picking. So from whatever sales you are, you will continue to see newly created backorder (previously if the SO was not part of that backorder you could not see it anymore from the SO). However, as soon as another SO is added in the backorder, that backorder will only be visible for the new concerned group of SO.

cc @jgrandguillaume @gurneyalex @sebalix

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
@@ -1,5 +1,5 @@
 # Copyright 2020 Camptocamp (https://www.camptocamp.com)
-# Copyright 2020 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# Copyright 2020-2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from itertools import groupby
@@ -46,13 +46,15 @@ class StockPicking(models.Model):
         return super().write(values)
 
     def action_cancel(self):
-        cancel_sale_id = self.env.context.get("cancel_sale_id")
-        if cancel_sale_id:
+        # When a SO is canceled, cancel only moves related to this SO and not
+        # all moves of the picking
+        cancel_sale_group_ids = self.env.context.get("cancel_sale_group_ids")
+        if cancel_sale_group_ids:
             moves = self.move_lines.filtered(
-                lambda m: cancel_sale_id in m.original_group_id.sale_ids.ids
+                lambda m: m.original_group_id.id in cancel_sale_group_ids
                 and m.state not in ("done", "cancel")
             )
-            moves.with_context(cancel_sale_id=False)._action_cancel()
+            moves.with_context(cancel_sale_group_ids=False)._action_cancel()
             return True
         else:
             return super().action_cancel()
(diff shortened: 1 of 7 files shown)


jbaudoux:
WIP: update of the tests

jbaudoux:
> WIP: update of the tests

done

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -46,13 +46,15 @@ def write(self, values):
...
     def action_cancel(self):
-        cancel_sale_id = self.env.context.get("cancel_sale_id")
-        if cancel_sale_id:
+        # When a SO is canceled, cancel only moves related to this SO and not
+        # all moves of the picking
+        cancel_sale_group_ids = self.env.context.get("cancel_sale_group_ids")
+        if cancel_sale_group_ids:
             moves = self.move_lines.filtered(
-                lambda m: cancel_sale_id in m.original_group_id.sale_ids.ids
+                lambda m: m.original_group_id.id in cancel_sale_group_ids
```
sebalix, line 54: Not sure to understand this line with the comment above. Aren't we canceling also moves not related to the canceled SO here?

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -65,45 +67,81 @@ def _create_backorder(self):
...
+
     def _prepare_merge_procurement_group_values(self, move_groups):
         sales = move_groups.sale_id
         return {
             "sale_ids": [(6, 0, sales.ids)],
-            "name": ", ".join(move_groups.mapped("name")),
+            "name": ", ".join(move_groups.sorted("name").mapped("name")),
         }
 
     def _merge_procurement_groups(self):
```
sebalix, line 94: A docstring is welcome

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -65,45 +67,81 @@ def _create_backorder(self):
...
+                backorder._update_merged_origin()
             backorders |= backorder
         return backorders
 
+    def _prepare_merged_origin(self):
+        """Concatenate all origin together.
+        Note that in standard, only max 5 are displayed"""
+        moves = self.move_lines.filtered(lambda m: m.state != "cancel")
+        origins = sorted(moves.filtered(lambda m: m.origin).mapped("origin"))
+        origins = list(dict.fromkeys(origins))
```
TDu: I don't think you need that line `mapped` already returns a list

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -65,45 +67,81 @@ def _create_backorder(self):
...
-            # different sale.
-            moves = picking.move_lines
-            moves_groups = moves.original_group_id
-            moves_sales = moves_groups.sale_id
-            group_sales = base_group.sale_ids
-            # if we have different sales, it means "_assign_picking" added
-            # moves from another SO in the picking
-            if moves_sales != group_sales:
-                # create a new joint group for the existing different groups
+        self.ensure_one()
```
sebalix, line 95: We are breaking something with this `self.ensure_one()`, before the `_merge_procurement_groups` was looping on `self`, which was not causing issue if there was no backorder to process. Now we get a singleton error. Maybe change the code calling this method?

sebalix (changes requested):

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
jbaudoux: I found it better to filter on procurement group. Each SO line gives the related procurement group, so I'm searching for moves with that procurement group.

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
jbaudoux: Indeed there could be no new backorder. I added a test to not call this method in this case

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
jbaudoux: It's to remove duplicates. I changed it to set and moved the sorted down on that line.

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
jbaudoux: done

sebalix (approved):

TDu (approved):

sebalix:
Working as expected on production, merging
/ocabot merge patch

PR #789, opened 2021-02-18
[13.0] [MIG] stock_picking_send_by_mail: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT27477

This PR superseds https://github.com/OCA/stock-logistics-workflow/pull/658 attending the required changes of @ValentinVinagre @sergio-teruel and @ernestotejeda 

please review it

PR #912, opened 2021-10-15
[14.0][MIG] stock_picking_send_by_mail: Migration to V14

ferran-S73:
Standard migration


rousseldenis (approved):

HaraldPanten (approved):

pedrobaeza:
Odoo since v13 allows to send an automatic email for the picking when validating it, so for me this module is not needed. The only advantage of it is the possibility of sending it manually, but I wouldn't create a specific template in that case, but used the core one.

HaraldPanten:
Well, the advantage is that user can decide which picking/s he/she wants to send by email instead of sending all of them once they're validated. Don't you think that this feature couldn't be interesting anymore?

You needed this one for V13 as well https://github.com/OCA/stock-logistics-workflow/pull/789 😝

pedrobaeza:
It's just as I have told about the manual possibility to have it, but I'm telling about reusing the same template. FYI, we have removed the module in v13 after teaching the users about the automatic flow, and they prefer that one.

ferran-S73:
> It's just as I have told about the manual possibility to have it, but I'm telling about reusing the same template. FYI, we have removed the module in v13 after teaching the users about the automatic flow, and they prefer that one.

@pedrobaeza I've added a commit removing the template and instead using core's template.

pedrobaeza:
We can't merge for now due to https://github.com/OCA/maintainer-quality-tools/issues/700

ValentinVinagre:
@pedrobaeza can it be merged ?

PR #914, opened 2021-10-18
[13.0][FIX] s_p_group_by_partner_by_carrier: fix printing of delivery slip

sebalix:
Two bugfixes:
- filter out canceled moves
- in case there is no sale order related to some moves (even if this should
  not happen), do not print an empty section line

Ref. 2967

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
@@ -131,7 +131,8 @@ class StockPicking(models.Model):
     def _get_sorted_moves(self):
         # Meant to be overriden
         self.ensure_one()
-        return self.move_lines.sorted(lambda m: m.sale_line_id.order_id.id)
+        moves = self.move_lines.filtered(lambda m: m.state != "cancel")
+        return moves.sorted(lambda m: m.sale_line_id.order_id.id)
 
     def _get_sorted_move_lines(self):
         # Meant to be overriden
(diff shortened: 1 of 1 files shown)


simahawk (approved):
seems legit

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -179,11 +180,12 @@ def get_delivery_report_lines(self):
...
                 for sale, sale_moves in grouped_moves:
-                    line_desc = sale.get_name_for_delivery_line()
-                    fake_record.update(
-                        {"description_picking": line_desc, "origin": sale.name}
-                    )
-                    sales_and_moves |= sales_and_moves.new(fake_record.copy())
+                    if sale:
+                        line_desc = sale.get_name_for_delivery_line()
+                        fake_record.update(
+                            {"description_picking": line_desc, "origin": sale.name}
```
simahawk, line 186: hmm we'll miss the keys? Is there a check or is going to fail in the report?

mmequignon (approved):

TDu (approved):

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
sebalix: This fake record dict is only used to append a section line with the name of the sale order, if there is no order we don't want to create such line.

PR #929, opened 2021-11-16
[14.0][ADD] stock_mass_scrap

jcoux:
Add wizard to mass scrap expired products:

* Using the wizard "Mass scrap" you can create a scrap for each expired quant.
* Using the server action, you can mass validate all selected scraps

### DIFF
--- a/stock_mass_scrap/wizard/mass_scrap.py
+++ b/stock_mass_scrap/wizard/mass_scrap.py
@@ -1,0 +1,90 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class MassScrap(models.TransientModel):
+    _name = "mass.scrap"
+    _description = "Mass scrap"
+
+    date_expiration_to_scrap = fields.Datetime(
+        default=lambda self: fields.Datetime.now(),
+        required=True,
+        string="Expiration date to scrap",
+        help="All the serial numbers/lots with an expiration date "
+        "inferior to this date will be scrapped",
+    )
+    scrap_location_id = fields.Many2one(
+        comodel_name="stock.location",
+        domain="[('scrap_location', '=', True)]",
+        default=lambda self: self.env["stock.location"].search(
+            [("scrap_location", "=", True)], limit=1
+        ),
+        required=True,
+    )
+    specific_warehouse_id = fields.Many2one(
+        comodel_name="stock.warehouse",
+        help="If you choose a warehouse here, only the serial numbers/lots "
+        "currently stored in this warehouse will be scrapped",
+    )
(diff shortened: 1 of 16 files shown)


omarie-ctc (approved):
After testing, fonctionnal part is ok.

AaronHForgeFlow:
LG. I'd like to have an option to exclude some of the lots that have expired, so the user can keep some of the lots.

stock_mass_scrap/wizard/mass_scrap.py
```diff
@@ -0,0 +1,91 @@
...
+        self.onchange_quants_to_scrap()
+
+        scrap_model = self.env["stock.scrap"]
+        scrap_ids = []
+        for quant in self.stock_quant_ids:
+            scrap = scrap_model.create(self._get_scrap_values(quant))
+            scrap_ids.append(scrap.id)
+
+        action = self.env.ref("stock.action_stock_scrap")
+        action_data = action.read()[0]
```
rousseldenis: Please change to _for_xml_id()

rousseldenis (changes requested):
LGTM. Just one change @jcoux

stock_mass_scrap/wizard/mass_scrap.py
jcoux: I looked the function you indicate:
https://github.com/odoo/odoo/blob/8a02942f0cee4559db4466a8b656eb4e21c662fd/odoo/addons/base/models/ir_actions.py#L127
But I don't understand how you want I use it.

My need here is to change the domain of the action to return only scraps I created.

But the function you indicate only give me the values of the action without the domain, and even if the domain was returned but the function, it will not help me to change the domain and returns the action customized with my specific domain.

Maybe I misunderstanding something?

rousseldenis: @jcoux The _for_xml_id is the core helper method to get the action. Moreover, it has sudo() in it too as some basic users don't have access to actions. So, please use it

ivantodorovich: You'd use it like this:

```py
    action = self.env["ir.actions.actions"]._for_xml_id("stock.action_stock_scrap")
    action["domain"] = [("id", "in", scrap_ids)]
    return action
```

ivantodorovich (approved):
It seems that there's only 1 comment missing.
Pre-approving

rousseldenis:
@jcoux

jcoux:
Thanks for reminder and explanation: I will fix it today! :)

jcoux:
@rousseldenis 

Fixed with this commit: https://github.com/OCA/stock-logistics-workflow/commit/57490253eede8c4a7f91824a029842a81b3f4820
Then rebased and squashed

rousseldenis (approved):

PR #936, opened 2021-12-01
[14.0] Add delivery_procurement_group_carrier. Extract feature from stock_picking_group_by_partner_by_carrier

jbaudoux:
The carrier in the procurement group that was added by stock_picking_group_by_partner_by_carrier is in fact also necessary for stock_dynamic_routing if you want to configure a routing by carrier.
This extracts the feature in a separate module so that stock_dynamic_routing can be installed without stock_picking_group_by_partner_by_carrier

cc @TDu


procurement_group_carrier/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Procurement Group Carrier",
+    "Summary": "Record the carrier on the procurement group",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "category": "Warehouse Management",
+    "depends": ["sale_stock", "delivery"],
```
rousseldenis: I would have named it sale_procurement_group_carrier or you need to put sale related stuff out of this.

rousseldenis (changes requested):
code review

procurement_group_carrier/__manifest__.py
sebalix: I agree

jbaudoux: Then sale_stock_procurement_group_carrier ?
It is more a stock feature than a sale one.
There is already a module "sale_stock_mto_as_mts_orderpoint", so i think that will be more in line with the naming convention.

jbaudoux:
module renamed from "procurement_group_carrier" to "delivery_procurement_group_carrier"

procurement_group_carrier/__manifest__.py
jbaudoux: The carrier is from the delivery module (that depends on sale_stock).
So I'm naming it delivery_procurement_group_carrier !

sebalix (approved):
LG

stock_picking_group_by_partner_by_carrier/models/procurement_group.py
```diff
@@ -4,7 +4,6 @@
 class ProcurementGroup(models.Model):
     _inherit = "procurement.group"
 
-    carrier_id = fields.Many2one("delivery.carrier", string="Delivery Method")
```
sebalix, line 7: We need a migration script for this field to move its metadata to the new addon otherwise the upgrade of this module will drop the column.

sebalix (changes requested):
I changed my mind: not good I think :)

stock_picking_group_by_partner_by_carrier/models/procurement_group.py
sebalix: @jbaudoux if you want I'll take care of this, give me access to your fork so I can push the fix directly

jbaudoux: @sebalix If you're maintainer, you should be able. Otherwise, make a PR on this and I'll include it

jbaudoux: @sebalix Is this https://github.com/OCA/stock-logistics-workflow/pull/936/commits/2d1e2590e7a8a26b329fccb40791004e3feefc33 solving the issue?

i-vyshnevska:
@jbaudoux as your repo is outside I cannot open a PR to your branch henсe I need access or you can pick commit https://github.com/camptocamp/stock-logistics-workflow/commit/1d5ced48940364801aafb98dcb4209d70f34c5aa
and we will be good to go

jbaudoux:
> @jbaudoux as your repo is outside I cannot open a PR to your branch henсe I need access or you can pick commit [camptocamp@1d5ced4](https://github.com/camptocamp/stock-logistics-workflow/commit/1d5ced48940364801aafb98dcb4209d70f34c5aa) and we will be good to go

@i-vyshnevska Why removing the new field from the view? For the migration of the data, I thought migration scripts were only executed when updating a module to a upper version, not when installing it (cc @sebalix )

i-vyshnevska:
@jbaudoux

remove from view - cos now it is presented in the new addon and duplicated in the form view
regards migration here we will have migration between v13-v14 so migration for 14.0.1.0.0 should run

stock_picking_group_by_partner_by_carrier/models/procurement_group.py
sebalix: @jbaudoux no, because if `stock_picking_group_by_partner_by_carrier` is upgraded before, Odoo will drop the database column post-upgrade so you will lose the data. You need to move your script in pre-migration (in `stock_picking_group_by_partner_by_carrier/migrations/`) even if the `delivery_procurement_group_carrier` doesn't exist in the `ir_module_module` table, that doesn't matter (based on my tests...). Then when the `delivery_procurement_group_carrier` addon will be installed Odoo won't re-create the column as it hasn't been drop (still referenced by a module in `ir_model_data`), so the data will be preserved.

sebalix: @i-vyshnevska I didn't see you move the script in `pre-migration`, but this would need to be in the `stock_picking_group_by_partner_by_carrier` addon instead

jbaudoux:
> @jbaudoux
> 
> remove from view - cos now it is presented in the new addon and duplicated in the form view regards migration here we will have migration between v13-v14 so migration for 14.0.1.0.0 should run

@i-vyshnevska Thanks, now I get it ;) I applied your changes

stock_picking_group_by_partner_by_carrier/models/procurement_group.py
i-vyshnevska: @sebalix we don't need to provide migration for stock_picking_group_by_partner_by_carrier, more of it we have errors when trying to move fields with the script as data already was created 
no migration scrips needed

TDu (approved):

stock_picking_group_by_partner_by_carrier/models/procurement_group.py
sebalix: I agree, no migration needed at time, my bad.

sebalix: @jbaudoux can you remove the migration script then? As `stock_picking_group_by...` depends on the current one, the migration path is already good, no need to do anything.

TDu: Sorry a little late to the party. But if we are talking about this one `delivery_procurement_group_carrier/migrations/14.0.1.0.0/pre-migration.py`
I think it make sense, no ?

jbaudoux: @sebalix Is it your final answer....

jbaudoux: `migrations/14.0.1.0.0/pre-migration.py` removed

sebalix: @TDu That what I thought initially, but after a test it is not needed. Indeed, this new addon is a dependency of the existing one, so Odoo will install this new addon first, which add a line in `ir_model_data` that references the `carrier_id` field, then the `stock_group_by...` addon is updated where the `carrier_id` isn't declared anymore so its reference is removed from `ir_model_data`, but as there is still a reference from `delivery_procurement_group_carrier` addon in `ir_model_data` then Odoo doesn't drop the column.

@jbaudoux thank you! Sorry for the noise...

sebalix (approved):

sebalix:
@rousseldenis are you OK with the current name?

rousseldenis (approved):

PR #939, opened 2021-12-09
[13.0][OU-FIX] stock_picking_send_by_mail: Migration script.

victoralmau:
Since migration https://github.com/OCA/stock-logistics-workflow/pull/789 some data needs to be updated to avoid `format_tz()` error when using mail template. 

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT33312


pedrobaeza:
Although this can be correct, shouldn't we switch to standard Odoo system?

pedrobaeza (approved):

CarlosRoca13 (approved):

HaraldPanten:
Shouldn't you remove "TT33312" from the commit message?

pedrobaeza:
No, that gives us the trace. Odoo does the same.

HaraldPanten (approved):

stock_picking_send_by_mail/migrations/13.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
...
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.load_data(
+        env.cr,
+        "stock_picking_send_by_mail",
+        "migrations/13.0.1.2.0/noupdate_changes.xml",
+    )
```
pedrobaeza, line 13: You should delete translations as well.

pedrobaeza (changes requested):

pedrobaeza:
I see that you have changed the source term in the translations. Are that translations completed? If so, you should change the translation as well with the correct expression.

victoralmau:
Previously, the term origin was not updated correctly, now the term origin is updated, and in the different translations the corresponding expression is added.

pedrobaeza (approved):

PR #940, opened 2021-12-12
[14.0][IMP] stock_partner_delivery_window: copy time window ids of partner

hailangvn:
When copying partner, if the partner's delivery time preference is time
window, time window ids should be copied as well.

### DIFF
--- a/stock_partner_delivery_window/models/res_partner.py
+++ b/stock_partner_delivery_window/models/res_partner.py
@@ -152,3 +152,22 @@ class ResPartner(models.Model):
                 )
             res[partner.id] = "\n".join(opening_times_description)
         return res
+
+    def copy_data(self, default=None):
+        result = super().copy_data(default=default)[0]
+        not_time_windows = self.delivery_time_preference != "time_windows"
+        not_copy_windows = not_time_windows or "delivery_time_window_ids" in result
+        if not_copy_windows:
+            return [result]
+        values = [
+            {
+                "time_window_start": window_id.time_window_start,
+                "time_window_end": window_id.time_window_end,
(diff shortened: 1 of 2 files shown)


stock_partner_delivery_window/models/res_partner.py
```diff
@@ -152,3 +152,23 @@ def short_format_time(time):
                 )
             res[partner.id] = "\n".join(opening_times_description)
         return res
+
+    def copy_data(self, default=None):
+        result = super().copy_data(default=default)
+        not_time_windows = self.delivery_time_preference != "time_windows"
+        not_copy_windows = not_time_windows or "delivery_time_window_ids" in default
```
petrus-v: We should probably manage the case where `not_time_windows` is `False` and `default` is `None` ? to avoid `TypeError: argument of type 'NoneType' is not iterable` error !

Not sure if it's makes sense could you add unit test to cover such case ?

petrus-v (changes requested):

stock_partner_delivery_window/models/res_partner.py
hailangvn: Thanks @petrus-v for your input. The test has been added to cover the case and the code is simplified. As expected result of `super().copy_data(default=default)` is of type `dict` (without any overriding: https://github.com/odoo/odoo/blob/14.0/odoo/models.py#L4540), the `result` is now used instead of `default`.

petrus-v: Thanks :+1:

petrus-v (approved):

rousseldenis (approved):

PR #942, opened 2021-12-13
[14.0][FIX] stock_putaway_by_route is not an application

andreampiovesana:
fix manifest because is not an application


legalsylvain:
I'm curious to know why it is a problem.

andreampiovesana:
it is a small fix, the reason is evident.

legalsylvain:
>  the reason is evident.

thank you for this very complete answer. ;-)

rousseldenis:
> > the reason is evident.
> 
> thank you for this very complete answer. ;-)

Arf. I think that the answer is the module is an 'accessory' module for a bigger application (e.g.: stock).

rousseldenis (approved):

PR #948, opened 2022-01-06
[14.0][ADD] stock_valuation_fifo_lot

newtratip:
This module is calculate fifo cost by lot

- [ ] Readme
- [ ] Unit Test

### DIFF
--- a/stock_valuation_fifo_lot/models/product.py
+++ b/stock_valuation_fifo_lot/models/product.py
@@ -1,0 +1,114 @@
+# Copyright 2022 Ecosoft Co., Ltd (https://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+from odoo.tools import float_is_zero
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _sort_by_all_candidates(self, all_candidates, sort_by):
+        """Hook function for other sort by"""
+        return all_candidates
+
+    def _get_all_candidates(self, company, sort_by=None):
+        all_candidates = (
+            self.env["stock.valuation.layer"]
+            .sudo()
+            .search(
+                [
+                    ("product_id", "=", self.id),
+                    ("remaining_qty", ">", 0),
+                    ("company_id", "=", company.id),
+                ]
+            )
+        )
+        if sort_by == "lot_create_date":
+            all_candidates = all_candidates.sorted(
+                lambda l: min(l.lot_ids.mapped("create_date"))
+            )
+        elif sort_by is not None:
+            all_candidates = self._sort_by_all_candidates(all_candidates, sort_by)
+        return all_candidates
+
+    def _run_fifo(self, quantity, company):
+        self.ensure_one()
+        move_id = self._context.get("used_in_move_id")
+        if self.tracking == "none" or not move_id:
+            vals = super()._run_fifo(quantity, company)
+        else:
+            move = self.env["stock.move"].browse(move_id)
+            move_lines = move._get_out_move_lines()
+            tmp_value = 0
+            tmp_remaining_qty = 0
+            for move_line in move_lines:
+                # Find back incoming stock valuation layers
+                # (called candidates here) to value `quantity`.
+                qty_to_take_on_candidates = move_line.product_uom_id._compute_quantity(
+                    move_line.qty_done, move.product_id.uom_id
+                )
+                candidates = self._get_all_candidates(company).filtered(
+                    lambda l: move_line.lot_id in l.lot_ids
+                )
+                for candidate in candidates:
+                    qty_taken_on_candidate = min(
+                        qty_to_take_on_candidates, candidate.remaining_qty
+                    )
+
+                    candidate_unit_cost = (
+                        candidate.remaining_value / candidate.remaining_qty
+                    )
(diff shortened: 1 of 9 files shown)


stock_valuation_fifo_lot/models/product.py
```diff
@@ -0,0 +1,112 @@
...
+                        precision_rounding=self.uom_id.rounding,
+                    ):
+                        break
+
+                if qty_to_take_on_candidates > 0:
+                    tmp_value += abs(candidate.unit_cost * -qty_to_take_on_candidates)
+                    tmp_remaining_qty += qty_to_take_on_candidates
+
+            # Calculate standard price (Sorted by lot created date)
+            all_candidates = self._get_all_candidates(company)
```
hailangvn:
```suggestion
            all_candidates = self._get_all_candidates(
                company,
                sort_by='lot_create_date',
            )
```

stock_valuation_fifo_lot/models/product.py
```diff
@@ -0,0 +1,112 @@
...
+                candidates = (
+                    self.env["stock.valuation.layer"]
+                    .sudo()
+                    .search(
+                        [
+                            ("product_id", "=", self.id),
+                            ("remaining_qty", ">", 0),
+                            ("company_id", "=", company.id),
+                        ]
+                    )
```
hailangvn: ... so that the method can be reused here.
```suggestion
                    self._get_all_candidates(company)
```

stock_valuation_fifo_lot/models/product.py
```diff
@@ -0,0 +1,112 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+from odoo.tools import float_is_zero
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _get_all_candidates(self, company):
```
hailangvn: IMHO, to increase reusability, we can refactor the `sorted` operation...
```suggestion
    def _get_all_candidates(self, company, sort_by=None):
```

stock_valuation_fifo_lot/models/product.py
```diff
@@ -0,0 +1,112 @@
...
+            .sudo()
+            .search(
+                [
+                    ("product_id", "=", self.id),
+                    ("remaining_qty", ">", 0),
+                    ("company_id", "=", company.id),
+                ]
+            )
+            .sorted(lambda l: l.lot_ids.create_date)
+        )
```
hailangvn: (BTW, we can resolve singleton issue here by combining `min` and `mapped`.)
```suggestion
        )
        if sort_by == 'lot_create_date':
            sort_by = lambda l: min(l.lot_ids.mapped('create_date'))
        if sort_by is not None:
            all_candidates = all_candidates.sorted(sort_by)
```

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,102 @@
...
+                )
+            layer = super(StockMove, move)._create_in_svl(
+                forced_quantity=forced_quantity
+            )
+            # Calculate standard price (Sorted by lot created date)
+            if (
+                move.product_id.cost_method == "fifo"
+                and move.product_id.tracking != "none"
+            ):
+                all_candidates = move.product_id._get_all_candidates(move.company_id)
```
hailangvn:
```suggestion
                all_candidates = move.product_id._get_all_candidates(
                    move.company_id,
                    sort_by='lot_create_date',
                )
```

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,102 @@
...
+        2. Change product standard price to first available lot price
+        """
+        layers = self.env["stock.valuation.layer"]
+        for move in self:
+            if len(move.lot_ids) > 1:
+                raise UserError(
+                    _(
+                        "Multiple lot on the stock move is not allowed for incoming transfer"
+                    )
+                )
```
hailangvn: The scenario to have more than 1 lot per move is common. So we should reconsider this check.
```suggestion
```

hailangvn (commented):
Thanks for the work. There are observation as in comment.

rousseldenis:
@newtratip Could you attend comments ?

Nantikan23 (approved):
functional tested  - Test with the case of returned products when the first lot returns, the lot will go back to the same sequence, and when delivering this lot again, will pick up that first lot to record the cost of goods sold.

Saran440:
@hailangvn Thank you for your review. it's very helpful for us.
I change function `_get_all_candidates` can sort_by and hook it for other sort by

@Nantikan23 Can you review it again, please

stock_valuation_fifo_lot/models/stock_move.py
Saran440: if not check, it affect with accounting or cost about it don't know how to accounting which lot

hailangvn: IIUC, the main purpose of the method is to find first lot that has transferred into current company and use its price as cost. So we can assume that all lots of the same delivery has same price for same product. Then IMHO current logic (`min` + `mapped`) is enough to deal with multiple lot on the stock move.

hailangvn: Thanks for the work anyway. :)

PR #952, opened 2022-01-12
[FIX] Permit autofill when avoid_lot_assignment is TRUE on picking_type

tgaullier:
Problem: auto_fill doesn't work when products are tracked (lot) and _avoid_lot_assignment_ is TRUE on picking_type.
If I understand well, I propose this fix


rousseldenis:
@tgaullier I think this was done with that purpose!

If you want to authorize auto lot assignment, uncheck the box...

PR #928, opened 2021-11-15
[14.0][FIX] stock_split_picking

xavier-bouquiaux:
previously, records without any done quantities were sent to the new picking

PR #961, opened 2022-01-31
[FIX] delete reserved move lines with qty done = 0 from source picking

Guenzn:
When splitting a move with many reserved quantity has many move lines (e.g. different locations) by mode qty_done the move lines with qty done = 0 will not be removed from source picking, which will lead to an overshippment. 

How to reproduce:
1. Configure warehouse having different locations
2. Create a product, update stock quantity at different locations loc1 -> quantity =1, loc2 -> quantity =1, loc3 -> quantity =1
3. Create an sales order add a line with quantity 3 of the product and confirm
4. Go to picking, you will see 3 move lines with the ordered product, one per location
5. Edit picking and set quantity done = 1 at first move line
image
6. Click on split and choose mode "Done quantites" and click split
image
7. Check source picking: the second and third line with quantity done zero will be kept and reservation will not be cancelled 
image
image
8. New picking with initial demand of 2 will be created, which will lead to an overshipment.
image

### DIFF
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -51,6 +51,8 @@ class StockPicking(models.Model):
                                     {'product_uom_qty': move_line.qty_done})
                             except UserError:
                                 pass
+                        elif not move_line.qty_done:
+                            move_line.unlink()
                     new_moves |= self.env['stock.move'].browse(new_move_id)
 
             # If we have new moves to move, create the backorder picking
--- a/stock_split_picking/tests/test_stock_split_picking.py
+++ b/stock_split_picking/tests/test_stock_split_picking.py
@@ -1,5 +1,6 @@
 # Copyright 2017 Tecnativa - Vicent Cubells <vicent.cubells@tecnativa.com>
 # Copyright 2018 Camptocamp SA - Julien Coux
+# Copyright 2021 sewisoft - Guenter Selbert <guenter.selbert@sewisoft.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo.tests.common import SavepointCase
@@ -13,9 +14,20 @@ class TestStockSplitPicking(SavepointCase):
         super(TestStockSplitPicking, cls).setUpClass()
 
         cls.src_location = cls.env.ref('stock.stock_location_stock')
+        cls.src_location_1 = cls.src_location.copy(
+            default={
+                'location_id': cls.src_location.id,
+                'name': cls.src_location.name + ' 1',
+            })
+        cls.src_location_2 = cls.src_location.copy(
+            default={
+                'location_id': cls.src_location.id,
+                'name': cls.src_location.name + ' 2',
+            })
         cls.dest_location = cls.env.ref('stock.stock_location_customers')
         cls.product = cls.env['product.product'].create({
             'name': 'Test product',
+            'type': 'product',
         })
         cls.partner = cls.env['res.partner'].create({
             'name': 'Test partner',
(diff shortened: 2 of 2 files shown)


Guenzn:
@nbessi @guewen @yvaucher

rousseldenis:
@Guenzn Thanks for your contribution.

Could you add a test that highlight your change ? Thanks

Guenzn:
@rousseldenis I don't think that #928 will fix my issue, because in my scenario, a new move will be created but the move lines with quantity done at the source move wan't be removed. See my video for a demo of the problem. I will see if I find some time to write a test for my case...

https://user-images.githubusercontent.com/24382867/152357756-c7adb288-bd80-49fe-8da0-f078f0d15f0a.mp4

Guenzn:
@rousseldenis I extended the testcase, that the requested demand will be reserved at two different locations.

rousseldenis:
> @rousseldenis I don't think that #928 will fix my issue, because in my scenario, a new move will be created but the move lines with quantity done at the source move wan't be removed. See my video for a demo of the problem. I will see if I find some time to write a test for my case...
> 
>  Split.Picking.Bug-2022-02-03_14.52.43.mp4

I think so. If you read the PR description, the explanation is the same.

Could you maybe try #928 one as it does not delete anything (which is better in terms of performances) ?

Guenzn:
> > @rousseldenis I don't think that #928 will fix my issue, because in my scenario, a new move will be created but the move lines with quantity done at the source move wan't be removed. See my video for a demo of the problem. I will see if I find some time to write a test for my case...
> > Split.Picking.Bug-2022-02-03_14.52.43.mp4
> 
> I think so. If you read the PR description, the explanation is the same.
> 
> Could you maybe try #928 one as it does not delete anything (which is better in terms of performances) ?

Sorry but i have to say that the test in #928 does not cover my scenario. In my scenario i create a picking with 10 units of **one** product, which is located on 2 different stock locations with a quantity of 5 at each location. After confirming the picking, the picking will have **1 single move with quantity reserved 10** having **2 move lines**. The first move line reserves 5 units at location 1 and the second move line reserves 5 units at location 2 but **of the same product**! And if I change the quantity done of the first **move line** (not move) to 4 and do a split on done quantity, the quantity reserved from the first move line will be reduced to 4 but the second *move line* where quantity done was 0 at split, won’t be removed from the source picking but was created at the backorder! So it will exists twice! Please check out my video again and keep in mind my explanation above. Maybe you can try my scenario with your code...

BTW: The move split function in v11 returns normally the id of a new created move. This was change as i see in the source code of version 14, where it will return the vals to create the move. Maybe the problem doesn't exists in v14 (I think that it will, but i can't test it today there is no runable version on runbot), but the main change #928 is that if move split does not return anything, you add the current move to new moves. When i modify my code this way, my problem persists anyway, because move split will return the new move. But it is required to modify the existing move respectively it's move lines.

Guenzn:
> > @rousseldenis I don't think that #928 will fix my issue, because in my scenario, a new move will be created but the move lines with quantity done at the source move wan't be removed. See my video for a demo of the problem. I will see if I find some time to write a test for my case...
> > Split.Picking.Bug-2022-02-03_14.52.43.mp4
> 
> I think so. If you read the PR description, the explanation is the same.
> 
> Could you maybe try #928 one as it does not delete anything (which is better in terms of performances) ?

@rousseldenis any news on that? I have tested my scenario on #928 but the failure exists on that PR too.

OSevangelist:
Hi @rousseldenis i have spoken with @Guenzn earlier today on Odoo Experience on this matter. Would it harm to merge this as it is all green and does apply to a rather old version of Odoo anyways. If you need further use cases added later, there would still be the possibility to further extend it, right? If nobody objects by tomorrow i'd prefer to merge that if you don't mind.

PR #962, opened 2022-02-03
[13.0][ADD] stock_scrap_cancel: New addon

victoralmau:
This module adds the cancel status to the scraps and allows the canceled ones to be converted into a draft.

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa TT34293


rousseldenis (approved):

chienandalu (commented):
I detected an issue on stock_lot_scrap: quick fix here: https://github.com/OCA/stock-logistics-workflow/pull/963 Anyway, nothing to do with this.

I've been testing and I find this behavior:

- I go to a reception picking and validate it.
- I get the pickings quantities in my stock.
- Then I go back to the picking and I scrap some units.
- My stock diminishes in the same amount of units as they're moved to the scrap location.
- Then I cancel the scrap and that move is erased.
- But my stock doesn't recover those units!

Shouldn't the stock be reverted from scrap to my inventory?

victoralmau:
> I detected an issue on stock_lot_scrap: quick fix here: #963 Anyway, nothing to do with this.
> 
> I've been testing and I find this behavior:
> 
>     * I go to a reception picking and validate it.
> 
>     * I get the pickings quantities in my stock.
> 
>     * Then I go back to the picking and I scrap some units.
> 
>     * My stock diminishes in the same amount of units as they're moved to the scrap location.
> 
>     * Then I cancel the scrap and that move is erased.
> 
>     * But my stock doesn't recover those units!
> 
> 
> Shouldn't the stock be reverted from scrap to my inventory?

Thanks for the review.
You are right, the stock should be increased again when canceling. Can you check again that this last change is correct?

chienandalu (approved):
Great works like charm now :slightly_smiling_face: :+1:

pedrobaeza (changes requested):
We have to check why a manual move is needed instead of only amending quantity. cc @carlosdauden

victoralmau:
> 

> We have to check why a manual move is needed instead of only amending quantity. cc @carlosdauden

I've applied a change to simplify everything which was probably what you had in mind.

stock_scrap_cancel/models/stock_scrap.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = "stock.scrap"
+
+    state = fields.Selection(selection_add=([("cancel", "Cancelled")]))
+
+    def action_draft(self):
+        return self.write({"state": "draft"})
+
+    def action_cancel(self):
+        old_scrap_qty = self.scrap_qty
+        self.move_id.move_line_ids.write({"qty_done": 0.0})
```
pedrobaeza, line 19: You should unlink the move after doing this.

pedrobaeza (approved):

PR #968, opened 2022-02-11
[14.0][ADD] stock_picking_invoice_link: add context key to get_stock_moves_link_invoice in order to allow to use it from external script

tafaRU:
This is useful when the module is installed in a database already containing invoices not in state cancel. In this way the method `get_stock_moves_link_invoice` can be also used from an external script in order to fill missing data useful i.e from https://github.com/OCA/account-invoice-reporting/tree/14.0/account_invoice_production_lot to show on their invoice lines the delivered production lots.

### DIFF
--- a/stock_picking_invoice_link/models/sale_order.py
+++ b/stock_picking_invoice_link/models/sale_order.py
@@ -10,11 +10,15 @@ class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
     def get_stock_moves_link_invoice(self):
+        skip_check_invoice_state = self.env.context.get(
+            "skip_check_invoice_state", False
+        )
         return self.mapped("move_ids").filtered(
             lambda mv: (
                 mv.state == "done"
                 and not (
-                    any(
+                    not skip_check_invoice_state
+                    and any(
                         inv.state != "cancel"
                         for inv in mv.invoice_line_ids.mapped("move_id")
                     )


tafaRU:
@pedrobaeza could you ask your opinion on this?
Thank you in advance.

stock_picking_invoice_link/models/sale_order.py
```diff
@@ -14,7 +14,8 @@ def get_stock_moves_link_invoice(self):
             lambda mv: (
                 mv.state == "done"
                 and not (
-                    any(
+                    self.env.context.get("check_invoice_state", True)
```
pedrobaeza: This context check should be non required, but the contrary to abort on case of finding something.

pedrobaeza (commented):
The main problem filling previous information is that is not going to be correct, as you only have the final result, not the intermediate one, so not able to deduce when multiple pickings and multiple invoices, which one is the source of the other.

tafaRU:
> The main problem filling previous information is that is not going to be correct, as you only have the final result, not the intermediate one, so not able to deduce when multiple pickings and multiple invoices, which one is the source of the other.

@pedrobaeza first of all thanks for your feedback!
In case the filling was restricted to only `serial`, do you think could be correct?

pedrobaeza:
Is there any trace for knowing which lots/serials are in the invoice for doing such constraint with warranty?

SimoneVagile (approved):

stock_picking_invoice_link/models/sale_order.py
```diff
@@ -14,7 +14,8 @@ def get_stock_moves_link_invoice(self):
             lambda mv: (
                 mv.state == "done"
                 and not (
-                    any(
+                    not self.env.context.get("skip_check_invoice_state", False)
```
rousseldenis: @tafaRU I'm wondering if this cannot be included in function signature as it's better to understand ?

tafaRU: Thanks for the suggestion but for my purpose (to use it in the script below) I'd prefer to keep it that way.

```
for inv_line in invoices.invoice_line_ids:
    product = inv_line.product_id
    if product.tracking == 'serial':
        for so_line in inv_line.sale_line_ids.with_context(skip_check_invoice_state=True):
            line_vals = so_line._prepare_invoice_line()
            move_line_ids = (
                    "move_line_ids" in line_vals and line_vals["move_line_ids"] or []
            )
            inv_line.write({"move_line_ids": move_line_ids})
```

tafaRU: @rousseldenis did you see my answer?

rousseldenis: Ok see

rousseldenis: @tafaRU Can you put context value in a variable externally to lambda in order to avoid calling it each time ?

tafaRU: @rousseldenis done

rousseldenis (approved):

tafaRU:
@rousseldenis thanks for the review, could it be merged?

tafaRU:
@rousseldenis could we proceed with the merge? Thank you in advance.

rousseldenis:
As comments have been attended

/ocabot merge minor

PR #969, opened 2022-02-15
[15.0] mig stock move line auto fill

robin5210:


rousseldenis:
@robin5210 Thanks for this. Could you solve commit user problems with your account? That will maybe solve CLA problem too.

rousseldenis:
@robin5210 What's the statys of this ?

PR #971, opened 2022-02-15
[10.0] stock_no_negative: backport allow_negative_stock on stock.location

alexis-via:
Rename variables in test: a recordset should not be designated by a variable that ends with "_id"!


rousseldenis (approved):

rousseldenis:
I think nobody will come here :sweat_smile:  , so merging.

/ocabot merge major

PR #975, opened 2022-03-04
[14.0] stock_picking_group_by_partner_by_carrier: deliveryslip report - fix SO sections

mmequignon:
https://github.com/odoo/odoo/blob/ef0320b82e679f2f6eb4b2ba8aa4530cda9a98a9/addons/stock/report/report_deliveryslip.xml#L77-L124

There's 4 different cases for the deliveryslip report:
 1) package and serial number
 2) package and no serial number
 3) no package but serial number
 4) no package and no serial number

For the 2nd an 4th case, odoo tries to aggregate the lines and therefore
shouldn't be groupped by SO, no fix needed here.

However, when there's serial numbers, we can group lines by SO, and we
should use `picking.get_delivery_report_lines()` to get the lines to
display. The 3rd case wasn't handled yet.

Also, when groupping by SO, the `get_delivery_report_lines` creates fake
`stock.move.line` for each SO section, with product_id = 1.
However, odoo tries to display the product, which is wrong for the 
SO sections. In such case, display the SO name instead.

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.xml
+++ b/stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.xml
@@ -52,6 +52,12 @@
         >
             <attribute name="t-foreach">o.get_delivery_report_lines()</attribute>
         </xpath>
+        <xpath
+            expr="//tbody/t[@t-else='']//tr[@t-as='move_line']"
+            position="attributes"
+        >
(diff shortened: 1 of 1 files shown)


sebalix (approved):

simahawk:
`s_p_group_by_partner_by_carrier: fix report` is till fine as a commit. You can provide details in the extended description.

simahawk (approved):
LG

PR #980, opened 2022-03-12
[14.0][IMP] stock_picking_warn_message: show blocking messages in red

dhx:
This allows the distinction between warn and block to be visible in the
 stock picking dialog by adding the additional picking_warn field to the
 stock.picking model containing what kind of message is to be shown.

### DIFF
--- a/stock_picking_warn_message/models/stock_picking.py
+++ b/stock_picking_warn_message/models/stock_picking.py
@@ -8,7 +8,27 @@ class StockPicking(models.Model):
 
     _inherit = "stock.picking"
 
+    picking_warn = fields.Text(compute="_compute_picking_warn")
     picking_warn_msg = fields.Text(compute="_compute_picking_warn_msg")
+
+    @api.depends(
+        "state",
+        "partner_id.picking_warn",
+        "partner_id.commercial_partner_id.picking_warn",
+    )
+    def _compute_picking_warn(self):
+        for rec in self:
+            picking_warn = "no-message"
+            if rec.state not in ["done", "cancel"] and rec.partner_id:
+                p = rec.partner_id.commercial_partner_id
+                if p.picking_warn == "block" or rec.partner_id.picking_warn == "block":
+                    picking_warn = "block"
+                elif (
+                    p.picking_warn == "warning"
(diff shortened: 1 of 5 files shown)


stock_picking_warn_message/models/stock_picking.py
```diff
@@ -8,8 +8,28 @@ class StockPicking(models.Model):
 
     _inherit = "stock.picking"
 
+    picking_warn = fields.Text(compute="_compute_picking_warn")
```
rousseldenis, line 11: It seems you change more than the color.

You should explain why you changed that standard behaviour.

dhx: Hi @rousseldenis, thanks for your feedback, to be able to distinguish between a warning and a blocking message (to be able to show it with a different color) this (new) computed field is used. As some logic is involved in the calculation, I guess it wouldn't make sense to put all that into the view.
Apart from the color the behaviour shouldn't change here as far as I can see (though in a later improvement it might make sense to also block the completion of the picking when there is a blocking message, see also: https://github.com/odoo/odoo/issues/51081 ).

rousseldenis:
@HviorForgeFlow Do you have an advice on this ?

rousseldenis (approved):
Seems to make sense

jonasbuchholz (approved):

dhx:
@rousseldenis thanks for the rebase, cla is signed in the meantime btw.

HviorForgeFlow (approved):
> @HviorForgeFlow Do you have an advice on this ?

After reviewing functionally this PR I see that when partner has a blocking message it is not possible to select that partner on the transfer. By the other hand if picking is generated once a SO is validated, it displays a blocking message but picking can be validated.

It is a good improvement, but also it would be nice to have a way that allow blocking the transfer for regular inventory users to be reviewed by a inventory administrator.

LGTM

rousseldenis:
> > @HviorForgeFlow Do you have an advice on this ?
> 
> By the other hand if picking is generated once a SO is validated, it displays a blocking message but picking can be validated.

I think it's out of scope as this is only to 'warn'.

PR #983, opened 2022-03-29
[14.0][MIG] stock_cost_method_last

alexis-via:


flotho (approved):
very usefull module

gurneyalex (changes requested):
The stock valuation is broken. 

I just tested this in the following way on runboat:

1. create a stockable product MyProduct, set the costing method to "last price"
2. purchase 3 units of MyProduct at $30 a piece, receive the products. Check stock valuation: OK ($90)
3. purchase 2 units of MyProduct at $40 a piece, receive the products. 
 
After this step, I expect the stock valuation to be $200 = 5 x $40, but I get $170 = 3 x $30 - $90 + $90 + 2 x $40 (from the stock valuation layers). 

If I make a similar test with a product configured with cost method = standard price, and I update  the standard price to the purchase price manually  *before* making the reception, I get the expected inventory valuation of $200.

Bilbonet (approved):
Hello @gurneyalex 

IMO this module works fine. 
This new method of cost is different of the method "Standar Price".
The method "Standar Price" update de stock value in our warehose if you change de "standar_pice" in the product form. This new method "Last Price" don't. Is a different point of view in the warehose stock valuation.

Maybe is convenient to clarify this in a better description of the module.

Thanks for the job @alexis-via

rousseldenis:
@alexis-via @gurneyalex Could you clarify ?

Bilbonet:
Hi, as I said, IMO this module is useful.
The concept that introduce for  the cost is interesting,  at least for a pair of my customer.
It would be interesting to have this module in OCA.
Thanks

gurneyalex:
@rousseldenis what is not clear in my comment? Either I totally misunderstand the purpose of the module (in which case, maybe an explanation in the README about its limitations would clarify things). I'm not as deeply in accounting as @alexis-via so maybe there are things that I'm missing. The customer to which I proposed this module was not happy with the resulting stock valuation he saw in the case I explain.

I'll get feedback from Camptocamp accounting folks (and revise my review in case they say it's fine).

gurneyalex:
@Bilbonet I don't know about your localisation. Can you check that the impact on the stock valuation is correct for you? as I said I have some serious issue with the stock valuation (both instant and at date) for product using the implementation as it is now, but this is something that could be fixed.

fclementic2c:
I agree on the business requirement. This module make sense when you produce 'à l'affaire' or in a context of high volatility....

However, technically,  if the stock_valuation_layer is not updated then the inventory valuation report will be wrong and, even further, if you activate the permanent accounting you will generate wrong journal entries...
@alexis-via Don't you agree on this ?

Thanks

Bilbonet:
According with the scenario:
1. create a stockable product MyProduct, set the costing method to "last price"
2. purchase 3 units of MyProduct at $30 a piece, receive the products. Check stock valuation: OK ($90)
3. purchase 2 units of MyProduct at $40 a piece, receive the products.
4. Stock valuation = $170

The clients are restaurants:
The previous day the clients ask for fish and other food.
The next morning buys fish and other food in kg in **diferente prices and wights in the same day**.

The stock is minimal and stays in warehouse 1 day or 2.

In this scenario the odoo user prefers the stock costing method "last price"

alexeirivera87:
Hello,

Any update on this PR ?

regards

andreampiovesana:
merge?

rousseldenis:
@alexis-via Could you answer @fclementic2c question ?

SoniaViciana:
This is LIFO (Last In, First Out), isn't it? We were interested in this addon but we have found that implicitly, this cost method is generally not recommended by international accounting standards. The reference to law in Europe:

https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32008R1126 -> Commission Regulation (EC) No 1126/2008 of 3 November 2008

image

image

In any case, I understand that in certain companies it may be justifiable to use this cost method. I just want to warn other interested colleagues that it shouldn't be used in a regular company without prior justification.

Thanks,

rousseldenis:
> This is LIFO (Last In, First Out), isn't it? We were interested in this addon but we have found that implicitly, this cost method is generally not recommended by international accounting standards. The reference to law in Europe:
> 
> https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32008R1126 -> Commission Regulation (EC) No 1126/2008 of 3 November 2008
> 
> image
> 
> image
> 
> In any case, I understand that in certain companies it may be justifiable to use this cost method. I just want to warn other interested colleagues that it shouldn't be used in a regular company without prior justification.
> 
> Thanks,

As 

> This is LIFO (Last In, First Out), isn't it? We were interested in this addon but we have found that implicitly, this cost method is generally not recommended by international accounting standards. The reference to law in Europe:
> 
> https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32008R1126 -> Commission Regulation (EC) No 1126/2008 of 3 November 2008
> 
> image
> 
> image
> 
> In any case, I understand that in certain companies it may be justifiable to use this cost method. I just want to warn other interested colleagues that it shouldn't be used in a regular company without prior justification.
> 
> Thanks,

@SoniaViciana In @fclementic2c comment, this can be justified for "by project" cost. And that complies with regulation (see §23)

andreampiovesana:
why not use last cost?

rousseldenis (approved):
Code review. @alexis-via Tests are welcome.

rousseldenis:
@gurneyalex Could you update your review ?

PR #984, opened 2022-03-29
Stock scrap cancel backport to 12.0

estevebadia:
Backport of the module **stock_scrap_cancel** from version 13.0 to version 12.0.

The exact same code seems to work for v12, so I just copyed the code in 13.0 branch (and changed some 13's for 12's).

I also took the opportunity to add another .po translation file for the catalan language.


rousseldenis:
@estevebadia Thanks for this. Could you preserve commits history ?

PR #988, opened 2022-04-11
[14.0] [ADD] stock_picking_estimate_value

paradoxxxzero:
This module adds amount of demand in term of product price to pickings.

It also adds reserved amount and reserved completion to outgoing pickings.


paradoxxxzero:
Also adds `stock_picking_estimate_value_order_prices` to use the average product price from the sale or purchase order

rousseldenis:
@paradoxxxzero Thanks for this. Could be great to add tests.

hparfr (approved):

PR #989, opened 2022-04-20
[ADD] new_module: additional functionality of stock_picking_invoice_link

matiasperalta1:
…ink module


ajaniszewska-dev:
@matiasperalta1  please fix pre-commit

TDu:
If this is a forward port from this https://github.com/OCA/stock-logistics-workflow/pull/816 ?  It is better to migrate it and keep the original commits.
Looking at the licences in the files this looks like a port from another version !

rousseldenis:
> If this is a forward port from this #816 ? It is better to migrate it and keep the original commits. Looking at the licences in the files this looks like a port from another version !

I think as the original code is in base module, keeping history is quite weird. But maybe, get commit that include new code and then a commit to move the code to new module.

rousseldenis:
> > If this is a forward port from this #816 ? It is better to migrate it and keep the original commits. Looking at the licences in the files this looks like a port from another version !
> 
> I think as the original code is in base module, keeping history is quite weird. But maybe, get commit that include new code and then a commit to move the code to new module.

At least adding @grindtildeath as contributor of this

grindtildeath (commented):
Thanks for the migration. However there are a few things to rework here:
- IMO, it's better to keep the commit from my PR in 12.0 and then have a commit to split the functionality into a dedicated module and remove the rest from original module.
- Then yes, please add me to the contributors since I did the original feature. :pray: 
- Please keep a unit test (need to be adapted from original PR)

Then, the breaking change with the stock_picking_invoice_link module in my original PR was the redefinition of field `account.move.picking_ids` to a computed field which was already done in v13.0, so I'm not sure why we'd still need an extra module for this feature now... But that's a question for @pedrobaeza

pedrobaeza:
If the breaking change was included in v13, indeed this module may not be applicable as something isolated.

PR #996, opened 2022-05-04
[ADD][12.0]  module stock_picking_filter_lot_qty

sergiocorato:
This module let user select only lot with available quantity in stock picking.


yostashiro:
@sergiocorato Have you seen this? https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_picking_filter_lot

sergiocorato:
> @sergiocorato Have you seen this? https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_picking_filter_lot

Yes, that one is to filter lot with any types of quant in location, this one is to filter lot wih positive sum of quants in location.

stock_picking_filter_lot_qty/models/stock_move_line.py
```diff
@@ -0,0 +1,28 @@
+from odoo import api, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = 'stock.move.line'
+
+    @api.onchange('product_id', 'product_uom_id')
+    def onchange_product_id(self):
+        res = super().onchange_product_id()
+        self.lot_id = False
```
rousseldenis: @sergiocorato I would say, as this is changing the standard behavior, it should be great to have an option on stock picking type. What do you think about ?

sergiocorato: Good option, I add it

sergiocorato: I added option to filter lot by location and/or by availability.

stock_picking_filter_lot_qty/models/stock_picking_type.py
```diff
@@ -0,0 +1,24 @@
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class StockPickingType(models.Model):
+    _inherit = 'stock.picking.type'
+
+    use_available_lots = fields.Boolean()
```
rousseldenis: A little help should be great

sergiocorato: Done, thanks!

PR #992, opened 2022-04-27
[14.0][MIG] stock_picking_show_backorder

JuanyDForgeflow:
This module provide a new field on stock pickings, allowing to see the
backorders associated to a picking.

PR #997, opened 2022-05-06
[14.0][FIX] stock_picking_invoice_link: test needs post install decorator + fix in test

AaronHForgeFlow:
Fixes 14.0 branch

Extracted from: https://github.com/OCA/stock-logistics-workflow/pull/992

Also the "partially_available" status does not exist in v14 in the picking: https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_picking.py#L265 it only exists in the stock moves. So a comparison in the test to that state makes no sense.

### DIFF
--- a/stock_picking_invoice_link/tests/test_stock_picking_invoice_link.py
+++ b/stock_picking_invoice_link/tests/test_stock_picking_invoice_link.py
@@ -4,12 +4,14 @@
 # Copyright 2021 Tecnativa - João Marques
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
+import odoo
 from odoo.exceptions import UserError
 from odoo.tests import Form
 
 from odoo.addons.sale.tests.common import TestSaleCommon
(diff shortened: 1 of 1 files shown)


stock_picking_invoice_link/tests/test_stock_picking_invoice_link.py
```diff
@@ -91,7 +93,7 @@ def test_00_sale_stock_invoice_link(self):
         )
         pick_1 = self.so.picking_ids.filtered(
             lambda x: x.picking_type_code == "outgoing"
-            and x.state in ("confirmed", "assigned", "partially_available")
```
pedrobaeza, line 94: This is doing more than only put post_install

AaronHForgeFlow: Sorry, I will update the description

rousseldenis (approved):
States were removed, so fast-tracking this

PR #998, opened 2022-05-09
[14.0] [ADD] stock_picking_reallocation

paradoxxxzero:
This module adds a reallocate button to dispatch a picking moves between several new pickings.

It uses a stored instance to allow saving current dispatching for later.


hparfr:
Create a reallocation, change the first date and the second.
Then the schedule date of the first and the second is the same.

And if possible, it's better if the backorder is always the second picking.

hparfr (approved):
LGTM thanks

mourad-ehm:
Hi @rousseldenis, could you reopen this PR ?

mourad-ehm (approved):
Code review

PR #1010, opened 2022-06-01
[14.0][IMP] stock_picking_invoice_link

ajaniszewska-dev:
port changes from https://github.com/OCA/stock-logistics-workflow/pull/816


stock_picking_invoice_link/models/account_move.py
```diff
@@ -16,6 +16,9 @@ class AccountMove(models.Model):
         string="Related Pickings",
         store=True,
         compute="_compute_picking_ids",
+        relation="account_invoice_stock_picking_rel",
+        column1="account_invoice_id",
```
grindtildeath: IMO we don't want to keep these values (using the defaults must be enough)

stock_picking_invoice_link/models/stock_picking.py
```diff
@@ -11,7 +11,13 @@ class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     invoice_ids = fields.Many2many(
-        comodel_name="account.move", copy=False, string="Invoices", readonly=True
+        comodel_name="account.invoice",
```
grindtildeath: Same as above, moreover this model does not exist in v14.0.

grindtildeath (approved):
LGTM :+1:

stock_picking_invoice_link/views/account_invoice_view.xml
```diff
@@ -3,7 +3,7 @@
     <!-- Copyright 2013-2014 Alexis de Lattre <alexis.delattre@akretion.com>
          License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
     <record id="invoice_form" model="ir.ui.view">
```
i-vyshnevska:
```suggestion
    <record id="view_move_form" model="ir.ui.view">
```

i-vyshnevska (approved):
add the complete name of addon to the PR name and commits message stock_picking_invoice_link

stock_picking_invoice_link/views/account_invoice_view.xml
ajaniszewska-dev: this record id already exists in: https://github.com/OCA/stock-logistics-workflow/blob/14.0/stock_picking_invoice_link/views/stock_view.xml#L24
so i get:

[W7902(duplicate-xml-record-id), ]  Duplicate xml record id "data/view_move_form_noupdate_0" in views/account_invoice_view.xml:5

i added "_inherit" postfix to avoid issues.

pedrobaeza (changes requested):
My fears of https://github.com/OCA/stock-logistics-workflow/pull/816#issuecomment-806443777 are still there

i-vyshnevska:
@pedrobaeza the problems emphasized in https://github.com/OCA/stock-logistics-workflow/pull/816#issuecomment-806443777 is reasonable, but this PR doesn't touch that part of the code, this improvement to other parts of the addon
as for v14 addon was already merged what would be the point to keep less effective?

pedrobaeza:
Please check latest changes on 13.0 version, that should cover all the possible cases. The latest one arrived with #750. We should forward-port that changes instead. If there's something still uncovered, please comment.

grindtildeath:
@pedrobaeza It's too bad I didn't got back to this PR during these OCA days to be able to talk about this issue together with you, and I'm not in Brussels but in the train on my way back to Switzerland.

I just tried the last changes that were merged in 14.0, but moves for product with ordered quantities are actually still not linked to the invoice.

About what you mentioned in https://github.com/OCA/stock-logistics-workflow/pull/816#issuecomment-806443777 , to me the biggest issue was the change to a computed field instead of having everything stored, but it was already done some times ago.

Then you guys have been adding considerations of return moves to credit notes in https://github.com/OCA/stock-logistics-workflow/pull/1100 which to me is the same kind of change of adding the support of ordered quantities.

Then AFAIU you also agreed in https://github.com/OCA/stock-logistics-workflow/pull/989#issuecomment-1109700392 that it doesn't make sense to keep this as a separated since the breaking change (i.e. the change to a computed field) was already merged.

Now, we will still need to rebase this PR and fix all the conflicts properly to consider the last changes that were done, but I don't want to spend any more time on this if the feature is going to stay in an unmerged PR.

Can we agree that we're going to merge this after rebase or should we better fix and merge https://github.com/OCA/stock-logistics-workflow/pull/989 ?

grindtildeath:
ping @pedrobaeza 
Can you please check my comment above? :pray:

pedrobaeza:
@CarlosRoca13 can you check?

grindtildeath:
ping @CarlosRoca13 :pray:

CarlosRoca13:
Hi @grindtildeath 

Sorry for not saying anything about it in this PR, We add the support to ordered quantities in this commit https://github.com/OCA/stock-logistics-workflow/commit/0bd063aad5874c83115c69c6bab87d4bc53c5186, you can check it and add any improvement that you need :smile:

grindtildeath:
@CarlosRoca13 No worries, at least the feature is now available out of the box in the module and we don't need to spend any more time on it.

@ajaniszewska-dev We can then close this PR.

PR #1012, opened 2022-06-06
[15.0][FIX] stock_picking_info_lot: no lot required check if no done quantities

dreispt:
- [FIX] stock_picking_info_lot: no lot required check if no done quantities

### DIFF
--- a/stock_picking_info_lot/views/stock_move_line_views.xml
+++ b/stock_picking_info_lot/views/stock_move_line_views.xml
@@ -44,7 +44,6 @@
                     name="lot_info"
                     attrs="{
                        'readonly': [('lot_info_usage', '=', 'no')],
-                       'required': [('lot_info_usage', '=', 'required')],
                        }"
                 />
             </xpath>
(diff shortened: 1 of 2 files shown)


stock_picking_info_lot/views/stock_move_line_views.xml
```diff
@@ -65,7 +64,6 @@
                     name="lot_info"
                     attrs="{
                        'readonly': [('lot_info_usage', '=', 'no')],
-                       'required': [('lot_info_usage', '=', 'required')],
```
rousseldenis, line 68: @dreispt 
Maybe you can set a field that compute the 'required' condition of lot_info_usage for form, so user can see if the field is required when he filled in a quantity ?

dreispt:
@rousseldenis Is any of the comments blocking for a merge? Thanks

rousseldenis (commented):
@dreispt Could you just add tests ?

rousseldenis:
@dreispt

PR #1014, opened 2022-06-09
[FIX] purchase with two steps can have dest moves on done state

ChrisOForgeFlow:
@ForgeFlow


rousseldenis (commented):
@ChrisOForgeFlow Thanks for this. Maybe this deserves a little test in order to avoid regression ?

PR #1015, opened 2022-06-10
[15.0][ADD]stock_picking_on_hold

dsolanki-initos:


stock_picking_on_hold/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    payment_method_id = fields.Many2one(
```
rousseldenis, line 7: IMHO, this has to go in another module

dsolanki-initos: @rousseldenis In which repository we should put this can you please suggest?

rousseldenis: In sale-workflow should be good.

rousseldenis:
@dsolanki-initos As it can be critical flow, could you add some tests ? Thanks

stock_picking_on_hold/models/sale_order.py
dsolanki-initos: I've created PR in sale-workflow you can refer this PR :- https://github.com/OCA/sale-workflow/pull/2073

PR #1004, opened 2022-05-20
[14.0][MIG] stock_valuation_layer_usage and stock_account_product_run_fifo_hook

ChrisOForgeFlow:
@ForgeFlow

Introduced new field to check original layer on outcome layers
Removed reservation case, because on core don't get any traceability data to assign candidate layer, only get by order and quantity available(field remaining_qty)

PR #1030, opened 2022-07-04
[13.0][FIX][stock_valuation_layer_usage]: look for true incoming moves

AaronHForgeFlow:
In case that the out move is connected with an internal transfer,
no valuation layer will be found, and then the system will create
a layer with negative quantity. We should try to look for the
true incoming move, if there is any.

cc @ForgeFlow

Backport from v14: https://github.com/OCA/stock-logistics-workflow/pull/1004


AaronHForgeFlow:
CI fails because of an import in other module :confused: 

```
workflow/stock_picking_group_by_partner_by_carrier_by_date/tests/__init__.py", line 1, in <module>
    from . import test_grouping_by_date
  File "/__w/stock-logistics-workflow/stock-logistics-workflow/stock_picking_group_by_partner_by_carrier_by_date/tests/test_grouping_by_date.py", line 3, in <module>
    from freezegun import freeze_time
ModuleNotFoundError: No module named 'freezegun'
```

MiquelRForgeFlow (approved):
LGTM

stock_valuation_layer_usage/models/stock_move.py
```diff
@@ -46,3 +46,12 @@ def _create_dropshipped_svl(self, forced_quantity=None):
                 }
             )
         return layers
+
+    def _get_in_moves(self):
```
rousseldenis, line 50: I'd prefer returning ids as better in terms of performances to do a set of ids and not using `|` operator

rousseldenis:
@AaronHForgeFlow

JordiMForgeFlow (approved):

PR #1032, opened 2022-07-05
[15.0][MIG] stock_return_request: Migration to 15.0

cesar-tecnativa:
@Tecnativa TT36575


pedrobaeza:
@cesar-tecnativa please check tests

rousseldenis:
@cesar-tecnativa

sergio-teruel:
tests fixed!!

stock_return_request/models/stock_return_request.py
```diff
@@ -267,7 +264,7 @@ def _action_confirm(self):
                     return_move.product_uom_qty += qty
                 done_moves.setdefault(move, self.env["stock.move"])
                 done_moves[move] += return_move
-                return_move._action_confirm()
+                return_move.state = "confirmed"
```
rousseldenis: Why _action_confirm() should not be used anymore ? At least a comment should be set

stock_return_request/models/stock_return_request.py
```diff
@@ -334,8 +331,9 @@ def _action_confirm(self):
             raise ValidationError(
                 _(
                     "It wasn't possible to assign stock for this returns:\n"
-                    "%s" % failed_moves_str
+                    "%(failed_moves_str)s"
                 )
+                % {"failed_moves_str": failed_moves_str}
```
rousseldenis: Usage of format() is better

stock_return_request/models/stock_return_request.py
```diff
@@ -535,15 +533,17 @@ def _get_returnable_move_ids(self):
...
-                            line.product_uom_id.name,
-                            line.product_id.display_name,
-                            qty_found,
-                        )
+                        "It wasn't possible to find enough moves to return "
+                        "%(line_quantity)f %(line_product_uom_id_name)s"
+                        "of %(line_product_id_displayname)s. A maximum of %(qty_found)f "
+                        "can be returned."
                     )
+                    % {
```
rousseldenis: Same as above and for all strings formatting

rousseldenis (changes requested):

stock_return_request/models/stock_return_request.py
pedrobaeza: Even for translatable content?

rousseldenis: Of course, you can do:

`_("My string {variable1}, {variable2}").format(variable1="Test", variable2="Test2")`

pedrobaeza: Is that preference expressed in any place of the guidelines?

rousseldenis: It's not an OCA one, but a python one

https://realpython.com/python-f-strings/

rousseldenis: But we can do a PR on OCA contributing guidelines

pedrobaeza: f-strings is something different than `.format`. What are the advantages of using `format` over `%`?

rousseldenis: For me, readability

pedrobaeza: Using dictionaries makes the readability fairly the same, and with `%`, you have less chances of overpassing line length, so not sure it's an advantage.

sergio-teruel:
All changes done!!

rousseldenis (approved):
LGTM

pedrobaeza (approved):

rousseldenis:
@cesar-tecnativa @pedrobaeza Could you do a fix to remove api.Environment.manage() in hooks.py to avoid warnings in test logs ? Thanks

pedrobaeza:
@rousseldenis done on #1134

stock_return_request/report/stock_return_report.xml
```diff
@@ -0,0 +1,141 @@
...
+
+    <template id="report_stock_return_request">
+        <t t-foreach="docs" t-as="request">
+            <t t-call="stock_return_request.report_stock_return_request_document" />
+        </t>
+    </template>
+
+    <record id="action_report_stock_return_request" model="ir.actions.report">
+        <field name="name">Stock Return Request Report</field>
+        <field name="model">report.stock.card.report</field>
```
CasVissers-360ERP, line 131: There is a dependency here on report.stock.card.report but this is not in the depends?

CasVissers-360ERP:
@cesar-tecnativa I was testing this module but it seems it is missing dependencies that are blocking report generation?

PR #1035, opened 2022-07-08
[15.0] [MIG] stock_move_assign_picking_hook

JasminSForgeFlow:
Standard Migration

@ForgeFlow


stock_move_assign_picking_hook/i18n/stock_move_assign_picking_hook.pot
```diff
@@ -0,0 +1,34 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* stock_move_assign_picking_hook
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 14.0\n"
```
MiquelRForgeFlow:
```suggestion
"Project-Id-Version: Odoo Server 15.0\n"
```

MiquelRForgeFlow (approved):

LoisRForgeFlow (changes requested):
I think this is not needed anymore, in v15 there is a hook already in place. https://github.com/odoo/odoo/commit/71ef4b728ea0fd3848208f83a05915e84e862b9a

Isn't it?

AaronHForgeFlow (commented):
I miss this part: https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L1043

Edit: indeed, the hook is already in Odoo standard, so the module is not needed.

JasminSForgeFlow:
Close PR because this hook cover in Odoo https://github.com/odoo/odoo/commit/71ef4b728ea0fd3848208f83a05915e84e862b9a

PR #1036, opened 2022-07-09
[13.0][FIX] stock_picking_invoice_link: Consider DS pickings

pedrobaeza:

- Create a sales order with Dropshipping route.
- Confirm the sale order and validate the DS picking.
- Invoice the purchase order.
- Invoice the sales order.

Current behavior:

The customer invoice is not linked to the DS picking

Expected behavior:

It's linked.

That's because current algorithm looks for existing invoices linked to the picking for discarding double links, but it's not having into
account that such invoices can be vendor ones.

We simply filter for customer invoices in the algorithm to fix it.

@Tecnativa TT37997

### DIFF
--- a/stock_picking_invoice_link/models/sale_order.py
+++ b/stock_picking_invoice_link/models/sale_order.py
@@ -15,7 +15,9 @@ class SaleOrderLine(models.Model):
                 and not (
                     any(
                         inv.state != "cancel"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@sergio-teruel @carlosdauden please review

carlosdauden (approved):

sergio-teruel (approved):

PR #1039, opened 2022-07-13
[15.0][MIG] stock_move_assign_picking_hook: Migrate 14.0 to 15.0

ellbristow:


theangryangel:
Don't believe this one is required under 15.0 due to https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L1033 being introduced?

ellbristow:
Cancelling as this module is no longer needed in 15.0
https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L1033

PR #1049, opened 2022-07-22
[14.0][FIX] stock_picking_filter_lot: if quantity changes location ids are not computed

lk-eska:
We had quantities in a location and weren't able to select it and we traced the problem to be this missing dependency.

### DIFF
--- a/stock_picking_filter_lot/models/stock_production_lot.py
+++ b/stock_picking_filter_lot/models/stock_production_lot.py
@@ -11,7 +11,7 @@ class StockProductionLot(models.Model):
         comodel_name="stock.location", compute="_compute_location_ids", store=True
     )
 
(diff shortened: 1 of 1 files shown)


lk-eska:
using quantity in this module is not a good idea at all.

dalonsod:
@lk-eska I've faced the same issue, in v13. You closed this indicating that using `quantity` shouldn't be a good idea but then... how did you solve this issue?

Code still doesn't include `quantity` in dependencies list in recent versions.

cc @fd-oerp @oihane @BernatPForgeFlow

lk-eska:
I didn't

dalonsod:
Well... implementing your fix is better that current code IMO, so I'll propose a PR cherry-picking your change anyway

PR #1050, opened 2022-07-25
[14.0] [FIX] delivery_total_weight_from_packaging: keep package weight on picking transfer

jbaudoux:
When a package weight is manually entered on a package, the value should not be lost when the package is transferred.

cc @sebalix @simahawk

### DIFF
--- a/delivery_total_weight_from_packaging/README.rst
+++ b/delivery_total_weight_from_packaging/README.rst
@@ -57,6 +57,7 @@ Contributors
 * Sébastien Alix <sebastien.alix@camptocamp.com>
 * `Trobz <https://trobz.com>`_:
 * Nguyen Hoang Hiep <hiepnh@trobz.com>
+* Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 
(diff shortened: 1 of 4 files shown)


sebalix:
@jbaudoux can you rebase?

simahawk:
@jbaudoux can you please elaborate on the use case?

jbaudoux:
> @jbaudoux can you please elaborate on the use case?

In a pick / pack / ship, when you pack the goods and create a parcel, you create a package and can provide the real measured shipping weight.
When you ship the parcel, the weight should not be erased and remain the one that was encoded during packing

sebalix:
> When you ship the parcel, the weight should not be erased and remain the one that was encoded during packing.

@jbaudoux can you put this comment in the compute method?

jbaudoux:
> > When you ship the parcel, the weight should not be erased and remain the one that was encoded during packing.
> 
> @jbaudoux can you put this comment in the compute method?

:+1: 
done

sebalix (approved):

PR #1052, opened 2022-08-01
[13.0][IMP] stock_picking_batch_extended_account: Add some tests. And more inheritable

sergio-teruel:
cc @Tecnativa TT36237

ping @carlosdauden @ernestotejeda

### DIFF
--- a/stock_picking_batch_extended_account/migrations/13.0.1.1.0/post-migration.py
+++ b/stock_picking_batch_extended_account/migrations/13.0.1.1.0/post-migration.py
@@ -1,0 +1,21 @@
+# Copyright 2022 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    batch_picking_auto_invoice_field = env.ref(
+        "stock_picking_batch_extended_account.field_res_partner__batch_picking_auto_invoice"
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE ir_property
+        SET value_text = CASE WHEN value_integer = 1 THEN 'yes' ELSE 'no' END,
+        type='char', value_integer=NULL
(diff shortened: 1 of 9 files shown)


carlosdauden (approved):

ernestotejeda (approved):
Code and functional review: :+1:

stock_picking_batch_extended_account/migrations/13.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE ir_property
+        SET value_text = CASE WHEN value_integer = 1 THEN 'yes' ELSE 'no' END,
```
pedrobaeza, line 16: Or is it `0`?

```suggestion
        SET value_text = CASE WHEN value_integer = 1 THEN 'yes' ELSE 'no' END,
            value_integer=NULL,
```

stock_picking_batch_extended_account/migrations/13.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+        type='char'
+        WHERE name='batch_picking_auto_invoice'
```
pedrobaeza: Please link by a stronger condition, as there can be another property field with the same name.

pedrobaeza (changes requested):

sergio-teruel:
@pedrobaeza Changes done!!

stock_picking_batch_extended_account/migrations/13.0.1.1.0/post-migration.py
pedrobaeza: Well, this is still vague. Please use `fields_id`.

sergio-teruel: Done!

pedrobaeza (approved):

PR #1062, opened 2022-08-09
[13.0][FIX] stock_picking_import_serial_number: Lot overwriting

ryanc-me:
The import wizard presents an option to "Overwrite Serial", which should
take all serials from the XLSX, and unconditionally write them into existing
SMLs.

However, the condition was not well formed, and so each XLSX line would
write to the same SML - effectively only importing the *last* XLSX line
into only the *first* SML.

PR #1065, opened 2022-08-10
[13.0][IMP] stock_picking_import_serial_number: Allow import S/N from file in internal and outgoing operations

sergio-teruel:
cc @Tecnativa TT38516

ping @pedrobaeza @chienandalu 

I have some doubt regarding internal pickings and ougoing...
At the moment, if the lot exists and not in sml reserved, the stock move line is created, but I don't check if there is availability...

@ryanc-me https://github.com/OCA/stock-logistics-workflow/pull/1062 fixed here

### DIFF
--- a/stock_picking_import_serial_number/README.rst
+++ b/stock_picking_import_serial_number/README.rst
@@ -39,9 +39,12 @@ Configuration
 To configure this module, you need to:
 
 #. Go to a *Inventory > Configuration > Settings*.
-#. Choose the product field which yo want to search products.
+#. Choose the product field which you want to search products.
 #. Choose the file column index where product reference is stored.
 #. Choose the file column index where serial number reference is stored.
+
+#. Go to a *Inventory > Operation types*.
+#. Check the field "Import lot from file" to allow to import S/N.
 
 Usage
 =====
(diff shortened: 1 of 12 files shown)


ryanc-me:
@sergio-teruel Looks good! I hope you don't mind my feedback below.

You make a good point regarding the availability check. Maybe one of these would work:

1. If the stock is not available, display a pop-up wizard showing the unavailable lines, and offering to Confirm or Cancel
2. Add an option to the initial import wizard (or company settings) so the user can choose whether to consider/ignore stock (e.g. "Check Availability" boolean field)
3. Ignore availability all together - this is an explicit import, much like entering SMLs manually via the detailed operations tab, so perhaps the user intent is to "import exactly what I say"?


Lastly, I think the availability-check logic should apply to incoming transfers also: If we have an inter-warehouse receipt, the picking type usage may be `incoming`, but the source location could actually be tracking stock (like `WH/Stock` -> `Transit` -> `WH2/Stock`, the last step is a receipt but we do know what stock is in `Transit`)

Maybe the [`location_src_id.usage`](https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_location.py#L37-L43) field is a good check instead? (i.e., ignore availability for any "Virtual" locations)

pedrobaeza (approved):
Let's finish with this:

/ocabot merge major

PR #1068, opened 2022-08-17
[13.0] [IMP] purchase_stock_picking_invoice_link: Use the same technique as in stock_picking_invoice_link

CarlosRoca13:
cc @Tecnativa TT38550

please @pedrobaeza @victoralmau review this


victoralmau (approved):
According to the logic of `stock_picking_invoice_link` it seems to make sense and everything is correct.

Remember to also apply these changes to v14 as well.

purchase_stock_picking_invoice_link/models/stock_move.py
```diff
@@ -8,6 +8,8 @@ class StockMove(models.Model):
     _inherit = "stock.move"
 
     def write(self, vals):
+        """Method used to associate the stock.move with the created account.move.line
```
pedrobaeza, line 11: For what is this needed?

pedrobaeza (commented):
Is the case of the credit note with "cancel + new invoice" option covered in this module?

purchase_stock_picking_invoice_link/models/stock_move.py
CarlosRoca13: Seeing this test, It's necesary to cover when we invoice the qty demanded before receive products from supplier. But the stock moves have to be linked with the corresponding invoice line.

https://github.com/OCA/stock-logistics-workflow/blob/92d19c57be418f1e53b773c4de232679808fc45a/purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py#L44-L68

CarlosRoca13:
> Is the case of the credit note with "cancel + new invoice" option covered in this module?

Yes, it is. I added a test to check it too

purchase_stock_picking_invoice_link/models/stock_move.py
pedrobaeza: OK, then please expand the docstring to say that the invoice was done before receiving the product. And docstrings end with the triple quote isolated in a new line if multi-line. See PEP257 for more information: https://peps.python.org/pep-0257/

pedrobaeza:
> > Is the case of the credit note with "cancel + new invoice" option covered in this module?
> 
> Yes, it is. I added a test to check it too

I don't see such test

CarlosRoca13:
> > > Is the case of the credit note with "cancel + new invoice" option covered in this module?
>>
> > Yes, it is. I added a test to check it too
>
> I don't see such test

https://github.com/OCA/stock-logistics-workflow/pull/1068/files#diff-a6f3c01f07d15948730d70d15d1f5f925d2677eb8fbeb0e71123c59e3e1b26aeR67-R97

purchase_stock_picking_invoice_link/models/stock_move.py
CarlosRoca13: Changes done

purchase_stock_picking_invoice_link/models/stock_move.py
```diff
@@ -8,6 +8,10 @@ class StockMove(models.Model):
...
+        when the invoicing method of the product is 'purchase' and the invoice is done
+        before receive the products.
```
pedrobaeza:
```suggestion
        before receiving the products.
```

pedrobaeza:
> > > > Is the case of the credit note with "cancel + new invoice" option covered in this module?
> > > 
> > > 
> > > Yes, it is. I added a test to check it too
> > 
> > 
> > I don't see such test
> 
> https://github.com/OCA/stock-logistics-workflow/pull/1068/files#diff-a6f3c01f07d15948730d70d15d1f5f925d2677eb8fbeb0e71123c59e3e1b26aeR67-R97

That's not the case I'm talking about. I'm talking about selecting this option in the "Add Credit Note" wizard:

Selección_007

CarlosRoca13:
@pedrobaeza test added and change done

pedrobaeza (approved):

pedrobaeza:
Please fw-port the changes.

/ocabot merge major

PR #1069, opened 2022-08-17
[14.0][FIX] stock_restrict_lot: do not merge stock move with different restrict_lot_id

petrus-v:
Odoo tries to avoid extra stock.move while generating stock.picking from
pull/push rules. We must tell odoo to not merge stock.move if they are restricted
different lot.

To reproduce this behaviour using sale_order_lot_selection you have to:
* setup inventory with multi-step routes
* setup warehouse with two steps outgoing shipments
* receive products with different lot
* create a sale order with different quantities per lot
* validate the sale order

Before this PR stock.move in the internal stock.picking pick order are merged

After this PR we have one line per selected lot in internal stock.picking pick order

### DIFF
--- a/stock_restrict_lot/models/stock_move.py
+++ b/stock_restrict_lot/models/stock_move.py
@@ -1,6 +1,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, exceptions, fields, models
+from odoo import _, api, exceptions, fields, models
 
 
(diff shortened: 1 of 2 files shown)


florian-dacosta (approved):

sebalix (approved):

sebalix:
@petrus-v ported to 15.0 here https://github.com/OCA/stock-logistics-workflow/pull/1078 , if you can review it :)

PR #1078, opened 2022-09-01
[15.0][FW] [14.0][FIX] stock_restrict_lot: do not merge stock move with different restrict_lot_id

sebalix:
Port of #1069 from 14.0 to 15.0.


sebalix:
The fix was already included in the migration commit there: https://github.com/OCA/stock-logistics-workflow/pull/921
I simply move the method at the same place + update the call to `super` (to reduce the chances of conflicts in the future), and the ported commit also add a unit test which is still useful.

petrus-v (approved):
LGTM Thanks taking care to FW unit test :heart:

PR #1084, opened 2022-09-07
[FIX] stock_picking_purchase_order_link: remove read access to ir.actions

BT-ojossen:
As odoo mentioned this changes here:
https://github.com/odoo/odoo/pull/53335


rousseldenis:
@BT-ojossen Thanks for this. Don't forget to install pre-commit locally before committing. Could you also provide a PR title more convenient for reviewers like "[FIX] stock_picking_purchase_order_link: remove read access to ir.actions" ?

rousseldenis (approved):

BT-fgarbely (approved):
Thanks for the fix

PR #1085, opened 2022-09-07
[13.0][ADD] stock_picking_product_assortment_availability_inline: Glue module to display availability quantities in pickings for a partner with a assortment defined

sergio-teruel:
cc @Tecnativa TT37820

The module stock_picking_product_assortment add a new field in picking operations to add an extra domain and the module stock_picking_product_availability_inline injects a context into product_id field.
This module injects the same context in the new field.

ping @chienandalu @carlosdauden


pedrobaeza:
Isn't better to intercept in the main module (`stock_picking_product_assortment`) the `fields_view_get` method and "copy" the other field for not requiring any glue module?

sergio-teruel:
I thin that this is the correct way.
I am sure that if If I had done it, you would have told me that the correct way is create a glue module... :smile:

pedrobaeza:
The border line between taking one approach or the other is clear for me, and I'll try to explain you for other times to take the same one:

- If the code is using introspection or any kind of hack like checking if there's a model/table or field/column, should go generally to use glue modules. Sometimes a exception is made if the overhead is minimal and the glue module will add very few value, but the main problem of this is that this part of the code won't be covered by unit tests and may lead to future problems when migrating to other versions.
- If the approach is taken for compatibility reasons, and don't use hacks like previous one, then no glue module is needed. I think this example is the best: you have an extra product field that shown instead of the normal one. The direct compatibility issue is that any module changing the original field extra attributes (attrs, etc) won't be reflected in the new field. Thus, a very good option for avoiding glue modules (or most of them), is to dynamically copy such attributes from one field to the other whatever they are, and this won't have any unit test penalty and the chances to get an incompatibility are lower, and if any, it will be discovered the same way as without this trick.

I hope that this explanation serves you for knowing how I would react to anything of this nature. And this is open also for debate if you don't agree with my reasoning.

Said that, and expecting more possible glue modules depending on the assortment things, don't you think the second approach would be the best?

pedrobaeza:
availavility > availability

pedrobaeza:
@CarlosRoca13 @carlosdauden if it's OK for you, please review it and let's merge it.

CarlosRoca13 (approved):

carlosdauden (approved):

PR #1086, opened 2022-09-09
[13.0][IMP] stock_picking_invoice_link: Show invoice and picking information only if the user has permissions.

victoralmau:
Show invoice and picking information only if the user has permissions.

@Tecnativa

### DIFF
--- a/stock_picking_invoice_link/views/account_invoice_view.xml
+++ b/stock_picking_invoice_link/views/account_invoice_view.xml
@@ -16,6 +16,7 @@
                     type="object"
                     attrs="{'invisible': [('picking_ids', '=', [])]}"
                     string="Pickings"
+                    groups="stock.group_stock_user"
                 >
                 </button>
             </div>
--- a/stock_picking_invoice_link/views/stock_view.xml
+++ b/stock_picking_invoice_link/views/stock_view.xml
@@ -5,6 +5,10 @@
         <field name="name">stock_picking_invoice_link.stock.picking.form</field>
         <field name="model">stock.picking</field>
         <field name="inherit_id" ref="stock.view_picking_form" />
+        <field
+            name="groups_id"
+            eval="[(6, 0, [ref('account.group_account_invoice')])]"
+        />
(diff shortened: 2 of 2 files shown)


stock_picking_invoice_link/views/account_invoice_view.xml
```diff
@@ -16,6 +16,7 @@
                     type="object"
                     attrs="{'invisible': [('picking_ids', '=', [])]}"
                     string="Pickings"
+                    groups="stock.group_stock_user"
```
pedrobaeza, line 19: This should be on the view, not here, or the same access errors will arise

stock_picking_invoice_link/views/stock_view.xml
```diff
@@ -25,6 +29,10 @@
         <field name="name">stock_picking_invoice_link.stock.move.form</field>
         <field name="model">stock.move</field>
         <field name="inherit_id" ref="stock.view_move_form" />
+        <field
+            name="groups_id"
+            eval="[(6, 0, [ref('account.group_account_invoice')])]"
```
pedrobaeza, line 34: There are multiple groups that gives access to invoices, so this is not true. Why this hasn't been a problem till now?

stock_picking_invoice_link/views/account_invoice_view.xml
victoralmau: Actually there is no access error, I think the smart-button should not be shown if the user does not have inventory permissions.

pedrobaeza: But is this needed then?

victoralmau: IMO yes, show a smart-button to see pickings if the user can't access, I think it can only confuse.

chienandalu: Billing users have explicit access to `stock.picking` so I'd this isn't needed.

pedrobaeza:
Closing then.

PR #1087, opened 2022-09-19
[13.0][IMP] stock_picking_invoice_link: Allow link more than one invoice

CarlosRoca13:
cc @Tecnativa TT38841

If we make partial invoices, we need that all partial invoices were linked
to the pickings related.


carlosdauden (approved):

pedrobaeza (changes requested):
This solution is incorrect and you can't restrict such modification. Please continue with the same approach as on purchases.

CarlosRoca13:
@pedrobaeza this is not the same use case

CarlosRoca13:
I will add the steps to reproduce and unit test for this use case

pedrobaeza:
Anyway, you can't restrict the change of the invoice quantity. And please expose the steps to reproduce the problem.

carlosdauden:
One way would be to add a configuration option to select this behavior

pedrobaeza:
I still need the why to do this for saying if this is legit.

carlosdauden:
Or you can add this restriction in account_invoice_report_grouped_by_picking

pedrobaeza:
I still don't see why you are going to rip the system to avoid a legit option.

CarlosRoca13:

1. Create a purchase/sale order buying/selling 10 units of some stockable product
2. Receive/send all
3. Create new invoice, invoicing just 8 units (This invoice will be linked to the picking)
4. Create new invoice, invoicing the remaining quantity
5. The invoice with 2 units won't be linked to the picking

pedrobaeza:
But restricting to change the invoice quantity doesn't solve this problem. You just forbid to do the steps 3 to 5 (and actually, you invoice and 10 units appear, but reduce the quantity to 8 afterwards, so the steps are not totally correct).

CarlosRoca13:
The PR is ready to be reviewed again :smile: please @carlosdauden @pedrobaeza

carlosdauden (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it

PR #1089, opened 2022-09-20
[13.0][FIX] purchase_stock_picking_invoice_link: Incorrect relation when backorders

CarlosRoca13:
cc @Tecnativa TT38840

Before this patch, when we receive products in a backorder and the other quantities received where invoiced yet, the picking of the backorder will be related to the invoice created for the other quantities.

With this patch, this situation will just happen when the invoicing policy of the product is based on quantities purchased.

Please @sergio-teruel @victoralmau review this

### DIFF
--- a/purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
+++ b/purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
@@ -23,14 +23,12 @@ class TestPurchaseSTockPickingInvoiceLink(common.SavepointCase):
         # Validate shipment
         picking = self.po.picking_ids[0]
         # Process pickings
-        picking.action_confirm()
         picking.move_lines.quantity_done = 1.0
         picking.button_validate()
         # Create and post invoice
         inv_action = self.po.action_view_invoice()
         inv_form = Form(self.env["account.move"].with_context(**inv_action["context"]))
         invoice = inv_form.save()
-        invoice.action_post()
         # Only one invoice line has been created
         self.assertEqual(len(invoice.invoice_line_ids), 1)
         line = invoice.invoice_line_ids
@@ -40,7 +38,7 @@ class TestPurchaseSTockPickingInvoiceLink(common.SavepointCase):
         # Invoices are set in pickings
         self.assertEqual(picking.invoice_ids, invoice)
 
-    def test_link_with_on_receive_quantities_policy(self):
+    def test_link_transfer_after_invoice_creation(self):
         self.product.purchase_method = "purchase"
         # Purchase order confirm
         self.po.button_confirm()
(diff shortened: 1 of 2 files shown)


sergio-teruel:
Review tests

sergio-teruel (approved):

victoralmau (approved):

pedrobaeza (changes requested):
I don't think this patch is correct, as the invoicing method shouldn't be linked to the stock moves we are finding.

You must always provide some steps to reproduce the problem this is trying to tackle, and you can start doing a regression test with such problem, and then, let's see the causes and how to solve it.

pedrobaeza:
> Before this patch, when we receive products in a backorder and the other quantities received where invoiced yet, the picking of the backorder will be related to the invoice created for the other quantities.

These are not clear steps to reproduce the problem. Please provide step per step in a list to get to the problem.

CarlosRoca13:

1. Create a purchase order buying 10 units of some stockable product
2. Receive 8 units of the product
3. Create back order
4. Invoice the received quantity (8 units)
5. Receive 2 units of the product in the created back order
6. You will see that the back order picking is related to the invoice with 8 units

pedrobaeza:
Thanks for the steps. I don't see with such steps how your patch solves this problem. And please add such case as a regression test.

CarlosRoca13:
Done @pedrobaeza

purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
```diff
@@ -123,3 +123,27 @@ def test_invoice_refund_modify(self):
...
         self.assertTrue(len(picking.invoice_ids) == 3)
         self.assertEqual(new_inv.picking_ids, picking)
+
+    def test_purchase_invoice_backorder_no_linked(self):
+        self.po.order_line.product_qty = 10
+        self.po.button_confirm()
+        picking = self.po.picking_ids[0]
+        picking.action_confirm()
+        picking.move_lines.quantity_done = 8.0
+        picking.button_validate()
```
pedrobaeza: Unneeded line.

purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
```diff
@@ -123,3 +123,27 @@ def test_invoice_refund_modify(self):
...
+        )
+        wiz.process()
+        inv_action = self.po.action_view_invoice()
+        inv_form = Form(self.env["account.move"].with_context(**inv_action["context"]))
+        invoice = inv_form.save()
+        invoice.action_post()
+        self.assertTrue(len(picking.invoice_ids) == 1)
+        self.assertEqual(invoice.picking_ids, picking)
+        backorder_picking = self.po.picking_ids.filtered(lambda p: p.state != "done")
+        backorder_picking.action_confirm()
```
pedrobaeza: Unneeded line

pedrobaeza:
The invoicing method is not explicitly set on the created product, and you have to test both cases, so the test suite must be expanded.

CarlosRoca13:
Done @pedrobaeza 

I have set explicitly the method on last test and added a new one for the other invoicing policy method.

pedrobaeza:
Please reflect the steps to reproduce the problem in the commit message, and there's still an unneeded line in the tests.

CarlosRoca13:
Added the steps and removed the line

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it to upper versions and apply the same to the other module and ping me when done.

PR #1096, opened 2022-09-29
[14.0] Add module stock_lot_product_qty_search

grindtildeath:
This module adds a search function on the Quantity (`product_qty`) field
of Lot/Serial Numbers (`stock.production.lot`) model, in order to search
on this field and provide a filter for records having a quantity on hand.

### DIFF
--- a/stock_lot_product_qty_search/__manifest__.py
+++ b/stock_lot_product_qty_search/__manifest__.py
@@ -1,0 +1,20 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock lot product qty search",
+    "summary": "Allows to search on Quantity field of Lot/Serial Numbers",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory/Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["grindtildeath"],
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "stock",
+    ],
+    "data": [
+        "views/stock_production_lot.xml",
+    ],
+}
--- a/stock_lot_product_qty_search/models/stock_production_lot.py
+++ b/stock_lot_product_qty_search/models/stock_production_lot.py
@@ -1,0 +1,66 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+from odoo.osv.expression import OR
+from odoo.tools import float_compare, float_is_zero
+
+from odoo.addons.stock.models.product import OPERATORS
+
+
+class ProductionLot(models.Model):
+    _inherit = "stock.production.lot"
+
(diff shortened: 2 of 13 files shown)


stock_lot_product_qty_search/models/stock_production_lot.py
```diff
@@ -0,0 +1,46 @@
...
+            ["lot_id", "quantity"],
+            ["lot_id"],
+            orderby="id",
+        )
+        lot_ids_with_quantity = {
+            group["lot_id"][0]: group["quantity"] for group in grouped_quants
+        }
+        lot_ids_without_qty = self.search(
+            [("id", "not in", list(lot_ids_with_quantity.keys()))]
+        ).ids
```
sebalix: Maybe add a comment saying this could have bad performance on huge DB?

sebalix (approved):

stock_lot_product_qty_search/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock lot product qty search",
+    "summary": "Allows to search on Quantity field of Lot/Serial Numbers",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory/Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["your-github-login"],
```
rousseldenis: Please change this

stock_lot_product_qty_search/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module adds a search function on the Quantity (`product_qty`) field
+of Lot/Serial Numbers (`stock.production.lot`) model, in order to search
+on this field and provide a filter for records having a quantity on hand.
```
rousseldenis, line 3: Ideally as @sebalix said, a line here to warn about performances problems.

rousseldenis (approved):

grindtildeath:
@sebalix @rousseldenis 
I just did profile the method on a 14.0 DB with 33K records and here's the result: https://gist.github.com/grindtildeath/39d04273fa1c515be6235c846a0d4ffa
IMO there's no need to warn about performance as it doesn't look bad at all.

rousseldenis:
As I see, you build modules that depends on this.

Isn't really a more elegant way to do this ?

grindtildeath:
@rousseldenis What do you mean by "more elegant"?
IMO, since the two modules provide distinct features (this one provides a search function and a filter based on this search function without altering anything, whereas the other one provides an extra configuration to alter results in M2o fields), I find it's good to have two separated modules.

rousseldenis:
> @rousseldenis What do you mean by "more elegant"? IMO, since the two modules provide distinct features (this one provides a search function and a filter based on this search function without altering anything, whereas the other one provides an extra configuration to alter results in M2o fields), I find it's good to have two separated modules.

The _search_product_qty() function will result in three searches. I'm just wondering if there is no other mean to do that

grindtildeath:
@rousseldenis I did add a condition to perform the search for lots without quantity, so it will avoid the unnecessary costly search if we're not looking for lots with a quantity of 0.
I guess we cannot do much better if we don't want to bypass the ORM.

simahawk (approved):

simahawk:
@grindtildeath can you squash?

PR #1097, opened 2022-09-29
[14.0] Add module stock_lot_on_hand_first

grindtildeath:
This module allows to display lots with a quantity on hand as first results
in the selection field of Detailed Operations (`stock.move.line`) in order
to avoid displaying old lots that are not in stock anymore.

Depends on:
 - [x] #1096

### DIFF
--- a/stock_lot_on_hand_first/__manifest__.py
+++ b/stock_lot_on_hand_first/__manifest__.py
@@ -18,5 +18,6 @@
         "views/stock_move_line.xml",
         "views/stock_move.xml",
         "views/stock_picking_type.xml",
+        "views/stock_picking.xml",
     ],
 }
--- a/stock_lot_on_hand_first/__manifest__.py
+++ b/stock_lot_on_hand_first/__manifest__.py
@@ -11,9 +11,7 @@
     "maintainers": ["grindtildeath"],
     "license": "AGPL-3",
     "installable": True,
-    "depends": [
-        "stock_lot_product_qty_search",
-    ],
+    "depends": ["stock_lot_product_qty_search", "base_view_inheritance_extension"],
     "data": [
         "views/stock_move_line.xml",
         "views/stock_move.xml",
--- a/stock_lot_on_hand_first/__manifest__.py
+++ b/stock_lot_on_hand_first/__manifest__.py
@@ -1,0 +1,22 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Lot On Hand First",
+    "summary": "Allows to display lots on hand first in M2o fields",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory/Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["grindtildeath"],
+    "license": "AGPL-3",
+    "installable": True,
+    "depends": [
+        "stock_lot_product_qty_search",
+    ],
+    "data": [
+        "views/stock_move_line.xml",
+        "views/stock_move.xml",
+        "views/stock_picking_type.xml",
+    ],
+}
--- a/stock_lot_on_hand_first/models/stock_move.py
+++ b/stock_lot_on_hand_first/models/stock_move.py
@@ -1,0 +1,14 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    display_lots_on_hand_first = fields.Boolean(
+        related="picking_type_id.display_lots_on_hand_first",
+        readonly=True,
+    )
--- a/stock_lot_on_hand_first/models/stock_picking.py
+++ b/stock_lot_on_hand_first/models/stock_picking.py
@@ -1,0 +1,14 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    display_lots_on_hand_first = fields.Boolean(
+        related="picking_type_id.display_lots_on_hand_first",
+        readonly=True,
+    )
--- a/stock_lot_on_hand_first/models/stock_picking_type.py
+++ b/stock_lot_on_hand_first/models/stock_picking_type.py
@@ -1,0 +1,23 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+
+
+class StockPickingType(models.Model):
+
+    _inherit = "stock.picking.type"
+
+    display_lots_on_hand_first = fields.Boolean(
+        compute="_compute_display_lots_on_hand_first",
+        store=True,
+        readonly=False,
+        help="When marked, lots that do not have a quantity in stock will be displayed"
+        " last in the selection",
+    )
+
+    @api.depends("use_existing_lots")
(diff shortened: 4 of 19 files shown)


stock_lot_on_hand_first/models/stock_picking_type.py
```diff
@@ -0,0 +1,23 @@
...
+        help="When marked, lots that do not have a quantity in stock will be displayed"
+        " last in the selection",
+    )
+
+    @api.depends("use_existing_lots")
+    def _compute_display_lots_on_hand_first(self):
+        """Reset display_lots_on_hand_first if use_existing_lots is set to False"""
+        for picking_type in self:
+            if not picking_type.use_existing_lots:
+                picking_type.display_lots_on_hand_first = False
```
i-vyshnevska, line 23:
```suggestion
                picking_type.display_lots_on_hand_first = False
            else:
                picking_type.display_lots_on_hand_first = True
```

stock_lot_on_hand_first/models/stock_production_lot.py
```diff
@@ -0,0 +1,40 @@
...
+                with_quantity_domain
+            )
+
+            if with_quantity_count >= limit:
+                name = with_quantity_domain
+            else:
+                with_quantity_ids = super()._name_search(
+                    name=name, args=with_quantity_domain, operator=operator, limit=limit
+                )
+                without_quantity_ids = super()._name_search(
```
simahawk, line 28: can't we sort result by qty instead of doing 2 searches?

stock_lot_on_hand_first/models/stock_picking_type.py
grindtildeath: No, in this case we don't want to force the value if the user don't set it by himself. The goal is only to set it to False if `use_existing_lots` was set to False, as an onchange would have done. :wink:

stock_lot_on_hand_first/models/stock_production_lot.py
grindtildeath: That's a debatable choice, but since we are here in a corner case where a limit is defined and we have less results with quantity than the limit, I'd prefer keeping it this way.

stock_lot_on_hand_first/views/stock_move_line.xml
```diff
@@ -0,0 +1,40 @@
...
+            name="inherit_id"
+            ref="stock.view_stock_move_line_detailed_operation_tree"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='lot_id']" position="attributes">
+                <attribute name="context">{
+                    'name_search_qty_on_hand_first': parent.display_lots_on_hand_first,
+                    'active_picking_id': picking_id,
+                    'default_company_id': parent.company_id,
+                    'default_product_id': parent.product_id,}</attribute>
```
grindtildeath:
```suggestion
                    'default_company_id': company_id,
                    'default_product_id': product_id,}</attribute>
```

stock_lot_on_hand_first/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Lot On Hand First",
+    "summary": "Allows to display lots on hand first in M2o fields",
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory/Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["your-github-login"],
```
sebalix: To change?
```suggestion
    "maintainers": ["your-github-login"],
```

stock_lot_on_hand_first/models/stock_production_lot.py
```diff
@@ -0,0 +1,40 @@
...
+
+class ProductionLot(models.Model):
+    _inherit = "stock.production.lot"
+
+    @api.model
+    def _name_search(self, name="", args=None, operator="ilike", limit=80):
+        """Move lots without a qty on hand at the end of the list"""
+
+        if self.env.context.get("name_search_qty_on_hand_first"):
+            with_quantity_domain = AND([args, [("product_qty", ">", 0)]])
```
sebalix, line 17: `args` could be `None`, should we copy the line from parent method?
https://github.com/odoo/odoo/blob/120ca0804f90b0782ee5075213a77f9d4f72ceef/odoo/models.py#L1810

sebalix (commented):
Dependency has been merged (commit to drop) + few comments

stock_lot_on_hand_first/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    display_lots_on_hand_first = fields.Boolean(
+        related="picking_type_id.display_lots_on_hand_first",
+        readonly=True,
```
ivantodorovich, line 13:
```suggestion
```
nitpicking: related is readonly by default ;)

stock_lot_on_hand_first/models/stock_picking.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    display_lots_on_hand_first = fields.Boolean(
+        related="picking_type_id.display_lots_on_hand_first",
+        readonly=True,
```
ivantodorovich, line 13:
```suggestion
```
nitpicking: related is readonly by default ;)

stock_lot_on_hand_first/models/stock_production_lot.py
ivantodorovich: Agree

stock_lot_on_hand_first/models/stock_production_lot.py
```diff
@@ -0,0 +1,40 @@
...
+                    name=name,
+                    args=AND([args, [("product_qty", "=", 0)]]),
+                    operator=operator,
+                    limit=limit - with_quantity_count,
+                )
+
+                return (
+                    self.browse(with_quantity_ids).ids
+                    + self.browse(without_quantity_ids).ids
+                )
```
ivantodorovich, line 42: Careful here, the original method doesn't return a list of ids.

The original method returns what `self._search` returns, and `_search` returns a `Query` instance (that evaluates to list of ids when it's being type casted, but it's more than that)

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo14-addon-stock_lot_product_qty_search@git+https://github.com/OCA/stock-logistics-workflow@refs/pull/1096/head#subdirectory=setup/stock_lot_product_qty_search
```
ivantodorovich: Already merged

ivantodorovich (changes requested):

stock_lot_on_hand_first/models/stock_production_lot.py
grindtildeath: Good catch

stock_lot_on_hand_first/models/stock_production_lot.py
grindtildeath: I added a comment to explain this. I guess it is fine to keep it this way?

grindtildeath:
@simahawk @sebalix @i-vyshnevska @ivantodorovich Thanks for the reviews, can you guys please update it after changes were attended? :pray:

ivantodorovich (approved):

rousseldenis (approved):
Code review

rousseldenis:
@grindtildeath Is this ready ?

grindtildeath:
@rousseldenis I guess so :wink:

rousseldenis:
@grindtildeath Maybe adding stock_lot_on_hand_first at the beginning of each commit message ?

rousseldenis:
@grindtildeath

grindtildeath:
@rousseldenis done

rousseldenis:
@simahawk @i-vyshnevska @sebalix ok for you ?

i-vyshnevska (approved):

sebalix (approved):

PR #1116, opened 2022-10-26
[14.0] Add stock_picking_progress

mmequignon:
This module adds a new `progress` field on `stock.picking` and `stock.move.line`.

On `stock.move.line`, this field represents the percentage of `qty_done` compared to the `product_uom_qty`.
On `stock.picking`, this field is the average progress of all move lines.
image

ref: rau-49

### DIFF
--- a/stock_picking_progress/hooks.py
+++ b/stock_picking_progress/hooks.py
@@ -1,0 +1,57 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+
+from odoo.tools.sql import column_exists, create_column
+
+logger = logging.getLogger(__name__)
+
+
+def setup_move_line_progress(cr):
+    table = "stock_move_line"
+    column = "progress"
+    _type = "float"
+    if column_exists(cr, table, column):
+        logger.info("%s already exists on %s, skipping setup", column, table)
+        return
+    logger.info("creating %s on table %s", column, table)
+    create_column(cr, table, column, _type)
+    fill_column_query = """
+        UPDATE stock_move_line
+        SET progress = CASE
+            WHEN (state = 'done') THEN 100
+            WHEN (product_uom_qty IS NULL or product_uom_qty = 0.0) THEN 0.0
+            ELSE (qty_done / product_uom_qty) * 100
+        END;
+    """
+    logger.info("filling up %s on %s", column, table)
+    cr.execute(fill_column_query)
+
+
+def setup_picking_progress(cr):
+    table = "stock_picking"
+    column = "progress"
+    _type = "float"
+    if column_exists(cr, table, column):
+        logger.info("%s already exists on %s, skipping setup", column, table)
+        return
+    logger.info("creating %s on table %s", column, table)
+    create_column(cr, table, column, _type)
+    fill_column_query = """
+        UPDATE stock_picking p
+        SET progress = subquery.avg_progress
+        FROM (
+            SELECT sml.picking_id, avg(sml.progress) as avg_progress
+            FROM stock_move_line sml
+            GROUP BY sml.picking_id
+        ) as subquery
+        WHERE p.id = subquery.picking_id;
+    """
+    logger.info("filling up %s on %s", column, table)
+    cr.execute(fill_column_query)
+
+
+def pre_init_hook(cr):
+    setup_move_line_progress(cr)
+    setup_picking_progress(cr)
--- a/stock_picking_progress/models/stock_move_line.py
+++ b/stock_picking_progress/models/stock_move_line.py
@@ -1,0 +1,27 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+from odoo.tools.float_utils import float_is_zero
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    progress = fields.Float(
+        compute="_compute_progress", store=True, group_operator="avg"
+    )
+
+    @api.depends(
+        "product_uom_qty", "product_uom_id", "product_uom_id.rounding", "qty_done"
+    )
+    def _compute_progress(self):
+        for record in self:
+            if record.state == "done":
+                record.progress = 100
+                continue
+            rounding = record.product_uom_id.rounding
+            if float_is_zero(record.product_uom_qty, precision_rounding=rounding):
+                record.progress = 0
+            else:
+                record.progress = (record.qty_done / record.product_uom_qty) * 100
--- a/stock_picking_progress/models/stock_picking.py
+++ b/stock_picking_progress/models/stock_picking.py
@@ -1,0 +1,25 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    progress = fields.Float(
+        compute="_compute_progress", store=True, group_operator="avg"
+    )
+
+    @api.depends("move_line_ids", "move_line_ids.progress")
+    def _compute_progress(self):
+        for record in self:
+            if record.state == "done":
+                record.progress = 100
+                continue
+            lines_progress = record.move_line_ids.mapped("progress")
+            # Avoid dividing by 0
+            if lines_progress:
+                record.progress = sum(lines_progress) / len(lines_progress)
+            else:
+                record.progress = 0
--- a/setup/stock_picking_progress/odoo/addons/stock_picking_progress
+++ b/setup/stock_picking_progress/odoo/addons/stock_picking_progress
@@ -1,0 +1,1 @@
+../../../../stock_picking_progress
--- a/setup/stock_picking_progress/setup.py
+++ b/setup/stock_picking_progress/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/stock_picking_progress/README.rst
+++ b/stock_picking_progress/README.rst
@@ -1,0 +1,1 @@
+wait 4 da bot
--- a/stock_picking_progress/__init__.py
+++ b/stock_picking_progress/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from .hooks import pre_init_hook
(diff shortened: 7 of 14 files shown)


rousseldenis:
@mmequignon Thanks for this. Nothing in a view ?

stock_picking_completion/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+{
+    "name": "Stock Picking Completion",
+    "summary": "Compute the stock.picking completion",
+    "version": "14.0.1.0.0",
+    "category": "Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "mmequignon, Camptocamp SA, Odoo Community Association (OCA)",
```
sebalix: I don't think you have to put your name in the `author` key, authorship belongs to the company (in case of software edition, at least in France).
```suggestion
    "author": "Camptocamp SA, Odoo Community Association (OCA)",
```

stock_picking_completion/hooks.py
```diff
@@ -0,0 +1,73 @@
...
+
+def column_exists(cr, args):
+    column_exists_query = """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name=%(table)s AND
+        column_name=%(column)s;
+    """
+    cr.execute(column_exists_query, args)
+    return cr.fetchone()
```
sebalix: I put that here: this function already exists in `openupgradelib`: https://github.com/OCA/openupgradelib/blob/94762d1a5a4172b680a7e019fe4d7db44b2bf7af/openupgradelib/openupgrade_tools.py#L36-L44

But I don't know if adding such dependency for this need is a good thing...

stock_picking_completion/hooks.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def setup_move_line_completion(cr):
+    table = "stock_move_line"
+    column = "completion"
+    args = {"table": table, "column": column}
+    if column_exists(cr, args):
```
sebalix: Nitpick, `args` is not obvious
```suggestion
    if column_exists(cr, table, column):
```

stock_picking_completion/hooks.py
```diff
@@ -0,0 +1,73 @@
...
+        SET completion = subquery.avg_completion
+        FROM (
+            SELECT sml.picking_id, avg(sml.completion) as avg_completion
+            FROM stock_move_line sml
+            GROUP BY sml.picking_id
+        ) as subquery
+        WHERE p.id = subquery.picking_id;
+    """
+    logger.info(f"filling up {column} on {table}")
+    cr.execute(fill_column_query)
```
sebalix: Did you test on a big DB how much time it takes?

stock_picking_completion/models/stock_move_line.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+from odoo.tools.float_utils import float_is_zero
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    completion = fields.Float(compute="_compute_completion", store=True)
```
sebalix: Why do we need this field stored BTW?

sebalix (commented):
Few comments.
As said we would like to rename this module `stock_picking_progress` as another module already this name (almost): [`stock_picking_completion_info`](https://github.com/OCA/wms/blob/14.0/stock_picking_completion_info/__manifest__.py)

stock_picking_completion/models/stock_move_line.py
mmequignon: So we don't have to compute all at once when loading a bunch of those.

stock_picking_completion/hooks.py
mmequignon: can't recall precisely, but I think it was between 1 and 2 minutes.

mmequignon:
> @mmequignon Thanks for this. Nothing in a view ?

Well, I was thinking about it and in the end, I think you can easily see what is done and compare it with what has to be done.
It will be used on the WMS stack, in `shopfloor_mobile`, in order to change the display color of the pickings / move lines depending on their progression.
I can't find a usecase where this field would be useful.

stock_picking_completion/models/stock_move_line.py
mmequignon: I.E. when retrieving pickings from `shopfloor_mobile`.
There's no point in this field if it's not stored, actually.

sebalix:
@mmequignon I think what @rousseldenis meant is, this module could display the progression on the transfer view, there is for instance the `widget="progressbar"` that could be used for that. This way your module is not anymore a technical module that could be used by other modules (like Shopfloor) but is useful on its own from a user POV.

stock_picking_progress/models/stock_move_line.py
```diff
@@ -0,0 +1,22 @@
...
+    @api.depends(
+        "product_uom_qty", "product_uom_id", "product_uom_id.rounding", "qty_done"
+    )
+    def _compute_progress(self):
+        for record in self:
+            rounding = record.product_uom_id.rounding
+            if float_is_zero(record.product_uom_qty, precision_rounding=rounding):
+                record.progress = 0.0
+            else:
+                record.progress = (record.qty_done / record.product_uom_qty) * 100
```
sebalix, line 27: In order to be used by the widget `progressbar`/`percentage`/`percentpie` you can't do that. Keep this operation when rendering the value for the user.
```suggestion
                record.progress = (record.qty_done / record.product_uom_qty)
```
Maybe rename this field `progress_ratio` then?

stock_picking_progress/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    progress = fields.Float(compute="_compute_progress", store=True)
+
+    @api.depends("move_line_ids", "move_line_ids.progress")
+    def _compute_progress(self):
+        for record in self:
+            lines_progress = record.move_line_ids.mapped("progress")
```
simahawk, line 20: the value on the move lines is stored: can't you do one select only w/ sum + count?

stock_picking_progress/hooks.py
```diff
@@ -0,0 +1,73 @@
...
+
+def column_exists(cr, args):
+    column_exists_query = """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name=%(table)s AND
+        column_name=%(column)s;
+    """
+    cr.execute(column_exists_query, args)
+    return cr.fetchone()
```
simahawk:
```suggestion
from odoo.tools.sql import column_exists
```

stock_picking_progress/hooks.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def setup_move_line_completion(cr):
+    table = "stock_move_line"
+    column = "completion"
+    args = {"table": table, "column": column}
+    if column_exists(cr, args):
+        logger.info(f"{column} already exists on {table}, skipping setup")
+        return
+    add_column_query = """
```
simahawk: you can use https://github.com/odoo/odoo/blob/14.0/odoo/tools/sql.py#L90

simahawk:
> > @mmequignon Thanks for this. Nothing in a view ?
> 
> Well, I was thinking about it and in the end, I think you can easily see what is done and compare it with what has to be done. It will be used on the WMS stack, in `shopfloor_mobile`, in order to change the display color of the pickings / move lines depending on their progression. I can't find a usecase where this field would be useful.

This is already done in the UI... what's missing?

simahawk (changes requested):
not sure we need this for the purpose you mentioned

stock_picking_progress/hooks.py
```diff
@@ -0,0 +1,73 @@
...
+        ALTER TABLE stock_move_line
+        ADD COLUMN completion float
```
JuMiSanAr: This should be progress right? It's not completion anymore.

JuMiSanAr (changes requested):
Maybe add a small explanation saying why this needs to be computed in the backend instead of in the frontend as before?

mmequignon:
> not sure we need this for the purpose you mentioned

We do. In order to display completion on a picking, you have to return all move lines.
When you want to list pickings to be done and add a color based on their progression, then you'll wait a loong time before it is displayed, which is why we had performance issues on a big database.
The fix here is, we don't need the move lines to only display de progress on a picking.
Just return the progress.

stock_picking_progress/hooks.py
mmequignon: good catch <3

simahawk:
@mmequignon ping

stock_picking_progress/models/stock_move_line.py
mmequignon: progressbar accepts values as percentages.

mmequignon:
@simahawk pr is updated

stock_picking_progress/models/stock_move_line.py
sebalix: I was wrong about the `* 100`, by default Odoo expects a %, not a ratio, for its different widgets (but we can override this through widget options)

stock_picking_progress/hooks.py
```diff
@@ -0,0 +1,56 @@
...
+    logger.info(f"creating {column} on table {table}")
+    create_column(cr, table, column, _type)
+    fill_column_query = """
+        UPDATE stock_move_line
+        SET progress = CASE
+            WHEN (product_uom_qty IS NULL or product_uom_qty = 0.0) THEN 0.0
+            ELSE (qty_done / product_uom_qty) * 100
+        END;
+    """
+    logger.info(f"filling up {column} on {table}")
```
simahawk:
```suggestion
    logger.info("filling up %s on %s", column, table)
```
Let it do its job and strings will be interpolated when the log is produced - if needed - and not at run time.

simahawk (commented):
what about @sebalix  comment on "ratio"?

sebalix (approved):

mmequignon:
> what about @sebalix comment on "ratio"?

dropped, as it is useless. The widget accepts percentages and shopfloor uses percentages as well.
To me it's pointless to use ratios here.

stock_picking_progress/hooks.py
mmequignon: modified this, testing the hook locally before pushing the change.

mmequignon: @simahawk updated

jbaudoux (approved):

simahawk (approved):

PR #1118, opened 2022-10-27
[FIX] stock_split_picking: confirm moves before assigning

i-vyshnevska:
as moves are copied they need to be confirmed first and then assigned, as on confirmation important steps that cannot be skipped


grindtildeath (approved):

TDu:
IMO the commit message should be rewritten to follow common conventions https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message
Too long title...

PR #1127, opened 2022-11-09
[16.0] [ADD] stock_picking_start: Add button start on picking form

lmignon:
The action_start, set the printed field to True and optionally fill the responsible field with the current user depending of the stock settings


lmignon:
cc @rousseldenis @jbaudoux

stock_picking_start/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Picking Start",
+    "summary": """
+        Add button to start picking""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
rousseldenis, line 10: You can add yourself to maintainers entry

rousseldenis (approved):
LGTM. Minor comment.

Should'nt we add a 'Stop' if people clicked on Start and do not want to ?

stock_picking_start/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+        if no_start_allowed:
+            raise UserError(
+                _(
+                    "The following picking(s) can't be started:\n" "%(names)s",
+                    names="\n".join(no_start_allowed.mapped("name")),
+                )
+            )
+
+    def _prepare_action_start_values(self, company):
+        value = {"printed": True}
```
jbaudoux, line 88: There should be a way to revert this. In old versions, this field was displayed but it's not since a few versions.

stock_picking_start/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+        if company.stock_picking_assign_operator_at_start:
```
jbaudoux, line 89: If we want to use that in shopfloor, then we need to ensure to configure it properly on the company. Not sure the option on the company is a good idea

rousseldenis: Of course it is as this changes the core default behaviour. It's a setting

lmignon:
@jbaudoux @rousseldenis The user XP is improved...
* started picking highlighted into the tree view 
* predefined filter for started pickings
* use badge on the form view to highlight a started picking
* add  "Cancel Started" button on started picking

image

image

image

sbejaoui (approved):
LGTM (code review)

jbaudoux (approved):

PR #1131, opened 2022-11-14
stock_scheduler_assignation_horizon: wider settings approach

i-vyshnevska:
scheduler ususally running in global mode when company is not set so settings defined on company level are not applied


simahawk:
s/approuche/approach
s/ususally/usually

Also, can you rephrase the description please?

stock_scheduler_assignation_horizon/models/res_config_settings.py
```diff
@@ -27,3 +40,13 @@ def _check_moves_assignation_horizon(self):
         for record in self:
             if record.moves_assignation_horizon < 0:
                 raise ValidationError(_("The assignation horizon cannot be negative"))
+
+    def get_global_is_moves_assignation_limited(self):
+        get_param = self.env["ir.config_parameter"].sudo().get_param
+        param = "s_scheduler_assignation_horizon.global_moves_assignation"
+        return safe_eval(get_param(param, False))
```
simahawk: no need for safe_eval, pls use `str2bool`

simahawk: Also, these methods should not stay here. I would move them to where you need them: on the proc group.

stock_scheduler_assignation_horizon/models/res_config_settings.py
```diff
@@ -3,11 +3,24 @@
...
 
 
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
+    global_moves_assignation = fields.Boolean(
+        "Use global Assignation Limit",
+        config_parameter="s_scheduler_assignation_horizon.global_moves_assignation",
+        readonly=False,
+        help="Check this box to allow limit settings " "for each company independent",
```
simahawk:
```suggestion
        help="Enable scheduler assignation horizon limit for all companies",
```

stock_scheduler_assignation_horizon/models/res_config_settings.py
```diff
@@ -3,11 +3,24 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tools.safe_eval import safe_eval
 
 
 class ResConfigSettings(models.TransientModel):
     _inherit = "res.config.settings"
 
+    global_moves_assignation = fields.Boolean(
```
simahawk:
```suggestion
    stock_horizon_move_assignation = fields.Boolean(
```

stock_scheduler_assignation_horizon/models/res_config_settings.py
```diff
@@ -3,11 +3,24 @@
...
+        "Use global Assignation Limit",
+        config_parameter="s_scheduler_assignation_horizon.global_moves_assignation",
+        readonly=False,
+        help="Check this box to allow limit settings " "for each company independent",
+    )
+    global_moves_assignation_horizon = fields.Integer(
```
simahawk:
```suggestion
    stock_horizon_move_assignation_limit = fields.Integer(
```

stock_scheduler_assignation_horizon/readme/DESCRIPTION.rst
```diff
@@ -1 +1,4 @@
-This module allows to set a time horizon that will prevent the scheduler from marking stock pickings as *ready* before the indicated number of days (horizon).
+This module allows to set a time horizon that will prevent the scheduler from marking stock
+pickings as *ready* before the indicated number of days (horizon).
+Options can be enabled globally and for each company separately, remember that to enable
+mode for the company scheduler needs to be modified to set the company correctly.
```
simahawk:
```suggestion

This behavior can be enabled globally via settings or for each company separately.
```
Can you please add a CONFIGURE.rst?

simahawk (changes requested):

stock_scheduler_assignation_horizon/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+
+    <record id="global_is_moves_assignation_limited" model="ir.config_parameter">
+        <field
+            name="key"
+        >s_scheduler_assignation_horizon.stock_horizon_move_assignation</field>
```
ivantodorovich:
```suggestion
        >stock_scheduler_assignation_horizon.stock_horizon_move_assignation</field>
```
IMO better not to abbreviate this. It's for the users

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -38,3 +44,13 @@ def _get_moves_to_assign_domain_assignation_limited(self, domain, company):
...
+
+    def get_global_is_moves_assignation_limited(self):
+        get_param = self.env["ir.config_parameter"].sudo().get_param
+        param = "s_scheduler_assignation_horizon.stock_horizon_move_assignation"
+        return str2bool(get_param(param, False))
+
+    def get_global_moves_limit_horizon(self):
+        get_param = self.env["ir.config_parameter"].sudo().get_param
+        param = "s_scheduler_assignation_horizon.stock_horizon_move_assignation_limit"
+        return safe_eval(get_param(param, 0))
```
ivantodorovich: If you expect an int, use `int(get_param(param, 0))`.

Maybe:

```py
        try:
            limit = int(get_param(param, 7))
        except ValueError:
            _logger.error("Wrong value for config parameter: %s", param)
            limit = 0
        return limit
```

grindtildeath: Nitpicking, but IMO the logger mentioned by @ivantodorovich should use a warning level and show it will use 0. e.g
```
        try:
            param_value = get_param(param, 0)
            limit = int(param_value)
        except ValueError:
            limit = 0
            _logger.warning("Wrong value for ir.config_parameter %s: %s . Using %s as default value", % (param, param_value, limit))
        return limit
```

grindtildeath:
Instead of adding another parameter at instance level, I'm wondering if we shouldn't have added this configuration on the picking type as is the case in Odoo 15... :thinking: 

Anyway IMO we better keep only the new setting as `ir.config_parameter` and remove the old one to avoid having a setting at company that doesn't do what is expected. (and since the `procurement_jit` module is not company dependent neither, I'm fine with this choice).

For this we still need a migration script though.

stock_scheduler_assignation_horizon/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Scheduler assignation horizon",
     "summary": "Set a timeframe limit to the delivery scheduler",
-    "version": "14.0.1.0.0",
+    "version": "14.0.1.0.1",
```
simahawk, line 6: can you set the dev status as alpha pls?

stock_scheduler_assignation_horizon/migrations/14.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    if column_exists(cr, "res_company", "is_moves_assignation_limited"):
+        query = "SELECT is_moves_assignation_limited FROM res_company;"
+        cr.execute(query)
+        res = cr.fetchall()
+
+        create_param = [True for x in res if x[0] is True]
```
simahawk, line 15: Can't you simply select one where is True?

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -2,31 +2,33 @@
...
     _inherit = "procurement.group"
 
     def _get_moves_to_assign_domain(self, company_id):
         domain = super()._get_moves_to_assign_domain(company_id)
-        company = self.env["res.company"].browse(company_id)
-        if company.is_moves_assignation_limited:
-            domain = self._get_moves_to_assign_domain_assignation_limited(
-                domain, company
-            )
+        global_limit = self.get_global_is_moves_assignation_limited()
```
simahawk: Pass the company. You keep the hook for whoever wants to customize the behavior by company.

simahawk (commented):
Minor remarks.

@grindtildeath @ivan can you update your reviews?

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -38,3 +40,24 @@ def _get_moves_to_assign_domain_assignation_limited(self, domain, company):
                 [("date", "<=", max_date)],
             ]
         )
+
+    def get_global_is_moves_assignation_limited(self):
```
ivantodorovich:
```suggestion
    def _get_global_is_moves_assignation_limited(self):
```
Should be private

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -38,3 +40,24 @@ def _get_moves_to_assign_domain_assignation_limited(self, domain, company):
...
+        get_param = self.env["ir.config_parameter"].sudo().get_param
+        param = "stock_scheduler_assignation_horizon.stock_horizon_move_assignation"
+        return str2bool(get_param(param, False))
+
+    def get_global_moves_limit_horizon(self):
```
ivantodorovich:
```suggestion
    def _get_global_moves_limit_horizon(self):
```
Should be private

stock_scheduler_assignation_horizon/migrations/14.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2022 Camtocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import SUPERUSER_ID, api
+from odoo.tools.sql import column_exists
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    if column_exists(cr, "res_company", "is_moves_assignation_limited"):
```
ivantodorovich, line 10: You can count on the column to exist, because it did before this new module version.
If the migration is executed, it means you're coming from a previous version: the column definitely exists

In any case, it doesn't hurt to check I guess.

I would avoid the nested `if` indentation by reverting your conditions.

```suggestion
    if not column_exists(cr, "res_company", "is_moves_assignation_limited"):
        return
    # ...
```

Alternatively, since the field already exist.. I think you can simply use the env rather than sql queries

```py
    if "is_moves_assignation_limited" not in env["res.company"]._fields:
        return
    companies = env["res.company"].search([])

    stock_horizon_move_assignation = any(c.is_moves_assignation_limited for c in companies)
    if not stock_horizon_move_assignation:
        return
    _create_ir_config_param(
        "stock_scheduler_assignation_horizon.stock_horizon_move_assignation",
        True,
        "stock_scheduler_assignation_horizon.is_moves_assignation_limited",
    )

    stock_horizon_move_assignation_limit = min(
        c.moves_assignation_horizon for c in companies
        if c.moves_assignation_horizon,
        default=0
    )
    if not stock_horizon_move_assignation_limit:
        return
    _create_ir_config_param(
       "stock_scheduler_assignation_horizon.stock_horizon_move_assignation_limit",
        stock_horizon_move_assignation_limit,
        "stock_scheduler_assignation_horizon.stock_scheduler_assignation_horizon",
    )

def _create_ir_config_param(key, value, xmlid):
    # ...
```

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -2,31 +2,33 @@
...
+        global_limit = self.get_global_is_moves_assignation_limited()
+        if global_limit:
+            delay = self.get_global_moves_limit_horizon()
+            domain = self._get_moves_to_assign_domain_assignation_limited(domain, delay)
         return domain
 
-    def _get_moves_to_assign_domain_assignation_limited(self, domain, company):
-        company_tz = pytz.timezone(company.partner_id.tz or "UTC")
+    def _get_moves_to_assign_domain_assignation_limited(self, domain, delay_days):
+        applied_tz = pytz.timezone("UTC")
```
grindtildeath: IMO no need to use a timezone to convert from UTC to UTC :upside_down_face: 

I would say we can define the max_date as `datetime.now() + timedelta(days=delay_days)` as you did, without  having to combine with max time.

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -2,31 +2,33 @@
...
         domain = super()._get_moves_to_assign_domain(company_id)
-        company = self.env["res.company"].browse(company_id)
-        if company.is_moves_assignation_limited:
-            domain = self._get_moves_to_assign_domain_assignation_limited(
-                domain, company
-            )
+        global_limit = self.get_global_is_moves_assignation_limited()
+        if global_limit:
+            delay = self.get_global_moves_limit_horizon()
+            domain = self._get_moves_to_assign_domain_assignation_limited(domain, delay)
```
grindtildeath: While we're in a big refactor, could we combine the domain here instead of inside the function? It would allow for easier override doing this:
```
horizon_domain = self._get_assignation_horizon_domain(delay)
domain = AND([domain, horizon_domain])
```

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -2,31 +2,33 @@
...
         return domain
 
-    def _get_moves_to_assign_domain_assignation_limited(self, domain, company):
-        company_tz = pytz.timezone(company.partner_id.tz or "UTC")
+    def _get_moves_to_assign_domain_assignation_limited(self, domain, delay_days):
```
grindtildeath:
```suggestion
    def _get_assignation_horizon_domain(self, delay_days):
```
IMO we can use a shorter function name :sweat_smile:

stock_scheduler_assignation_horizon/models/procurement_group.py
grindtildeath: Since we don't have any company settings anymore, a shorter version as `_is_assignation_horizon_limited` would also be nice :innocent:

stock_scheduler_assignation_horizon/models/procurement_group.py
grindtildeath: `_get_assignation_horizon_limit` ? :innocent:

grindtildeath (commented):
Small things, mostly naming of things that can be improved to make the code more readable.

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -2,39 +2,48 @@
...
-        return expression.AND(
-            [
-                domain,
-                [("date", "<=", max_date)],
-            ]
+    def _get_assignation_horizon_domain(self, delay_days):
+        max_date = datetime.now() + timedelta(days=delay_days)
+        return [("date", "<=", max_date)]
+
+    def _get_global_is_moves_assignation_limited(self):
```
grindtildeath:
```suggestion
    def _is_assignation_horizon_limited(self):
```
As asked in previous comment :wink:

stock_scheduler_assignation_horizon/models/procurement_group.py
```diff
@@ -2,39 +2,48 @@
...
     _inherit = "procurement.group"
 
     def _get_moves_to_assign_domain(self, company_id):
         domain = super()._get_moves_to_assign_domain(company_id)
-        company = self.env["res.company"].browse(company_id)
-        if company.is_moves_assignation_limited:
-            domain = self._get_moves_to_assign_domain_assignation_limited(
-                domain, company
-            )
+        global_limit = self._get_global_is_moves_assignation_limited()
```
grindtildeath:
```suggestion
        global_limit = self._is_assignation_horizon_limited()
```

grindtildeath (approved):
Can you please check the comments on the migration script? Pre approving anyway.

grindtildeath:
@simahawk Can you please update your review and remove stale label?

rousseldenis (approved):
Code review

simahawk (approved):

PR #1134, opened 2022-11-16
[15.0][FIX] Update copier template + deprecated method

pedrobaeza:
- Avoid the Flake8 Gitlab problem.
- Remove deprecated method api.Environment.manage().

@Tecnativa

### DIFF
--- a/stock_return_request/hooks.py
+++ b/stock_return_request/hooks.py
@@ -41,43 +41,42 @@ def pre_init_hook(cr):
 
 def post_init_hook(cr, registry):
     """Set moves returnable qty on hand"""
-    with api.Environment.manage():
-        env = api.Environment(cr, SUPERUSER_ID, {})
-        moves_draft = env["stock.move"].search([("state", "in", ["draft", "cancel"])])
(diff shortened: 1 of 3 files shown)


stock_return_request/hooks.py
```diff
@@ -41,7 +41,7 @@ def pre_init_hook(cr):
 
 def post_init_hook(cr, registry):
     """Set moves returnable qty on hand"""
-    with api.Environment.manage():
+    if SUPERUSER_ID:  # Fake if for not having a big diff
```
rousseldenis: @pedrobaeza Is the condition justified for a 'diff' problem ?

pedrobaeza: I think so for avoiding 40 lines of diff (the + and -).

rousseldenis: Yes, but what that matters is the result. And the condition is unwanted

pedrobaeza: OK, as you prefer

rousseldenis (approved):

PR #1128, opened 2022-11-10
[13.0][IMP] stock_picking_invoice_link: Add support to order invoice policy

CarlosRoca13:
cc @Tecnativa 

ping @pedrobaeza @sergio-teruel

PR #1140, opened 2022-11-25
[13.0][FIX] stock_picking_invoice_link: inventory users cannot validate picking

augusto-weiss:
Due to #1128 an user with only inventory permissions cannot validate a picking

Steps to reproduce error:
- Create a sale order with a product (should set invoice policy: ordered quantities)
- Confirm the order and invoiced it
- With a user that only have "User" permission on Inventory, try to validate the transfer.


pedrobaeza (approved):

CarlosRoca13 (approved):
Thanks!

pedrobaeza:
/ocabot merge patch

This has to be fw-port it to upper versions.

PR #1143, opened 2022-11-30
[MIG] stock_landed_costs_currency: Migration to 15.0

CoraVanga:


stock_landed_costs_currency/models/stock_landed_cost.py
```diff
@@ -9,13 +9,19 @@ class LandedCost(models.Model):
...
+        related="company_id.currency_id",
+    )
+    amount_total = fields.Monetary(
+        currency_field="company_currency_id",
+    )
 
     @api.onchange("account_journal_id")
     def _onchange_account_journal_id(self):
-        if self.account_journal_id and self.account_journal_id.currency_id:
+        if self.account_journal_id.currency_id:
```
rousseldenis, line 24: @DinhTien24 Shouldn't the currency_id field changed into computed readonly False one and the onchange function converted in compute one?

PR #1006, opened 2022-05-24
[MIG] stock_picking_invoice_link: Migration to 15.0

JoelZilli:

PR #1147, opened 2022-12-02
[15.0][MIG] stock_picking_invoice_link: Migration to 15.0

CarlosRoca13:
cc @Tecativa TT35793

This PR supersed https://github.com/OCA/stock-logistics-workflow/pull/1006
I just add the FIX for the migration done by @JoelZilli

please review this @pedrobaeza @sergio-teruel


stock_picking_invoice_link/models/account_move.py
```diff
@@ -75,7 +75,6 @@ class AccountMoveLine(models.Model):
         column2="move_id",
         string="Related Stock Moves",
         readonly=True,
-        copy=False,
```
pedrobaeza: @CarlosRoca13 why are you removing this? A copied invoice shouldn't be linked to the same pickings/moves. If you need it for reversal operations, it have to be re-added in their prepare methods.

pedrobaeza (changes requested):

CarlosRoca13:
Yes, you are right, I overwrite copy_data method instead, to keep the link just when we make a refund of type modify. Just because odoo use this data for create the new invoice: https://github.com/odoo/odoo/blob/8986330a14000da7f88d2787ff54965d71ca02b7/addons/account/wizard/account_move_reversal.py#L139

pedrobaeza (approved):

pedrobaeza:
As the other PR is totally approved and this one fixes the remaining thing, I'm merging:

/ocabot merge nobump

pedrobaeza:
@CarlosRoca13 on v14 the `copy=False` is also removed. Did you modify it the same way?

CarlosRoca13:
> @CarlosRoca13 on v14 the `copy=False` is also removed. Did you modify it the same way?

I think that it was added at 15.0 migration...

pedrobaeza:
OK, but anyway it should be needed for v14 for avoiding the same effect. Please backport it.

PR #1150, opened 2022-12-07
[14.0][ADD] stock_picking_default_planned_transfer

Kev-Roche:
Small module to set Planned Transfer as Default from Inventory Views
cc @hparfr

### DIFF
--- a/stock_picking_default_planned_transfer/hooks.py
+++ b/stock_picking_default_planned_transfer/hooks.py
@@ -1,0 +1,11 @@
+# Copyright (C) 2022 Akretion (<http://www.akretion.com>).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID
+from odoo.api import Environment
+
+
+def no_default_immediate_tranfer_create(cr, registry):
+    env = Environment(cr, SUPERUSER_ID, {})
+    env["ir.config_parameter"].set_param("stock.no_default_immediate_tranfer", "True")
--- a/setup/stock_picking_default_planned_transfer/odoo/addons/stock_picking_default_planned_transfer
+++ b/setup/stock_picking_default_planned_transfer/odoo/addons/stock_picking_default_planned_transfer
@@ -1,0 +1,1 @@
+../../../../stock_picking_default_planned_transfer
--- a/setup/stock_picking_default_planned_transfer/setup.py
+++ b/setup/stock_picking_default_planned_transfer/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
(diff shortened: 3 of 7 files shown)


rousseldenis:
Hi @Kev-Roche. Thanks for this.

IMHO, changing a default behaviour should be configurable (that's why existing tests are failing).

Why not adding a configuration entry in stock settings for that ?

hparfr:
>IMHO, changing a default behaviour should be configurable (that's why existing tests are failing).

It's a bit difficult. 

This module only sets a parameter defined by odoo ! 

```py
    def _get_action(self, action_xmlid):
        action = self.env["ir.actions.actions"]._for_xml_id(action_xmlid)
        if self:
            action['display_name'] = self.display_name

        default_immediate_tranfer = True
        if self.env['ir.config_parameter'].sudo().get_param('stock.no_default_immediate_tranfer'):
            default_immediate_tranfer = False
```
https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_picking.py#L237

Should the module with the failing test work in this legit setup ? May be. I didn't read yet the failing case.
 
If someone want to deactivate this feature, he just has to toggle the ir.parameter to false.

stock_picking_default_planned_transfer/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+    )
+    if config_parameter and config_parameter.value != "True":
+        config_parameter.write({"value": "True"})
+    else:
+        env["ir.config_parameter"].create(
+            {
+                "key": "stock.no_default_immediate_tranfer",
+                "value": "True",
+            }
+        )
```
petrus-v: I think you could make the code easier using [set_param](https://github.com/OCA/OCB/blob/16.0/odoo/addons/base/models/ir_config_parameter.py#L80)

```suggestion
    env["ir.config_parameter"].set_param("stock.no_default_immediate_tranfer", "True")
```

Kev-Roche: @petrus-v , I m not sure because stock.no_default_immediate_tranfer is called in Odoo native but not existing, we must create it IMO

rousseldenis: @Kev-Roche That's exactly what Odoo does: https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_config_parameter.py#L95

Kev-Roche: ho, did not know that! So, yes @petrus-v change is better. Thank you @rousseldenis

florian-dacosta (approved):
Not sure why the test of the other module were failing since it is all green now.
But I think this module is ok, as it basically change a param which has impact of some views but don't really change logic in Odoo

LGTM

hparfr (approved):

rousseldenis (approved):

PR #1158, opened 2022-12-15
[MIG][16.0] delivery_procurement_group_carrier

rousseldenis:
I'm not sure all code should remain in this module (sale action_draft() and carrier_id on procurement_group), so, I'm not sure about module name.


delivery_procurement_group_carrier/models/sale_order_line.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Camptocamp (https://www.camptocamp.com)
+# Copyright 2020-2021 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
lmignon: @rousseldenis `s/http/https`

delivery_procurement_group_carrier/tests/test_carrier.py
```diff
@@ -1,45 +1,48 @@
...
-        so = self.env["sale.order"].create(sale_order_vals)
-        so.action_confirm()
-        self.assertTrue(so.picking_ids)
-        self.assertEqual(so.procurement_group_id.carrier_id, so.carrier_id)
+
+        with Form(self.env["sale.order"]) as sale_form:
+            sale_form.partner_id = self.partner
+            with sale_form.order_line.new() as line_form:
+                line_form.product_id = product
+        sale = sale_form.save()
```
lmignon: @rousseldenis You should no more use `Form` to create a SO. Odoo core has been modified to use computed non readonly fields for required fields.

lmignon (changes requested):

delivery_procurement_group_carrier/models/sale_order_line.py
rousseldenis: @lmignon Done

lmignon (approved):
LGTM (Code + functionnal)

rousseldenis:
@jbaudoux

petrus-v (approved):
LGTM (code review)

jbaudoux (approved):

jbaudoux:
> I'm not sure all code should remain in this module (sale action_draft() and carrier_id on procurement_group), so, I'm not sure about module name.

When the SO is put back to draft, you can change the carrier. That's why the procurement group is removed as it could be invalid after confirm.

sebalix (approved):

PR #1159, opened 2022-12-15
[16][stock_restrict_lot] Add Restrict lot in move list views

florian-dacosta:
Since the field restricts the reservation, I guess it is good to be displayed to help the user understand what's going on.
It is still optional though.
Also added a way to search picking depending on its lots which can be really helpfull when using this module

### DIFF
--- a/stock_restrict_lot/models/stock_picking.py
+++ b/stock_restrict_lot/models/stock_picking.py
@@ -1,0 +1,12 @@
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    # this is a technical field, making possible to search a picking by any restrict lot
+    # defined on its line, from the web search bar.
+    restrict_lot_id = fields.Many2one(
+        related="move_ids.restrict_lot_id", string="Restrict Lot"
+    )
--- a/stock_restrict_lot/views/stock_picking.xml
+++ b/stock_restrict_lot/views/stock_picking.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+        <record id="custom_main_stock_picking_form_view" model="ir.ui.view">
+            <field name="model">stock.picking</field>
+            <field name="inherit_id" ref="stock.view_picking_form" />
+            <field name="arch" type="xml">
+                <xpath
+                expr="//field[@name='move_ids_without_package']/tree/field[@name='product_id']"
+                position="after"
+            >
+                    <field
+                    name="restrict_lot_id"
+                    readonly="1"
+                    optional="show"
+                    groups="stock.group_production_lot"
+                />
+                </xpath>
+            </field>
+        </record>
+
+        <record id="custom_stock_picking_search_view" model="ir.ui.view">
+            <field name="model">stock.picking</field>
+            <field name="inherit_id" ref="stock.view_picking_internal_search" />
+            <field name="arch" type="xml">
+                <field name="product_id" position="after">
+                    <field
+                    name="restrict_lot_id"
+                    string="Lot"
+                    groups="stock.group_production_lot"
+                />
+                </field>
+            </field>
+        </record>
+
+</odoo>
--- a/stock_restrict_lot/__manifest__.py
+++ b/stock_restrict_lot/__manifest__.py
@@ -9,4 +9,5 @@
     "license": "AGPL-3",
     "installable": True,
     "depends": ["stock"],
+    "data": ["views/stock_move_views.xml", "views/stock_picking.xml"],
 }
--- a/stock_restrict_lot/models/__init__.py
+++ b/stock_restrict_lot/models/__init__.py
@@ -1,2 +1,3 @@
 from . import stock_move
 from . import stock_rule
+from . import stock_picking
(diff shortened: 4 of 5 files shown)


florian-dacosta:
Small improvment to get the field in view (optional)
Review appreciated @petrus-v @sebalix @newtratip @bizzappdev

bizzappdev:
IMO searching on restrict_lot_id is a nice idea. but having it on list view will confuse people as there might be multiple lines and in the case of related fields it will display only the first one. which may lead to confusion.
so my suggestion is either to change the caption to First Restrict Lot or just limit that fields to the search level only or have a computed method which displays all or something like that.
@florian-dacosta

stock_restrict_lot/models/stock_picking.py
```diff
@@ -0,0 +1,10 @@
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    restrict_lot_id = fields.Many2one(
```
petrus-v, line 10: Shouldn't be a o2m so called `restrict_lot_ids` ?

petrus-v (commented):
That interesting, I would probably hide the field by default. Not sure displayed all restricted lot are really meaningful with large picking but like the feature of searching picking concerned by a given lot.

florian-dacosta:
@bizzappdev 
No, actually, if a picking contains multiple lines with each one a different restrict lot, you can search and find the picking searching with any of the lot, not just the first one. 
There are already filters of this kind in Odoo, it is like searching a sale order with a product, etc
I use it like this and it works fine.

stock_restrict_lot/models/stock_picking.py
florian-dacosta: The only purpose is to be able to search, and it works fine this way.
Here is the same kind of stuff : https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_picking.py#L410

petrus-v: Even odoo is doing that, i would prefer a proper definition if it works as well, if some one else (a developer) is reusing this related field it makes hard to expect it returns a recordset instead a record.

bizzappdev:
> @bizzappdev No, actually, if a picking contains multiple lines with each one a different restrict lot, you can search and find the picking searching with any of the lot, not just the first one. There are already filters of this kind in Odoo, it is like searching a sale order with a product, etc I use it like this and it works fine.

Search and filter will work perfectly. I was just concerned about showing this field in the list and form view. as it will represent partial information

florian-dacosta:
> I was just concerned about showing this field in the list and form view. as it will represent partial information

@bizzappdev But it is not on picking form/tree view, it is on `stock.move` tree/form view.
The override of picking form view is in order to be able to include it in the "operations" tab, where we can see the s`tock.move`

stock_restrict_lot/models/stock_picking.py
florian-dacosta: I usually avoid doing stuff differently than Odoo SA when it works well to avoid unexpected issues.
I could add a comment to indicates that this is a technical field which purpose is to be able to search a picking by restrict lot.
Actually I don't even know how it works technically for the search to work with this m2o not stored field.

Anyway, I just made a test with a compute o2m (the goal is to keep a not stored field since we don't need this field for any other purpose that the search, at least for now) and the filter does not work.

`2023-03-08 09:31:30,426 165 ERROR db odoo.osv.expression: Non-stored field stock.picking.restrict_lot_ids cannot be searched.`

I think it is reasonable to go with the current solution for a start, do you agree?

petrus-v: ok then, good to know thanks for testing that, leaving a comment seems a good idea !

petrus-v (approved):

bizzappdev:
> > I was just concerned about showing this field in the list and form view. as it will represent partial information
> 
> @bizzappdev But it is not on picking form/tree view, it is on `stock.move` tree/form view. The override of picking form view is in order to be able to include it in the "operations" tab, where we can see the s`tock.move`

Ahh My bad.

bizzappdev (approved):
LGTM

stock_restrict_lot/models/stock_picking.py
sebalix: I agree, leaving a comment (or "help") about the role of this field is very welcome.

stock_restrict_lot/views/stock_picking.xml
```diff
@@ -0,0 +1,36 @@
...
+                <xpath
+                expr="//field[@name='move_ids_without_package']/tree/field[@name='product_id']"
+                position="after"
+            >
+                    <field
+                    name="restrict_lot_id"
+                    readonly="1"
+                    optional="show"
+                    groups="stock.group_production_lot"
+                />
```
sebalix, line 17: Could we use `column_invisible` in `attrs` to display this field only if `parent.restrict_lot_id` is defined?
Having it displayed on all stock transfers is probably useless.

florian-dacosta: @sebalix It does not seem to work, not sure why.

florian-dacosta: >  agree, leaving a comment (or "help") about the role of this field is very welcome.

Done

sebalix (approved):

rousseldenis:
@florian-dacosta Could you rebase ?

florian-dacosta:
@rousseldenis Done

PR #1160, opened 2022-12-15
[16.0] Add module stock_quant_package_move_wizard

alexis-via:
This is a re-write of the module "stock_quant_packages_moving_wizard" from odoomrp https://github.com/akretion/odoomrp-wip/tree/10-mig-quant_moving_wiz/stock_quant_packages_moving_wizard

This is a DRAFT PR. It is working well to move a quant without package.
Left to do: add ability to move a package.

### DIFF
--- a/setup/stock_quant_package_move_wizard/odoo/addons/stock_quant_package_move_wizard
+++ b/setup/stock_quant_package_move_wizard/odoo/addons/stock_quant_package_move_wizard
@@ -1,0 +1,1 @@
+../../../../stock_quant_package_move_wizard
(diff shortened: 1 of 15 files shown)


alexis-via:
It has several improvements compared to the version of odoomrp:

- if you select a picking type in the wizard, it will create a picking with all the stock moves
- ability to move a fraction of the quant, not just the full quant (by default, it moves the full quant)

rousseldenis:
@alexis-via Have you tried https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_move_location ?

As IMHO, those two modules cover the same functionality

alexis-via:
@rousseldenis you are certainly right ! But I'll continue to maintain this module, because it's much easier and faster. So I'll maintain outside OCA. It is now available here https://github.com/akretion/odoo-usability/tree/16.0/stock_quant_package_move_wizard (with FR translation !)

PR #1162, opened 2022-12-19
[14.0][ADD] stock_lot_production_date

sebalix:
This module adds the possibility to set a *Production Date* on lots/serial numbers, similar to `product_expiry` with its *Expiration Date*.

### Lot form
lot


stock_lot_production_date/models/stock_move_line.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def _assign_production_lot(self, lot):
+        super()._assign_production_lot(lot)
+        if lot.product_id.use_production_date:
+            lot.production_date = fields.Datetime.now()
```
sebalix: I still don't know if this auto-assignment should be done only when validating a MO or not.
With this implementation, the production date will be set even when we create a lot through a normal stock transfer.

sebalix: This part has been removed and will be handled by https://github.com/OCA/manufacture/pull/934

stock_lot_production_date/models/product_template.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    use_production_date = fields.Boolean(
+        string="Production Date",
+        help=(
+            "When this box is ticked, the date of production "
+            "will be set automatically on lot/serial numbers."
```
grindtildeath: AFAICS nothing is going to set the production date automatically in this module. IMO as the module depends only on `stock` it could either provide the field `production_date` on the lot without anything else, or make the field `production_date` required for lots of such products that would be received from a supplier. (but then it would need to be displayed on the stock move line when the picking type has `use_create_lots` set to True, and it needs to be "propagated" when the `stock.production.lot` is created)

sebalix: Yes, I forgot to update this tooltip now we want this module to display only this new field.
Regarding the reception workflow to invite the user to set this production date could be an improvement.

sebalix: I removed the option on `product.template` to get only the field available on the lot.

rousseldenis (approved):
LGTM. Great addition. Wondering why nobody thought about it as it is common...

ivantodorovich (approved):

simahawk (approved):

PR #1164, opened 2022-12-20
[REF] stock_picking_package_preparation: Website link

SirTakobi:
In the last commit I updated the repo template (see https://github.com/OCA/oca-addons-repo-template) because it was failing.


SirTakobi:
@OCA/logistics-maintainers can someone please have a look?
It should be quick, this is only fixing a broken link and updating CI configuration.
If you want me to split the commits in many PRs it is not a problem.

rousseldenis (approved):

PR #1167, opened 2022-12-21
[FIX] stock_picking_invoice_link: Fix Invoice Group on Stock.Picking

joshkreud:
This PR adds the accounting group to the `stock.picking` form view. 

Without the group attribute you can get an ACL error when trying to access a picking with linked invoices without having access to the invoices.

### DIFF
--- a/stock_picking_invoice_link/views/stock_view.xml
+++ b/stock_picking_invoice_link/views/stock_view.xml
@@ -5,9 +5,13 @@
         <field name="name">stock_picking_invoice_link.stock.picking.form</field>
         <field name="model">stock.picking</field>
         <field name="inherit_id" ref="stock.view_picking_form" />
+        <field
+            name="groups_id"
+            eval="[(6, 0, [ref('account.group_account_invoice')])]"
+        />
         <field name="arch" type="xml">
             <div name="button_box" position="inside">
-                <field name="invoice_ids" invisible="1" />
+              <field name="invoice_ids" invisible="1" />
                 <button
                     name="action_view_invoice"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You are not preventing it at field level. Please include it at view level.

pedrobaeza:
It isn't still at view level...

joshkreud:
Heya, 

Yeah I first fixed pre-commit. Hadn't installed it previously in this Repo. 

What do you mean by view level in this case? I changed it in the View XML. 

I just tried to also add the groups  in the Model but that leads to an error, because then odoo completely removes `invoice_ids` from the view: 

```
Error: While parsing modifiers for button: for modifier "invisible": Unknown field invoice_ids in domain
```

If i just add the `groups` attribute on the view level in XML, it works as expected. 

On another note: The failing CI test. 
That error comes from another module. 
So i guess we have to wait until this is fixed in the 14.0 Branch right?

joshkreud:
Ooh I think I just figured out what you mean by view level. 
Still a bit of testing to do, but i'm on it! 

Converted the PR to draft for the time being.

joshkreud:
Okay I reviewed it a bit more and am now closing this PR. 
Even though, we lock stock users out of some Pickings, when there are invoices attached, the Default ACLs do sometimes permit  Users without the accounting group to view certain Invoices. 

If we would add the `account.group_account_invoice` group to the view, we would hide Invoices even though the user could sometimes have access to some. 

It really depends on the ACL setup people are using. 

I don't think imposing the Accounting group would be right here.

PR #1168, opened 2022-12-21
[15.0][MIG] stock_move_change_source_location

MiquelRForgeFlow:
Standard migration.


stock_move_change_source_location/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows you to change the source location of stock moves from the picking
```
rousseldenis, line 1: @MiquelRForgeFlow Could you explain a little your use case ?

MiquelRForgeFlow: In https://github.com/OCA/stock-logistics-workflow/pull/868, you found this module great hehe

A client asked us to be able to change the location where you get the stock in a delivery, not wanting the default one coming from the rule. And as we migrate this client to newer versions, we still need to migrate this module.

AaronHForgeFlow (approved):
Functional tests

stock_move_change_source_location/readme/DESCRIPTION.rst
rousseldenis: Ok. I wanted to know because some things have been done in wms if that correspond to your flow (it seems at first sight).

In some situations, it could be great not to generate all stock moves (mainly the internal ones). Then, at a point of time, release the stock flow.

e.g.: Generate only the OUT > Customer move after sale order confirmation.

I suppose this is the case you want to solve when you are not sure at sale confirmation from where should come the products.

https://github.com/OCA/wms/pull/506

setup/stock_move_change_source_location/odoo/addons/stock_move_change_source_location
```diff
@@ -0,0 +1 @@
+../../../../stock_move_change_source_location
```
rousseldenis, line 1: For next time, run pre-commit before doing the migration commit

stock_move_change_source_location/wizards/stock_move_change_source_location_wizard.py
```diff
@@ -0,0 +1,139 @@
...
+        warehouse = picking.location_id.warehouse_id
+        return [
+            ("id", "child_of", warehouse.view_location_id.id),
+            ("usage", "=", "internal"),
+        ]
+
+    def _get_allowed_states(self):
+        return ["waiting", "partially_available", "confirmed", "assigned"]
+
+    warehouse_view_location_id = fields.Many2one(
```
rousseldenis, line 32: Could you put fields on top ?

rousseldenis (changes requested):
Minor changes

setup/stock_move_change_source_location/odoo/addons/stock_move_change_source_location
MiquelRForgeFlow: My philosophy is that if the pre-commit commit will not have any changes (except the setup folder thing) then let's go directly to the migration commit to avoid adding extra commits hehe

stock_move_change_source_location/wizards/stock_move_change_source_location_wizard.py
MiquelRForgeFlow: Done

stock_move_change_source_location/tests/test_stock_move_change_source_location.py
```diff
@@ -66,33 +71,23 @@ def setUpClass(cls):
...
+                            "location_dest_id": cls.warehouse.wh_output_stock_loc_id.id,
                         },
                     ),
                 ],
             }
         )
 
     def qty_on_hand(self, product):
-        wiz = self.env["stock.inventory"].create(
+        self.env["stock.quant"].create(
```
rousseldenis: Better to add 'inventory_mode=True' in context.

See: https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_quant.py#L177

MiquelRForgeFlow: Done.

MiquelRForgeFlow: With this change, the tests are broken!

rousseldenis (approved):
Code review.

Apart of that, I'm wondering the flow you have.

As I suppose the aim of this is to change the source location when the move has been created too early. So, the whole stock state can have changed before beginning the move operation.

FYI, there is a set of modules around the stock release channel flows: https://github.com/OCA/wms/tree/16.0/stock_release_channel

It allows to accumulate pickings in a particular channel (e.g. output -> customers for a carrier X) from sales confirmations, then, when the stock manager decides to release (we are near to the deliver time), all stock operations (pickings) are created and take into account the stock state at that time.

stock_move_change_source_location/tests/test_stock_move_change_source_location.py
MiquelRForgeFlow: Solved.

rousseldenis: Mmmh, one of the field is not in the allowed ones. 

https://github.com/odoo/odoo/blob/6c7385c74fafbe84042261e29565a77e2ca0c03c/addons/stock/models/stock_quant.py#L803

Weird

rousseldenis: Ok, you've got it

MiquelRForgeFlow:
It's green 🍏

PR #1171, opened 2022-12-22
[14.0][IMP] stock_picking_progress: compute progress for moves

JuMiSanAr:
This improvement makes the "progress" computed field also available for stock moves.

### DIFF
--- a/stock_picking_progress/hooks.py
+++ b/stock_picking_progress/hooks.py
@@ -3,12 +3,14 @@
 
 import logging
 
+from odoo import SUPERUSER_ID, api
 from odoo.tools.sql import column_exists, create_column
 
 logger = logging.getLogger(__name__)
 
 
 def setup_move_line_progress(cr):
+    """Update the 'progress' column for move lines."""
     table = "stock_move_line"
     column = "progress"
     _type = "float"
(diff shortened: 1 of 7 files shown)


stock_picking_progress/hooks.py
```diff
@@ -29,6 +31,32 @@ def setup_move_line_progress(cr):
     cr.execute(fill_column_query)
 
 
+def setup_move_progress(cr):
```
sebalix, line 34:
```suggestion
def setup_move_progress(cr):
    """Update 'progress' column for not-started or already done moves."""
```

stock_picking_progress/hooks.py
```diff
@@ -29,6 +31,32 @@ def setup_move_line_progress(cr):
...
+            WHEN (state = 'done') THEN 100
+            WHEN (state in ('draft', 'confirmed')) THEN 0.0
+            WHEN (product_uom_qty IS NULL or product_uom_qty = 0.0) THEN 0.0
+        END;
+    """
+    logger.info("filling up %s on %s", column, table)
+    cr.execute(fill_column_query)
+
+
+def setup_move_progress_post(env):
```
sebalix, line 57:
```suggestion
def setup_move_progress_post(env):
    """Update 'progress' column for moves in progress.

    Need to be processed in post as the computation depends on the computed `quantity_done` field.
    """
```

sebalix (commented):
`pre-commit` to run to fix the builb + small remarks

sebalix (approved):

stock_picking_progress/models/stock_move.py
```diff
@@ -0,0 +1,31 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+from odoo.tools.float_utils import float_is_zero
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    progress = fields.Float(
```
sebalix, line 11: That makes me rethink about what we said regarding the computation of the `stock.picking.progress` field: IMO it should be computed from `stock.move.progress` instead of `stock.move.line.progress` so we include qty that can't be reserved (no `stock.move.line` created).

Should it be a drawback for Shopfloor or not?

sebalix: My take: `picking.progress` field should be computed from the moves if its delivery policy is "all at once", and from move lines if the policy is "as soon as possible". That way we got results that the user could expect.
ping @jbaudoux 
I won't block for now as this PR doesn't change the computation at the picking level but it's something we need to check later.

PR #1172, opened 2022-12-26
[12.0] [ADD] `stock_production_lot_traceability`

ivantodorovich:
**Description**
Easily see which Lots/Serial Numbers were consumed or produced by/from another one.

**Usage**
Go to the form view of any Lot/Serial Number that has either been consumed or that has produced other Lots/Serial Numbers, and click on the "Consumed" or "Produced" smart buttons to navigate to these records. It will show all records, directly or indirectly consumed or produced by it.


damdam-s (approved):

jcoux (approved):
Technically seems good :+1:

rousseldenis (approved):

PR #1178, opened 2023-01-03
[14.0][IMP] stock_picking_progress: calculate picking progress from moves

JuMiSanAr:
With this PR:

- The picking progress is calculated from the moves. This is done to avoid issues with transfers which include moves at different states, as the lines won't be available for some.
- The progress is set to 100 (instead of 0) if there is nothing left to do.

### DIFF
--- a/stock_picking_progress/hooks.py
+++ b/stock_picking_progress/hooks.py
@@ -23,7 +23,7 @@ def setup_move_line_progress(cr):
         UPDATE stock_move_line
         SET progress = CASE
             WHEN (state = 'done') THEN 100
-            WHEN (product_uom_qty IS NULL or product_uom_qty = 0.0) THEN 0.0
+            WHEN (product_uom_qty IS NULL or product_uom_qty = 0.0) THEN 100.0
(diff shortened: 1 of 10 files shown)


JuMiSanAr:
@sebalix do you think we need a new version with its corresponding migration scripts for this?

mmequignon:
Since this is a two parts PR, it would make sense to split this in two commits

stock_picking_progress/hooks.py
```diff
@@ -80,7 +80,7 @@ def setup_picking_progress(cr):
         SET progress = subquery.avg_progress
         FROM (
             SELECT sml.picking_id, avg(sml.progress) as avg_progress
-            FROM stock_move_line sml
+            FROM stock_move sml
```
mmequignon:
```suggestion
            FROM stock_move sm
```

mmequignon:
> @sebalix do you think we need a new version with its corresponding migration scripts for this?

Yes. If this module is already installed, then progress might be wrong and has to be fixed.

sebalix (approved):

PR #1187, opened 2023-01-10
[16.0][ADD] stock_picking_group_by_base

rousseldenis:
Related to pickings grouping modules, it appears we needed index creation. This implements that need.

PR #1188, opened 2023-01-11
[16.0][ADD] stock_picking_group_by_max_weight

rousseldenis:
Depends on:

- [ ] https://github.com/OCA/stock-logistics-workflow/pull/1187


stock_picking_group_by_max_weight/models/stock_move.py
```diff
@@ -0,0 +1,24 @@
...
+    def _search_picking_for_assignation_domain(self):
+        """
+        Group pickings per customer
+        """
+        domain = super()._search_picking_for_assignation_domain()
+        if self.picking_type_id.group_pickings_maxweight:
+            domain = AND(
+                [
+                    domain,
+                    [("weight", "<=", self.picking_type_id.group_pickings_maxweight)],
```
lmignon: @rousseldenis The `weight` of the picking candidates should not make the current weight exceed the `group_pickings_maxweight`
```suggestion
                    [("weight", "<=", self.picking_type_id.group_pickings_maxweight - self.picking.weight)],
```

lmignon (changes requested):
Thank you @rousseldenis. Nevertheless IMO the logic is not right...

stock_picking_group_by_max_weight/models/stock_move.py
rousseldenis: In fact, the field evaluated on the picking should be the authorized maximum weight. So we needed to store that.

lmignon (approved):
LGTM (technical + functional)

FrancoMaxime (approved):
LGTM

lmignon:
@OCA/logistics-maintainers This one is ready to merge and has been rebased to remove the dependency on previously unmerged addon

stock_picking_group_by_max_weight/views/stock_picking_type.xml
```diff
@@ -0,0 +1,20 @@
...
+    <record model='ir.ui.view' id="view_picking_type_form">
+        <field name="name">Operation Types</field>
+        <field name="model">stock.picking.type</field>
+        <field name="inherit_id" ref="stock.view_picking_type_form" />
+        <field name="arch" type="xml">
+            <field name="show_reserved" position="after">
+                <field
+                    name="group_pickings_maxweight"
+                    groups="base.group_no_one"
+                    attrs="{'invisible': [('code', '!=', 'outgoing')]}"
```
jbaudoux: Why? This is more for internal picking than delivery. I would remove this invisible line

jbaudoux (changes requested):

hparfr (changes requested):
Can you add a dependency to stock_logistics_uom ?

lmignon:
> stock_logistics_uom

Why? it's useless

stock_picking_group_by_max_weight/views/stock_picking_type.xml
lmignon: Indeed -> Done

jbaudoux (approved):

hparfr:
> > Why? it's useless
 
No. Because you are comparing weight without units.

jbaudoux:
> > > Why? it's useless
> 
> No. Because you are comparing weight without units.

In standard, the weight on the picking is always in kg (or lbs), coming from `env['product.template']._get_weight_uom_name_from_ir_config_parameter()`
If you use the module "product_logistics_uom" to set a different uom weight on the product, I don't see a module changing the weight uom of the picking and I expect it is still always in kg.
Moreover, it looks like it leads to many wrong computations, even the module product_packaging_dimension is not doing the right conversions while summing and there are lots of module depending on this one.

I find it legit this module assumes the picking weight is always in kg and the max weight is also in kg. So it is fine for me to compare without caring about the uom.

lmignon:
@hparfr In odoo you can assume that a `weight` field without any companion `weight_uom_id` field is always expressed into the default weight uom defined in the odoo's config parameters. In a lot of place where the weight is displayed into the UI, the name of the default uom is displayed next to the field. The `product_logistics_uom` should be modified to not store the weight nor the volume into another uom than the one defined into the odoo configuration otherwise it will introduce inconsistencies. In its current implementation `product_logistics_uom`  will corrupt your data.

lmignon:
@hparfr I created an issue for `product_logistics_uom` https://github.com/OCA/product-attribute/issues/1312

hparfr:
>I find it legit this module assumes the picking weight is always in kg and the max weight is also in kg. So it is fine for me to compare without caring about the uom.

I'm ok with that. My request was more "did you considerate to add this as a dependency" than "do it". :-)


> The product_logistics_uom should be modified to not store the weight nor the volume into another uom than the one defined into the odoo configuration otherwise it will introduce inconsistencies.

The goal of the product_logistics_uom is to be explicit about units.
AND If you need to change some units along the way (like if you have products less than 10g or you want to express volumes in cl instead of m³), you are sure everything works well.
It allows to express unit in the right unit for the user, be it g or cm.

Odoo is very naive about this. Not a long ago, volumes were hardcoded in m³ and there was no m³ uom !

If you do not want to introduce a dependency to product_logisitcs_uom , no problem for that. I may add one glue module at some point if I need it.

hparfr (approved):

PR #1189, opened 2023-01-11
[15.0][FIX] stock_picking_filter_lot: Update location_ids correctly when creating a new inventory adjustment

BernatPForgeFlow:
When we create a new inventory adjustment for a product/location/lot-serial number, if that Lot/Serial Number has never been at that location, it creates the Stock Quant but the Lot/Serial Number doesn't create the link to the location and subsequently, when it is needed in a picking, it cannot be located.

It is necessary to link each of the locations to the Lot/Serial Number for it to be accessible. To do so, we will recompute the location when we change quantity. Because when we firstly create the Stock Quant, quantity is 0 and it does not find that location.

### DIFF
--- a/stock_picking_filter_lot/models/stock_production_lot.py
+++ b/stock_picking_filter_lot/models/stock_production_lot.py
@@ -11,7 +11,7 @@ class StockProductionLot(models.Model):
         comodel_name="stock.location", compute="_compute_location_ids", store=True
     )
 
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
Makes sense!

LoisRForgeFlow (approved):
LGTM, thanks for the detailed explanation :+1:

PR #1190, opened 2023-01-13
[MIG] stock_move_backdating

SirTakobi:
Migration from `10.0` (https://github.com/OCA/stock-logistics-workflow/tree/499a379ca11c047cf4cce9ab10ddf9e777c54597/stock_move_backdating).

This needed some changes, mainly because Warehouse management is changed a lot between `10.0` and `12.0`, the biggest change is probably https://github.com/odoo/odoo/commit/b3180c841101510081ee8ef9c52d205497efdd4f.


stock_move_backdating/models/stock_picking.py
```diff
@@ -8,21 +8,50 @@
...
+
+    # Disable pylint to match the exact signature in super
+    # pylint: disable=dangerous-default-value
+    @api.multi
+    def _create_backorder(self, backorder_moves=[]):
+        # When a move needs backdating,
+        # we are processing the moves of a picking one by one,
+        # so we don't have to create a backorder if a move is missing
+        if 'date_backdating' not in self.env.context:
+            super()._create_backorder(backorder_moves=backorder_moves)
```
eLBati: `return`?

stock_move_backdating/models/stock_quant.py
```diff
@@ -0,0 +1,20 @@
...
+
+class StockQuant (models.Model):
+    _inherit = 'stock.quant'
+
+    def _update_available_quantity(
+        self, product_id, location_id, quantity,
+        lot_id=None, package_id=None, owner_id=None, in_date=None,
+    ):
+        date_backdating = self.env.context.get('date_backdating')
+        if date_backdating is not None and in_date is not None:
```
eLBati: Why if `in_date is not None`?

eLBati (changes requested):

stock_move_backdating/models/stock_quant.py
SirTakobi: Because in `10.0` only the dates of `quant_ids` were updated and the field `quant_ids` only contains the quants in the destination location:
image
(this is the move of a delivery)

that is why, seeing how this method is called in https://github.com/odoo/odoo/blob/3f4654f9cb0a05c65195745c78f494c228e09549/addons/stock/models/stock_move_line.py#L442-L450, I only wanted to change the date of the quant in the destination (last line, with parameter `in_date`) but not the date of the quant in the source location (first line, without parameter `in_date`).

On a second thought, it might make more sense to also change the quant in the source location so that the dates stay aligned; I added this change and its test in their own commit.

stock_move_backdating/models/stock_picking.py
SirTakobi: Thanks for the review! Added

eLBati (approved):
Thanks

stock_move_backdating/models/stock_move.py
```diff
@@ -0,0 +1,61 @@
...
+        """Process the moves one by one, backdating the ones that need to."""
+        for move in self:
+            move_line = first(move.move_line_ids)
+            date_backdating = move_line.date_backdating
+            if date_backdating:
+                move = move.with_context(
+                    date_backdating=date_backdating,
+                )
+
+            moves_todo |= super(StockMove, move)._action_done()
```
rousseldenis:
```suggestion
            moves_todo_ids = set()
            (...)

            move_todo_ids.add(super(StockMove, move)._action_done().ids)
            (...)
            
            return self.browse(moves_todo_ids)
```

Better for performances

rousseldenis (approved):
Seems great.

stock_move_backdating/models/stock_move.py
SirTakobi: Thanks for having a look!
So this is for managing `ids` as much as possible, then `browse` the records only once: I guess this might be faster.
Do you have insight on _how much_ faster this is?

rousseldenis: I think we did the search few years ago. But I don't have the metrics anymore. Maybe @simahawk can help.

rousseldenis: @SirTakobi 

I've just wrote a little script :

```python

#!/usr/bin/env python
import logging

import click
import click_odoo

_logger = logging.getLogger(__name__)


@click.command()
@click_odoo.env_options(default_log_level='error')
def main(env):
    _initialize_records(env)
    _test_records(env)
    _test_ids(env)


def _initialize_records(env):
    values = list()
    for x in range(10000):
        values.append({
            "name": "Test %s" % str(x),
        })
    env["res.partner"].create(values)


def _test_records(env):
    _logger.info("BEGIN MERGE RECORDS")
    partners = env["res.partner"].browse()
    for partner in env["res.partner"].search([]):
        partners |= partner
    _logger.info("END MERGE RECORDS")


def _test_ids(env):
    _logger.info("BEGIN MERGE IDS")
    partners = set()
    for partner in env["res.partner"].search([]):
        partners.add(partner.id)
    env["res.partner"].browse(partners)
    _logger.info("END MERGE IDS")


if __name__ == "__main__":
    main()
```


And here is the output for the 10 000 records:

image


So, merging records take ~4 seconds and browsing ids 20 ms :smiley:

rousseldenis: And I've just 'base' module installed

SirTakobi: Very interesting, thanks!

PR #1195, opened 2023-01-20
[13.0][FIX] stock_return_request: missing cache

chienandalu:
Ensure that all the records get their value

cc @Tecnativa TT41230

please review @pedrobaeza @victoralmau

### DIFF
--- a/stock_return_request/models/stock_move.py
+++ b/stock_return_request/models/stock_move.py
@@ -24,6 +24,7 @@ class StockMove(models.Model):
     def _compute_qty_returnable(self):
         """Looks for chained returned moves to compute how much quantity
            from the original can be returned """
+        self.qty_returnable = 0
         for move in self.filtered(lambda x: x.state not in ["draft", "cancel"]):
             if not move.returned_move_ids:
                 if move.state == "done":


rvalyi (approved):

stock_return_request/models/stock_move.py
```diff
@@ -24,6 +24,7 @@ class StockMove(models.Model):
     def _compute_qty_returnable(self):
         """Looks for chained returned moves to compute how much quantity
            from the original can be returned """
+        self.qty_returnable = 0
```
rousseldenis, line 27:
```suggestion
        moves_to_update = self.filtered(lambda x: x.state not in ["draft", "cancel"])
        (self - moves_to_update).qty_returnable = 0
        for move in moves_to_update:
        (...)
```

rousseldenis: This avoid to trigger update on the whole recordset if not necessary

pedrobaeza: AFAIK, since v13, this is not a problem, as this doesn't retrigger a recomputation.

pedrobaeza (approved):

PR #1197, opened 2023-01-23
[16.0][ADD] stock_move_manage_priority

glitchov:
Enable to manage priority at stock move level


stock_move_manage_priority/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    stock_move_manage_priority = fields.Boolean(
+        related="company_id.stock_move_manage_priority",
+        readonly=False,
```
lmignon, line 12: store=True

stock_move_manage_priority/models/stock_move.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @api.depends("company_id.stock_move_manage_priority", "is_locked", "state")
+    def _compute_is_priority_editable(self):
+        user_has_group = self.env.user.has_group(
+            "stock_move_manage_priority.group_stock_move_priority_manager"
+        )
+        for rec in self:
+            rec.is_priority_editable = (
+                rec.company_id.stock_move_manage_priority
+                and user_has_group
```
lmignon: @glitchov You should check for `user_has_group` first to avoid to load the company info if this first condition is already false... When you write a boolean expression you must have in mind to always check condition with the less performance impact first.... In this case, the group condition, then conditions on the record itself and finally condition involving the load of linked model (the company info)

stock_move_manage_priority/models/stock_move.py
```diff
@@ -0,0 +1,33 @@
...
+                and not rec.is_locked
+                and rec.state not in ("done", "cancel")
+            )
+
+    def _inverse_priority(self):
+        if not self.is_priority_editable:
```
lmignon: @glitchov An inverse is an api.multi method. You must iterate on self.... Moreover, if I'm not wrong, I think that the original value is available on the `_origin` attribute of the record...

```python
for rec in self:
    if not rec.is_priority_editable:
        rec.priority = rec._origin.priority
```

lmignon (changes requested):

stock_move_manage_priority/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module enables the capability to manage the priority of a stock move
+directly on the move itself. The goal is to promote/demote specific moves.
```
lmignon, line 2: @glitchov An important think to have in mind when we write a description is to explain the 'Why' and the 'Who'...

```
Into the reserving process, the moves are sorted by reservation date, then priority,
then date and finally by sequence. By default, the only way to change the priority
of a move is to change the priority of the picking.  If this default behavior fits
for the most of the cases, we can imagine that in some cases, we want to change the
priority of a move without changing the priority of the picking since the stock is
limited for this particular product and we want to be sure that the move will be
reserved before others created before this one for example. This addon allows you
to it by displaying the priority field into the Operations tree of the picking form
view.
```

stock_move_manage_priority/models/stock_move.py
```diff
@@ -0,0 +1,34 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    is_priority_editable = fields.Boolean(
+        compute="_compute_is_priority_editable",
+    )
+    priority = fields.Selection(
+        inverse="_inverse_priority",
+        readonly="False",
```
lmignon: @glitchov `readonly=False`

lmignon (changes requested):
The readonly attribute is not properly defined on the priority field. Therefore the field is never editable into the UI

stock_move_manage_priority/views/stock_picking_views.xml
```diff
@@ -0,0 +1,28 @@
...
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='move_ids_without_package']/tree"
+                position="inside"
+            >
+                <field name="is_priority_editable" invisible="1" />
+                <field
+                    name="priority"
+                    attrs="{'readonly': [('is_priority_editable', '=', False)]}"
+                    decoration-success="priority == '-1'"
```
lmignon: @glitchov `-1` is not a valid value for priority...

lmignon (changes requested):

lmignon (approved):
LGTM (Codre review + functional)

lmignon:
ping @rousseldenis

rousseldenis (approved):
Code review. LGTM

PR #1198, opened 2023-01-24
[16.0][MIG] purchase_stock_picking_invoice_link: Migration to 16.0

JavierIniesta:
Issue: #1101


JavierIniesta:
We have done functional tests and it works correctly.

rousseldenis (approved):
Code review

purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
```diff
@@ -0,0 +1,205 @@
+# Copyright 2019 Vicent Cubells <pedro.baeza@tecnativa.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from odoo import fields
+from odoo.tests import Form, common, tagged
+
+
+@tagged("-at_install", "post_install")
```
lmignon, line 7: useless...  but I don't want to reopen an endless debate

lmignon (approved):
Code review

purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
pedrobaeza: Not this time, Laurent. For having a proper CoA, this should be put on post install. It's a requirement put on everything depending on `account`. Please see as something related (but only partial history) https://github.com/odoo/odoo/commit/f34d8632572dfe8911b62471882eae60f6e6ae72

lmignon: Thank you @pedrobaeza for the explanation. :+1:

PR #1199, opened 2023-01-25
[16.0][ADD] stock_picking_backorder_reason

rousseldenis:
In some flows, one may want to let the user have the choice on something more understandable in his context. So, instead choosing a backorder strategy, let him choose a 'reason'.

Let define some backorder reasons and associate the strategy on those reasons.

We define also a new strategy which is 'Use partner option' to delegate it on partner level.


stock_picking_backorder_reason/models/stock_backorder_reason.py
```diff
@@ -0,0 +1,22 @@
...
+    _description = "Stock Backorder Reason"
+
+    name = fields.Char(required=True, translate=True)
+    backorder_action_to_do = fields.Selection(
+        selection=[
+            ("create", "Create backorder"),
+            ("cancel", "Cancel backorder"),
+            ("use_partner_option", "Use partner option"),
+        ],
+        default="picking_type",
```
lmignon: @rousseldenis `picking_type` is not into the selection list

stock_picking_backorder_reason/demo/stock_backorder_reason.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 ACSONE SA/NV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
+
+    <!-- TODO
+    <record model="stock.backorder.reason" id="stock_backorder_reason_demo_1">
+       <field name="name">...</field>
+    </record>
+    -->
```
lmignon: @rousseldenis Do we need-it?

stock_picking_backorder_reason/models/stock_picking_type.py
```diff
@@ -0,0 +1,57 @@
...
+        readonly=False,
+        store=True,
+        help="Check this in order to consider pickings in this type as "
+        "Purchase ones for backorder reason cancellation.",
+    )
+
+    @api.depends("backorder_reason")
+    def _compute_backorder_reason_sale(self):
+        activated_types = self.filtered("backorder_reason_sale")
+        to_deactivate = activated_types.filtered(lambda t: not t.backorder_reason)
```
lmignon: a double loop.... should be done in one loop.

stock_picking_backorder_reason/models/stock_picking_type.py
```diff
@@ -0,0 +1,57 @@
...
+        to_deactivate.update({"backorder_reason_sale": False})
+
+    @api.depends("backorder_reason")
+    def _compute_backorder_reason_purchase(self):
+        activated_types = self.filtered("backorder_reason_purchase")
+        to_deactivate = activated_types.filtered(lambda t: not t.backorder_reason)
```
lmignon: a double loop.... should be done in one loop.

lmignon (changes requested):

stock_picking_backorder_reason/models/stock_backorder_reason.py
rousseldenis: Done

stock_picking_backorder_reason/models/stock_picking_type.py
rousseldenis: Done

stock_picking_backorder_reason/demo/stock_backorder_reason.xml
rousseldenis: I've added some

rousseldenis:
@lmignon Could you update your review ?

lmignon (approved):
Thank you @rousseldenis LGTM (Core review + functional tests)

lmignon (changes requested):
@rousseldenis The licence must be LGPL since this addon will be mixed with helpdesk.

lmignon (approved):

jbaudoux:
@rousseldenis @lmignon I don't like the module name and how the module solves the initial requirement.

Basically, we want to configure on the supplier if he supports backorder in his system or not, and this in order to not keep a reception backorder open in odoo for nothing as you will never receive the missing goods.

So mainly we need to configure a backorder strategy on the supplier.
Then we need to be able to process a reception, make intermediate validation while receiving (like when a pallet is received and you want to be able to put it away before continuing receiving, or like when you want to urgently receive a single product from the reception and continue later) and when everything is received, apply the supplier backorder strategy on the remaining backorder.
When you validate the picking, in standard you get a wizard asking if you want a backorder or not. Instead of asking if we want a backorder or not, you could ask if all received goods have been processed and if not then create a backorder (this could even be configured on the reception picking as backorder strategy to always), otherwise apply the supplier backorder strategy. If no supplier strategy is configured on the supplier, then we even don't need to ask anything, we just apply standard.
The reason basically serves this purpose of asking if all goods have been received.

Besides this, in my opinion, you cannot mirror this feature for sales in the same way. So all the code to try to be generic and cover sales looks useless to me.

I would simplify a lot this module to just keep the basic need regarding supplier and rename the module to reflect that it basically adds a supplier backorder strategy.

rousseldenis:
> Besides this, in my opinion, you cannot mirror this feature for sales in the same way. So all the code to try to be generic and cover sales looks useless to me.

We have flows where customer has particular requirement on this, so we need to keep that.

> I would simplify a lot this module to just keep the basic need regarding supplier and rename the module to reflect that it basically adds a supplier backorder strategy.

From my preceding remark, I agree partly on this as this could be renamed 'backorder_partner_stragey' or 'backorder_third_party_strategy' to reflect a module centered on the 'partner' strategy, letting the 'reason' as accessory and not the main functionnality.

rousseldenis:
> I would simplify a lot this module to just keep the basic need regarding supplier and rename the module to reflect that it basically adds a supplier backorder strategy.

We have the case for customers too.

stock_picking_backorder_reason/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Stock Picking Backorder Reason",
+    "summary": """
+        Allows to add a layer on top of backorder confirmation in order to choose
+        a reason. That reason will contain the strategy to apply.""",
+    "version": "16.0.1.0.0",
+    "license": "LGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
```
jbaudoux, line 11: Can you add maintainers also?
```suggestion
    "author": "ACSONE SA/NV,BCIM,Odoo Community Association (OCA)",
```

PR #1202, opened 2023-01-27
[ADD] - stock_move_propagate_first_move

sbejaoui:
### Propagate the first move reference to all moves in a multi-step route.

Sometimes when you process your move, you need to know which move has
been at the origin of the procurement chain that lead to the creation
of the move. This information could be useful for example to know
if the move is part of an outgoing or incoming process without having
to look at the chained moves.

This module add the fields `first_move_id` and `first_picking_type_id` on the
move line to give this information in a transparent way without having
to wonder about the direction in which the move chain should be navigated
to get the information.


sbejaoui:
cc/ @lmignon

stock_move_picking_type_origin/models/stock_rule.py
```diff
@@ -0,0 +1,43 @@
...
+        if move_dest:
+            original_picking_type_id = move_dest[0].original_picking_type_id.id
+        else:
+            original_picking_type_id = stock_move_values.get("picking_type_id")
+        stock_move_values["original_picking_type_id"] = original_picking_type_id
+        return stock_move_values
+
+    def _push_prepare_move_copy_values(self, move_to_copy, new_date):
+        values = super()._push_prepare_move_copy_values(move_to_copy, new_date)
+        values["original_picking_type_id"] = move_to_copy.original_picking_type_id.id
```
lmignon: @sbejaoui 
????
```suggestion
        values["original_picking_type_id"] = move_to_copy.original_picking_type_id.id or move_to_copy.picking_type_id.id
```
?????

lmignon (changes requested):
@sbejaoui The original_picking_type_id seems to not be set properly on the first move in a push chain 
image

lmignon (approved):
LGTM (Code review + functional tests)

stock_move_propagate_first_move/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This addon propagate the first move reference to all moves in a multi-step route.
```
lmignon: @sbejaoui you should explain the `why`... something like

```
Sometimes when you process your move, you need to know which move has
been at the origin of the procurement chain that lead to the creation
of the move. This information could be useful for example to know
if the move is part of an outgoing or incoming process without having
to look at the chained moves.

This module add the fields first_move_id and first_picking_type_id on the
move line to give this information in a transparent way without having
to wonder about the direction in which the move chain should be navigated
to get the information.
```

(feel free to adapt or correct the proposed description)

sbejaoui: very helpful, thanks

lmignon (approved):
Code review + functional

rousseldenis (approved):
Code review

stock_move_propagate_first_move/models/stock_move.py
```diff
@@ -0,0 +1,23 @@
...
+        readonly=True,
+        help="The original move which generated this one.",
+    )
+    first_picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        string="Original Operation Type",
+        related="first_move_id.picking_type_id",
+        store=True,
+        help="Picking type of the original move which generated this one.",
+    )
```
lmignon, line 23: @sbejaoui  IMO we should always initialize the first_move_id with self if not provided at create. In this way it's not required to always check if first_move_id is set...

```suggestion
    )
    
    @api.model_create_multi
    def create(self, vals_list):
        res = super().create(vals_list)
        for res in res:
            if res.first_move_id:
                continue
            res.first_move_id = res
        return res
```

lmignon (changes requested):

PR #1205, opened 2023-01-30
[16.0][FIX] Update copier template

pedrobaeza:
To avoid the isort problem (PyCQA/isort#2077)

@Tecnativa


rousseldenis:
@pedrobaeza Already don in #1203

PR #1206, opened 2023-01-30
[14.0][ADD] stock_internal_resupply_stock_in

Kev-Roche:
depends on https://github.com/OCA/stock-logistics-warehouse/pull/1546

This module allows to automate resupply between warehouses within the same company, by using the "stock in" picking type.

### DIFF
--- a/stock_internal_resupply_stock_in/models/stock_picking.py
+++ b/stock_internal_resupply_stock_in/models/stock_picking.py
@@ -1,0 +1,60 @@
+# Copyright 2023 Akretion (https://www.akretion.com).
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def action_confirm(self):
+        for rec in self:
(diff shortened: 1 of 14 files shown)


stock_internal_resupply_stock_in/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+from odoo import fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def action_confirm(self):
+        for rec in self:
+            if rec.picking_type_id == rec.picking_type_id.warehouse_id.stock_in_type_id:
+                rec.move_lines.procure_method = "make_to_order"
```
petrus-v, line 14: I would probably prefer a method that let user (the one configuring stock rules) know  that with such picking type it will force the procurement method. I think this can be done adding a new field on `stock.picking.type` like `force_procure_method` and could be used like this or from:

 ```suggestion
            if rec.picking_type_id.force_procure_method:
                rec.move_lines.procure_method = rec.picking_type_id.force_procure_method
```

Kev-Roche: What will be another case if not "make_to_order" ?

petrus-v: make_to_stock and I'm not sure if there are other community module which could add new types, if you prefer a boolean that could do the trick !

PR #1214, opened 2023-02-09
[16.0][ADD] - add stock_override_procurement

sbejaoui:
In odoo core modules there is often a method to provide the procurement values but not the attributes like location, product_id, etc. This technical module makes it possible to override them if they are passed in the values.


sbejaoui:
cc/ @lmignon

stock_override_procurement/models/procurement_group.py
```diff
@@ -0,0 +1,42 @@
...
+        ]
+
+    @api.model
+    def _get_override_procurement(self, procurement):
+        values = procurement.values
+
+        def _get_override_value(field_name):
+            return (
+                values.get(field_name)
+                if values.get(field_name)
```
lmignon:
```suggestion
                if field_name in values
```
required to allows to unset a value by override

lmignon (changes requested):

lmignon (changes requested):
Thank you @sbejaoui A useful addon. A little change required for a specific case...

lmignon (approved):
LGTM (Code review)
Thank you @sbejaoui

rousseldenis (approved):
Code review. LGTM

Hackish but will work great!

rvalyi (approved):
LGTM

PR #1232, opened 2023-02-23
[16.0][MIG] stock_picking_operation_quick_change

florian-dacosta:
Standard Migration.
I am wondering if we could change some things though.

1) Do we really need the button for outgoing pickings ? It seems to me it is more usefull to change dest location on reception, or even internal, but not on delivery ?
I would hide it on outgoing pickings

2) The button is available for all stock users... But it changes the location, to be consistent, it should be available for the group Manage Multiple Stock Locations only IMO

3) The button is only in the operation tab. When picking is configured to show the detailed operations, I think the button should be there, and then hiden from the operation tab ?

I can change the module accordingly, what do you think ?
@duyanhk15 @leemannd @yankinmax @cyrilmanuel @pedrobaeza @sergio-teruel @LoisRForgeFlow


LoisRForgeFlow:
@florian-dacosta My opinion for each of the points:

1. I agree with you, but since the README explicitly describes that use case, I would check with the original authors because maybe they had a case that really needed the feature in deliveries. It is an old module, maybe it did make sense at the time but it doesn't anymore.
2. Yes, hiding the button when multilocations are not activated seems a good idea.
3. I'm more hesitant with this one. The detailed operations (stock.move.lines) may have more specific locations applied (due to putaway strategies) and this modules changes de locations of the stock.moves. Users may think that the feature is meant to adjust the bin destination location, but I think if the wizard affects stock.moves is more suitable to change de "general" demand location on the stock move. Just something to have in mind if you decide to change it.

florian-dacosta:
> 3\. I'm more hesitant with this one. The detailed operations (stock.move.lines) may have more specific locations applied (due to putaway strategies) and this modules changes de locations of the stock.moves. Users may think that the feature is meant to adjust the bin destination location, but I think if the wizard affects stock.moves is more suitable to change de "general" demand location on the stock move. Just something to have in mind if you decide to change it.

No, AFAIK, it does change the location on `stock.move.line` and not stock.move.

LoisRForgeFlow:
> > 3. I'm more hesitant with this one. The detailed operations (stock.move.lines) may have more specific locations applied (due to putaway strategies) and this modules changes de locations of the stock.moves. Users may think that the feature is meant to adjust the bin destination location, but I think if the wizard affects stock.moves is more suitable to change de "general" demand location on the stock move. Just something to have in mind if you decide to change it.
> 
> No, AFAIK, it does change the location on `stock.move.line` and not stock.move.

You are [right](https://github.com/OCA/stock-logistics-workflow/blob/15.0/stock_picking_operation_quick_change/wizards/stock_picking_wizard.py#L108), my bad. Then your suggestion makes total sense.

Thanks!

florian-dacosta:
I did the changes 2 and 3 of the PR description and not the one since I am not sure it will be ok for every one.
This PR is ready

florian-dacosta:
Test failure seems unrelated and is already failing in main 16.0 branch

rousseldenis:
> Test failure seems unrelated and is already failing in main 16.0 branch

@florian-dacosta Yes, it seems this is caused by: https://github.com/odoo/odoo/commit/ca9aab977d6aa7d3565c2b69f1cddd3edcb7fc9f

I need to take care of it. :face_in_clouds:

rousseldenis (approved):

florian-dacosta:
I  havejust included last fix on v15 from https://github.com/OCA/stock-logistics-workflow/pull/1331

@sergiocorato @sergio-teruel @CarlosRoca13 @Ricardoalso @gbrito @FrancescoBallerini 

Any chance to have a review to get this finally merged ?
Thanks

Ricardoalso (approved):

PR #1240, opened 2023-03-08
[14.0] stock_picking_send_by_mail: able to send the email directly

sebalix:
Add a new option on delivery operation types to send automatically an e-mail to the customer's shipping contact:
image

And add a hook to override the template used.

### DIFF
--- a/stock_picking_send_by_mail/models/stock_picking.py
+++ b/stock_picking_send_by_mail/models/stock_picking.py
@@ -8,9 +8,12 @@ from odoo import _, models
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    def _get_picking_send_email_template(self):
+        return self.company_id.stock_mail_confirmation_template_id
+
     def action_picking_send(self):
         self.ensure_one()
-        template = self.company_id.stock_mail_confirmation_template_id
+        template = self._get_picking_send_email_template()
         compose_form = self.env.ref(
             "mail.email_compose_message_wizard_form",
             False,
--- a/stock_picking_send_by_mail/models/stock_picking.py
+++ b/stock_picking_send_by_mail/models/stock_picking.py
@@ -2,30 +2,39 @@
 # Copyright 2017 Tecnativa - Vicent Cubells
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, models
+from odoo import _, fields, models
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    delivery_email_sent = fields.Boolean()
+
     def _get_picking_send_email_template(self):
         return self.company_id.stock_mail_confirmation_template_id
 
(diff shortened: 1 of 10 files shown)


stock_picking_send_by_mail/models/stock_picking.py
```diff
@@ -8,21 +8,39 @@
...
         ctx = dict(
             default_model="stock.picking",
             default_res_id=self.id,
             default_use_template=bool(template),
-            default_template_id=template and template.id or False,
+            default_template_id=template_id,
             default_composition_mode="comment",
             user_id=self.env.user.id,
         )
+        if send:
```
simahawk: I'd move this to its own method and I would enable this w/ a flag on the picking type

simahawk (changes requested):

stock_picking_send_by_mail/models/stock_picking.py
sebalix: forgot this part!

stock_picking_send_by_mail/models/stock_picking_type.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    send_delivery_email = fields.Boolean(
+        string="Send Delivery E-mail",
+        help="Send an e-mail on validation to the customer with the delivery slip.",
```
simahawk, line 12: mention that you can configure the email template in the settings

stock_picking_send_by_mail/tests/test_stock_picking_send_by_mail.py
```diff
@@ -29,8 +34,23 @@ def setUpClass(cls):
...
         )
 
-    def test_send_mail(self):
+    def test_send_mail_action(self):
         self.picking.action_confirm()
         self.picking.action_assign()
         result = self.picking.action_picking_send()
         self.assertEqual(result["name"], "Compose Email")
+
+    def test_send_mail_direct(self):
```
simahawk, line 43: hmm I see no automatic send :smile:

stock_picking_send_by_mail/tests/test_stock_picking_send_by_mail.py
```diff
@@ -10,8 +11,12 @@ def setUpClass(cls):
         super(TestStockPickingSendByMail, cls).setUpClass()
         cls.product = cls.env["product.product"].create({"name": "Test product"})
         cls.picking_type = cls.env.ref("stock.picking_type_out")
+        cls.customer = cls.env["res.partner"].create(
```
simahawk, line 14: while you are at it: can you disable tracking pls?

simahawk (changes requested):

sebalix:
We probably don't need all of this, it's a std feature of Odoo through `stock.picking._send_confirmation_email` method...

PR #1250, opened 2023-03-13
[16.0][FIX] stock_picking_start: Allows picking manual creation

lmignon:
Before this change it was no more possible to create a picking from the UI since the 'start' field is provided into the dict of values and a check was done to determine is this value could be set to False. The problem was that the check didn't take into account the original value to detect if the provided value will update the record...

ping @jbaudoux ping @rousseldenis


rousseldenis (approved):
Code review. LGTM

jbaudoux (approved):

PR #1251, opened 2023-03-14
[16.0][ADD] sale_stock_restocking_fee_invoicing

glitchov:
On demand charge restocking fee for accepting returned goods.
Transferred from https://github.com/OCA/account-invoicing/pull/1380


sale_stock_restocking_fee_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,66 @@
...
+    _inherit = "sale.order"
+
+    def _get_restocking_fee_line_name(self, stock_move):
+        """Return the SO line description for the restocking fee. This method should
+        be called with the customer lang into the context to get the description into
+        the right language
+        """
+        lang = self.partner_id.lang
+        return _(
+            f"Restocking fee for {stock_move.product_uom_qty} "
```
rousseldenis: You cannot use format like that as values will be reformatted before applying translation.

Either, use `_(%(<variable>)s) % dict(variable=variable) `or` _({variable}).format(variable=variable)`

sale_stock_restocking_fee_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    is_customer_return = fields.Boolean(compute="_compute_is_customer_return")
```
rousseldenis: Those fields name can enter in conflict with other modules as too generic.

Either, put them in a separate module, either rename them (with maybe the 'fee' word)

sale_stock_restocking_fee_invoicing/wizards/stock_return_picking_line.py
```diff
@@ -0,0 +1,22 @@
...
+
+    _inherit = "stock.return.picking.line"
+
+    charge_restocking_fee = fields.Boolean(
+        help="Tick this box if you wish to charge your customer a fee in "
+        "case of return of goods",
+        default=False,
+    )
+
+    is_customer_return = fields.Boolean(compute="_compute_is_customer_return")
```
rousseldenis: Same as above

rousseldenis (changes requested):
Code review. Some changes but looks promising

sale_stock_restocking_fee_invoicing/models/sale_order.py
lmignon: > You cannot use format like that as values will be reformatted before applying translation.
> 
> Either, use `_(%(<variable>)s) % dict(variable=variable) `or` _({variable}).format(variable=variable)`

The right way is 
`_("%(variable)s", variable="value")`

sale_stock_restocking_fee_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    is_fee_customer_return = fields.Boolean(compute="_compute_is_fee_customer_return")
+
+    is_fee_customer_out = fields.Boolean(compute="_compute_is_fee_customer_out")
```
lmignon: @glitchov IMO this change is no appropriate. We should no prefix these fields to avoid name collision. These fields define a wider concept. It add to new fields to easily know if pickings refers to a customer out or a customer_return. We can put this functionality into a dedicated addon. @rousseldenis What can you propose as name since you are at the origin of this initial change.

lmignon (changes requested):

lmignon (approved):
LGTM (Code + Functional)

lmignon:
@rousseldenis Can you update your review PLZ?

sale_stock_restocking_fee_invoicing/models/stock_move.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    picking_kind = fields.Selection(related="picking_id.picking_kind", readonly=True)
+
+    def _action_done(self, cancel_backorder=False):
+        result = super()._action_done(cancel_backorder=cancel_backorder)
+        move_ids_to_charge_by_so = defaultdict(list)
+        for move in self:
+            if not move._is_restocking_fee_chargeable():
+                continue
+            so = move.sale_line_id.order_id
```
rousseldenis, line 26: This can be void.

Either you check if order_id is set here, either you add the condition in the _is_restocking_fee_chargeable().

It causes a stack trace :

image

rousseldenis (changes requested):
@glitchov Could you fix and add a test case for a delivery/return without sale order ?

lmignon:
@glitchov  https://github.com/OCA/stock-logistics-workflow/pull/1254 is merged. Can you remove the dependency on this PR into the test-requirements

rousseldenis (approved):
Code review

PR #1258, opened 2023-03-23
[16.0][ADD] stock_move_line_serial_unique

kanda999:
@qrtl


yostashiro (commented):
@kanda999 Please add the business rationale why this module is needed.

stock_quant_unique_serial/models/stock_quant.py
```diff
@@ -0,0 +1,37 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    @api.constrains("product_id", "lot_id")
+    def _check_stock_quant_unique(self):
+        for record in self:
```
rousseldenis: @kanda999 Couln't you call the original Odoo method instead  (as it seems more complex than your code) ?

https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_quant.py#L1011

rousseldenis: Have you ever thought about suggesting this to Odoo directly, maybe by adding a setting to enforce an error and not just a warning ?

kanda999: @rousseldenis I adjusted to use original Odoo method.

I thought about changing Odoo directly, but there is a button for displaying warnings in the stock.quant model.
Some users may have the same serial with different owners.

stock_quant_unique_serial/models/stock_picking.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _action_done(self):
+        self = self.with_context(location_dest_id=self.location_dest_id)
+        return super(StockPicking, self)._action_done()
```
yostashiro: I think it'd be more appropriate to do this in stock.move.line, to cover cases from manufacturing orders as well.

stock_quant_unique_serial/models/stock_quant.py
```diff
@@ -0,0 +1,30 @@
...
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    @api.constrains("product_id", "lot_id")
+    def _check_stock_quant_unique(self):
+        dest_location = self.env.context.get("location_dest_id")
+        if dest_location:
+            location = dest_location
+        else:
+            location = self.location_id
```
yostashiro:
```suggestion
        location = self.env.context.get("location_dest_id")
        if not location:
            location = self.location_id
```

stock_quant_unique_serial/models/stock_quant.py
```diff
@@ -0,0 +1,30 @@
...
+        for record in self:
+            if (
+                record.lot_id
+                and record.product_id.tracking == "serial"
+                and location.usage == "internal"
```
yostashiro:
```suggestion
                and location.usage in ("internal", "transit")
```

stock_quant_unique_serial/models/stock_quant.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class StockQuant(models.Model):
+    _inherit = "stock.quant"
+
+    @api.constrains("product_id", "lot_id")
+    def _check_stock_quant_unique(self):
```
yostashiro:
```suggestion
    def _check_stock_quant_serial_unique(self):
        """This method intends to check whether the stock for specifield serial already exists when stock is being created (e.g. a purchase receipt is being processed), to avoid duplicates.
        """
```

stock_quant_unique_serial/models/stock_quant.py
```diff
@@ -0,0 +1,30 @@
...
+            location = dest_location
+        else:
+            location = self.location_id
+        for record in self:
+            if (
+                record.lot_id
+                and record.product_id.tracking == "serial"
+                and location.usage == "internal"
+            ):
+                message, dummy = self.env["stock.quant"]._check_serial_number(
```
yostashiro:
```suggestion
                message, dummy = self._check_serial_number(
```

stock_quant_unique_serial/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,8 @@
+When creating or updating a stock.quant record with serial management products ,
+the constraint checks if serial number already exists. If a duplicate is found, the user get a warning message and the record is not saved.
+
+Standard behavior fails to prevent the creation of multiple records
+with the same serial number when different owners are specified for the same product.
+
+This can lead to warnings in the Inventory Adjustment view, which may be not desirable.
+So adds the ability to prevent the creation of multiple records with the same serial number, even when different owners are specified for the same product.
```
yostashiro:
```suggestion
This module adds a constraint on stock.quant to make sure that no stock exists for the specified serial when stock is being created (e.g. purchase receipt is being processed) to avoid the duplicates.

Standard behavior fails to prevent the duplicates, for example, when there is stock for a serial with no owner, and new stock comes in with the same serial with an owner. This module intends to cater to this case.
```

yostashiro (changes requested):
I prefer the module name to be `stock_quant_serial_unique`.

yostashiro (approved):
Code review. LGTM 👍

yostashiro:
My review wasn't complete.  We are experiencing some issues and this PR will be updated.

rousseldenis:
@kanda999 Why removing constraint on stock quant ? Moreover, could you fix tests?

kanda999:
> @kanda999 Why removing constraint on stock quant ? Moreover, could you fix tests?

@rousseldenis 
Thank you very much for your review.  I reviewd a test script and I fixed it.

Standard behavior fails to prevent the duplicates, for example, when there is stock for a serial with no owner, and new stock comes in with the same serial with an owner.  This module intends to cater to this case.
I thought it would be better not to get errors in Stock Adjustment view.  Users may temporarily register duplicate serials when adjusting inventory from the Stock Adjustment view.
For example, you want to change the owner as follows
ーStock Adjustmentー
ProductA/001/OwnerA
↓
ProductA/001/OwnerB
ーー
In these cases, it may be more convenient to be able to register temporarily.
So stock quant constraint is removed.

rousseldenis (approved):
Code and functional review

yostashiro (approved):
Code review. Tested in a local environment as well. 👍

PR #1261, opened 2023-03-29
[13.0][IMP] stock_split_picking: hook to validate split

JaumeBforgeFlow:


stock_split_picking/models/stock_picking.py
```diff
@@ -12,9 +12,13 @@ class StockPicking(models.Model):
...
+    def _validate_split(self):
+        pass
+
     def split_process(self):
         """Use to trigger the wizard from button with correct context"""
         for picking in self:
+            self._validate_split()
 
             # Check the picking state and condition before split
             if picking.state == "draft":
```
LoisRForgeFlow: If the hook is meant to add more custom validations, wouldn't it make sense to include the standard ones already inside the hook?

JordiMForgeFlow: I agree with @LoisRForgeFlow. I would also suggest to use the word check instead of validate for the method, like _check_split_process. WDYT?

stock_split_picking/models/stock_picking.py
```diff
@@ -12,20 +12,22 @@ class StockPicking(models.Model):
...
-            if picking.state == "draft":
-                raise UserError(_("Mark as todo this picking please."))
-            if all([x.qty_done == 0.0 for x in picking.move_line_ids]):
-                raise UserError(
-                    _(
-                        "You must enter done quantity in order to split your "
-                        "picking in several ones."
-                    )
-                )
+            self._check_split_process()
```
JordiMForgeFlow:
```suggestion
            picking._check_split_process()
```

stock_split_picking/models/stock_picking.py
JaumeBforgeFlow: @LoisRForgeFlow @JordiMForgeFlow Both of your feedback make sense

JordiMForgeFlow (approved):
LGTM 👍🏼

JuanyDForgeflow (approved):

LoisRForgeFlow:
@JaumeBforgeFlow could you fwport the change?

PR #1262, opened 2023-03-30
[16.0][IMP] stock_move_propagate_first_move: Add init hook

lmignon:
Add init hook to initialize values for columns added by the moduel. To avoid freezing the database by updating millions on records, the values are initialised for the stock_move created for the last 12 months. There is a high probability that these columns are not required by your code since these old records should never be involved by your current activies.


lmignon:
In draft while not fully functionally validated on our migrated DB.

PR #1264, opened 2023-04-02
[ADD] - add stock_move_propagate_first_move_cancel

sbejaoui:
This module is based on `stock_move_propagate_first_move` module, which allows to track the first move that initiated the procurement chain.

With this module installed, and when the first move is canceled, all related moves will also be canceled.


stock_move_propagate_first_move_cancel/models/stock_move.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import models
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    def _action_cancel(self):
+        res = super()._action_cancel()
+        moves_to_cancel = self.search([("first_move_id", "in", self.ids)]) - self
```
rousseldenis: Could you add ('id', 'not in', self.ids) in domain instead of substract ?

rousseldenis (approved):
Code review

stock_move_propagate_first_move_cancel/models/stock_move.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import models
+
+
+class StockMove(models.Model):
+
+    _inherit = "stock.move"
+
+    def _action_cancel(self):
+        res = super()._action_cancel()
+        moves_to_cancel = self.search(
```
rousseldenis, line 15: @sbejaoui Thinking on this, shouldn't it use the stock.rule 'propagate _cancel' attribute and add that in the domain?

lmignon:
@rousseldenis Can you reopen it?

lmignon:
:+1:  Code review

PR #1268, opened 2023-04-04
[ADD] stock_picking_is_completed

rousseldenis:


rousseldenis:
This maybe needs a module to implement a search on move_ids_without_package field as that one is needed in depends() to make working the client side without saving the picking and generates a warning in Odoo logs

lmignon:
@rousseldenis The concept is useful but I'm a little bit afraid by the performance cost on large pickings since on every operation, the system will need to reload all the operations to compute the field. We'll see how it works on our DB...

rousseldenis:
> @rousseldenis The concept is useful but I'm a little bit afraid by the performance cost on large pickings since on every operation, the system will need to reload all the operations to compute the field. We'll see how it works on our DB...

Yes, I had that in mind too. But, I think this should be correct as it has no impact on operations that can be done outside the interface if the field is not loaded and in the interface, we are limited to the first 80 move lines.

Let's go testing

PR #934, opened 2021-11-26
[14.0][MIG] product_expiry_simple

alexis-via:
Migration from v12 (PR https://github.com/OCA/stock-logistics-workflow/pull/532) to v14.

PR #1271, opened 2023-04-06
[14.0][MIG] product_expiry_simple #tryagain

bealdav:
supersed #934


.github/workflows/test.yml
```diff
@@ -36,9 +36,18 @@ jobs:
...
           - container: ghcr.io/oca/oca-ci/py3.6-ocb14.0:latest
+            include: "product_expiry_simple,product_expiration_date"
+            name: test with OCB
+          - container: ghcr.io/oca/oca-ci/py3.6-odoo14.0:latest
+            exclude: "product_expiry_simple,product_expiration_date"
+            makepot: "true"
+            name: test with Odoo
+          - container: ghcr.io/oca/oca-ci/py3.6-ocb14.0:latest
+            exclude: "product_expiry_simple,product_expiration_date"
             name: test with OCB
```
bealdav, line 58: Please @rousseldenis could you help to fix configuration to get it works
by comparison of https://github.com/OCA/server-tools/blob/14.0/.github/workflows/test.yml#L36
it seems ok, but in real not

rousseldenis: Yes, I saw your tries.

But, the problem here I think is the mutual exclusion in the manifest. I think the product_expiry module is installed in the first step as required by other modules.

I think you should exclude them also and put in the rebel modules: stock_mass_scrap, stock_move_quick_lot. Could you try ?

@sbidoul Any advice?

sbidoul: Looking at the error I see "product_expiration_date does not exist". But indeed this module does not exist in this repo?

bealdav: You're right, madre mia, I shouldn't work after 22h, caramba !

rousseldenis:
And @bealdav the question at 10.000 dollars. Isn't there a mean to make this (the module) configurable to allow it being installed along with product_expiry ?

bealdav:
@rousseldenis I suppose you suggest a new module product_expiry_simple_but_not_so_easy_finally ?

rousseldenis:
> @rousseldenis I suppose you suggest a new module product_expiry_simple_but_nor_so_easy_finally ?

No, I wanted to see if both can coexist. Maybe by hiding the standard fields when installing this

bealdav:
Yes, I understood, sorry for my joke ;-)
I think @alexis-via  might be opposite to this option ?

But you're right, in this case runboat fails because of  ```
    raise UserError(msg % (module.shortdesc, exclusion.exclusion_id.shortdesc))
odoo.exceptions.UserError: Modules "Product Expiry Simple" and "Products Expiration Date" are incompatible. ```

bealdav:
What are the options to make runboat works even if product_expiry_simple is not installed ?

rousseldenis:
> What are the options to make runboat works even if product_expiry_simple is not installed ?

I think this is not possible as exclude key in manifest raises an Exception:

image

Maybe to have its own repo should be a solution

sbidoul:
> What are the options to make runboat works even if product_expiry_simple is not installed ?

At the moment, runboat does not support conflicting modules that prevent simultaneous installation of all modules in the repo. Suggestions are welcome in https://github.com/sbidoul/runboat/issues/36

bealdav:
Probably, but runboat install addons of the whole repo with dependency.
Could we consider runboat install the whole repo except rebels ones ?

alexis-via:
I don't believe in a solution where product_expiry_simple depend on product_expiry ; I think it will bring more problems than solutions. The goal of product_expiry_simple is to have a easier/simpler drop-in replacement to product_expiry. Then we need to adapt the OCA tests system to allow the development of such alternative modules.

bealdav:
@sbidoul @rousseldenis could you add `no stale` label to this PR. Thanks.

rousseldenis:
@bealdav Following https://github.com/sbidoul/runboat/issues/36, is this needing the rebel module configuration ?

bealdav:
I suppose this not needed, then we can merge this PR as soon as there reviews. Could you approve the PR @rousseldenis ?

alexis-via (approved):

florian-dacosta (approved):
LGTM

francesco-ooops:
@bguillot @bealdav @rousseldenis this has broken runboat for other PRs, eg: https://github.com/OCA/stock-logistics-workflow/pull/1383

bealdav:
Oops, I forgot https://github.com/sbidoul/runboat/issues/36 to deal with rebel modules.

rousseldenis:
> Oops, I forgot [sbidoul/runboat#36](https://github.com/sbidoul/runboat/issues/36) to deal with rebel modules.

Yes. This will prevent testing on runboat but not merges. We will need to manage it during OCA days. Remotely :-(

francesco-ooops:
@rousseldenis any plan to fix runboat for 14? thank you!

francesco-ooops:
@bealdav gentle reminder as we're out of runboat on v14 :)

bealdav:
The problem is that I don't really understand runboat machinery to make a pertinent PR. ;-(
Really sorry

rousseldenis:
@bealdav Solved. Read https://github.com/orgs/OCA/discussions/126#discussioncomment-7624191

PR #1273, opened 2023-04-12
[FIX][15.0] stock_picking_mass_action - missing button_validate_picking_ids context

theangryangel:
Missing `button_validate_picking_ids` context means pickings aren't actually transferred if action_immediate_transfer_wizard or _action_generate_backorder_wizard is called.

Validation is gated by the context: https://github.com/odoo/odoo/blob/15.0/addons/stock/wizard/stock_backorder_confirmation.py#L58

And as part of the standard process of `button_validate` added here: https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_picking.py#L1012-L1014

@GlodoUK GH105825

### DIFF
--- a/stock_picking_mass_action/wizard/mass_action.py
+++ b/stock_picking_mass_action/wizard/mass_action.py
@@ -74,6 +74,9 @@ class StockPickingMassAction(TransientModel):
                     lambda m: m.state not in ("done", "cancel")
                 )
             )
+            assigned_picking_lst = assigned_picking_lst.with_context(
+                button_validate_picking_ids=assigned_picking_lst.ids
+            )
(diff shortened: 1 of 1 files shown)


stock_picking_mass_action/wizard/mass_action.py
```diff
@@ -74,6 +74,9 @@ def mass_action(self):
                     lambda m: m.state not in ("done", "cancel")
                 )
             )
+            assigned_picking_lst = assigned_picking_lst.with_context(
```
rousseldenis, line 77: @theangryangel Thanks. Could you add a test for that behavior ?

PR #1284, opened 2023-05-01
[16.0][IMP] stock_no_negative: add option to skip negative qty check

AungKoKoLin1997:
Depended by https://github.com/OCA/manufacture/pull/1025
@qrtl

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -12,6 +12,11 @@ class StockQuant(models.Model):
 
     @api.constrains("product_id", "quantity")
     def check_negative_qty(self):
+        # To provide an option to skip the check when necessary.
+        # e.g. mrp_subcontracting_skip_no_negative - passes the context
+        # for subcontracting receipts.
+        if self.env.context.get("skip_negative_qty_check"):
+            return
         p = self.env["decimal.precision"].precision_get("Product Unit of Measure")
         check_negative_qty = (
             config["test_enable"] and self.env.context.get("test_stock_no_negative")


stock_no_negative/models/stock_quant.py
```diff
@@ -12,6 +12,10 @@ class StockQuant(models.Model):
 
     @api.constrains("product_id", "quantity")
     def check_negative_qty(self):
+        # If there is stock.move for subcontracting receipts, the context will be passed
+        # from mrp_subcontracting_skip_no_negative and pass the check for finished products.
```
yostashiro: Can we adjust the comment to avoid limiting the expected usage to the subcontracting receipt scenario? I'm not sure myself there is any other case where people might want to use this context, but let's keep the possibility open.

```suggestion
        # To provide an option to skip the check when necessary. e.g. mrp_subcontracting_skip_no_negative
        # passes the context for subcontracting receipts.
```

stock_no_negative/models/stock_quant.py
```diff
@@ -12,6 +12,10 @@ class StockQuant(models.Model):
...
+        if self.env.context.get("skip_negative_qty_check"):
```
rousseldenis, line 18: I'm not in favor putting another context key.

Could you remove config["test_enable"], keep only the "test_stock_no_negative" check and use it in your module?

Thanks

yostashiro: I don't think that works out nicely as it basically deactivates the negative quantity check except for tests, or am I missing something?

rousseldenis: Mmmh, yes I re-read that.

In fact, IMHO, those kind of behaviors should be deactivated by default. And activated in tests when wanting it by configuration.

So, my recommendation is to add a configuration parameter, then we can keep just the "skip_negative_check".

yostashiro: Thanks for the suggestion. I think I get the idea, however it would require another step of activating the function upon installation of the module, which might be seen as an unnecessary inconvenience by users. We would also need to add a migration script, I suppose. Therefore I am not very convinced with the approach of adding a config parameter. Would be good to hear opinions of other members.

rousseldenis: > Thanks for the suggestion. I think I get the idea, however it would require another step of activating the function upon installation of the module, which might be seen as an unnecessary inconvenience by users. We would also need to add a migration script, I suppose. Therefore I am not very convinced with the approach of adding a config parameter. Would be good to hear opinions of other members.

The problem with modules that change standard behavior is that can cause other modules 'not aware of those changes' to fail. This can lead to such 'hacky' code that rely on test flow ('test_enable' stuff).

This is antiprogramatic way of doing things.

IMHO a configuration parameter requires one click step(done once) but you can for instance activate it in your project managment stuff if you have it and users won't complain.

pedrobaeza: I agree with current approach and `config["test_enable"]` is a good compromise between integration and convenience, so please don't remove it.

yostashiro: Thanks for the comments. I'm personally more on the side of being pragmatic and keeping `config["test_enable"]` (I don't really see risks/inconveniences with this approach), though I also understand how it's seen as "anti-programmatic".

That said, can we leave this discussion outside the scope of this PR as it's not related to the point of the proposed improvement?

yostashiro (approved):

pedrobaeza (approved):

AungKoKoLin1997:
@pedrobaeza Can we get this PR merged?

PR #1289, opened 2023-05-10
[IMP][14.0][stock_plit_picking] Warn user if no spliting will happen

flotho:
if all the qty done are the same as the iniial, no spliting is done with the option on quantity done. A bit confusing.
This message inform the user of this behaviour

### DIFF
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -24,6 +24,19 @@ class StockPicking(models.Model):
                 )
             )
 
+        if all(
+            [
+                x.qty_done == x.product_uom_qty
+                for x in self.move_line_ids.filtered(lambda m: m.qty_done != 0.0)
+            ]
+        ):
+            raise UserError(
+                _(
+                    "All done quantities are equal to initial ones. You must choose "
+                    "another option in order to split your picking in several ones."
+                )
+            )
+            
     def split_process(self):
         """Use to trigger the wizard from button with correct context"""
         for picking in self:


stock_split_picking/models/stock_picking.py
```diff
@@ -24,6 +24,19 @@ def _check_split_process(self):
                 )
             )
 
+        if all(
```
rousseldenis, line 27: @flotho Could you add a test case for that ? Thanks

stock_split_picking/models/stock_picking.py
```diff
@@ -24,6 +24,19 @@ def _check_split_process(self):
...
+            [
+                x.qty_done == x.product_uom_qty
```
damdam-s, line 29: as `qty_done` and `product_uom_qty` are `float`, I would use the `float_compare` method from [odoo.tools](https://github.com/odoo/odoo/blob/14.0/odoo/tools/float_utils.py#L127)

stock_split_picking/models/stock_picking.py
```diff
@@ -24,6 +24,19 @@ def _check_split_process(self):
...
+                for x in self.move_line_ids.filtered(lambda m: m.qty_done != 0.0)
```
damdam-s, line 30: same remark here but for `float_is_zero` https://github.com/odoo/odoo/blob/14.0/odoo/tools/float_utils.py#L102

stock_split_picking/models/stock_picking.py
damdam-s: not blocking for this 😄

stock_split_picking/models/stock_picking.py
damdam-s: not blocking for this 😄

PR #1292, opened 2023-05-16
[14.0][IMP] delivery_procurement_group_carrier: Set Carrier on new picking from procurement.group

mt-software-de:
If run a procurement where the carrier is set.
I would be helpful, if the carrier on a new create picking is automatically set.


mt-software-de:
@jbaudoux could you do the review?

delivery_procurement_group_carrier/models/stock_move.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _get_new_picking_values(self):
+        values = super()._get_new_picking_values()
+        carrier = self.group_id.carrier_id
+        if carrier and self.picking_type_id.code in ["outgoing", "incoming"]:
+            values["carrier_id"] = carrier.id
```
rousseldenis, line 12: @mt-software-de Is it a real behavior you want? As Odoo behavior will be weird as it consider it should be only set on delivery pickings.

Shouldn't you use this instead ? https://github.com/OCA/delivery-carrier/blob/14.0/stock_picking_delivery_link/models/stock_picking.py#L24

mt-software-de: Yes it is a real behavior, we need for return pickings from a customer to our warehouse.

mt-software-de:
ping @jbaudoux

jbaudoux (approved):

PR #1297, opened 2023-05-21
[16.0][FIX] stock_picking_auto_create_lot: use standard method for lot val prep

yostashiro:
Before this commit, lot value preparation was not using the standard method, which would cause problems when there was another module trying to add more elements to vals extending the standard method.

For example, `product_expiry` wouldn't work with auto lot creation by this module.

https://github.com/odoo/odoo/blob/c3eae07259cb38656aa910ad98d0da04f4fc7edb/addons/product_expiry/models/stock_move_line.py#L66-L75

@qrtl


AungKoKoLin1997 (approved):
LGTM.

PR #1303, opened 2023-06-05
[FIX] stock_picking_start: Don't evaluate void browse records to boolean value

rousseldenis:
e.g.: If user_id is set to False and was not set before, the condition to check if record is modified would return True as res.users() != False


rousseldenis:
@lmignon

lmignon (approved):

sbejaoui (approved):

hparfr:
oops

/ocabot merge minor

PR #1316, opened 2023-06-23
[16.0][MIG] stock_picking_line_sequence

bosd:
Standard migration
stock_picking_line_sequence


rousseldenis (approved):
Code review

AungKoKoLin1997 (commented):
The function is not working as expected.
When we create transfer and then add new product line, the sequence is 1 for first line. But after clicked save button, it changes to 2.

This is before we save.

Screen Shot 2023-08-02 at 08 24 07

This is after we saved.

Screen Shot 2023-08-02 at 08 24 19

stock_picking_line_sequence/models/stock.py
```diff
@@ -0,0 +1,98 @@
...
+    max_line_sequence = fields.Integer(
+        string="Max sequence in lines", compute="_compute_max_line_sequence"
+    )
+
+    def _reset_sequence(self):
+        for rec in self:
+            current_sequence = 1
+            for line in rec.move_ids_without_package:
+                line.sequence = current_sequence
+                current_sequence += 1
```
AungKoKoLin1997: As far as I checked, calling `_reset_sequence` from `create` of stock_move will get two moves (one is virtual record, the other is real record). So, that will increase +1 to the sequence for read move record. May be this can be solved but not sure it is good way or not.
```
       for line in rec.move_ids_without_package:
            # Check if the record ID is an integer (real ID) or a string (virtual ID)
            if isinstance(line.id, int):
                line.sequence = current_sequence
                current_sequence += 1
```

bosd: Incorporated these changes.. it is fixed!

AungKoKoLin1997 (approved):

bosd:
Can we get a merge? 🙏

rousseldenis:
@bosd Could you amend your last commit message to add the module name and a better summary ?

image


Thanks

bosd:
tests are broken. It seems we need this now.
https://github.com/OCA/stock-logistics-workflow/pull/1329

AungKoKoLin1997:
@rousseldenis Can we get this PR merged?

stock_picking_line_sequence/models/stock.py
```diff
@@ -84,8 +84,10 @@ def _reset_sequence(self):
         for rec in self:
             current_sequence = 1
             for line in rec.move_ids_without_package:
-                line.sequence = current_sequence
-                current_sequence += 1
+                # Check if the record ID is an integer (real ID) or a string (virtual ID)
+                if isinstance(line.id, int):
```
rousseldenis, line 88: @bosd FYI, usually, we use NewId class instead.

So:

`if not isinstance(line.id, NewId):`

bosd:
@CasVissers-360ERP It's merged 😉

PR #1317, opened 2023-06-23
[15.0] Create module stock_move_consu_location_from_putaway

grindtildeath:
This module allows, for consumable products, to use the putaway location
as a source location of the stock move line, when the stock move
use a parent location of the putaway location as its source location.

As Odoo allows to define putaway rules for consumable products, incoming moves
will have the putaway applied to define the destination location of the
move lines, allowing to display this location in the Picking Operations report.

On outgoing moves though, as the stock levels are obviously not managed for
consumable products, there is no quantity to reserve in any location and the
Picking Operations report can only display the source location of the move.

With this module however, as the location of the putaway will be used as
the source location of the move line, the Picking Operations report will
display this location where the consumable product is supposed to be stored.

### DIFF
--- a/stock_move_consu_location_from_putaway/models/stock_move.py
+++ b/stock_move_consu_location_from_putaway/models/stock_move.py
@@ -1,0 +1,19 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
+        res = super()._prepare_move_line_vals(
+            quantity=quantity, reserved_quant=reserved_quant
+        )
+        if (
+            self.product_id.detailed_type == "consu"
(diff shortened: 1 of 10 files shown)


stock_move_consu_location_from_putaway/models/stock_move.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def _prepare_move_line_vals(self, quantity=None, reserved_quant=None):
+        res = super()._prepare_move_line_vals(
+            quantity=quantity, reserved_quant=reserved_quant
+        )
+        if (
+            self.product_id.detailed_type == "consu"
+            and self.location_id.usage == "internal"
+        ):
+            putaway_loc = self.location_id._get_putaway_strategy(self.product_id)
```
sebalix, line 17: We always get a result here? No need to check it?

sebalix (approved):
LG, one question however

stock_move_consu_location_from_putaway/models/stock_move.py
TDu: It looks like we do.

TDu (approved):

PR #1319, opened 2023-06-29
[16.0][ADD] stock_lot_price_unit

AungKoKoLin1997:
This module adds to the lot model the unit price field, which takes the value from the corresponding field of the stock move that has created the lot. 
The unit price of the lot is just for reference only. It does not reflect miscellaneous adjustments (e.g. variances between the receipt price and the vendor bill price, landed costs, etc.) that are reflected in the stock valuation layer.

@qrtl


yostashiro (approved):
Been in production use for a while now.

rousseldenis:
@AungKoKoLin1997  @yostashiro Mmmmh, I'm not sure you are in the right direction with this.

For me, it is the contrary. You can't change the quant price unit to a computed method based on lot value. This will lead to wrong stock valuations and migration problems.

As the lot can be concerned by several quants (and different values - on the quant), you can have several values for the same lot.

If you want to get the price of the first move (and so, the first quant created), you should compute it on the lot level instead (and name it 'first_price_unit' or something like that).

AungKoKoLin1997:
> For me, it is the contrary. You can't change the quant price unit to a computed method based on lot value. This will lead to wrong stock valuations and migration problems.

@rousseldenis If I am not mistaken, there is no price unit field in the quant and we added price unit field in this module.

> If you want to get the price of the first move (and so, the first quant created), you should compute it on the lot level instead (and name it 'first_price_unit' or something like that).

Yes. I agree with you.

PR #1322, opened 2023-06-30
[Add] stock_picking_account_date: new module that add accounting date…

alan196:
… to picking and pass it to the related account moves


stock_picking_account_date/models/stock_move.py
```diff
@@ -0,0 +1,77 @@
...
+            credit_account_id,
+            debit_account_id,
+            journal_id,
+            qty,
+            description,
+            svl_id,
+            cost,
+        )
+
+    def _get_price_unit(self):
```
yostashiro, line 35: Maybe you can do like this:

```python
    def _get_price_unit(self):
        self.ensure_one()
        accounting_date = self.picking_id.accounting_date
        if accounting_date:
            self = self.with_context(accounting_date=accounting_date)
        return super()._get_price_unit()
```

and then extend `_convert()` of `res.currency` in such a way that it overrides the date with `accounting_date` when it's available.

Wasn't sure if `date` of stock moves should also be the same as `accounting_date`.

stock_picking_account_date/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "name": "Stock Picking Account Date",
+    "summary": "Define account date in pickings",
+    "version": "15.0.1.0.0",
+    "category": "Stock",
+    "author": "Jarsa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "license": "LGPL-3",
+    "depends": [
+        "stock_account",
+        "purchase_stock",
```
yostashiro, line 14: Why should it depend on `purchase_stock`?

stock_picking_account_date/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2022, Jarsa Sistemas, S.A. de C.V.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
yostashiro, line 2: LGPL or AGPL?

yostashiro (commented):
There is https://github.com/OCA/stock-logistics-warehouse/pull/1711 for 16.0 for the same purpose.  We should find a way to make them consistent.

The first thing first, should the module be added in this repo or in stock-logistics-warehouse?  Wasn't very sure myself, but this repo might be more appropriate given that https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_move_backdating is in here.

cc @AungKoKoLin1997

stock_picking_account_date/__manifest__.py
alan196: You are correct, it's not neccesary

alan196: The dependency is necessary, for the method _get_price_unit if the purchase order is in a currency that is different from the company currency, the price must be updated to the exchange rate of the accounting_date.

I think this feature is not a must for this module, maybe we can split this functionality in another module "stock_picking_account_date_purchase" what do you thing @yostashiro ?

stock_picking_account_date/__manifest__.py
alan196: LGPL, I have updated the license in all files

alan196:
> There is [OCA/stock-logistics-warehouse#1711](https://github.com/OCA/stock-logistics-warehouse/pull/1711) for 16.0 for the same purpose. We should find a way to make them consistent.
> 
> The first thing first, should the module be added in this repo or in stock-logistics-warehouse? Wasn't very sure myself, but this repo might be more appropriate given that https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_move_backdating is in here.
> 
> cc @AungKoKoLin1997

I think the module belongs to stock-logistics-workflow repo, I already check the module of https://github.com/OCA/stock-logistics-warehouse/pull/1711, and it's almost the same with little changes, I'll add the feature to show the accounting_date only in incoming and outgoing moves 👍 

@AungKoKoLin1997 if you are ok, I can make a PR to your current PR branch in order to be consistent.

AungKoKoLin1997:
> I'll add the feature to show the accounting_date only in incoming and outgoing moves +1
> 
> @AungKoKoLin1997 if you are ok, I can make a PR to your current PR branch in order to be consistent.

@alan196 This feature is already done in my PR. Please check it https://github.com/qrtl/stock-logistics-warehouse/blob/6c86163dce7aca1b4b1934e53172d7598a0e2c3e/stock_picking_accounting_date/models/stock_picking.py#L16

AungKoKoLin1997:
@alan196 If you come across any potential improvements, please feel free to submit a Pull Request to our branch.

PR #1350, opened 2023-08-09
[15.0][ADD] stock_valuation_fifo_lot

ps-tubtim:
This module is used to calculate FIFO cost by lot.


manuelregidor:
@ps-tubtim I have checked the module and it works fine to me. However, I have found and issue that may be interesting to address. It happens with products that are in stock prior to the module installation.

1. Before installing the module, I have created a storable product with a unique serial number tracking and a FIFO costing method. I have purchased 4 units of this product for 2€ each. See a screenshot of the On hand product quantity and a screenshot of the Inventory valuation for this product.
on_hand
stock_valuation

2. I have installed the module.

3. I have created a sales order with the 4 units of the product, validated it and generate the picking. The picking contains the right serial numbers.
stock_picking

4. Once the picking is validate, the inventory valuation is updated. However, some numbers are not right:

a) Total Value is still 8 but it should be 0. 
stock_valuation_a

b) Remaining Qty in the incoming stock.valuation.layer is still 4 but it should be 0, which means that this layer will always be found as available when it is actually out of stock.
stock_valuation_b

c) Both the Unit Value and and the Total Value of the outgoing stock.valuation.layer are set to 0, but the values should be 2 and 8 respectively.
stock_valuation_c

I am aware that this situation only happens with products in stock before installing the module. The module works totally fine otherwise, but I was wondering if you are planning to fix this situation in the current version of the module.

Thank you very much for this contribution, it is truly appreciated.

manuelregidor (approved):
LGTM

ValentinVinagre (approved):
Nice work 😄

PR #1355, opened 2023-08-14
[15.0][ADD] stock_force_date

ps-tubtim:
This module allows to force inventory date.


rousseldenis:
@ps-tubtim Thanks for this. Have you looked at this https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_move_backdating ?

ps-tubtim:
> @ps-tubtim Thanks for this. Have you looked at this https://github.com/OCA/stock-logistics-workflow/tree/12.0/stock_move_backdating ?

@rousseldenis I think these 2 modules are quite similar. So, I have migrated the module stock_move_backdating instead.

ps-tubtim:
Move to https://github.com/OCA/stock-logistics-workflow/pull/1358

PR #1357, opened 2023-08-17
[16.0][IMP] stock_quant_package_dimension - include packaging weight

mmequignon:
Also include packaging's weight in package's estimated weight

### DIFF
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -20,7 +20,7 @@ class StockQuantPackage(models.Model):
     )
     estimated_pack_weight_kg = fields.Float(
         "Estimated weight (in kg)",
-        digits="Product Unit of Measure",
+        digits="Stock Weight",
         compute="_compute_estimated_pack_weight_kg",
         help="Based on the weight of the product.",
     )
--- a/stock_quant_package_dimension/models/stock_quant_package.py
+++ b/stock_quant_package_dimension/models/stock_quant_package.py
@@ -180,11 +180,12 @@ class StockQuantPackage(models.Model):
             picking_id
         )
         for package in self:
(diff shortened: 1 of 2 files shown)


stock_quant_package_dimension/models/stock_quant_package.py
```diff
@@ -180,11 +180,12 @@ def _compute_estimated_pack_weight_kg(self):
             picking_id
         )
         for package in self:
-            weight = 0.0
+            # Note: also include packaging weight
+            weight = package.product_packaging_id.weight
```
jbaudoux, line 184:
```suggestion
            weight = package.product_packaging_id.base_weight
```
It is `base_weight` on v16 https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_package_type.py#L22

jbaudoux (changes requested):

stock_quant_package_dimension/models/stock_quant_package.py
sebalix: It is `package.package_type_id.base_weight` or `package.product_packaging_id.package_type_id.base_weight` (NOTE: `package_type_id` is not required)

jbaudoux: For a package that you deliver, the type of package (chosen during the put in pack) is `package.package_type_id` and it's weight is given by `base_weight`

What is `package.product_packaging_id` ?

serCliff: Hello everyone,

After a thorough review of the issue, I am confident that the solution lies in utilizing `package.package_type_id.base_weight`. This is because the weight that needs to be added to the total weight corresponds to the weight of the specific package itself. For instance, if a pallet weighs 5 kgs, you must include this 5 kg in the total weight calculation.

PR #1358, opened 2023-08-17
[15.0][MIG] stock_move_backdating: Migration to 15.0

ps-tubtim:
- Move feature from `stock.inventory` and `stock.inventory.line` to `stock.quant` because `stock.inventory` and `stock.inventory.line` are removed
- Edit view of `stock.move.line` for show `date_backdating` field
- Allow create backorder


AungKoKoLin1997:
@ps-tubtim Can you update tests cases?

PR #926, opened 2021-11-04
[13.0][ADD] stock_receipt_lot_info

LoisRForgeFlow:
Description
---------------

This module allows to encode more information to create lot/serial numbers
while processing an incoming shipment.

By default, you can only provide a lot/serial number name
when receiving products tracked by lot/serial numbers, with this module you can also
provide the following information of the lot:

* Lot/Serial End of Life Date.
* Lot/Serial Best before Date.
* Lot/Serial Removal Date.
* Lot/Serial Alert Date.

To do list:
------------

* [ ] : Depends on this hook proposed to odoo: https://github.com/odoo/odoo/pull/79381.
* [ ] : Add tests

@ForgeFlow

PR #1363, opened 2023-08-24
[15.0][MIG] stock_receipt_lot_info

SergiCForgeFlow:
Supersedes [#982](https://github.com/OCA/stock-logistics-workflow/pull/982)


rousseldenis:
@SergiCForgeFlow Thanks for this. As @LoisRForgeFlow marked it here: https://github.com/OCA/stock-logistics-workflow/pull/926#issue-1044924600

Isn't it time to add tests ?

LoisRForgeFlow:
> @SergiCForgeFlow Thanks for this. As @LoisRForgeFlow marked it here: [#926 (comment)](https://github.com/OCA/stock-logistics-workflow/pull/926#issue-1044924600)
> 
> Isn't it time to add tests ?

@rousseldenis We did it in v16 https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_receipt_lot_info/tests/test_receipt_lot_info.py. v13 and v15 won't ever be merged becouse of the dependency to odoo PR. However, I'd like to keep PRs open because I use this module in this versions.

LoisRForgeFlow:
So, I just found Odoo merged another PR with the needed hook for this module... I  have updated the name of the method and added some tests. This is now ready to merge

JoanSForgeFlow (approved):
Works as expected!

rousseldenis:
@SergiCForgeFlow @LoisRForgeFlow IMHO, the module name is too vague as the USAGE.rst.

You say that this allows to enter 'info' about lots. But that's not true. This is used to enter the different dates (removal, alert, best before).

In addition, I would say those information should be computed automatically depending on the configuration on the product level if the product is based on expiration date (checkbox).

What I understand about the need of this is in some implementations, it is not possible to configure all the product alerts times and you want to fill it in on reception step. That should be explained either in CONTEXT or DESCRIPTION fragment.

Thanks

LoisRForgeFlow:
> @SergiCForgeFlow @LoisRForgeFlow IMHO, the module name is too vague as the USAGE.rst.
> 
> You say that this allows to enter 'info' about lots. But that's not true. This is used to enter the different dates (removal, alert, best before).
> 
> In addition, I would say those information should be computed automatically depending on the configuration on the product level if the product is based on expiration date (checkbox).
> 
> What I understand about the need of this is in some implementations, it is not possible to configure all the product alerts times and you want to fill it in on reception step. That should be explained either in CONTEXT or DESCRIPTION fragment.
> 
> Thanks

Hi @rousseldenis 

I have aligned the readme to be more accurate in the comparison with current standard behavior.

Does this make more sense now?

JordiMForgeFlow (approved):
Code review LGTM

PR #1364, opened 2023-08-24
[15.0][ADD] stock_no_negative_move_line

gabriel-grinspan:


stock_no_negative_move_line/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
rousseldenis: @gabriel-grinspan The manifest has wrong format.

See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

rousseldenis:
@gabriel-grinspan Could you fix pre-commit ? I suggest you to install it locally before committing.

gabriel-grinspan:
@rousseldenis I ran the tests in Odoo locally and it seems to have worked. How do I run the pre-commit check locally?

AungKoKoLin1997:
> How do I run the pre-commit check locally

You can reference from this https://pre-commit.com/ .

pedrobaeza:
I created a specific OCA page for that: https://github.com/OCA/maintainer-tools/wiki/Install-pre-commit

AungKoKoLin1997:
@gabriel-grinspan  This module is the same feature with [stock_no_negative](https://github.com/qrtl/stock-logistics-workflow/tree/15.0/stock_no_negative), isn't?

rousseldenis:
> @gabriel-grinspan This module is the same feature with [stock_no_negative](https://github.com/qrtl/stock-logistics-workflow/tree/15.0/stock_no_negative), isn't?

@gabriel-grinspan Following your module description, this seems indeed the same behavior as stock_no_negative.

So, you should be able to explain what is the advantage of this module (as if I understand well, is an extension of stock_no_negative).

Thanks

gabriel-grinspan:
@AungKoKoLin1997 @rousseldenis
The purpose of this module is to add an additional validation to the flow.
Currently, when just using the stock_no_negative app, the user is blocked on validation of a picking. When using this app, the check is added to just setting the done quantity on the move line, no need to wait for validation. This will allow the user to know immediately when they have made an error.

gabriel-grinspan:
> @gabriel-grinspan Could you fix pre-commit ? I suggest you to install it locally before committing.

@rousseldenis It looks like the pre-commit is removing the tests file from the init. This seems like an error to me. I see that other OCA modules have the same import. Additionally, it seems that this is why the checks indicate that my code is not mostly covered by the tests.

AungKoKoLin1997:
@gabriel-grinspan Can you please run pre-commit in your local env? Then, please fix some suggestions from pre-commit log.
In your case, pre-commit is not removing the tests file from the init, you need to add newline at the end of file (read the suggestions carefully). Most of the code format will fix automatically by running pre-commit and some are needed to fix manually.

PR #1375, opened 2023-09-07
[14.0][ADD] add stock_picking_auto_create_lot_qty

aayartsev:
Allow create lots for every n count of product


stock_picking_auto_create_lot_qty/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+        string="Create lot every",
+        help="Create lot every n uom - here user can"
+        " set a value and select one of the UoMs available"
+        " for purchase (if value = 0, behavior stays the"
+        " same as base module)",
+    )
+
+    only_multiples_allowed = fields.Boolean(
+        string="Only multiples allowed",
+        help="This flag determines behavior in picking (more on this later)",
```
HekkiMelody:
```suggestion
        help="Blocks automatic creation of lots if quantity received is not a multiple of the selected value",
```

stock_picking_auto_create_lot_qty/models/stock_move_line.py
```diff
@@ -0,0 +1,35 @@
...
+            else:
+                lots_by_product[lot.product_id.id] += lot
+        for line in self:
+            lot = first(lots_by_product[line.product_id.id])
+            lots_by_product[line.product_id.id] = (
+                lots_by_product[line.product_id.id] - lot
+            )
+            line.lot_id = lot
+            if lot.product_id.tracking == "serial":
+                lots_by_product[line.product_id.id] -= lot
```
HekkiMelody: This is almost the same as `stock_picking_auto_create_lot` except for lines 30-32.

Is there a reason I'm not seeing for not running super() and adding only these lines at the end?

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
...
+        pickings = self.filtered(lambda p: p.picking_type_id.auto_create_lot)
+        lines = pickings.mapped("move_line_ids").filtered(
+            lambda x: (
+                not x.lot_id
+                and not x.lot_name
+                and x.product_id.tracking != "none"
+                and x.product_id.auto_create_lot
+            )
+        )
+        new_lines = pickings.mapped("move_line_ids").filtered(lambda x: (x.qty_done))
```
HekkiMelody:
```suggestion
        new_lines = pickings.mapped("move_line_ids").filtered(lambda x: x.qty_done)
```
IMO this improves clarity (at first I thought this was a tuple)

stock_picking_auto_create_lot_qty/tests/common.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+
+class CommonStockPickingAutoCreateLot(object):
```
HekkiMelody:
```suggestion
class CommonStockPickingAutoCreateLotQty(object):
```

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo.exceptions import UserError
+from odoo.tests import SavepointCase
+
+from .common import CommonStockPickingAutoCreateLot
+
+
+class TestStockPickingAutoCreateLot(CommonStockPickingAutoCreateLot, SavepointCase):
```
HekkiMelody:
```suggestion
class TestStockPickingAutoCreateLotQty(CommonStockPickingAutoCreateLotQty, SavepointCase):
```

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
...
+                            " not allow for automatic lot creation"
+                        )
+                    )
+                if (
+                    product_id.create_lot_every_n
+                    and not product_id.only_multiples_allowed
+                ):
+                    every_n = product_id.product_tmpl_id.create_lot_every_n
+
+                    count_of_lots = int(current_product_qty // every_n)
```
aleuffre, line 50:
```suggestion
                    count_of_lots = current_product_qty // every_n
```

Floor division already returns an integer

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _set_auto_lot(self):
```
HekkiMelody, line 112: Once again, unless there's a strong reason not to, I think it'd be best to split `move_line_ids` by quantity here and then call super() in order to better integrate with other modules.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
...
+            )
+        )
+        new_lines = pickings.mapped("move_line_ids").filtered(lambda x: (x.qty_done))
+        if not new_lines:
+            count_by_product = {}
+            for line in lines:
+                if line.product_id not in count_by_product:
+                    count_by_product[line.product_id] = line.product_uom_qty
+                else:
+                    count_by_product[line.product_id] += line.product_uom_qty
```
HekkiMelody: An easier way of doing this would be

```suggestion
            count_by_product = dict.fromkeys(lines.mapped("product_id"), 0)
            for line in lines:
                count_by_product[line.product_id] += line.product_uom_qty
```

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
...
+                    product_id.batch_uom_id or product_id.uom_id,
+                    round=False,
+                )
+                if product_id.create_lot_every_n and product_id.only_multiples_allowed:
+                    raise UserError(
+                        _(
+                            "The quantity received for the product does"
+                            " not allow for automatic lot creation"
+                        )
+                    )
```
HekkiMelody: Shouldn't this error only be raised if the `current_product_qty` is not a multiple of `create_lot_every_n`?

HekkiMelody (changes requested):

stock_picking_auto_create_lot_qty/models/stock_move_line.py
aayartsev: If I do as you say, we will get following situation:
- in super will created lots
- for all stock move lines will be assigned only one lot, other lots will be not assgned to any stock move line
- i can not find reliable way to find other lots that are not assigned to stock move line wich are created in super, may be you can give me a tip with it.

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: `    for _new_line_count in range(count_of_lots):
TypeError: 'float' object cannot be interpreted as an integer`

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: If this function worked with single record i can accept your suggestion, but it is always work with multiple records, and if i will call super(), we will have undisireable result

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: your suggestion will work only in case were we already have any value in dict by this key. If we create key we need to set value first and then we can add something to this value.

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: I got this case from task.

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: done

stock_picking_auto_create_lot_qty/tests/common.py
aayartsev: done

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
aayartsev: done

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: fixed

HekkiMelody (changes requested):
Functional comment:

- Auto creation of lots every N should only be enabled if we're tracking by lot and not by unique serial number. 
  - if trying to assign the same serial number to multiple objects, an error is always raised.
  - when changing the `tracking` field an `onchange` should clean the field `create_lot_every_n` accordingly
    - if `none` or `serial`: create_lot_every_n = 0 or empty
    - if `lot`: any value is accepted
  - the fields you added to the product page should only be showed if `tracking` ==`lot`

stock_picking_auto_create_lot_qty/models/stock_move_line.py
HekkiMelody: The more I think about it the more I think we can probably just fix this function in the base module directly. I've also done some light testing of this change, and it seems fine. It also doesn't break any automated tests.

When `_set_auto_lot` is executed, we already have the appropriate `stock.move.line` objects (split in individual objects in case they're tracked by serial. So really what we want to do is just assign one lot to each line, in order.

In a separate commit, you can change these two lines by removing the if (and indenting appropriately)

https://github.com/OCA/stock-logistics-workflow/blob/14.0/stock_picking_auto_create_lot/models/stock_move_line.py#L38-L39

This should allow you to completely remove this entire file in your PR.

stock_picking_auto_create_lot_qty/models/stock_picking.py
HekkiMelody: After some exploration, I realized that the method responsible for splitting a picking into separate move lines is `_action_assign`

https://github.com/OCA/OCB/blob/14.0/addons/stock/models/stock_move.py#L1310

So the splitting should be done there IMO. It should be easier to split the moves you need (only if they're tracked by lot and meet other conditions) and then call super() so as to be less destructive.

stock_picking_auto_create_lot_qty/models/stock_picking.py
HekkiMelody: Isn't that what `dict.fromkeys` would do? Anyway, I thought it would be a bit cleaner, but it's not a big deal regardless.

HekkiMelody:
When you're finished, please use `oca-gen-addon-readme` to generate the readme for your module. https://github.com/OCA/maintainer-tools/blob/master/README.md#readme-generator

Though I wonder if this could be an [IMP] to the base module instead of its own module...

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: done

stock_picking_auto_create_lot_qty/models/stock_picking.py
aayartsev: done

stock_picking_auto_create_lot_qty/models/stock_move_line.py
aayartsev: done

stock_picking_auto_create_lot_qty/models/product_template.py
```diff
@@ -0,0 +1,41 @@
...
+        help="Blocks automatic creation of lots if quantity"
+        " received is not a multiple of the selected value",
+    )
+
+    @api.onchange("tracking")
+    def _onchange_tracking(self):
+        if self.tracking and self.tracking != "lot":
+            self.create_lot_every_n = False
+        if not self.tracking:
+            self.create_lot_every_n = False
```
HekkiMelody:
```suggestion
        if self.tracking != "lot":
            self.create_lot_every_n = False
```

HekkiMelody (commented):
Code review: 

- You can do it all in the end, but do remember to only have one commit per module you're changing (so first you'd have one commit with the fix to "stock_picking_auto_create_lot" and one commit for ""stock_picking_auto_create_lot_qty" and to follow OCA conventions in the commit message https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

- As pointed out earlier, the readme is also still missing (but again, you can do it all in the end if you prefer)

- Code looks ok except the one point below. I also still have doubts about the overwriting of `_set_auto_lot` and I don't necessarily think that's the correct place for that operation. Maybe someone else who is familiar with inventory and stock can comment on it and point us in the right direction

GabbasovDinar (approved):
please, squash commits by modules

stock_picking_auto_create_lot_qty/models/stock_picking.py
francesco-ooops: @dessanhemrayev @GabbasovDinar what about this?

dessanhemrayev: the "_action_assign" method is launched at the very end, and before that the "set_lot_auto" method and its subtypes are triggered, so I think everything is correct here

GabbasovDinar (approved):

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        self.assertEqual(len(lot), 2)
+
+    def test_multiples_allowed_true(self):
+        self._create_picking()
+        self._create_move(
+            product=self.product_serial_auto_qty_5_multiples_allowed_true, qty=8.0
+        )
+        # check for error
+        self.assertRaises(UserError, self.picking.button_validate)
```
HekkiMelody: Please add to this test a move with qty 15 and check that it works correctly

dessanhemrayev: Thanks for the feedback, I will of course add a test so that all cases are covered by tests.

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
```diff
@@ -58,12 +58,22 @@ def test_multiples_allowed_false(self):
...
+
+    def test_multiples_allowed_true_2(self):
+        self._create_picking()
+        self._create_move(
+            product=self.product_serial_auto_qty_5_multiples_allowed_true, qty=15.0
+        )
+        # check for error
+        # reminder = 0
         self.assertRaises(UserError, self.picking.button_validate)
```
HekkiMelody: question: Maybe I'm misunderstanding something, but

- multiples of 5 are allowed
- the quantity is 15
- this shouldn't raise any error

no?

dessanhemrayev: You're right 
Thanks, I'll fix it

HekkiMelody (approved):
Code review, LGTM

dessanhemrayev:
@francesco-ooops 
@rousseldenis 
@pedrobaeza
please review)

Bearnard21 (changes requested):
Whenever it's non-multiple to the settings amounts of goods are being ordered it's impossible to validate the delivery. 

Please try the  following tests:

Test 1
Create a new product and set the following properties:
* Create lot every 5 units
* Only multiples allowed = False
Create a purchase order for this product with a quantity of 8. Confirm the purchase order and validate the picking.
Expected behaviour:
Two lots will be created: one for 5 units and one for 3 units.
Actual result:
You will get User Error: You need to supply a Lot/Serial number for products "porduct_id".


Test 2
Create a new product and set the following properties:
* Create lot every 5 units
* Only multiples allowed = True
Create a purchase order for this product with a quantity of 8. Confirm the purchase order and validate the picking.
Expected behaviour:
An error will be raised stating that the quantity received for the product does not allow for automatic lot creation.
Actual result:
You will get User Error: You need to supply a Lot/Serial number for products "porduct_id".

Bearnard21 (approved):
Functionality LGTM

stock_picking_auto_create_lot_qty/models/product_template.py
```diff
@@ -0,0 +1,39 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    batch_uom_id = fields.Many2one(
+        string="Uom",
+        help="Uom",
```
geomer198: Uninformative help description.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,68 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _set_auto_lot(self):
+        """
+        Allows to be called either by button or through code
+        """
+        pickings = self.filtered(lambda p: p.picking_type_id.auto_create_lot)
+        lines = pickings.mapped("move_line_ids").filtered(
```
geomer198: I think better use one cycle for preparing value for `lines` and `new_lines` variables.
```py
lines = self.env[...].browse()
new_lines = self.env[...].browse()
for line in pickings.mapped("move_line_ids"):
    if not line.lot_id and not line.lot_name and line.product_id.tracking != 'none' and line.product.auto_create_lot:
        lines |= line
    if line.qty_done:
        new_lines |= line
````
This solution use only one cycle and your solution use two cycle.

File: stock_picking_auto_create_lot_qty/models/stock_picking.py [hunk 15]
```diff
@@ -0,0 +1,68 @@
...
+            lambda x: (
+                not x.lot_id
+                and not x.lot_name
+                and x.product_id.tracking != "none"
+                and x.product_id.auto_create_lot
+            )
+        )
+        new_lines = pickings.mapped("move_line_ids").filtered(lambda x: x.qty_done)
+        if not new_lines:
```
geomer198: I think here better use this solution:
```
if new_lines:
    return
count_by_product = dict.fromkeys(lines.mapped("product_id"), 0)
...
```

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _set_auto_lot(self):
```
geomer198, line 112: I think this function is too complex!
Could split method logic in separate methods?

stock_picking_auto_create_lot_qty/tests/common.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+class CommonStockPickingAutoCreateLotQty(object):
```
geomer198: Why would inheriting from `object`?
Class can inheriting from SavepointCase.

stock_picking_auto_create_lot_qty/tests/common.py
```diff
@@ -0,0 +1,93 @@
...
+
+        cls.stock_serial_sequence_2 = cls.env.ref(
+            "stock.sequence_production_lots"
+        ).copy()
+        cls.stock_serial_sequence_2.prefix = "TEST"
+        cls.company_1 = cls.env.ref("base.main_company")
+        cls.company_2 = cls.env.ref("stock.res_company_1")
+
+    @classmethod
+    def _create_product(
```
geomer198, line 28: Could you add docstrings?

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
```diff
@@ -0,0 +1,153 @@
...
+        # * tracking "lot" to "none"
+        cls.product_no_tracking = cls._create_product(
+            tracking="lot", auto=True, every_n=5, multiple_allow=False
+        )
+        # * tracking every_n = 0
+        cls.product_0_every_n = cls._create_product(
+            tracking="lot", auto=True, every_n=0, multiple_allow=False
+        )
+
+    def test_multiples_allowed_false(self):
```
geomer198, line 34: Could you add docstrings?

geomer198 (changes requested):
Please fix code by my comments

stock_picking_auto_create_lot_qty/models/stock_picking.py
dessanhemrayev: Thanks for hints) I'll fix it

stock_picking_auto_create_lot_qty/models/stock_picking.py
dessanhemrayev: Thanks for hints) I'll fix it

stock_picking_auto_create_lot_qty/models/product_template.py
dessanhemrayev: I'll fix it

stock_picking_auto_create_lot_qty/models/stock_picking.py
dessanhemrayev: I'll fix it

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
dessanhemrayev: I'll fix it

stock_picking_auto_create_lot_qty/tests/common.py
dessanhemrayev: I'll fix it

stock_picking_auto_create_lot_qty/tests/common.py
dessanhemrayev: I'll fix it

stock_picking_auto_create_lot/models/stock_move_line.py
```diff
@@ -35,5 +35,4 @@ def set_lot_auto(self):
         for line in self:
             lot = first(lots_by_product[line.product_id.id])
             line.lot_id = lot
-            if lot.product_id.tracking == "serial":
-                lots_by_product[line.product_id.id] -= lot
+            lots_by_product[line.product_id.id] -= lot
```
rousseldenis: @aayartsev Wasn't this case covered by a test ?

stock_picking_auto_create_lot_qty/models/product_template.py
```diff
@@ -8,7 +8,7 @@ class ProductTemplate(models.Model):
 
     batch_uom_id = fields.Many2one(
         string="Uom",
-        help="Uom",
+        help="Units of measure",
```
geomer198: In help should be contain description about destination this field.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -1,68 +1,134 @@
...
-from odoo import _, models
+from odoo import _, api, models
 from odoo.exceptions import UserError
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    @api.model
+    def _get_stock_move_lines(self, pickings):
```
geomer198: Method name is not shown what do you do in this method.
Which lines you get? Why do you return tupple of lines and new lines?

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -1,68 +1,134 @@
...
+        """Get stock move lines for the given pickings, and the new ones
```
geomer198: Doc string too not shown what is method do.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -1,68 +1,134 @@
...
+            pickings (recordset): Recordset of stock pickings for which to retrieve
+                                  stock move lines
+
+        Returns:
+            tuple (lines, new_lines): A tuple containing two recordsets - lines and new_lines.
+                 - lines: Existing stock move lines associated with the pickings.
+                 - new_lines: New stock move lines that need to be created.
+
+        """
+        lines = new_lines = self.env["stock.move.line"].browse()
```
geomer198: line and new_line has the same link to object or separate links?
If links is the same then this solution isn't correct. Please add tests for the method.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -1,68 +1,134 @@
...
+            line (stock.move.line): Stock move line to use as a template.
+            current_product_qty (int): Quantity of the product to generate lots for.
+
+        Returns:
+            lines (recordset): A set of new stock move lines representing the
+                                   generated lots.
+        """
+        lines = self.env["stock.move.line"].browse()
+        every_n = product_id.product_tmpl_id.create_lot_every_n
+        count_of_lots = int(current_product_qty // every_n)
```
geomer198, line 94: If `every_n` is zero then raised ZeroDivisionError. Please fix it and add tests for this method.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -1,68 +1,134 @@
...
+            count_by_product (dict): Dictionary with product IDs as keys and number of lots
+                                    to generate as values.
+        """
+        count_by_product = dict.fromkeys(lines.mapped("product_id"), 0)
+        for line in lines:
+            count_by_product[line.product_id] += line.product_uom_qty
+        return count_by_product
+
+    @api.model
+    def _generate_lot_every_n(self, product_id, line, current_product_qty):
```
geomer198: The same problem. I don't understand what is method do by name and doc string.

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -1,68 +1,134 @@
...
-        lines = pickings.mapped("move_line_ids").filtered(
-            lambda x: (
-                not x.lot_id
-                and not x.lot_name
-                and x.product_id.tracking != "none"
-                and x.product_id.auto_create_lot
+        lines, new_lines = self._get_stock_move_lines(pickings)
+        if new_lines:
+            return
+        count_by_product = self._compute_count_by_product(lines)
```
geomer198: Why method name `_compute`.  Better use name `_prepare_count_by_products`,

stock_picking_auto_create_lot_qty/tests/common.py
```diff
@@ -1,8 +1,10 @@
 # Copyright (C) 2023 Cetmix OÜ
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from odoo.tests import SavepointCase
 
-class CommonStockPickingAutoCreateLotQty(object):
+
+class CommonStockPickingAutoCreateLotQty(SavepointCase):
```
geomer198: Class name should by contain `Common` in class name.
For example:
```py
class CommonStockPickingCommon(...):
```

geomer198 (changes requested):
Please fix code by my comments, add tests for new methods.

stock_picking_auto_create_lot/models/stock_move_line.py
dessanhemrayev: Thanks for feedback
I checked and opened new PR #1566 )

geomer198 (approved):
LGTM!

Bearnard21 (approved):
Functionality Ok

Bearnard21 (approved):
LGTM

stock_picking_auto_create_lot_qty/models/stock_picking.py
```diff
@@ -135,9 +117,9 @@ def _set_auto_lot(self):
         lines, new_lines = self._split_stock_move_lines(pickings)
         if new_lines or not lines:
             return
-        count_by_product = self._prepare_count_by_products(lines)
-        line = lines[0]
-        for product_id, product_qty in count_by_product.items():
+        for line in lines:
+            product_id = line.product_id
```
geomer198, line 121: Better use `product` variable name because `product_id` is  id for product.

stock_picking_auto_create_lot_qty/tests/test_stock_picking_auto_create_lot_qty.py
```diff
@@ -203,12 +203,10 @@ def test_prepare_stock_move_lines_for_lots_2(self):
         for creating lots, including validation checks and assertions on the generated lots.
         """
         self._create_picking()
-        product = self.product_serial_auto_qty_5_multiples_allowed_false
+        product_id = self.product_serial_auto_qty_5_multiples_allowed_false
```
geomer198, line 206: The same

geomer198 (approved):
LGTM

francesco-ooops:
@rousseldenis good for you? :)

francesco-ooops:
@pedrobaeza can it be merged? thanks!

stock_picking_auto_create_lot_qty/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Picking Auto Create Lot Quantity",
+    "summary": "Auto batch generation by quantity",
+    "version": "14.0.1.0.0",
+    "development_status": "Production/Stable",
+    "category": "stock",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Ooops, Cetmix OÜ, Odoo Community Association (OCA)",
```
pedrobaeza, line 11: `Cetmix OÜ`?

francesco-ooops: @pedrobaeza that's their legal name: https://www.odoocompanies.com/c-cetmix-o

pedrobaeza:
Merging blindly:

/ocabot merge nobump

PR #1377, opened 2023-09-08
[16.0][ADD] stock_move_zone_location_source

rousseldenis:


stock_move_zone_location_source/models/stock_location.py
```diff
@@ -0,0 +1,25 @@
...
+        help="Check this to consider this picking type as the source one for "
+        "the moves later in the flow.",
+    )
+
+    @api.depends("is_zone")
+    def _compute_is_considered_as_source(self):
+        # Reset value if is_zone field is set to False
+        self.filtered(lambda location: location.is_zone).update(
+            {"is_considered_as_source": False}
+        )
```
jbaudoux, line 25: In fact, you reset value when is_zone is set to True. You inverted the filter domain

jbaudoux (changes requested):

jbaudoux:
From your explanation, you try to build a reporting based on a package type category and not from where it has been picked.
You have a package type on a delivery package and you can also set a package type on a product (https://github.com/OCA/wms/blob/16.0/stock_storage_type/models/product_template.py)
In your approach, you cannot have multiple picking zones for a same product/package category (e.g. food)

rousseldenis:
> From your explanation, you try to build a reporting based on a package type category and not from where it has been picked. You have a package type on a delivery package and you can also set a package type on a product (https://github.com/OCA/wms/blob/16.0/stock_storage_type/models/product_template.py) In your approach, you cannot have multiple picking zones for a same product/package category (e.g. food)

@jbaudoux Nope. Nothing to do with a package type category.

This is only to find the REAL origin (for a particular move) of a product move. This will retrieve only the location origin.

To find it, we look behind and take the first location we find in the chain that has these characteristics:

- It is a zone location
- It is marked 'Considered as source'

So, if the product comes at T time from Food A zone, it will be marked coming from there. If product configuration has changed and comes from Food B zone, the next shipping will mention that one.

jbaudoux:
Yes I understand what your module is doing. But I don't think the report you build should be based on this module.

Because the information serves a report that is for the transporter. And I think he doesn't care from where in your stock the product is coming. He needs to know what type of package/product to load in the truck and unload at customer. E.g. 2 boxes of drugs, 3 bags of food and 1 box of hardware. He is interested by the package type as this is what he handles.

Just a design remark ;)

PR #1332, opened 2023-07-20
[16.0][mig] stock_partner_delivery_window

aktiv-heli-kantawala:

PR #1378, opened 2023-09-19
[16.0][ADD] stock_partner_delivery_window

nguyenminhchien:
To replace:
* https://github.com/OCA/stock-logistics-workflow/pull/1332


cyrilmanuel (approved):
not an expert but LGTM

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -98,61 +95,6 @@ def is_in_delivery_window(self, date_time):
                     return True
         return False
 
-    def _get_delivery_time_format_string(self):
-        return _("From %s to %s")
-
-    def get_delivery_time_description(self):
```
jbaudoux: Why have you dropped those 2 methods ?

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo-test-helper
+odoo-addon-stock_move_free_reservation_reassign @ git+https://github.com/OCA/stock-logistics-workflow.git@refs/pull/1379/head#subdirectory=setup/stock_move_free_reservation_reassign
```
jbaudoux: This can now be dropped

jbaudoux (changes requested):
The 2 dropped methods on `res_partner` should be recovered

stock_partner_delivery_window/models/res_partner.py
nguyenminhchien: @jbaudoux i removed them because i found no place call them. please confirm if you still want to recover them.

nilshamerlinck: this is used in some customer modules for v14, better keep it indeed as likely to be needed when migrating to v16

test-requirements.txt
nguyenminhchien: @jbaudoux : dropped

jbaudoux (approved):

simahawk (approved):
LG

PR #1379, opened 2023-09-20
[16.0][FIX] cannot re-assign move after applying inventory

nguyenminhchien:
Related issue: https://github.com/OCA/stock-logistics-workflow/issues/1380

### DIFF
--- a/stock_move_free_reservation_reassign/models/stock_quant.py
+++ b/stock_move_free_reservation_reassign/models/stock_quant.py
@@ -25,5 +25,7 @@ class StockQuant(models.Model):
         ]._get_move_free_reservation_ids() as move_to_reassign_ids:
             res = super()._apply_inventory()
         if move_to_reassign_ids:
-            self.env["stock.move"].browse(move_to_reassign_ids)._action_assign()
+            self.env["stock.move"].browse(move_to_reassign_ids).with_context(
+                inventory_mode=False
+            )._action_assign()
         return res


rousseldenis:
@nguyenminhchien LGTM. Could you maybe add a test that explain that change ?

nguyenminhchien:
@rousseldenis : updated.

stock_move_free_reservation_reassign/models/stock_quant.py
```diff
@@ -25,5 +25,7 @@ def _apply_inventory(self):
         ]._get_move_free_reservation_ids() as move_to_reassign_ids:
             res = super()._apply_inventory()
         if move_to_reassign_ids:
-            self.env["stock.move"].browse(move_to_reassign_ids)._action_assign()
+            moves = self.env["stock.move"].browse(move_to_reassign_ids)
+            if not self.env.context.get("assign_in_inventory_mode"):
```
rousseldenis: @nguyenminhchien I don't get why you need a second context parameter.

The ir.config_parameter is sufficent.

nguyenminhchien: it's painful to add the context "assign_in_inventory_mode". Remove inventory_mode (`inventory_mode=False`) is a must to keep this module works, making the current tests green are enough to explain the change.
So, i will revert this change, no need more unittest, what do you think?

rousseldenis (approved):

lmignon (approved):

PR #1385, opened 2023-09-28
[16.0][IMP] stock_owner_restriction: increase method coverage

AungKoKoLin1997:
Extend following methods for the scenario where owner restriction should be forced:

- _update_available_quantity() (stock.quant)
- _update_reserved_quantity() (stock.quant)

Split lines to identify the owner_restriction setting into separate method for better extensibility.

Fix read_group() (stock.quant) which was incorrectly assigning the owner record in the domain instead of the id.

@qrtl

### DIFF
--- a/stock_owner_restriction/models/stock_quant.py
+++ b/stock_owner_restriction/models/stock_quant.py
@@ -1,5 +1,6 @@
 # Copyright 2020 Carlos Dauden - Tecnativa
 # Copyright 2020 Sergio Teruel - Tecnativa
+# Copyright 2023 Quartile Limited
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import api, models
 from odoo.osv import expression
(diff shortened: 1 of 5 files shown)


rousseldenis:
@AungKoKoLin1997 Seems weird that the coverage has decreased... 

See https://app.codecov.io/gh/OCA/stock-logistics-workflow/pull/1385?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=OCA

Could you add tests cases for those?

AungKoKoLin1997:
After added tests case, the coverage has increased a little bit.

stock_owner_restriction/models/stock_quant.py
```diff
@@ -25,20 +26,33 @@ def _gather(
             owner_id=owner_id,
             strict=strict,
         )
-        restricted_owner_id = self.env.context.get("force_restricted_owner_id", None)
-        if owner_id is None or restricted_owner_id is None:
+        restricted_owner = self.env.context.get("force_restricted_owner_id", None)
```
rousseldenis, line 29: @AungKoKoLin1997 I am wondering why this context key has been added. As there is one in stock module: "owner_id".

IMHO, if we want to force the owner, we should use that one instead. This has the advantage of removing a lot of not necessary code.

Could you explain ? Thanks

yostashiro: > @AungKoKoLin1997 I am wondering why this context key has been added. As there is one in stock module: "owner_id".

@rousseldenis I was also thinking about this with no definite answer. My tentative interpretation is to avoid triggering unexpected behaviors by sharing the key with the standard logic, but I'm not sure it's a valid concern.

@sergio-teruel Would be able to explain how you decided to use the `force_restricted_owner_id` context key?

AungKoKoLin1997: @rousseldenis 
In my opinion, adding force_restricted_owner_id alongside owner_id is a strategic choice for improving how we manage stock. owner_id simply identifies the owner, but force_restricted_owner_id gives us a nuanced way to control actions based on ownership, enhancing security. Plus, keeping these keys separate avoids confusion with other parts of the system using owner_id.

rousseldenis:
@AungKoKoLin1997 Could you squash your commits ? Thanks

AungKoKoLin1997:
@rousseldenis  Done!

stock_owner_restriction/models/stock_quant.py
```diff
@@ -25,20 +26,33 @@ def _gather(
...
+                        domain,
+                        [
+                            (
+                                "owner_id",
+                                "=",
+                                restricted_owner and restricted_owner.id or False,
+                            )
+                        ],
+                    ]
+                )
```
yostashiro: Let's save a few lines.

```suggestion
                owner_domain = [
                    ("owner_id", "=", restricted_owner and restricted_owner.id or False),
                ]
                domain = expression.AND([domain, owner_domain])
```

rousseldenis:
I need to fix the tests in another module as Odoo changed behaviors in core...

rousseldenis:
@AungKoKoLin1997 Could you rearrange a little bit your commit messages in order to have them significant?

yostashiro (approved):
Code review.

PR #1388, opened 2023-10-02
[IMP] stock_picking_invoice_link: add pre-migration script to speed up installation.

JordiBForgeFlow:


stock_picking_invoice_link/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2023 ForgeFlow, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    logger = logging.getLogger(__name__)
```
pedrobaeza, line 9: Instead of populating a logger, just use `openupgrade.logged_query`

stock_picking_invoice_link/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,39 @@
...
+    logger.info("Creating column: delivery_count for table account_move")
+    cr.execute(
+        """
+    ALTER TABLE account_move
+        ADD COLUMN IF NOT EXISTS delivery_count INTEGER;
+    """
+    )
+    logger.info("Populating values for column delivery_count in table account_move")
+    cr.execute(
```
pedrobaeza, line 18: Please use `openupgrade.logged_query`. It's very important to log the number of changes done for traceability.

pedrobaeza (changes requested):
These are OpenUpgrade migration scripts, not general ones, so `openupgradelib` is mandatory in that cases. If you want to use them with enterprise, just add the library when doing the post steps.

legalsylvain (commented):
Pre migration script are not executed when installing module. 

To speed up installation you have to put a pre init script.

Dont you think ?

PR #1391, opened 2023-10-03
[ADD] new module stock_low_show_qty_location

HekkiMelody:
This module simply adds more info to the tree view of the lot.


HekkiMelody:
~~/ocabot merge patch~~

EDIT: I typed in the wrong window...

HekkiMelody:
I've added camptocamp to the authors since this module copies a few things from `stock_lot_product_qty_search`

francesco-ooops:
> I've added camptocamp to the authors since this module copies a few things from `stock_lot_product_qty_search`

@grindtildeath maybe this can interest you :)

stock_lot_show_qty_location/models/stock_production_lot.py
```diff
@@ -0,0 +1,95 @@
...
+        grouped_quants = self.env["stock.quant"].read_group(
+            [("lot_id", "!=", False), ("location_id", "in", quants_locations.ids)],
+            ["lot_id", "quantity", "reserved_quantity"],
+            ["lot_id"],
+            orderby="id",
+        )
+        lot_ids_with_quantity = {
+            group["lot_id"][0]: group["quantity"] - group["reserved_quantity"]
+            for group in grouped_quants
+        }
```
grindtildeath, line 84: Here you risk having results with a quantity = 0.

grindtildeath (commented):
Hello @aleuffre 
Thanks for your contribution.

I'm wondering if starting with defining a search function for stock.quant.available_quantity wouldn't make things simpler to be able to search the available qty of a lot. (IMO this could be done in a separate module call stock_quant_available_quantity_search)

By defining such a search function, you won't need to do the substraction of `quantity - reserved_quantity` in the search function on `stock.production.lot`, and you would allow  searching for it directly on quants.

Then, instead of having to copy the code and potentially having a same issue in two different modules due to copied code, maybe we can refactor my module to provide a either a search mixin for stock quantities or an extendable function, where you'd only have to pass the `available_quantity` field as an argument.

Please let me know what do you think and if you need any help.

HekkiMelody:
> Hello @aleuffre Thanks for your contribution.
> 
> I'm wondering if starting with defining a search function for stock.quant.available_quantity wouldn't make things simpler to be able to search the available qty of a lot. (IMO this could be done in a separate module call stock_quant_available_quantity_search)
> 
> By defining such a search function, you won't need to do the substraction of `quantity - reserved_quantity` in the search function on `stock.production.lot`, and you would allow searching for it directly on quants.
> 
> Then, instead of having to copy the code and potentially having a same issue in two different modules due to copied code, maybe we can refactor my module to provide a either a search mixin for stock quantities or an extendable function, where you'd only have to pass the `available_quantity` field as an argument.
> 
> Please let me know what do you think and if you need any help.

Thanks for your feedback! It makes total sense. I'll come back to this as soon as I have some time and follow your suggestions

francesco-ooops:
@grindtildeath this module is not needed anymore for our customer's implementation, so feel free to pick it up and bring it to completion and we'll assist with testing and review

PR #1330, opened 2023-07-16
[16.0][MIG] - stock_picking_progress

sbejaoui:

PR #1394, opened 2023-10-06
[16.0][ADD] stock_picking_operation_loss_quantity

rousseldenis:
Depends on:

- #1330


lmignon:
depends on #1330

rousseldenis:
What this module needs is to make the user access inventory easily through the Loss picking menu

lmignon:
ping @rousseldenis This one is used in our project and in production for years...

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,173 @@
...
+        new_loss_move_vals = {
+            "name": self.product_id.display_name,
+            "product_id": self.product_id.id,
+            "product_uom_qty": unprocessed_qty,  # This is the "demand"
+            "product_uom": self.product_uom_id.id,
+            "location_id": self.location_id.id,
+            "location_dest_id": loss_pick_type.default_location_dest_id.id,
+            "lot_ids": [Command.set(self.lot_id.ids)] if self.lot_id else False,
+        }
+
```
lmignon: @nicolas-delbovier-acsone For flexibility and readability this code should be put into a `_prepare_loss_move_vals(self, qty)` method `

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,173 @@
...
+            ("owner_id", "=", self.owner_id.id),
+            ("state", "not in", ("done", "cancel")),
+            ("picking_type_id", "=", loss_pick_type.id),
+            (
+                "location_dest_id",
+                "=",
+                loss_pick_type.default_location_dest_id.id,
+            ),
+        ]
+        similar_loss_lines = self.env["stock.move.line"].search(search_domain)
```
lmignon: IMO This code should also be put into a dedicated method and the domain also provided by a specific method

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,173 @@
...
+            if quants_available_quantity > 0:
+                new_loss_move = self.env["stock.move"].create(
+                    {
+                        **new_loss_move_vals,
+                        "product_uom_qty": quants_available_quantity,
+                        "picking_id": loss_picking.id,
+                    }
+                )
+                new_loss_move._action_confirm(merge=False)
+
```
lmignon, line 141: I can see a problem with this approach. In the end, if the product hasn’t been found, we’re left with n identical moves plus a final move with the remaining quantities in the location. For the operator, who only sees the transactions, this will pose a problem. In my view, it would be preferable to merge the moves and retain the number of loss declarations already made for the same product/batch in the same location.

Image

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+            # Use merge = False so that the number of move lines = the number
+            # of loss declared for this quant
+            new_loss_move._action_confirm(merge=False)
+        else:
+            loss_picking = self.env["stock.picking"].create(
+                {
+                    "picking_type_id": loss_pick_type.id,
+                    "location_id": self.location_id.id,
+                    "location_dest_id": loss_pick_type.default_location_dest_id.id,
```
lmignon: IMO the information required to create the picking should be provided by a prepare method `_prepare_loss_picking_vals` ....

lmignon (changes requested):
@nicolas-delbovier-acsone Thank you for the improvements. However, I do have some concerns about the usability of the approach. See my comment.

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,185 @@
...
+            ("state", "not in", ("done", "cancel")),
+            ("picking_type_id", "=", loss_pick_type.id),
+            (
+                "location_dest_id",
+                "=",
+                loss_pick_type.default_location_dest_id.id,
+            ),
+        ]
+
+    def _find_loss_picking(self):
```
lmignon, line 72: @nicolas-delbovier-acsone You search for a `stock.move.line` not for a picking. This method should be renamed.

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,185 @@
...
+                unprocessed_qty, 0, precision_rounding=self.product_uom_id.rounding
+            )
+            <= 0
+        ):
+            raise ValidationError(
+                _("You try to create a Loss picking without any loss quantity!")
+            )
+
+        # Search for an already existing LOSS picking for this quant
+        loss_picking = self._find_loss_picking()
```
lmignon, line 107: This variable name is confusing `loss_picking` seems to refer to a picking but in fact it's a stock.move.line....

lmignon (changes requested):
@nicolas-delbovier-acsone  Some little comments about methods and variable naming but it looks good 😏

stock_picking_operation_loss_quantity/models/stock_move_line.py
nicolas-delbovier-acsone: No, I am indeed looking for a picking.

I look for a picking of type "LOSS" that contains moves that match with my new one I want to create: 
```python
def _find_loss_picking(self):  
    similar_loss_lines = self.env["stock.move.line"].search(
        self._find_loss_picking_moves_domain()
    )
    loss_picking = first(similar_loss_lines.picking_id) # <- LOOK HERE
    return loss_picking
```

stock_picking_operation_loss_quantity/models/stock_move_line.py
nicolas-delbovier-acsone: It is a picking, see my comment above

stock_picking_operation_loss_quantity/models/stock_move_line.py
lmignon: My bad 😕

lmignon (approved):
Code review + Functional tests

stock_picking_operation_loss_quantity/models/stock_move.py
```diff
@@ -0,0 +1,29 @@
...
+        self.ensure_one()
+        if (
+            float_compare(
+                self.product_uom_qty,
+                self.reserved_availability,
+                precision_rounding=self.product_uom.rounding,
+            )
+            > 0
+        ):
+            self.with_context(_loss_ignored_locations=ignored_location)._action_assign()
```
jbaudoux: @nicolas-delbovier-acsone You need to give `force_qty` otherwise the reservation will be skipped if the line was partially processed as the move is in partially_available

jbaudoux (changes requested):

stock_picking_operation_loss_quantity/models/stock_move.py
nicolas-delbovier-acsone: I've just tested this out but it breaks `test_loss_quantity_auto_reallocation` unit test with:
```python
>       self.assertEqual(new_line.reserved_uom_qty, 5.0)
E       AssertionError: 1.0 != 5.0
```

I updated the unit test to show that the code handles partial qty gracefully "as is". Can you have a look and tell me if the updated unit test covers what you had in mind?

jbaudoux: Can you make a test where you set a quantity done on the second move line and then declare a loss on the first move line. For that use case, it requires the force_qty to equal to the move line quantity that was unlinked just before. However, I see this requirement is for odoo >= 17.0.
https://github.com/odoo/odoo/blob/17.0/addons/stock/models/stock_move.py#L1691

nicolas-delbovier-acsone: Done! Is this new test what you expected? Note that I did not need to add the "force_qty"

jbaudoux: Yes thanks for the test. Indeed you don't need force_qty in 16.0 but it will be required in next versions. To make it migration friendly, you could add it :)

nicolas-delbovier-acsone: No I can not add it since it breaks the test as I mentioned above:

>I've just tested this out but it breaks test_loss_quantity_auto_reallocation unit test with:

nicolas-delbovier-acsone:
I had to change the logic for reallocation a bit in order to allow odoo to reserve stock from the same location but from another lot.

The logic changed from 
> Try to reserve stock from another **location**

to
>Try to reserve stock from another **quant**

stock_picking_operation_loss_quantity/models/stock_move_line.py
jbaudoux: Still `_find_loss_picking_moves_domain` needs to be renamed to `_find_loss_move_line_domain`

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _prepare_loss_move_vals(self, unprocessed_qty):
+        loss_pick_type = self.location_id.warehouse_id.loss_type_id
+        return {
+            "name": self.product_id.display_name,
+            "product_id": self.product_id.id,
+            "product_uom_qty": unprocessed_qty,  # This is the "demand"
+            "product_uom": self.product_uom_id.id,
+            "location_id": self.location_id.id,
+            "location_dest_id": loss_pick_type.default_location_dest_id.id,
```
jbaudoux:
```suggestion
            "picking_type_id": loss_pick_type.id,
            "location_id": self.location_id.id,
            "location_dest_id": loss_pick_type.default_location_dest_id.id,
```

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+                != rec.location_id.warehouse_id.loss_type_id
+            )
+
+    def action_lose_quantity(self):
+        if any(not rec.is_action_lose_quantity_allowed for rec in self):
+            raise UserError(_("You are not allowed to declare loss quantities"))
+        return self._lose_quantity()
+
+    def _prepare_loss_move_vals(self, unprocessed_qty):
+        loss_pick_type = self.location_id.warehouse_id.loss_type_id
```
jbaudoux: loss_pick_type should be an argument of the method

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+        loss_picking = self._find_loss_picking()
+        if not loss_picking:
+            loss_picking = self.env["stock.picking"].create(
+                self._prepare_loss_picking_vals()
+            )
+
+        new_loss_move_vals = self._prepare_loss_move_vals(unprocessed_qty)
+        new_loss_move = self.env["stock.move"].create(
+            {**new_loss_move_vals, "picking_id": loss_picking.id}
+        )
```
jbaudoux: Set picking type reservation to "manual" and you should not bother of the picking anymore
```suggestion
        new_loss_move_vals = self._prepare_loss_move_vals(unprocessed_qty, loss_pick_type)
        new_loss_move = self.env["stock.move"].create(new_loss_move_vals)
```

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+        new_loss_move._action_confirm()
+        loss_picking.loss_declaration_count += 1
```
jbaudoux: the count should be on the move. The call to `_action_confirm` returns the new move after merge

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+        ]
+
+    def _find_loss_picking(self):
+        similar_loss_lines = self.env["stock.move.line"].search(
+            self._find_loss_picking_moves_domain()
+        )
+        loss_picking = first(similar_loss_lines.picking_id)
+        return loss_picking
+
+    def _create_loss_move_line(self, unprocessed_qty: float):
```
jbaudoux: Nice to have.
Move to `stock.quant` and call it `_lock_quant(picking_type, quantity)`
A base method that locks a quant by placing a reservation on it. There could be different types of locks using different picking types (loss, quality lock...)
That could even move to a separate base module `stock_quant_lock`

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+            {**new_loss_move_vals, "picking_id": loss_picking.id}
+        )
+        new_loss_move._action_confirm()
+        loss_picking.loss_declaration_count += 1
+
+        if (
+            self.location_id.warehouse_id.loss_auto_clear_threshold
+            and loss_picking.loss_declaration_count
+            >= self.location_id.warehouse_id.loss_auto_clear_threshold
+        ):
```
jbaudoux: You could make this check before calling the method to give the right quantity to the created move. No need to create 2 times a move.
Would be more clear if the method was on the quant

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+                new_loss_move = self.env["stock.move"].create(
+                    {
+                        **new_loss_move_vals,
+                        "product_uom_qty": quants_available_quantity,
+                        "picking_id": loss_picking.id,
+                    }
+                )
+                new_loss_move._action_confirm()
+
+        loss_picking.action_assign()
```
jbaudoux: Bug !

You gather specific quant, create a move for the quantity but don't create the move line.
The reservation could lock another quant on the same location+product.

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _lose_quantity(self):
+        """
+        This is the main function to call in order to declare a loss.
+
+        It will check if operation is in progress (if operator has found the
+        whole quantity, do not allow to declare a loss).
+
+        Then, lock the quant that should be reserved by the loss picking and
+        create that loss picking.
```
jbaudoux: Ah, it's written here :) There definitely should be somewhere a `quant._lock_quant(picking_type, quantity)`

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        loss_lines = self.filtered(lambda line: line.progress != 100.0)
```
jbaudoux: Overkill to depend on `stock_picking_progress` module just to compare if the done quantity == to do quantity.
Remove that dependency

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+
+        loss_picking.action_assign()
+        return loss_picking
+
+    def _unreserve_unprocessed_qty(self) -> float:
+        self.ensure_one()
+        unprocessed_qty = self.reserved_uom_qty - self.qty_done
+        # Free the quantity that the operator was not able to do
+        self.reserved_uom_qty = self.qty_done
+        return unprocessed_qty
```
jbaudoux: Convert to product uom

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+    def _prepare_loss_move_vals(self, unprocessed_qty):
+        loss_pick_type = self.location_id.warehouse_id.loss_type_id
+        return {
+            "name": self.product_id.display_name,
+            "product_id": self.product_id.id,
+            "product_uom_qty": unprocessed_qty,  # This is the "demand"
+            "product_uom": self.product_uom_id.id,
+            "location_id": self.location_id.id,
+            "location_dest_id": loss_pick_type.default_location_dest_id.id,
+            "lot_ids": [Command.set(self.lot_id.ids)] if self.lot_id else False,
```
jbaudoux: That's for serial numbers, not for reserving a specific lot

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+        if not loss_picking:
+            loss_picking = self.env["stock.picking"].create(
+                self._prepare_loss_picking_vals()
+            )
+
+        new_loss_move_vals = self._prepare_loss_move_vals(unprocessed_qty)
+        new_loss_move = self.env["stock.move"].create(
+            {**new_loss_move_vals, "picking_id": loss_picking.id}
+        )
+        new_loss_move._action_confirm()
```
jbaudoux: This may fail.
You don't control the reserved lot / owner.
You must create manually the move line.

stock_picking_operation_loss_quantity/models/stock_move_line.py
```diff
@@ -0,0 +1,199 @@
...
+                location_id=line.location_id,
+                lot_id=line.lot_id,
+                package_id=line.package_id,
+                owner_id=line.owner_id,
+            )
+            quants._lock_quants_for_loss()
+
+            unprocessed_qty = line._unreserve_unprocessed_qty()
+            loss_picking = line._create_loss_move_line(unprocessed_qty)
+            loss_picking._schedule_loss_activity()
```
jbaudoux: To rework. Something like:

```suggestion
            quants._lock_quants_for_loss()
            loss_pick_type = self.location_id.warehouse_id.loss_type_id

            unprocessed_qty = line._unreserve_unprocessed_qty()
            # todo : search for previous loss move, check count / threshold
            ....
            loss_pickings = self.env["stock.picking"]
            for quant in quants:
                lock_quantity = unprocessed_qty or quant.quantity
                move = quant._lock_quant(lock_quantity, loss_pick_type)
                loss_pickings |= move.picking_id
                # todo : decrease unprocessed_qty
                ...
            loss_pickings._schedule_loss_activity()
```

jbaudoux (changes requested):

lmignon:
* [x] Depends on #2449

lmignon:
Do not merge. I found a bug I'm fixing.

lmignon:
@jbaudoux In my last commit, I modified the `_lose_quantity` method on `stock.move.line` to `unlink` the lines if they are no longer relevant after the operation.

IMO, this is safer and ensures that a line always represents something that needs to be done or has already been picked. More importantly, we don't end up presenting the user with operations where there is nothing left to do. WDY?

PR #1397, opened 2023-10-11
[ADD] stock_picking_batch_grouped

SirAionTech:


SirAionTech:
Rebased on `16.0`.

SirAionTech:
Rebased on `16.0`.

stock_picking_batch_grouped/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Manage product quantities in batch transfers.
```
rousseldenis, line 1: @SirAionTech Could you update a little bit this (as this is not clear at first sight) ?

rousseldenis:
@SirAionTech Could you explain a lttile bit what is added in terms of functionalities as there are already detailed operations:

This module:

image


Standard:

image

stock_picking_batch_grouped/readme/DESCRIPTION.rst
SirAionTech: Thanks for having a look!
I explained what the new `Products` tab is, and added a few lines on how it works

rousseldenis: Maybe renaming the module `stock_picking_batch_grouped_by_product` or something like that should be great to reflect the functionality (because `grouped`alone is a little bit vague).

SirAionTech: While implementing your suggestion I realized/remembered that products is not the only thing the module is grouping by: it is also grouping by lot/serial, source location and destination location.

Moreover in the module it is possible to override some parts in order to customize the grouping, for instance the fields listed above are exposed in a dedicated method: https://github.com/OCA/stock-logistics-workflow/blob/3060c9ace643323cfda439275fab6c1dcb5cccad/stock_picking_batch_grouped/models/stock_picking_batch_line.py#L63-L70
I added a note about this in the README

SirAionTech:
> @SirAionTech Could you explain a lttile bit what is added in terms of functionalities as there are already detailed operations:
> 
> This module:
> 
> image
> 
> Standard:
> 
> image

The README has been improved in the meanwhile, now it is:

> In the batch transfer there will be a new tab that shows the products moved by the transfer.
>
>Each line in the `Products` tab represents a group of one or more move lines in the batch transfer: these grouped lines share the values shown in each `Product` line (`From`, `To`, ...).
>
>The fields to group by can be extended by other modules.
>
>Usage
>=====
>
>Set the `Done Quantity` in a line of the new `Products` tab: this will propagate the input value to the linked move lines.
>
>The input value is split so that the `Done Quantity` in the grouped move lines is filled sequentially from the oldest.

Could you please be more specific about what should be better explained?

PR #1352, opened 2023-08-10
[13.0][ADD] product_cost_price_avco_sync_mrp: New module to process result svls from a production

CarlosRoca13:
cc @Tecnativa TT44658

PR #1398, opened 2023-10-13
[13.0][FIX] product_cost_price_avco_sync: Avoid infinite loop writing current SVL

pedrobaeza:
Needed for https://github.com/OCA/stock-logistics-workflow/pull/1352

The preprocess hook serves for evaluating for the current SVL being synced if there's something that conditionate its amount/quantity, and avoiding the regular process, or if it modifies another SVLs.

The problem is that if you need to write the same SVL that is being processed, it provokes an infinite loop.

With this fix, we return the expected values to be written in the current SVL, and store it later at the end of the regular process, but being cached in our SVL dict values for also being taking into account on the rest of the process.

We also cache the SVL before doing any process for not having cache-misses, and simplify one method for not passing a computable extra arg.

@Tecnativa TT44658

### DIFF
--- a/product_cost_price_avco_sync/models/stock_valuation_layer.py
+++ b/product_cost_price_avco_sync/models/stock_valuation_layer.py
@@ -134,9 +134,8 @@ class StockValuationLayer(models.Model):
         svl_dic["remaining_value"] = svl_dic["remaining_qty"] * svl_dic["unit_cost"]
 
     @api.model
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As this fix is blocking the other PR and neutral for this module alone, let's fast-track it as the CI is green:

/ocabot merge patch

PR #1401, opened 2023-10-17
[14.0][IMP] update copier template

sergiocorato:


sergiocorato:
runboat check is failing (as in other 14.0 PRs) because it's not aware of rebels module, I think.
Is there a way to fix this problem?

rousseldenis:
> runboat check is failing (as in other 14.0 PRs) because it's not aware of rebels module, I think. Is there a way to fix this problem?

@sergiocorato See my comment here : https://github.com/OCA/stock-logistics-workflow/pull/1271#issuecomment-1764742065

rousseldenis:
@sergiocorato It seems pre-commit should be fixed

sergiocorato:
> @sergiocorato It seems pre-commit should be fixed

Done, thanks!

francesco-ooops:
@rousseldenis merge?

PR #1402, opened 2023-10-18
[16.0][MIG] stock_landed_costs_mrp_subcontracting

JasminSForgeFlow:
Standard Migration

@ForgeFlow


stock_landed_costs_mrp_subcontracting/models/stock_landed_cost.py
```diff
@@ -0,0 +1,53 @@
...
+                [tuple(self.company_id.ids)],
+            )
+            for res in self.env.cr.fetchall():
+                valued_picking_ids_per_company[res[1]].append(res[0])
+        for cost in self:
+            picking_ids = valued_picking_ids_per_company[cost.company_id.id]
+            if picking_ids:
+                additional_pickings = self.env["stock.picking"].browse(picking_ids)
+                cost.allowed_picking_ids += additional_pickings
+        return res
```
AungKoKoLin1997, line 53: Why we need to return in compute function in 16.0(not added in 15.0)?

stock_landed_costs_mrp_subcontracting/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 ForgeFlow, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Stock Landed Costs MRP Subcontracting",
+    "summary": "When using Subcontracting, use the incoming shipment in "
+    "landed costs",
+    "version": "16.0.1.1.0",
```
AungKoKoLin1997:
```suggestion
    "version": "16.0.1.0.0",
```

JasminSForgeFlow: It is fixed.
Thanks

stock_landed_costs_mrp_subcontracting/models/stock_landed_cost.py
JasminSForgeFlow: if we do not add a return it will fail pre-commit as below

image

AungKoKoLin1997: @JasminSForgeFlow FYI, you can use `# pylint: disable=missing-return` for this kind of case. But I am not against using return.

AungKoKoLin1997 (approved):

AlexPForgeFlow (approved):

AlexPForgeFlow:
several months in production

xaviedoanhduy (approved):

stock_landed_costs_mrp_subcontracting/models/stock_landed_cost.py
```diff
@@ -0,0 +1,53 @@
...
+        for move in moves:
+            mo = move.move_orig_ids.production_id[-1:]
+            if mo and mo.picking_type_id.code == "mrp_operation":
+                moves_to_remove += move
+                moves_to_add += move.move_orig_ids
+        moves -= moves_to_remove
+        moves += moves_to_add
+        return moves
+
+    def _compute_allowed_picking_ids(self):
```
xaviedoanhduy, line 24:
```suggestion
    def _compute_allowed_picking_ids(self):
```

sorry for the noise, but since `_compute_allowed_picking_ids` has been removed from this commit https://github.com/odoo/odoo/commit/f329de26

xaviedoanhduy (changes requested):

PR #1406, opened 2023-10-20
[16.0] [FIX] stock_picking_batch_extended_account: Remove `stock_picking_batch_extended` dependency

Shide:
Removed unnecessary dependency: `stock_picking_batch_extended`.

Nothing of these modules is used on `stock_picking_batch_extended_account` nor `stock_picking_batch_extended_account_sale_type`, so we can remove dependency and work in the same way.

Button has moved before `state` field, like `stock_picking_batch_extended` does.

Tests has been modified to not use `stock_picking_batch_extended` methods and use the standard ones.

Separated in 2 commits to improve migrations.

MT-3611 @moduon @rafaelbn @fcvalgar @gelojr @ernestotejeda please review if you want 😸

### DIFF
--- a/stock_picking_batch_extended_account/__manifest__.py
+++ b/stock_picking_batch_extended_account/__manifest__.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - Ernesto Tejeda
+# Copyright 2023 Moduon Team - Eduardo de Miguel
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 {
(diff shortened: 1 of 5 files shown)


rafaelbn (approved):
Great! 👍🏼 

Question of renaming:

- [ ] stock_picking_batch_extended_account -> stock_picking_batch_account

Next versión 17?

rafaelbn:
We would like to have the review of @sergio-teruel or @carlosdauden  Thank you! 😄 ❤️

sergio-teruel (approved):

sergio-teruel:
in fact, the module stock_picking_batch_extended  in 16 could be erased.

rafaelbn (approved):
/ocabot merge patch

ping @RodrigoBM

PR #1417, opened 2023-11-01
[16.0][FIX] stock_move_line_auto_fill: to fix installation error when mrp & stock_barcode apps exist on the db

rinaldifirdaus:
This commit is to ensure that this module can be installed when the database already has the 'mrp' and 'stock_barcode' modules installed. both of the modules are inheriting the stock_picking_type_views.xml.

@qrtl


yostashiro (approved):
Makes sense.

The tag probably should be 'FIX', IMO.

AungKoKoLin1997 (approved):

rousseldenis (approved):

PR #1419, opened 2023-11-02
[16.0] stock_picking_invoice_link: fix string

alexis-via:


alexis-via:
For me, the current string doesn't make sense. So I propose to change it.

pedrobaeza (approved):
/ocabot merge patch

PR #1365, opened 2023-08-25
[16.0][MIG] delivery_total_weight_from_packaging: Migrate to 16.0

QuocDuong1306:
Depend on:
- https://github.com/OCA/product-attribute/pull/1404

PR #1420, opened 2023-11-06
[14.0][IMP] delivery_total_weight_from_packaging: add base weight to delivery weight

mmequignon:
Adds the packaging weight to the delivery weight.
This is built-in in odoo 16.0.

### DIFF
--- a/delivery_total_weight_from_packaging/readme/ROADMAP.rst
+++ b/delivery_total_weight_from_packaging/readme/ROADMAP.rst
@@ -1,0 +1,2 @@
+* In odoo core 16.0, there's a new `base_weight` field on new model `stock.package.type`.
+  Remove this field and write a migration for it.
--- a/delivery_total_weight_from_packaging/__manifest__.py
+++ b/delivery_total_weight_from_packaging/__manifest__.py
@@ -17,5 +17,7 @@
         # OCA/product-attribute
         "product_total_weight_from_packaging",
     ],
-    "data": [],
+    "data": [
+        "views/product_packaging.xml",
+    ],
 }
--- a/delivery_total_weight_from_packaging/models/__init__.py
+++ b/delivery_total_weight_from_packaging/models/__init__.py
@@ -1,3 +1,4 @@
+from . import product_packaging
 from . import stock_move
 from . import stock_picking
 from . import stock_quant_package
--- a/delivery_total_weight_from_packaging/models/product_packaging.py
+++ b/delivery_total_weight_from_packaging/models/product_packaging.py
@@ -1,0 +1,9 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models, fields
+
+class ProductPackaging(models.Model):
+    _inherit = "product.packaging"
+
+    base_weight = fields.Float(string='Base Weight', help='Weight of the packaging')
--- a/delivery_total_weight_from_packaging/models/stock_quant_package.py
+++ b/delivery_total_weight_from_packaging/models/stock_quant_package.py
@@ -33,8 +33,8 @@ class StockQuantPackage(models.Model):
             )
 
     def _get_weight_from_packaging(self):
+        weight = self.packaging_id.base_weight or 0.0
         # NOTE: code copied/pasted and adapter from `delivery`
-        weight = 0.0
         if self.env.context.get("picking_id"):
             current_picking_move_line_ids = self.env["stock.move.line"].search(
                 [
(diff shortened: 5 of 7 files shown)


francesco-ooops:
@mmequignon thanks for the PR! Can you set the correct name of module in PR title and commit?

mmequignon:
> @mmequignon thanks for the PR! Can you set the correct name of module in PR title and commit?

Well, I would usually have done this, but since it's like 50 chars long, I would avoid this in order to have a meaningful description.

francesco-ooops:
@sebalix any input on this? 

@pedrobaeza do we have guidelines on this? I also wondered what to do in case of long name modules

@mmequignon we should probably merge this first: https://github.com/OCA/stock-logistics-workflow/pull/1365

delivery_total_weight_from_packaging/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+* In odoo core 16.0, there's a new `base_weight` field.
```
sebalix:
```suggestion
* In odoo core 16.0, there's a new `base_weight` field on new data model `stock.package.type`.
```

sebalix:
PR title and commit message are not the same (obviously).
For contributors, the exact name of the module should be put in the PR title to find it easily (I updated it), and for commit message it could be renamed this way: `delivery_total_weight_from_packaging: add base_weight` as the maximum lenght of the commit header line is 72 characters on GitHub/GitLab and other tools (50 is the default Vim configuration when writing commit, very conservative :) )

sebalix (approved):
LG to me.

A better solution would have been to add one dedicated module that extends base `delivery` module to add the `base_weight`, and `delivery_total_weight_from_packaging` depending on it to re-use this field to do its computation (as it is breaking the inheritance sadly). So users not using `delivery_total_weight_from_packaging` could still leverage this `base_weight` field.

Now, all of our Odoo instances are using the current module I guess, so it's OK for us to host this field here... what's your opinion @francesco-ooops ?

EDIT: and avoiding an extra addon will ease the migration to 16.0+ as we'll probably still need `delivery_total_weight_from_packaging` that can host the migration script, while the extra addon will be uninstalled.

pedrobaeza:
> @pedrobaeza do we have guidelines on this? I also wondered what to do in case of long name modules

The commit message line length is just a recommendation. I give more priority to put the complete module name, and overpass such limit when needed. The guideline IMO is to be "short" on the message, not to rip the essential elements.

rousseldenis:
> > @pedrobaeza do we have guidelines on this? I also wondered what to do in case of long name modules
> 
> The commit message line length is just a recommendation. I give more priority to put the complete module name, and overpass such limit when needed. The guideline IMO is to be "short" on the message, not to rip the essential elements.

Agree with @pedrobaeza. Those kind of thoughts has led to commit messages with module names like 'x_y_z: A sho. messag.' and I exagerate a little.

Meaningful message should prevail on Github screen restrictions.

My two cents

pedrobaeza:
Big news! Denis agrees with something I have said! :stuck_out_tongue_closed_eyes: 

I hope you are enjoying Odoo XP. We'll see next year... in Tenerife :wink:

francesco-ooops:
> > > @pedrobaeza do we have guidelines on this? I also wondered what to do in case of long name modules
> > 
> > 
> > The commit message line length is just a recommendation. I give more priority to put the complete module name, and overpass such limit when needed. The guideline IMO is to be "short" on the message, not to rip the essential elements.
> 
> Meaningful message should prevail on Github screen restrictions.

I agree with both of you and I think it's worth being more specific in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#naming about this, as so often in reviews commit naming comes up and there's room for clarification in guidelines (eg: https://github.com/OCA/purchase-workflow/pull/2086#issuecomment-1805297177 )

Same thing for PR titles, as there are no guidelines on that, and it's not ideal for anyone involved to ask contributors to respect standards that are not written on guidelines

@sebalix I tagged you as AFAIU you're working on a project that relies on PR titles, so if you want to add a few lines of clarification to guidelines I can assist :)

For the PR itself, not very familiar with this module sorry!

mmequignon:
Thanks guys for your input. I updated the PR, not only the commit name, but also the overridden view.
Instead of putting this new fields on the packaging form view, I'm putting it on the delivery packages.
This makes more sense to me.

francesco-ooops:
> @mmequignon we should probably merge this first: #1365

sorry, I got misled from initial message and thought this was a 16.0 PR! my bad

PR #1422, opened 2023-11-08
[16.0][ADD] stock_picking_move_package_to_package

angelinaanaki:


stock_picking_move_package_to_package/views/stock_package_level_view.xml
```diff
@@ -0,0 +1,20 @@
...
+        model="ir.ui.view"
+    >
+        <field name="name">Package Level Tree Picking Move Package To Package</field>
+        <field name="model">stock.package_level</field>
+        <field name="inherit_id" ref="stock.package_level_tree_view_picking" />
+        <field name="arch" type="xml">
+            <field name="package_id" position="after">
+                <field
+                    name="package_dest_id"
+                    attrs="{'readonly': [('state', 'in', ('confirmed', 'assigned', 'done', 'cancel'))]}"
```
rousseldenis, line 14: @angelinaanaki Could you explain a little bit your use case as it is a little bit too restrictive.

I can't set a destination package in a normal flow (tested on runboat).

image

angelinaanaki: @rousseldenis It is important that the user can add the destination package in all the states where he can select a source one. In this case, the source package is also read-only.

rousseldenis: @angelinaanaki Ok, so, USAGE and DESCRIPTION fragments should be a little bit more explicits. Thanks

ivs-cetmix: @rousseldenis thank you for mentioning this! Done, please check

ivs-cetmix (approved):
Functional review LGTM

rousseldenis:
@angelinaanaki @ivs-cetmix I need a little bit more information about the use case for this.

Could you explain a little bit more when you need this ?

On runbot, when I have a PICK when I have put the product in a Package (PACK0001) with put in pack. On the next step, I can decide to change to a new one with a new Put in Pack action. So, the product is in a different package.

Moreover, you need to explain why your mechanism is only available in draft state as it is quite nonsense in day-to-day operations.


Thanks

ivs-cetmix:
> @angelinaanaki @ivs-cetmix I need a little bit more information about the use case for this.
> 
> Could you explain a little bit more when you need this ?
> 
> On runbot, when I have a PICK when I have put the product in a Package (PACK0001) with put in pack. On the next step, I can decide to change to a new one with a new Put in Pack action. So, the product is in a different package.
> 
> Moreover, you need to explain why your mechanism is only available in draft state as it is quite nonsense in day-to-day operations.
> 
> Thanks

@rousseldenis good question! It's time for the OCA "Context.md" magic) I have added this useful file and updated the readme. Please check if this became clearer now.

ivs-cetmix (changes requested):
Found an issue when using this module with "lot"-tracked products @angelinaanaki set an activity for you

pandoc-3.1.9-x86_64-macOS.pkg
ivs-cetmix, line 1: This should be removed

ivs-cetmix (changes requested):

angelinaanaki:
> Found an issue when using this module with "lot"-tracked products @angelinaanaki set an activity for you

@ivs-cetmix Fixed it. Please check.

stock_picking_move_package_to_package/tests/test_move_package_to_package.py
```diff
@@ -0,0 +1,112 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo.tests import TransactionCase
+
+
+class TestStockPicking(TransactionCase):
+    def setUp(self):
```
ivs-cetmix: Use  setUpClass instead of the setUp in tests.
Disable tracking unless you need it for functional purposes

    @classmethod
    def setUpClass(cls):
        super().setUpClass()

stock_picking_move_package_to_package/tests/test_move_package_to_package.py
```diff
@@ -0,0 +1,112 @@
...
+            {
+                "name": "New Stock",
+                "location_id": self.stock_location.warehouse_id.view_location_id.id,
+                "company_id": self.company.id,
+            }
+        )
+        self.picking_type = self.stock_location.warehouse_id.int_type_id
+        self.picking_type.write({"show_entire_packs": True})
+
+    def test_move_package_to_package(self):
```
ivs-cetmix: We need to test cases for:
- Products that are not tracked by lot
- Products that are tracked by lot/serial number

ivs-cetmix (changes requested):
Functional OK but check comments pls

angelinaanaki:
> Functional OK but check comments pls

@ivs-cetmix Done. Please check and share your feedback.

ivs-cetmix (approved):
LGTM

ivs-cetmix:
@rousseldenis would appreciate your review when you have time. Thank you in advance!

stock_picking_move_package_to_package/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
DemchukM: Please replace the writing with license in all files where it is indicated

now listed is http://www.gnu.org/licenses/lgpl
but it is necessary
http://www.gnu.org/licenses/agpl

angelinaanaki: Thanks for noticing. I've fixed it. Please check.

DemchukM (approved):
LGTM

ivs-cetmix:
Hey @OCA/logistics-maintainers  would be great to have this moved forward 😄

jbaudoux (changes requested):
Thanks for your contribution but the approach is ugly and must be reworked

The move is based on the move lines. So you cannot process the moves and then modify the quants. This is very dangerous, bypassing some checks and could lead to unexpected results if your moves are chained. You cannot do this !

The destination package is set on the move lines. You need to update those with the wanted destination package.

Side note: Please put a description in your PR

angelinaanaki:
> Thanks for your contribution but the approach is ugly and must be reworked
> 
> The move is based on the move lines. So you cannot process the moves and then modify the quants. This is very dangerous, bypassing some checks and could lead to unexpected results if your moves are chained. You cannot do this !
> 
> The destination package is set on the move lines. You need to update those with the wanted destination package.
> 
> Side note: Please put a description in your PR

@jbaudoux 

Thank you for your helpful feedback—I really appreciate it!
The PR mentioned (https://github.com/OCA/stock-logistics-workflow/pull/1463), regarding the code, implements the logic that adds the content to a stock transfer depending on the selected package. Which may partially help in the problem that should be solved by this module:

https://github.com/cetmix/oca-stock-logistics-workflow/blob/16.0-t2954-stock_picking_move_package_to_package-new_module/stock_picking_move_package_to_package/readme/CONTEXT.md ->

> Let's imagine you have a parcel forwarding business. You receive small
> packages from your customer pack them into bigger ones and forward
> afterwards.
> 
> To complete this operation using the standard Odoo flow:
> -  Create a new picking
> -  Add each item of the "Source package" to the list of operations
> -  Manually set the "Destination package" for each line
> -  Validate
> -  Items from the source package are moved to the destination package
>
> 
> This is really time consuming and not convenient because you just want
> to move the entire package at once. Especially if you want to move
> multiple packages at once each of them containing multiple items.


But again the Destination Package should be set manually for every line. Maybe one idea is to inherit this module and if the content is added from the package then with help of a wizard to add the destination package for the given source packages from the lines.

Or, I've come up with another idea. Instead of altering the current module workflow, I propose a change in functionality during validation. Take a look at the updated [stock_picking.py](https://gist.github.com/angelinaanaki/8a82a49f50a9e4ecb2785ac161babc5b). Before validation, the logic assigns the destination package to all move lines sharing the same source as in the package level. This approach is functional and passes the tests and it follows the standard logic. However, a problem arises: the standard package field at the package level mirrors the destination package values from the move lines after validation, resulting in both fields showing the same value at the package level.

Maybe making the destination package field invisible(if possible in the tree view) when it has the same value as the standard package at the package level and using a text field to indicate the package movement could solve this. That way, it would show that the picking's package was moved from another package. 

I'd appreciate your thoughts on this or any alternative ideas you may have.

yelizariev:
Some ideas about a possible solution

1. Changing quants (current implementation) or move lines (proposal in gist) looks too risky
https://gist.github.com/angelinaanaki/8a82a49f50a9e4ecb2785ac161babc5b

2. Alternative PR use another approach: they temporary change `package_level_id` for `stock.move` during assigning process. This seems better, but still looks a bit hacky
https://github.com/OCA/stock-logistics-workflow/pull/1463/

3. If we want to change field `result_package_id` in move lines, maybe we should try to do it inside the process and not before/after.
Somewhere here maybe:
https://github.com/odoo/odoo/blob/fd0d3d1e045794bbd0b8b0b5d968463abb525a63/addons/stock/models/stock_picking.py#L998-L1006
https://github.com/odoo/odoo/blob/fd0d3d1e045794bbd0b8b0b5d968463abb525a63/addons/stock/models/stock_package_level.py#L76-L77

Extending those methods is a bit tricky. We could first try copy-paste whole method and check if it works as expected. If it's really what we need, then extending could be done in the following tricky way:
First, extend `_check_entire_pack` (or `_set_is_done`_ and call super with a new attribute in context.
Second, extend `write` (or `create`) method and check that context attribute to replace value for `result_package_id`

Does it make any sense to you?

ivs-cetmix:
> Some ideas about a possible solution
@jbaudoux could you please have a look at this idea?

ivs-cetmix:
Hi @OCA/logistics-maintainers @jbaudoux sorry for bothering but any chances to have this merged?

jbaudoux:
Sorry for the long delay.

When you validate the transfer, odoo process the move lines. I think it is just a matter of putting the right result package on each move line, then odoo will do the right packing. So you would just need a convenient way to set this value on many lines at once.

What about making `package_dest_id` on the package level a computed field with inverse method.
* inverse : When you set a value on package_dest_id of the package level, you propagate on linked move lines to the result_package_id. Note that you need to have the package level flagged as done to ensure the related move lines are there. Either, the package_dest_id is readonly if level is not flagged as done, or you flag it as done at the start of this inverse method when you put a value.
* compute: When all related lines have the same result_package_id, then you see the value on package_dest_id of the package level.

angelinaanaki:
> Sorry for the long delay.
> 
> When you validate the transfer, odoo process the move lines. I think it is just a matter of putting the right result package on each move line, then odoo will do the right packing. So you would just need a convenient way to set this value on many lines at once.
> 
> What about making `package_dest_id` on the package level a computed field with inverse method.
> 
> * inverse : When you set a value on package_dest_id of the package level, you propagate on linked move lines to the result_package_id. Note that you need to have the package level flagged as done to ensure the related move lines are there. Either, the package_dest_id is readonly if level is not flagged as done, or you flag it as done at the start of this inverse method when you put a value.
> * compute: When all related lines have the same result_package_id, then you see the value on package_dest_id of the package level.

@jbaudoux Thank you for the review! I've made the suggested changes. Could you please take another look and let me know if everything is good now? Thanks again for your time!

stock_picking_move_package_to_package/models/stock_package_level.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    @api.depends("move_line_ids.result_package_id")
+    def _compute_package_dest_id(self):
+        """
+        Compute `package_dest_id` based on `result_package_id` of related move lines.
+        Set `package_dest_id` if all move lines share the same `result_package_id`.
+        """
+        for level in self:
+            result_package_ids = level.move_line_ids.mapped("result_package_id")
```
jbaudoux: I think there is no need for mapped
```suggestion
            result_package_ids = level.move_line_ids.result_package_id
```

stock_picking_move_package_to_package/models/stock_package_level.py
```diff
@@ -0,0 +1,48 @@
...
+        "Destination Package",
+        compute="_compute_package_dest_id",
+        inverse="_inverse_package_dest_id",
+        store=True,
+        readonly=False,
+        check_company=True,
+        domain="[('location_id', 'child_of', parent.location_dest_id), "
+        "'|', "
+        "('company_id', '=', False), "
+        "('company_id', '=', company_id)]",
```
jbaudoux: What if you want to move to an empty package? Then there is no location. You should allow this case also.
```suggestion
        domain="[('company_id', 'in', [False, company_id]), "
        "'|', ('location_id', 'child_of', parent.location_dest_id), "
        "('location_id', '=', False)]",
```

stock_picking_move_package_to_package/tests/test_move_package_to_package.py
```diff
@@ -0,0 +1,363 @@
...
+        incoming_picking1.action_put_in_pack()
+        incoming_picking1.action_confirm()
+        incoming_picking1.button_validate()
+
+        internal_picking1 = self.StockPicking.search(
+            [("origin", "=", incoming_picking1.name)]
+        )
+        package_level1 = self.PackageLevel.search(
+            [("picking_id", "=", internal_picking1.id)]
+        )
```
jbaudoux:
```suggestion
        package_level1 = internal_picking1.package_level_ids
```

stock_picking_move_package_to_package/tests/test_move_package_to_package.py
```diff
@@ -0,0 +1,363 @@
...
+        incoming_picking2.action_confirm()
+        incoming_picking2.button_validate()
+
+        internal_picking2 = self.StockPicking.search(
+            [("origin", "=", incoming_picking2.name)]
+        )
+
+        package_level2 = self.PackageLevel.search(
+            [("picking_id", "=", internal_picking2.id)]
+        )
```
jbaudoux:
```suggestion
        package_level2 = internal_picking2.package_level_ids
```

stock_picking_move_package_to_package/tests/test_move_package_to_package.py
```diff
@@ -0,0 +1,363 @@
...
+            {
+                "package_id": package2.id,
+                "package_dest_id": package1.id,
+                "picking_id": internal_picking2.id,
+                "company_id": self.company.id,
+                "location_id": self.input_location.id,
+                "location_dest_id": self.stock_location.id,
+            }
+        )
+        package_level2.write({"is_done": True})
```
jbaudoux:
```suggestion
        # Move package2 to package1
        package_level2.package_dest_id = package1
        assertTrue(package_level2.is_done)
```

stock_picking_move_package_to_package/tests/test_move_package_to_package.py
```diff
@@ -0,0 +1,363 @@
...
+        # Check the available product qty of product1 and product2 after the operations
+        self.assertEqual(
+            product1.qty_available, 3.0, "Total product quantity is not as expected."
+        )
+        self.assertEqual(
+            product2.qty_available, 2.0, "Total product quantity is not as expected."
+        )
+
+    def test_move_package_to_package_with_tracking_products(self):
+        # Create a products
```
jbaudoux, line 197: Same remarks as above

jbaudoux (commented):
Looking already way better. Thanks for the improvements

angelinaanaki:
@jbaudoux Thank you for the review! Could you please take another look when you have a moment?

jbaudoux (approved):
Thanks for the fixes.
Can you squash the commits to only keep one?

angelinaanaki:
> Thanks for the fixes. Can you squash the commits to only keep one?

@jbaudoux Thank you! The commits have been squashed into one.

ivs-cetmix:
Wow! @OCA/logistics-maintainers  What a great day to ~~save the world~~ merge this PR! Isn't it? 😄

rousseldenis (approved):

rousseldenis:
@angelinaanaki @ivs-cetmix In addition to this, I would have added a configuration parameter on picking type level to activate the field

ivs-cetmix:
Hello @rousseldenis, thank you for merging! However there is an issue with HTML description which was not generated properly for some reason. Here is a PR that fixes it, would appreciate you reviewing and merging it: https://github.com/OCA/stock-logistics-workflow/pull/1675

Regarding an additional parameter let me check this. Sounds like a good idea, just need to find the best way to implement it 👍

PR #1423, opened 2023-11-10
[16.0][FIX] sale_order_global_stock_route

aktiv-mohit-chauhan:
Fixed write method to update all sale order line records, addressing the first-record-only issue in the absence of for loop.

### DIFF
--- a/sale_order_global_stock_route/models/sale_order.py
+++ b/sale_order_global_stock_route/models/sale_order.py
@@ -28,7 +28,8 @@ class SaleOrder(models.Model):
             lines = self.mapped("order_line").filtered(
                 lambda l: l.route_id.id != vals["route_id"]
             )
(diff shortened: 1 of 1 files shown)


rousseldenis:
@mohitaktiv Thanks for this. Could you explain the reason of that change ?

sckevin98:
@rousseldenis 
In our system route changes on change of delivery address. When this occured in a sales order we noticed only first product in sales order lines get updated with the new route of the changed delivery address. This caused issues of creating different delivery records (dropshipment and normal delivery pickings) for the same sales order, since it depends on the routes mapped in sales order lines.
With @mohitaktiv update it changes the route for every sale order line item, avoiding having different routes for the same sales order.

aktiv-mohit-chauhan:
Hello @rousseldenis,
Could you please review this FIX?

PR #1424, opened 2023-11-12
[16.0][ADD] stock_picking_limit_return_qty

ivs-cetmix:
Limit quantity of items that can be returned in the same picking

Before this PR:

User can return any amount of products in picking return. Even if this amount is more tha initial amount in the original picking.

After this PR:

if user tries to return bigger qty ot a product than it was in the original picking ValidationError is raised. This includes quantities already returned.


stock_picking_limit_return_qty/tests/test_stock_picking_return.py
```diff
@@ -0,0 +1,86 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestStockReturnPicking(TransactionCase):
+    post_install = True
+    at_install = True
+
```
em230418: Old style test tagging should not work in 16.0

stock_picking_limit_return_qty/wizard/stock_picking_return.py
```diff
@@ -0,0 +1,46 @@
...
+    @api.constrains("quantity")
+    def _constraints_quantity(self):
+        for rec in self:
+            if rec.quantity and rec.quantity_max and rec.quantity > rec.quantity_max:
+                raise ValidationError(
+                    _(
+                        "You can return only %(max_qty)s of %(product)s",
+                        max_qty=rec.quantity_max,
+                        product=rec.product_id.name,
+                    )
```
em230418, line 26: You need to wrap only "You can return only %(max_qty)s of %(product)s" argument.

stock_picking_limit_return_qty/tests/test_stock_picking_return.py
```diff
@@ -0,0 +1,86 @@
...
+            msg = ""
+            try:
+                line.quantity = line.quantity + 1
+            except ValidationError as e:
+                error = str(e)
+                msg = "You can return only %s of %s" % (
+                    line.quantity_max,
+                    line.product_id.name,
+                )
+            self.assertEqual(error, msg)
```
em230418: Maybe self.assertRaises(ValidationError) instead?

em230418 (changes requested):

stock_picking_limit_return_qty/wizard/stock_picking_return.py
ivs-cetmix: Hi @em230418 , thank you for your feedback! Actually current approach should work as well. Like it's described [here](https://www.odoo.com/documentation/15.0/es/developer/howtos/translations.html#variables)

Translation terms are exported correctly as well
image

em230418: I see. Confused usage of % operator.

stock_picking_limit_return_qty/tests/test_stock_picking_return.py
ivs-cetmix: Fixed

stock_picking_limit_return_qty/tests/test_stock_picking_return.py
ivs-cetmix: Fixed, please check

ivs-cetmix:
@em230418 we have pushed the updates. Could you please do a code review again? Thank you in advance!

em230418 (approved):
Code looks fine. Didn't check functionality

stock_picking_limit_return_qty/wizard/stock_picking_return.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_stock_return_picking_form_subcontracting" model="ir.ui.view">
+        <field name="name">Return lines</field>
+        <field name="model">stock.return.picking</field>
+        <field name="inherit_id" ref="stock.view_stock_return_picking_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='product_return_moves']/tree/field[@name='quantity']"
```
robinkeunen, line 9: I only see one `quantity` in the view, why not use the simpler _field_ selector ?

```xml
<field name="quantity" position="after" />
        <field name="quantity_max" invisible="1" />
</field>
```
or even
```xml
<form position="inside" />
        <field name="quantity_max" invisible="1" />
</form>

**Review by robinkeunen — CHANGES_REQUESTED** [review]
Mostly good. View definition could be simpler if I'm not mistaken.

can you take a look at my PR? https://github.com/OCA/stock-logistics-workflow/pull/1432

**Review by Bearnard21 — APPROVED** [review]
Functionality Ok

File: stock_picking_limit_return_qty/wizard/stock_picking_return.xml (hunk 4 above)
Comment (ivs-cetmix) [reply stock_picking_limit_return_qty/wizard/stock_picking_return.xml:9]: Hi @robinkeunen thank you for your feedback! We opted for keeping them more explicit in case this module would be installed on some system which already has some modifications done in the same views.
To make it more capable of survival in a harsh module conditions)

Comment (robinkeunen) [reply stock_picking_limit_return_qty/wizard/stock_picking_return.xml:9]: Well, I think it's actually more robust to be more generic : `<form>` or `<field name="quantity">` have more chance to survive the future than `//field[@name='product_return_moves']/tree/field[@name='quantity']`.

But ok :-)

**Review by robinkeunen — APPROVED** [review]

File: stock_picking_limit_return_qty/wizard/stock_picking_return.py [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+
+    @api.model
+    def _prepare_stock_return_picking_line_vals_from_move(self, stock_move):
+
+        res = super(
+            ReturnPicking, self
+        )._prepare_stock_return_picking_line_vals_from_move(stock_move)
+
+        # Store maximum quantity that is possible to return
+        res.update(
```
rousseldenis, line 41: As this is breaking the default Odoo's behaviour, I would have added a configuration parameter to activate this.

Either a global one, either one on stock.picking.type object (I don't know if it is defined on stock move values at this point)

rousseldenis (changes requested):

stock_picking_limit_return_qty/wizard/stock_picking_return.py
ivs-cetmix: @rousseldenis thank you for mentioning this. Sounds reasonable. The only thing we didn't add it is because this is actually the only feature this module is adding. So this setting is actually if the module installed or not. Don't want this behaviour? Just uninstall it.
However the concern was that this module will be installed on the runboat bu default. So it might cause some confusion.  It could be a good idea if this module could be installed as using checkbox in settings like this is done with some core Odoo modules. However I didn't manage to find to do the same for the third party modules.
Would be great if you could give some hints.

PR #1426, opened 2023-11-14
[14.0][FIX] stock_picking_inter_warehouse: multiple fixes

HekkiMelody:
- Compatibility with multi-company:
  Previously the route was only created for the default company, and various checks and tests were missing in order to make the module work in a multi-company environment.

  Now an inter-warehouse route is created for each company, including on installation via an init_hook. A small migration script is provided as well.

- Constraint on Receipt Partner on Warehouse:
  the partner defined on the warehouse for interwarehouse transfers must be unique.

- Restricted domain on picking by company
  Do not allow user to select partners related to warehouses of other companies during an intercompany picking


francesco-ooops (approved):
Functional ok!

PR #1428, opened 2023-11-15
[16.0][ADD] stock_quant_package_fast_move

angelinaanaki:
This PR is dependent on changes introduced in [PR #1422](https://github.com/OCA/stock-logistics-workflow/pull/1422), specifically on the module **stock_picking_move_package_to_package**.


rousseldenis:
@angelinaanaki Thanks for this. Couldn't you use a batch picking flow instead (several picking at once) - and maybe extend that process ?

ivs-cetmix:
> @angelinaanaki Thanks for this. Couldn't you use a batch picking flow instead (several picking at once) - and maybe extend that process ?

Hi @rousseldenis just wondering why should we use the batch picking flow if we can transfer multiple package in a singe picking using package level?

rousseldenis:
> > @angelinaanaki Thanks for this. Couldn't you use a batch picking flow instead (several picking at once) - and maybe extend that process ?
> 
> Hi @rousseldenis just wondering why should we use the batch picking flow if we can transfer multiple package in a singe picking using package level?

Ok, need testing the flow

stock_quant_package_fast_move/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+* Cetmix <https://cetmix.com/>
```
ivs-cetmix: Please use MD format for links

ivs-cetmix (approved):
Code review LGTM. Please fix comment regarding link format in the /readme files and squash commits afterwards

Bearnard21 (approved):
All features work as expected

Aldeigja (approved):
Functional LGTM

geomer198 (approved):
LGTM!

DemchukM (commented):
In some files, you have an incorrect license reference.
The link is http://www.gnu.org/licenses/lgpl and you need http://www.gnu.org/licenses/agpl

Please review the files and correct them.

Everything else is fine

DemchukM (approved):
LGTM

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Returns:
+        - bool: True if the move is successful.
+
+        Raises:
+        - exceptions.UserError: If the destination location is the same as the current location
+        or if the destination package does not belong to the specified location.
+        """
+        # Check if the location is different from the current location
+        if location == self[0].location_id and not destination_package:
```
DemchukM, line 100: Why do you only check the location_id in the first package?

What will happen if packages from different location_ids are selected?

stock_quant_package_fast_move/wizard/stock_quant_package_fast_move_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        package_ids = self._context.get("active_ids")
+        packages = self.env["stock.quant.package"].browse(package_ids)
+        if any(package.location_id != packages[0].location_id for package in packages):
+            raise exceptions.UserError(
+                _("All packages must have the same source location.")
+            )
+
+        if self.put_in_new_package:
+            destination_package_id = packages.create_new_package(self.location_dest_id)
```
DemchukM: You pass 1 argument to the "create_new_package function", but it takes no arguments

stock_quant_package_fast_move/models/res_company.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    picking_type_id = fields.Many2one(
+        "stock.picking.type", string="Package Move Operation"
+    )
```
yelizariev: Maybe make a more unique name for this field? To avoid possible conflicts with other custom modules

stock_quant_package_fast_move/wizard/stock_quant_package_fast_move_wizard.py
```diff
@@ -0,0 +1,39 @@
...
+    def action_move(self):
+        """
+        Execute the package movement based on the specified parameters.
+        """
+        package_ids = self._context.get("active_ids")
+        packages = self.env["stock.quant.package"].browse(package_ids)
+        if any(package.location_id != packages[0].location_id for package in packages):
+            raise exceptions.UserError(
+                _("All packages must have the same source location.")
+            )
```
yelizariev: Maybe make a separate method for this check, so it can be used in `_move_to_location` ?

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def create_new_package(self):
+        """
+        Create a new stock package.
+
+        Returns:
+        - package (recordset of stock.quant.package): The newly created package object.
+        """
+        package = self.env["stock.quant.package"].create({})
+        return package
```
yelizariev, line 165: Maybe just create a package inline?

ivs-cetmix: > Maybe just create a package inline?

This is in case someone would need to implement some custom flow here

stock_quant_package_fast_move/models/stock_quant_package.py
angelinaanaki: Before this, validation ensures the source location is the same for all packages. - [https://github.com/OCA/stock-logistics-workflow/pull/1428/files#diff-8d3622c031c6c7da6edf3c7e1d6e83021cacb13df6f5cc2f25b518262fae59deR47](https://github.com/OCA/stock-logistics-workflow/pull/1428/files#diff-8d3622c031c6c7da6edf3c7e1d6e83021cacb13df6f5cc2f25b518262fae59deR47)

stock_quant_package_fast_move/wizard/stock_quant_package_fast_move_wizard.py
angelinaanaki: Fixed! Thanks for the catch!

stock_quant_package_fast_move/models/res_company.py
angelinaanaki: Fixed! Appreciate the suggestion!

stock_quant_package_fast_move/wizard/stock_quant_package_fast_move_wizard.py
angelinaanaki: Done!

DemchukM (approved):
LGTM

yelizariev (approved):
LGTM

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -89,9 +91,10 @@ def _move_to_location(self, location, destination_package=None):
             package_level = self.env["stock.package_level"].create(package_level_vals)
             package_level.write({"is_done": True})
 
-        # Validate the picking
         picking.action_confirm()
-        picking.button_validate()
+        if no_validate:
```
ivs-cetmix: Should't it be `if not no_validate` ?

`If you check 'Don't Validate', the created internal picking won't be validated. Otherwise, it will be validated and the package will directly moved to the specified destination location`

ivs-cetmix (changes requested):

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -27,15 +27,15 @@ def check_source_location(self):
...
+    def _move_to_location(self, location, destination_package=None, validate=False):
         """
         Move packages to a specified location.
 
         Parameters:
         - location (recordset of stock.location): The destination location.
         - destination_package (recordset of stock.quant.package, optional):
         Optional destination package. If provided, it must belong to the specified location.
-        - no_validate (boolean, optional):
+        - validate (boolean, optional):
```
ivs-cetmix, line 86: If set to True, the created picking will be validated. Otherwise it will remain in the "Ready" state.

angelinaanaki: Thank you for your review. Could you please check it again?

ivs-cetmix (approved):
LGTM

stock_quant_package_fast_move/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2023 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Fast Package Move Between Locations",
+    "version": "16.0.1.0.0",
+    "category": "Inventory/Inventory",
+    "summary": "Move packages between locations directly from the 'Package' menu",
+    "depends": ["stock_picking_move_package_to_package"],
```
ivs-cetmix, line 8: Hi @angelinaanaki why do we need to depend on this module?
Are we using any of its functions?

angelinaanaki: Hi @ivs-cetmix, the dependency is needed because of the extension on picking validation in [stock_picking_move_package_to_package](https://github.com/OCA/stock-logistics-workflow/pull/1422/files#diff-2a830cea3f020868b61e6114f83c8f3714e66c4d26724dc4bfa88d8a0253c697R26) necessary for package movements by setting the [package_dest_id](https://github.com/OCA/stock-logistics-workflow/pull/1422/files#diff-2c9d091a57bda97f2f06c4b038df5eb9351fd1b814efddfb09986a899f6a98ccR10) field in [_move_to_location](https://github.com/OCA/stock-logistics-workflow/pull/1428/files#diff-8d3622c031c6c7da6edf3c7e1d6e83021cacb13df6f5cc2f25b518262fae59deR84).

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,110 @@
...
+            "target": "new",
+            "view_id": self.env.ref(
+                "stock_quant_package_fast_move.stock_quant_package_fast_move_wizard_view_form"
+            ).id,
+            "view_mode": "form",
+            "context": self.env.context,
+        }
+
+    def check_source_location(self):
+        if any(package.location_id != self[0].location_id for package in self):
```
ivs-cetmix: Please remove this restriction because this scenario is possible.

ivs-cetmix (changes requested):

ivs-cetmix (approved):
LGTM

GabbasovDinar (approved):
LGTM

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,166 @@
...
+        adds them to the Batch Transfer, and confirms the transfer. If the "validate"
+        option is enabled, the batch is also validated.
+
+        Parameters:
+        - location (recordset of stock.location): The destination location.
+        - destination_package (recordset of stock.quant.package, optional):
+        Optional destination package. If provided, it must belong to the specified location.
+        - validate (boolean, optional):
+        If set to True, the created picking will be validated.
+        Otherwise it will remain in the "Ready" state. Defaults to False.
```
hoangtrann: `validate` is required in args, is this docstring accurate?

hoangtrann (approved):

hoangtrann:
Tested and it's working as expected.

However, I noticed that when I choose not to `Validate` in the pop-up wizard, regardless of choosing Destination Package or Put in new Pack, the Package Content is blank.

image

Validating the picking and go back to Package Content, it's now showing.

angelinaanaki:
> Tested and it's working as expected.
> 
> However, I noticed that when I choose not to `Validate` in the pop-up wizard, regardless of choosing Destination Package or Put in new Pack, the Package Content is blank.
> 
> image
> 
> Validating the picking and go back to Package Content, it's now showing.

@hoangtrann Thank you very much for your review! The issue was due to the moves being created only at the package level while in the 'draft' state, as seen [here](https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_picking.py#L865). I resolved this by adding the 'done' state after the picking confirmation. Could you please review it again and let me know your thoughts?

hoangtrann (approved):
Looking good, package is now showing content if not validate immediately

test-requirements.txt
ivs-cetmix, line 1: stock_picking_move_package_to_package was merged, please remove it from the`test-requirements`

angelinaanaki: Done. I’ve removed stock_picking_move_package_to_package from the **test-requirements**.

ivs-cetmix:
Hey @OCA/logistics-maintainers would be great to have this PR merged. Not only it's already "ready for merge" but also has "fast" in the module name! 😄

stock_quant_package_fast_move/models/stock_quant_package.py
jbaudoux: I can't spot where is that check? And why is this necessary that they are all in the same source location?
```suggestion
        if location in self.location_id and not destination_package:
```

stock_quant_package_fast_move/models/res_company.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    package_move_picking_type_id = fields.Many2one(
+        "stock.picking.type",
+        string="Package Move Operation",
+        domain="[('show_entire_packs', '=', True)]",
+    )
+    use_batch_transfers = fields.Boolean()
```
jbaudoux, line 15: Is there any use for having this option?

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,165 @@
...
+            "res_model": "stock.quant.package.fast.move.wizard",
+            "target": "new",
+            "view_id": self.env.ref(
+                "stock_quant_package_fast_move.stock_quant_package_fast_move_wizard_view_form"
+            ).id,
+            "view_mode": "form",
+            "context": self.env.context,
+        }
+
+    def create_batch_transfer(
```
jbaudoux, line 24:
```suggestion
    def _move_with_batch_transfer(
```

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,165 @@
...
+            raise exceptions.UserError(
+                _("The destination package does not belong to the specified location.")
+            )
+
+        active_company = self.env.company
+        picking_type_id = active_company.package_move_picking_type_id
+        use_batch_transfers = active_company.use_batch_transfers
+
+        if use_batch_transfers:
+            self.create_batch_transfer(
```
jbaudoux, line 121:
```suggestion
            self._move_with_batch_transfer(
```

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,165 @@
...
+                location, destination_package, validate, picking_type_id
+            )
+        else:
+            # Create a picking
```
jbaudoux, line 125: If you create a method for above option, create also a method for this option like
`_move_with_picking(location, destination_package, validate, picking_type_id)`
to be consistent

stock_quant_package_fast_move/models/stock_quant_package.py
```diff
@@ -0,0 +1,165 @@
...
+            # Set the is_done flag to True for all package levels
+            # associated with the created picking
+            picking.package_level_ids.write({"is_done": True})
+            if validate:
+                # Validate the picking
+                picking.button_validate()
+
+        return True
+
+    def create_new_package(self):
```
jbaudoux, line 157: If you need this hook, then move it inside the wizard as this serves the wizard only

stock_quant_package_fast_move/wizard/stock_quant_package_fast_move_wizard.py
```diff
@@ -0,0 +1,40 @@
...
+    package_dest_id = fields.Many2one(
+        "stock.quant.package",
+        "Destination Package",
+        domain="[('location_id', '=', location_dest_id)]",
+    )
+    location_dest_id = fields.Many2one(
+        "stock.location", "Destination Location", required=True
+    )
+    put_in_new_package = fields.Boolean("Put in New Package")
+    validate = fields.Boolean()
```
jbaudoux, line 20: Is there any use for not validating the movement?

jbaudoux (commented):
Just minor remarks. I wonder if all the options are necessary (validate & batch)?

What happens if the moved package was reserved by another operation? Do you want to allow the move?

stock_quant_package_fast_move/readme/USAGE.md
```diff
@@ -0,0 +1,15 @@
+- Go to **Inventory -> Products -> Packages**.
+
+- Open a package or select several packages in the list view. **Important**: All selected packages must be located in the same location.
```
jbaudoux, line 3: The important restriction was dropped. This need to be removed then. And the location check here to be fixed https://github.com/OCA/stock-logistics-workflow/pull/1428/files#r1444030377

stock_quant_package_fast_move/wizard/stock_quant_package_fast_move_wizard.py
ivs-cetmix: @jbaudoux yes, sometimes we create a picking and then validate it later when goods a physically moved.

ivs-cetmix:
> Just minor remarks. I wonder if all the options are necessary (validate & batch)?
> 
> What happens if the moved package was reserved by another operation? Do you want to allow the move?

Hi @jbaudoux , first of all - thank you for your review!  Regarding your question - sounds reasonable. Let us check this.

PR #1429, opened 2023-11-16
[15.0][IMP] stock_return_request: Improve performance and simplicity.

carlosdauden:
Avoid overload normal flows to cover a non usual flow.

@Tecnativa


carlosdauden:
ping @chienandalu @sergio-teruel

rousseldenis:
@carlosdauden As you remove a part of functionality, maybe you can a little bit explain why. Thanks

stock_return_request/models/stock_move.py
```diff
@@ -30,9 +21,9 @@ def _compute_qty_returnable(self):
                 if move.state == "done":
                     move.qty_returnable = move.quantity_done
                 else:
+                    # TODO: Review if this option has sense
                     move.qty_returnable = move.reserved_availability
```
chienandalu, line 25: Maybe not so much. If there was a bussiness logic I can't remember anymore...

chienandalu (approved):
:+1: Agreed overall

chienandalu:
> @carlosdauden As you remove a part of functionality, maybe you can a little bit explain why. Thanks

Well, the functionality is not being lost. The main purpose is to turn a computed stored field into a non stored one so we can release the load of its compute outside of its intended flows

stock_return_request/models/stock_return_request.py
```diff
@@ -502,6 +501,7 @@ def _get_returnable_move_ids(self):
             moves = stock_move_obj.search(
                 line._get_moves_domain(), order=line.request_id.return_order
             )
+            moves = moves.filtered(lambda m: m.qty_returnable > 0.0)
```
rousseldenis, line 504: You should use float_compare()

carlosdauden: Changes done

stock_return_request/models/stock_move.py
```diff
@@ -11,17 +11,8 @@ class StockMove(models.Model):
         string="Returnable Quantity",
         compute="_compute_qty_returnable",
         readonly=True,
-        store=True,
-        recursive=True,
     )
 
-    @api.depends(
```
pedrobaeza, line 18: **question:** Why removing `depends`? Although being non stored, the depends indicates when to mark the result as dirty if cached AFAIK.

pedrobaeza (commented):
I have rebased and fix conflict, but there's one doubt inline.

pedrobaeza (approved):
Well, there's no answer on my question, but the optimization remains interesting, so let's merge it:

/ocabot merge patch

Please fw-port it to upper versions

PR #1430, opened 2023-11-16
[15.0][FIX] stock_return_request: Fix string to translate

pedrobaeza:
The string to translate shouldn't be put with triple quote, or it becomes a hell to be translated. The Spanish translation was also incorrect, provoking the new PO linter to fail.

@Tecnativa


rousseldenis (approved):

chienandalu (approved):

carlosdauden (approved):

pedrobaeza:
@sbidoul @gurneyalex it seems the bot is still not up.

PR #1440, opened 2023-11-27
[ADD] stock_restrict_lot_domain: products can ignore lot restriction

HekkiMelody:
New module that adds a domain in the settings. Only products included in the domain have their lot restricted.

To do:

- [ ] Automated tests
- [ ] Documentation


florian-dacosta:
Hello,
Thanks for the contribution.

Could you explain the use case ?
Why is it not possible to just not fill the `restrict_lot_id` field for the products that don't match the company domain added here ?

I am not very favorable to complexify this basic module for very specific cases, I think it would be best to try to implement it in a submodule, and eventually just adding a hook in this one do you think it is doable?

HekkiMelody:
> Hello, Thanks for the contribution.
> 
> Could you explain the use case ? Why is it not possible to just not fill the `restrict_lot_id` field for the products that don't match the company domain added here ?
> 
> I am not very favorable to complexify this basic module for very specific cases, I think it would be best to try to implement it in a submodule, and eventually just adding a hook in this one do you think it is doable?

Hello, thank you for taking the time to look at it.
The use case is that the customer wants to be able to "suggest" a lot to the logistics department without it actually becoming a hard constraint.

I do not think the added complexity is much, and I'd like to avoid adding yet another module. (and potentially a bridge module for every descendant of this one as well). Of course if you really don't think this is a good fit for this module I'll go down the route of having an additional module. Please do let me know!

florian-dacosta:
> Hello, thank you for taking the time to look at it. The use case is that the customer wants to be able to "suggest" a lot to the logistics department without it actually becoming a hard constraint.

I am not sure I get it, at least in this module. Here, the proposed change will just stop the propagation of the `restrict_lot_id` in case of chained stock moves. How does it help to suggest something to the logistics department ?

On the sale_order_lot_selection module, you already stop the propagation of the lot from the sale order to the stock move, I don't see why you should do the same here.
For now IMHO, I strongly feel it should be part of another module.

HekkiMelody:
> > Hello, thank you for taking the time to look at it. The use case is that the customer wants to be able to "suggest" a lot to the logistics department without it actually becoming a hard constraint.
> 
> I am not sure I get it, at least in this module. Here, the proposed change will just stop the propagation of the `restrict_lot_id` in case of chained stock moves. How does it help to suggest something to the logistics department ?
> 
> On the sale_order_lot_selection module, you already stop the propagation of the lot from the sale order to the stock move, I don't see why you should do the same here. For now IMHO, I strongly feel it should be part of another module.

That's fair enough, I definitely see where you're coming from. I'll move it to another module, maybe just adding a hook here if needed. Thank you for your time!

francesco-ooops:
@aleuffre can you update PR name?

HekkiMelody:
> @aleuffre can you update PR name?

Sorry, I thought I had.

HekkiMelody:
I think I'll find some other way of doing this.

PR #1443, opened 2023-11-29
[14.0][IMP] stock_restrict_lot: validate lot moved is correct

HekkiMelody:
Adds a check on picking's validation that the lot indicated in the move line is actually the one in the restriction.

TODO:
- [x] Tests


francesco-ooops:
@florian-dacosta should be ready for review now

stock_restrict_lot/models/stock_move.py
```diff
@@ -92,3 +93,26 @@ def _split(self, qty, restrict_partner_id=False):
...
+        res = super()._action_done(cancel_backorder=cancel_backorder)
+        self._check_lot_consistent_with_restriction()
+        return res
+
+    def _check_lot_consistent_with_restriction(self):
+        """
+        Check that the lot set on move lines
+        is the same as the restricted lot set on the move
+        """
+        for move in self.filtered(lambda x: x.restrict_lot_id and x.move_line_ids):
```
geomer198: In this function has 2 cycle, I think better do checks in main cycle without `filtered` function.
```py
for move in self:
    if not (move.restrict_lot_id and move.move_line_ids):
        continue
    move_line_lot = move.mapped("move_line_ids.lot_id")
    ...
```

stock_restrict_lot/models/stock_move.py
```diff
@@ -92,3 +93,26 @@ def _split(self, qty, restrict_partner_id=False):
...
+            move_line_lot = move.mapped("move_line_ids.lot_id")
+            if move.restrict_lot_id != move_line_lot:
+                raise UserError(
+                    _(
+                        "The lot(s) %(move_line_lot)s being moved is "
+                        "inconsistent with the restriction on "
+                        "lot %(move_restrict_lot)s set on the move",
+                        move_line_lot=", ".join(x.display_name for x in move_line_lot),
```
geomer198: Here you can use `mapped` function without cycle:
```py
move_line_lot=", ".join(move_line_lot.mapped('display_name')),
```
I think this code looks more clean.

stock_restrict_lot/tests/test_restrict_lot.py
```diff
@@ -271,3 +272,40 @@ def test_compute_quantites(self):
         product.invalidate_cache()
         product = product.with_context(lot_id=lot2.id)
         self.assertEqual(product.outgoing_qty, 1)
+
+    def test_move_validation_inconsistent_lot(self):
```
geomer198, line 276: Could you add docstring for this function?

geomer198 (changes requested):
Please fix the code by my comments.

HekkiMelody:
> Please fix the code by my comments.

Thank you for your review @geomer198 , I've addressed all your comments

stock_restrict_lot/readme/CONTRIBUTORS.rst
```diff
@@ -1,2 +1,8 @@
 * Florian da Costa <florian.dacosta@akretion.com>
 * Michael Tietz (MT Software) <mtietz@mt-software.de>
+
+* Ooops404
```
rousseldenis, line 4: @aleuffre @francesco-ooops The contributor's name should be visible

francesco-ooops: Hi @rousseldenis , once again I think we should clarify what does "contributor" mean, I would be happy to receive your feedback in https://github.com/OCA/maintainer-tools/issues/592 

this is fixed anyway

rousseldenis: A contributor is a physical individual.

francesco-ooops: @rousseldenis added the physical individual, myself

geomer198 (approved):
LGTM!

GSLabIt (approved):

francesco-ooops:
@florian-dacosta looks good to you? :)

florian-dacosta (approved):
LGTM
But which is the case where we could an inconsistency between the lot of the `stock.move.line` and the `stock.move`?

Is it manual mistak choosing the wrong lot in the UI? In that case, I guess it could also be solved by adding a domain on the `stock.move.line` `lot_id` depending on the `restrict_lot_id` of the stock.move.

Another possibility, to block the user before the validation, would be to replace the `action_done` override by an `api.constrains` method to check the lot.

Anyway, ok for me the way it is.

HekkiMelody:
> LGTM But which is the case where we could an inconsistency between the lot of the `stock.move.line` and the `stock.move`?
> 
> Is it manual mistak choosing the wrong lot in the UI? In that case, I guess it could also be solved by adding a domain on the `stock.move.line` `lot_id` depending on the `restrict_lot_id` of the stock.move.
> 
> Another possibility, to block the user before the validation, would be to replace the `action_done` override by an `api.constrains` method to check the lot.
> 
> Anyway, ok for me the way it is.

Thanks for the approval!

I didn't want to add a constraint because `stock.move` and `stock.move.line` are not written to at the same time, and there could be times just in the normal functioning of the `stock` module where the fields are not in sync (maybe the move line has been created but the lot has not been selected/reserved yet, etc), so having just a check in `._action_done()` seemed safer to me.

francesco-ooops:
@pedrobaeza go with this?

francesco-ooops:
or actually also @florian-dacosta can merge :)

PR #1444, opened 2023-11-30
[14.0] [FIX] stock_picking_inter_warehouse: proper merge picking managing

renda-dev:
Before this PR the "Disable Merge Picking Moves" option on the picking type would have split the picking lines into multiple pickings.

The correct behavior is that the picking lines from the same picking should be aggregated, while the picking lines from different pickings should not be aggregated between them.


francesco-ooops (approved):
Functional ok!

PR #1445, opened 2023-11-30
[ADD][12.0] product_weighable_default_weight : new module to guess weight field from products UoM, for weighable products

legalsylvain:
This module extends the functionality of product module to set default value on weight field for weighable products.

Once installed, when a user create a new weighable product (``uom_id.measure_type == "weight"``), the ``weight`` field will be computed, depending on the UoM.

- if UoM is "kg", weight will be 1.0
- if UoM is "ton", weight will be 1000.0

This module is interesting if you use other modules based on the ``weight`` field. (Ex: ``sale_order_weight`` in OCA/sale-workflow repo).

### DIFF
--- a/product_weighable_default_weight/i18n/fr.po
+++ b/product_weighable_default_weight/i18n/fr.po
@@ -1,0 +1,27 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#   * product_weighable_default_weight
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 12.0\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2024-02-12 15:39+0000\n"
+"PO-Revision-Date: 2024-02-12 15:39+0000\n"
+"Last-Translator: <>\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: product_weighable_default_weight
+#: model:ir.model,name:product_weighable_default_weight.model_product_product
+msgid "Product"
+msgstr "Article"
+
+#. module: product_weighable_default_weight
+#: model:ir.model,name:product_weighable_default_weight.model_product_template
+msgid "Product Template"
+msgstr "Modèle d'article"
+
--- a/product_weighable_default_weight/models/product_product.py
+++ b/product_weighable_default_weight/models/product_product.py
@@ -1,0 +1,35 @@
+# Copyright (C) 2023 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, api
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.onchange("uom_id", "uom_po_id")
+    def _onchange_uom(self):
+        res = super()._onchange_uom()
+        new_weight = self.product_tmpl_id._get_weight_from_uom(self.uom_id)
+        if new_weight:
+            self.weight = new_weight
+        return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            new_weight = self.env["product.template"]._get_weight_from_uom_id(
+                vals.get("uom_id", False)
+            )
+            if new_weight:
+                vals["weight"] = new_weight
+        return super().create(vals_list)
+
+    def write(self, vals):
+        new_weight = self.env["product.template"]._get_weight_from_uom_id(
+            vals.get("uom_id", False)
+        )
+        if new_weight:
+            vals["weight"] = new_weight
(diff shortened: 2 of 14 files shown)


product_weighable_default_weight/i18n/fr.po
```diff
@@ -0,0 +1,49 @@
...
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: nplurals=2; plural=n != 1;\n"
+"X-Generator: Weblate 3.10\n"
+
+#. module: pos_default_empty_image
+#: model:ir.model.fields,field_description:pos_default_empty_image.field_pos_config__iface_fixed_font_size
+#: model_terms:ir.ui.view,arch_db:pos_default_empty_image.pos_config_view_form
+msgid "Fixed Font Size"
+msgstr "Tamaño de fuente fijo"
```
quentinDupont: para mi humilde conocimiento del francés, esto es español :smile:

product_weighable_default_weight/models/product_template.py
```diff
@@ -0,0 +1,46 @@
...
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.onchange('uom_id')
+    def _onchange_uom_id(self):
+        res = super()._onchange_uom_id()
+        new_weight = self._get_weight_from_uom(self.uom_id)
+        if new_weight:
+            self.weight = new_weight
+        return res
```
quentinDupont, line 17:
```suggestion
        else:
            self.weight = 0
        return res
```

product_weighable_default_weight/models/product_product.py
```diff
@@ -0,0 +1,35 @@
...
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.onchange("uom_id", "uom_po_id")
+    def _onchange_uom(self):
+        res = super()._onchange_uom()
+        new_weight = self.product_tmpl_id._get_weight_from_uom(self.uom_id)
+        if new_weight:
+            self.weight = new_weight
+        return res
```
quentinDupont, line 17:
```suggestion
        else:
            self.weight = 0
        return res
```

quentinDupont (changes requested):
My review :
- French translation instead of spanish :) 
- And my changes are about a non expected behaviour : if you choose "ton", it write "1000" with onchange, then if you decide to choose "Hour" or "Unit" (UoM not weighable), the value stays at "1000" → need to be set as default, and default is 0 if I don't say crap

product_weighable_default_weight/i18n/fr.po
legalsylvain: Ni siquiera entiendo cómo pude haber hecho eso...

legalsylvain:
>  if you choose "ton", it write "1000" with onchange, then if you decide to choose "Hour" or "Unit" (UoM not weighable), the value stays at "1000" → need to be set as default, and default is 0 if I don't say crap

I'm not sure about that point. I mean, **indeed** in your use case, the expected behaviour is to reset to 0. but in the following use case, I don't think so : 
- user click on "create a new product"
- He set **first** the weight of the product. (digamos: 0,5 kilogramos) 
- Then he set the uom field as "Unit".
- total : it will reset the "weight" field, while the value is legitimate.

My perception of onchange (and of compute + store=False in recent versions) is that it should change a field if the old value is "reasonably" the bad one, and the new one is "reasonably" the good one. Not sure if it's a real "odoo rule" though.

product_weighable_default_weight/i18n/fr.po
quentinDupont: es porque eres demasiado bilingüe mientras yo uso Deepl jaja

quentinDupont:
> > if you choose "ton", it write "1000" with onchange, then if you decide to choose "Hour" or "Unit" (UoM not weighable), the value stays at "1000" → need to be set as default, and default is 0 if I don't say crap
> 
> I'm not sure about that point. I mean, **indeed** in your use case, the expected behaviour is to reset to 0. but in the following use case, I don't think so :
> 
>     * user click on "create a new product"
> 
>     * He set **first** the weight of the product. (digamos: 0,5 kilogramos)
> 
>     * Then he set the uom field as "Unit".
> 
>     * total : it will reset the "weight" field, while the value is legitimate.
> 
> 
> My perception of onchange (and of compute + store=False in recent versions) is that it should change a field if the old value is "reasonably" the bad one, and the new one is "reasonably" the good one. Not sure if it's a real "odoo rule" though.

yes, UX stuff.. maybe there is no good answer ! But i quite agree with your use cases, better than my example ! so it's a  :heavy_check_mark: for me !

quentinDupont (approved):

product_weighable_default_weight/i18n/fr.po
legalsylvain: Fixed !

bealdav (approved):
LGTM

hparfr (approved):

legalsylvain:
Hi @hparfr could you merge this approved one ?
thanks !

legalsylvain:
Hi @OCA/stock-weighing-maintainers somebody could merge this PR approved by 3 people since 4 monthes ? 

thanks !

rvalyi:
merging based on approvals.

/ocabot merge nobump

PR #1446, opened 2023-12-01
[17.0][MIG] stock_picking_invoice_link: Migration to 17.0

lethuthao2403:
ref: [BSRD-699](https://jira.camptocamp.com/browse/BSRD-699)


gurneyalex (approved):

gurneyalex (approved):

loicgrifnee:
There seems to be an error in the way of overriding the reverse_moves method. 


```python
def reverse_moves(self, is_modify=False):
     action = super(
           AccountMoveReversal, self.with_context(force_copy_stock_moves=True)
     ).reverse_moves()
``` 

By not passing the is_modify=is_modify parameter, this prevents proper functioning.

ferran-S73:
> There seems to be an error in the way of overriding the reverse_moves method.
> 
> ```python
> def reverse_moves(self, is_modify=False):
>      action = super(
>            AccountMoveReversal, self.with_context(force_copy_stock_moves=True)
>      ).reverse_moves()
> ```
> 
> By not passing the is_modify=is_modify parameter, this prevents proper functioning.

Fixed in 
- https://github.com/OCA/stock-logistics-workflow/pull/1703

@lethuthao2403 @gurneyalex could you review?

PR #1448, opened 2023-12-04
[14.0] Create module sale_purchase_stock_line_note

grindtildeath:
This module allows to propagate the value of note field from the sale
order lines to the stock moves and from the stock moves to the purchase
order lines in case of MTO.

### DIFF
--- a/sale_purchase_stock_line_note/tests/test_sale_stock_purchase_line_note.py
+++ b/sale_purchase_stock_line_note/tests/test_sale_stock_purchase_line_note.py
@@ -1,0 +1,37 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo.tests.common import Form, SavepointCase
+
+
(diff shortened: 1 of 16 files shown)


mmequignon (approved):

TDu (approved):

sale_purchase_stock_line_note/tests/test_sale_stock_purchase_line_note.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo.tests.common import Form, SavepointCase
+
+
+class TestSalePurchaseStockLineNote(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 9:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

simahawk (approved):
LG

PR #1450, opened 2023-12-06
[14.0] [FIX] stock_picking_inter_warehouse: picking type permission error

renda-dev:
Trying to create a picking will write the picking type `default_location_dest_id` and would raise an error if the user creating the picking has user access to the inventory.

With this fix it will write on the picking type only when it's strictly necessary with sudo access to avoid the error.


HekkiMelody (approved):
The fix is okay.

That `_onchange_partner_id` is suspect in the first place and I think it should be fixed or removed entirely, though, possibly in a different PR.

PR #1451, opened 2023-12-12
[16.0][IMP] stock_picking_group_by_max_weight: Split if max weight exceed

lmignon:


jbaudoux (changes requested):
I don't like the approach. It doesn't look to work properly when moves are added. Post processs is called before action_confirm, so you break the moves merging mechanism of odoo.

You better not split the moves in multiple pickings at this stage.

I would have implemented this at the batch picking creation based on the picking device limit. If the picking weight exceeds the picking device max weight, create a split order (you have the method in shopfloor) containing a subset of the moves within the weight limit.
In your PR here you already have defined the concept of max weight: https://github.com/OCA/wms/pull/534/files#diff-b737ddc38c458efb5b613c33ee65b4cefb4a0fc5b0dc46e93b7712ec34cda478R22
Add the split picking creation as mentioned here : https://github.com/OCA/wms/pull/534/files#r1428768425

jbaudoux (approved):
As the module already exists and this improves it, let's approve this PR.

I still think this should be approached differently...

rousseldenis (approved):

PR #1452, opened 2023-12-14
[16.0][IMP] delivery_procurement_group_carrier: Set Carrier on new picking from procurement.group and group new  moves per carrier

sbejaoui:
forward-port of #1292


jbaudoux (approved):
Thanks for the fw-port and v16 improvements

Pay attention to the length of your first line commit message...

PR #1454, opened 2023-12-15
[16.0][FIX] group_by_partner_by_carrier: Group with carrier propagation

lmignon:
If the stock rule is configured to propagate the carrier, the domain to group pickings must add the criteria on the carrier. Before this change, in a warehouse configured in multi steps deliveries with the carrier propagation set to true, it was not possible to group pickings except the outgoring ones.


lmignon:
cc @jbaudoux @rousseldenis @simahawk @cyrilmanuel

rousseldenis (approved):

jbaudoux (changes requested):
Doesn't make sense.
Grouping cannot be enabled on non outgoing moves, and this on purpose

simahawk:
I agree w/ JE. Can you elaborate a bit on the use case?

PR #1458, opened 2023-12-19
[14.0] [IMP] stock_picking_inter_warehouse: sync picking on validate

renda-dev:
Instead of syncing pickings on confirm sync them on validate.

That helps including every changes made to the picking before validating it, which it would not be possible if the user creates a wrong picking as an "Immediate Transfer".


francesco-ooops (approved):
Functional ok!

HekkiMelody (approved):
Code review

PR #1463, opened 2023-12-22
[WIP][ADD] stock_picking_by_package: Select Packages to add their content to a stock transfer

dreispt:
https://github.com/OCA/stock-logistics-workflow/blob/7b83c22eb9181403e8d46bd2ea32ac1f9702161b/stock_picking_by_package/README.rst


rousseldenis:
@dreispt Have you looked at https://github.com/OCA/stock-logistics-workflow/pull/1422 ?

Isn't it the same flow ?

dreispt:
@rousseldenis Good catch, probably two perspectives for the same problem.

PR #1472, opened 2024-01-08
[17.0] [IMP] stock_picking_invoice_link: Added pickings to notebook sheet

NICO-SOLUTIONS:


NICO-SOLUTIONS:
dont want to break workflow with other modules... is it ok to add the returns to the pickings? 
i am tryoing to improve account_invoice_report_grouped_by_picking, to reflect invoice lines even if the returned quantity is equal to the delierever quantity. Also i´d like to implement the possiblity to have all information on invoice when returned quantity is greater then delivered. At the moment the initial delivered goods that are returned will lose its relation to the delivery in that case....

So if anybody has problems going on with it... please let me know

NICO-SOLUTIONS:
feel free to share your thoughts

PR #1476, opened 2024-01-12
[FIX] simplify the name of merged procurment groups

sbejaoui:
When merging procurement groups for numerous sales orders, we often end up with an excessively long procurement group name that includes all the associated sale orders. 

it's ugly and useless, considering that we can access the linked orders directly from the group. This fix addresses the issue by simply mentioning the related partner name.


lmignon (approved):

rousseldenis:
@sbejaoui Could you include module name in commit message ?

rousseldenis (approved):
Code review

/ocabot merge minor

PR #1480, opened 2024-01-16
[14.0][FIX] stock_landed_costs_delivery: Filtering the correct Landed Cost

victoralmau:
It is necessary to filter the correct Landed Cost (from the correct picking) to set the cost lines.

Example use case:
- PO without carrier.
- Picking A without carrier done > LC A (draft) without cost lines.
- Picking B with carrier done > LC B (done) with cost lines.

@Tecnativa TT45656

### DIFF
--- a/stock_landed_costs_delivery/models/stock_picking.py
+++ b/stock_landed_costs_delivery/models/stock_picking.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Tecnativa - Víctor Martínez
+# Copyright 2023-2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from odoo import models
 from odoo.tools import config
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
LGTM :+1:

stock_landed_costs_delivery/models/stock_picking.py
```diff
@@ -7,6 +7,12 @@
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    def _add_delivery_cost_to_po(self):
+        """We add the context to identify in _create_delivery_line() method the picking."""
+        self.ensure_one()
+        self = self.with_context(from_picking_id=self.id)
```
pedrobaeza: You can pass the recordset instead of the ID, and it's best practice.

victoralmau:
Changes done.

stock_landed_costs_delivery/models/stock_picking.py
```diff
@@ -7,6 +7,12 @@
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
+    def _add_delivery_cost_to_po(self):
+        """We add the context to identify in _create_delivery_line() method the picking."""
+        self.ensure_one()
+        self = self.with_context(from_picking_id=self)
```
pedrobaeza: If you pass a recordset, be consistent with the context key name:

```suggestion
        self = self.with_context(from_picking=self)
```

pedrobaeza (approved):
/ocabot merge patch

PR #1484, opened 2024-01-18
[15.0][FIX] stock_landed_costs_delivery: Filtering the correct Landed Cost

victoralmau:
FWP from 14.0: https://github.com/OCA/stock-logistics-workflow/pull/1480

It is necessary to filter the correct Landed Cost (from the correct picking) to set the cost lines.

Example use case:
- PO without carrier.
- Picking A without carrier done > LC A (draft) without cost lines.
- Picking B with carrier done > LC B (done) with cost lines.

@Tecnativa TT45656


pedrobaeza (approved):
Fast-tracking the fw-port as merged in previous version:

/ocabot merge patch

PR #1488, opened 2024-01-23
[14.0][ADD] stock_rule_buy_check_stock_before

TDu:


rousseldenis:
@TDu Have you looked at https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_mts_mto_rule ?

It seems this is a duplicate

TDu:
Yes looks similar, thanks for the link.
But the idea behind this one is to use it with the dropshipment route, which I do not think it is possible with that other one. Or I am mistaken ?

stock_rule_buy_check_stock_before/models/procurement_group.py
```diff
@@ -0,0 +1,34 @@
...
+from odoo.osv import expression
+
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    def _is_product_available(self, product_id, quantity):
+        if quantity == 0:
+            return False
+        return product_id.free_qty >= quantity
```
jbaudoux, line 14: Use `immediately_usable_qty` instead of `free_qty`
Give the warehouse in context to support multi-wh
Use float_compare

jbaudoux (changes requested):

stock_rule_buy_check_stock_before/models/procurement_group.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def _get_rule(self, product_id, location_id, values):
+        disable_dropshiping = False
+        sale_line_id = values.get("sale_line_id")
+        if sale_line_id:
+            sale_line = self.env["sale.order.line"].browse(sale_line_id)
+            warehouse = values.get("warehouse_id")
+            if sale_line and self.with_context(
+                warehouse=warehouse.id
+            )._is_product_available(product_id, sale_line.product_uom_qty):
```
jbaudoux, line 24: Would be cleaner to give the warehouse in argument and let the method set the context

stock_rule_buy_check_stock_before/models/stock_rule.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    disable_if_stock_exists = fields.Boolean(string="Bypass if stock exists")
```
jbaudoux, line 10:
```suggestion
    disable_if_stock_exists = fields.Boolean(string="Bypass if available stock")
```

stock_rule_buy_check_stock_before/models/procurement_group.py
jbaudoux, line 24: This computation in `_get_rule` seems to early. I think the approach of mentioned module of Denis is better. You cannot compute this for all deliveries

jbaudoux:
@Tdu Can you test by adapting https://github.com/OCA/stock-logistics-warehouse/tree/14.0/stock_mts_mto_rule ?
You need:
* to be able to set a location on the rule
* to be able to configure if you want partial qty
* to extract the method to compute the available stock and use `immediately_usable_qty`

TDu:
@rousseldenis Thanks for pointing me to that other module.
This one is a fail attempt...closing it.

PR #1489, opened 2024-01-23
[15.0][IMP] stock_picking_import_serial_number: Add package compatibility

victoralmau:
Adds the possibility to set the package name (`stock.quant.package`) in the import file to be auto-created with that name.

Please @carlosdauden and @sergio-teruel can you review it?

@Tecnativa TT47271


stock_picking_import_serial_number/wizard/import_serial_number_wizard.py
```diff
@@ -41,6 +42,9 @@ def default_get(self, fields):
         string="Column index for product", default=0
     )
     sn_serial_column_index = fields.Integer(string="Column index for S/N", default="1")
+    sn_package_column_index = fields.Integer(
+        string="Column index for Package", default="2"
```
chienandalu: Is it right to put a string as a default in an `Integer`?

carlosdauden (approved):

chienandalu (approved):
:+1:

victoralmau:
This would be ready to merge, right?

stock_picking_import_serial_number/wizard/import_serial_number_wizard.py
```diff
@@ -67,14 +71,37 @@ def action_import(self):
                 self._import_serial_number(xl_sheet, move_lines, picking)
         self.data_file = False
 
+    def _search_or_create_package(self, name):
+        """Return the package (if it exists) or create a new one."""
+        package_model = self.env["stock.quant.package"]
+        package = package_model.search([("name", "=", name)], limit=1)
```
pedrobaeza: Packages have company?

victoralmau: Right, added.

stock_picking_import_serial_number/wizard/import_serial_number_wizard.py
```diff
@@ -67,14 +71,39 @@ def action_import(self):
                 self._import_serial_number(xl_sheet, move_lines, picking)
         self.data_file = False
 
+    def _search_or_create_package(self, name):
+        """Return the package (if it exists) or create a new one."""
+        package_model = self.env["stock.quant.package"]
+        package = package_model.search(
+            [("name", "=", name), ("company_id", "=", self.env.company.id)], limit=1
```
pedrobaeza: `company_id` is required?

stock_picking_import_serial_number/wizard/import_serial_number_wizard.py
```diff
@@ -67,14 +71,39 @@ def action_import(self):
...
+        )
+        return package or package_model.create({"name": name})
```
pedrobaeza, line 85: Missing `company_id`

victoralmau: `company_id` field is a compute store.

I think it should be taken into account for search but should be omitted in create because it will be automatically set with the quants information (https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_quant.py#L1047)

pedrobaeza: Uhm, then maybe we should ignore it also in search or sophisticate the search, as if the package is empty, it won't match correctly.

victoralmau: Ok, ignored the field in search.

pedrobaeza: I think the search should be `"|", ("company_id", "=", False), ("company_id", "=", self.picking_ids.company_id.id)`

victoralmau:
Changes done.

pedrobaeza (approved):
/ocabot merge major

PR #1491, opened 2024-01-25
[15.0][ADD] stock_picking_auto_revert

AaronHForgeFlow:
Module to correct mistakes in pickings the fast way. Instead of cancel this module will create a return and then will recreate the issued picking.

Based on https://github.com/OCA/stock-logistics-workflow/pull/593 Originally it was discarded because of the stock_cancel module.

Similar approach Odoo uses for correcting mistakes in invoices. that is, do a reverse and recreate the original invoice in draft state so the user can edit it.

cc @ForgeFlow

This PR Replaces: https://github.com/OCA/stock-logistics-warehouse/pull/1918


stock_picking_auto_revert/models/stock_move.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "stock.move"
+
+    def _check_restrictions(self):
+        # Restrictions before remove quants
+        if self.returned_move_ids:
+            raise exceptions.UserError(
+                _("Action not allowed. Move splited / with returned moves.")
+            )
+        if self.move_dest_ids or self.search([("move_dest_ids", "in", self.ids)]):
+            raise exceptions.UserError(_("Action not allowed. Chained move."))
```
rousseldenis: A better user message could be great.

`You cannot revert this stock picking. Its stock moves are linked to other moves. Origin pickings: %(origin_pickings)s. Destination pickings: %(destination_pickings)%`

stock_picking_auto_revert/models/stock_move.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _check_restrictions(self):
+        # Restrictions before remove quants
+        if self.returned_move_ids:
+            raise exceptions.UserError(
+                _("Action not allowed. Move splited / with returned moves.")
```
rousseldenis:
```suggestion
                _("You cannot revert this stock picking. Move splited / with returned moves.")
```

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+        default_data.update({"location_id": pick.location_id.id})
+        return_wiz = StockReturnPicking.create(default_data)
+        return_wiz._onchange_picking_id()
+        for rm in return_wiz.product_return_moves:
+            sm = pick.move_lines.filtered(lambda x: x.product_id.id == rm.product_id.id)
+            if rm.quantity < sm.product_uom_qty:
+                raise exceptions.UserError(msg)
+        try:
+            res = return_wiz.create_returns()
+        except exceptions.UserError as e:
```
rousseldenis: Do you need the try/except block as you do nothing?

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+
+        return_pick = self.env["stock.picking"].browse(res["res_id"])
+
+        # Validate picking
+        return_pick.action_set_quantities_to_reservation()
+        return_pick._action_done()
+        new_pick = pick.copy()
+        new_pick.action_assign()
+        action = self.env.ref("stock.action_picking_tree_all")
+        result = action.read()[0]
```
rousseldenis:
```suggestion
        result = self.env["ir.actions.act_window"]._for_xml_id("stock.action_picking_tree_all")
```

stock_picking_auto_revert/tests/test_stock_picking.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2019 ForgeFlow
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestStockPicking(TransactionCase):
+    def setUp(self):
```
rousseldenis: No need of this override

rousseldenis (changes requested):
Code review. Seems great!

GuillemCForgeFlow (commented):
functional and code review 👍🏿 
I'm not sure if it would be useful to add the reference of the original transfer on the Source Document for the newly created, just so the user has some more traceability on what has happened.

stock_picking_auto_revert/models/stock_move.py
AaronHForgeFlow: Much better! thanks

stock_picking_auto_revert/models/stock_move.py
AaronHForgeFlow: Suggestion applied.

stock_picking_auto_revert/models/stock_picking.py
AaronHForgeFlow: I think this was legacy code, removed.

stock_picking_auto_revert/models/stock_picking.py
AaronHForgeFlow: Opps, did not see that. I just changed it.

AaronHForgeFlow:
Thanks @GuillemCForgeFlow I added the original picking name between brackets, like this:

image

GuillemCForgeFlow (approved):
LGTM! thanks for adding the suggestion 👍🏿

stock_picking_auto_revert/tests/test_stock_picking.py
AaronHForgeFlow: You are right. I just removed it.

rousseldenis:
@AaronHForgeFlow Maybe testing exceptions could be great

AaronHForgeFlow:
@rousseldenis tests added :)

rousseldenis:
> @rousseldenis tests added :)

A module that passed several versions :-)

stock_picking_auto_revert/tests/test_stock_picking.py
```diff
@@ -0,0 +1,123 @@
+# Copyright 2019 ForgeFlow
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.exceptions import UserError
+from odoo.tests import Form
+from odoo.tests.common import TransactionCase
+
+
+class TestStockPicking(TransactionCase):
+    def setUp(self):
```
rousseldenis: Just one more thing, prefer `classmethod` here.

stock_picking_auto_revert/view/stock_view.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
rousseldenis: The folder name should be `views`

stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -0,0 +1,63 @@
...
+    _inherit = "stock.picking"
+
+    def _check_restrictions(self):
+        # returned_move_ids in stock.move
+        # split_from in stock.move
+        for move in self.move_lines:
+            move._check_restrictions()
+
+    def action_revert_recreate(self):
+        self.ensure_one()
```
rousseldenis, line 17: And one more question.

Should'nt it be more interesting to do a wizard for that flow ?

AaronHForgeFlow: Perhaps. I did it this way because the users just want less clicks as possible. I prefer not to change it, if not necessary.

stock_picking_auto_revert/models/stock_move.py
```diff
@@ -0,0 +1,33 @@
...
+            origin_pickings = ", ".join(self.mapped("picking_id.name"))
+            destination_pickings = ", ".join(
+                self.mapped("move_dest_ids.picking_id.name")
+            )
+            raise exceptions.UserError(
+                _(
+                    "You cannot revert this stock picking. Its stock moves are linked to "
+                    "other moves. Origin pickings: {origin_pickings}. "
+                    "Destination pickings: {destination_pickings}."
+                ).format(
```
rousseldenis: Ooops, don't use `format()` but use translation function arguments.

AaronHForgeFlow: Yes, sorry for this. It is now fixed.

dreispt (approved):

dreispt:
Related to #603

PR #1492, opened 2024-01-25
[16.0][FIX] stock_receipt_lot_info: hook merged into odoo

LoisRForgeFlow:
A hook was introduced in odoo (15.0 and 16.0), but with a different name `_get_value_production_lot`.

Tests in version 16 were incorrect... not testing what was needed, in this commit this is fixed and they test the functionality properly.

Forwardport of last 3 commits in https://github.com/OCA/stock-logistics-workflow/pull/1363


JoanSForgeFlow (approved):
Works as expected!

JoanSForgeFlow (approved):

PR #1494, opened 2024-01-30
[14.0][MIG] stock_move_backdating

toita86:


stock_move_backdating/models/stock_move_line.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+def check_date(date):
+    now = fields.Datetime.now()
+    if date and date > now:
+        raise UserError(
+            _("You can not process an actual " "movement date in the future.")
+        )
```
PicchiSeba: **question**: given that this function is not specific to `stock.move.line` and it's used in many places, what do you think about moving it to a common file? They're tipically named `utils.py` and they're located in the module's root directory.

You could also merge these strings because the line doesn't exceed 88 characters.
```suggestion
def check_date(date):
    now = fields.Datetime.now()
    if date and date > now:
        raise UserError(
            _("You can not process an actual movement date in the future.")
        )
```

stock_move_backdating/models/stock_inventory.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def post_inventory(self):
+        no_backdate_inventories = self.env["stock.inventory"].browse()
+        for inventory in self:
+            date_backdating = inventory.date_backdating
+            if date_backdating:
+                inventory_ctx = inventory.with_context(
+                    date_backdating=date_backdating,
+                )
+                super(StockInventory, inventory_ctx).post_inventory()
```
PicchiSeba, line 31: **suggestion**: I think doing it this way works the same and it's clearer

```suggestion
                inventory_ctx.post_inventory()
```

stock_move_backdating/models/stock_inventory.py
```diff
@@ -0,0 +1,36 @@
...
+            else:
+                no_backdate_inventories |= inventory
+        res = super(StockInventory, no_backdate_inventories).post_inventory()
```
PicchiSeba, line 34: **suggestion**: same as before
```suggestion
        res = no_backdate_inventories.post_inventory()
```

stock_move_backdating/models/stock_inventory.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @api.onchange(
+        "date_backdating",
+    )
+    def onchange_date_backdating(self):
+        self.ensure_one()
+        check_date(self.date_backdating)
+
+    def post_inventory(self):
+        no_backdate_inventories = self.env["stock.inventory"].browse()
```
PicchiSeba, line 24: That empty `browse` is not necessary, you get the same result by referencing the model

```suggestion
        no_backdate_inventories = self.env["stock.inventory"]
```

stock_move_backdating/models/stock_inventory_line.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import models
+
+
+class StockInventoryLine(models.Model):
+    _inherit = "stock.inventory.line"
+
+    def _get_move_values(self, qty, location_id, location_dest_id, out):
+        res = super()._get_move_values(qty, location_id, location_dest_id, out)
+        self.ensure_one()
```
PicchiSeba: **suggestion**: Let's switch them. So we don't have unnecessary calls when using this method on recordsets having more than 1 element

```suggestion
        self.ensure_one()
        res = super()._get_move_values(qty, location_id, location_dest_id, out)
```

PicchiSeba (commented):
Code Review: A few observations

**polish**: remove "[14.0]" from the last commit message

stock_move_backdating/models/stock_inventory.py
toita86: I tried implementing this change but it breaks the module. The same goes for the change at line `31`.

PicchiSeba: nvm then, we can ignore it

PicchiSeba: does it give you any specific error?

toita86: By changing line `31`, first it shutsdown odoo and on the logs I have
```py
  File "/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_trace_dispatch_regular.py", line 359, in __call__
    is_stepping = pydev_step_cmd != -1
RecursionError: maximum recursion depth exceeded in comparison
Fatal Python error: Cannot recover from stack overflow.
```

PicchiSeba: Oh right, I just realized we're in the `post_inventory` method already.
Clearly if we call it again we enter in an infinite call loop. You did it correctly the first time :D

PicchiSeba: You can mark these issues as resolved

PicchiSeba (approved):
Code Review: LGTM

stock_move_backdating/models/stock_inventory.py
toita86: I understand, thank you for the explanation!

francesco-ooops (approved):
Functional ok!

francesco-ooops:
@rousseldenis good to go? thanks!

stock_move_backdating/models/stock_inventory.py
```diff
@@ -21,7 +21,7 @@ def onchange_date_backdating(self):
         check_date(self.date_backdating)
 
     def post_inventory(self):
-        no_backdate_inventories = self.env["stock.inventory"].browse()
+        no_backdate_inventories = self.env["stock.inventory"]
```
rousseldenis: @toita86 The former writing was the good one.

stock_move_backdating/models/stock_inventory_line.py
```diff
@@ -8,8 +8,8 @@ class StockInventoryLine(models.Model):
     _inherit = "stock.inventory.line"
 
     def _get_move_values(self, qty, location_id, location_dest_id, out):
-        res = super()._get_move_values(qty, location_id, location_dest_id, out)
         self.ensure_one()
+        res = super()._get_move_values(qty, location_id, location_dest_id, out)
```
rousseldenis, line 11: @toita86 As super method calls ensure_one(), you can remove it here.

stock_move_backdating/models/stock_move.py
```diff
@@ -27,6 +27,7 @@ def _backdating_account_moves(self):
             )
             move_account_moves.update(
                 {
+                    "name": False,
```
rousseldenis, line 30: Why this ?

stock_move_backdating/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_fill_date_backdating,access_fill_date_backdating,model_fill_date_backdating,base.group_user,1,0,0,0
```
rousseldenis: Shouldn't it be stock.group_stock_user instead ?

rousseldenis (changes requested):
Code review

stock_move_backdating/models/stock_inventory.py
PicchiSeba: Why is it? We get the same result

rousseldenis: Yes, but formally this is better as instance is not really the same as a browse record set. Nevertheless, this one is not blocking

stock_move_backdating/models/stock_picking.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _action_done(self):
+        result = super()._action_done()
+        for picking in self:
+            picking._backdating_update_picking_date()
+            picking._backdating_update_account_moves_date()
+        return result
+
+    # Disable pylint to match the exact signature in super
+    # pylint: disable=dangerous-default-value
```
HekkiMelody: **chore**: these two lines can be removed, it's no longer necessary to disable pylint

stock_move_backdating/models/stock_move.py
HekkiMelody: Not entirely sure, but I think this is suggested by this error message

https://github.com/Odoo/odoo/blob/14.0/addons/account/models/sequence_mixin.py#L68-L75

Without it, this constraint fails as there is a mismatch between the date and the name's implied date, checked by that method. Do you have other suggestions for how to deal with it?

stock_move_backdating/models/stock_inventory.py
toita86: Thank you for the review, I implemented almost all the changes you suggested.

stock_move_backdating/models/stock_inventory_line.py
toita86: Done

francesco-ooops:
@newtratip could you take a look? thanks!

HekkiMelody (approved):
Code review, LGTM

francesco-ooops:
@rousseldenis ready for your review, thanks!

francesco-ooops:
@rousseldenis good for merge?

rousseldenis (approved):

PR #1496, opened 2024-02-05
[16.0][ADD] stock_picking_partner_note

santostelmo:
Allow to record a message for the person in charge of order preparation at the level of the customer, 
then have it as a note on the picking transfer.

Example: Max pallet height 1m20

Note: This will allow to remove from shopfloor the shopfloor.packing.info and replace it by this more generic module.


stock_picking_partner_note/i18n/fr.po
```diff
@@ -0,0 +1,54 @@
...
+msgstr "Nom"
+
+#. module: stock_picking_partner_note
+#: model:ir.model.fields,field_description:stock_picking_partner_note.field_stock_picking_note__note_type_id
+msgid "Note Type"
+msgstr "Type de note"
+
+#. module: stock_picking_partner_note
+#: model:ir.ui.menu,name:stock_picking_partner_note.stock_picking_note_types_menu_menu
+msgid "Type de note"
```
mmequignon: is this the english string?

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    note = fields.Html(compute="_compute_note", store=True)
+
+    @api.depends("partner_id")
```
mmequignon, line 13:
```suggestion
    @api.depends("picking_type_id.partner_note_type_ids", "partner_id.stock_picking_note_ids.name/description")
```
We're missing a few depends here

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    note = fields.Html(compute="_compute_note", store=True)
```
mmequignon, line 11: as this is not a "regular" note, I would've named it "preparation_note" or smth like that.

stock_picking_partner_note/models/stock_picking_note.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+from odoo import fields, models
+
+
+class StockPickingNote(models.Model):
+    _name = "stock.picking.note"
+    _order = "name"
+
+    name = fields.Char(required=True)
```
mmequignon:
```suggestion
    name = fields.Char(required=True)
    description = fields.Text(required=True)
```
`fields.Char` isn't multi line

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+                "picking_type_id.partner_note_type_ids"
+            )
+            partner_picking_note_ids = picking.mapped(
+                "partner_id.stock_picking_note_ids"
+            )
+            picking_notes = []
+
+            for note in partner_picking_note_ids:
+                if note.note_type_id in pickig_type_note_type_ids:
+                    picking_notes += [note.name]
```
mmequignon:
```suggestion
                    picking_notes += [note.description]
```
See comment below

stock_picking_partner_note/views/stock_picking.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record model='ir.ui.view' id="view_picking_form_inherit">
+        <field name="name">Stock Picking Form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <field name="show_reserved" position="after">
+                <field name="partner_note_type_ids" widget="many2many_tags" />
```
mmequignon:
```suggestion
                <field name="note" />
```
I suppose we want to display the html here

stock_picking_partner_note/i18n/fr.po
santostelmo: it should be in EN. Changing...

stock_picking_partner_note/models/stock_picking.py
santostelmo: We don't want to update note when `picking_type_id.partner_note_type_ids` change

stock_picking_partner_note/models/stock_picking_note.py
santostelmo: The  `name` field should be of type Text indeed.
But we don't need description.

stock_picking_partner_note/views/stock_picking.xml
santostelmo: Actually we don't need this views. I'll drop it. `note` is a standard field.

mmequignon: Do we want the customer to see the note on the report?
I thought is was a field for the WH employees.

stock_picking_partner_note/models/stock_picking.py
mmequignon: Then why do we rely on it in the compute method?
Also, I would depend on the state, as when picking is "done", we do not want to update the note anymore.

stock_picking_partner_note/views/stock_picking.xml
santostelmo: If the customer has access to the delivery documents then he can see the note.

santostelmo: I just tested: 
On Picking Operations (for logistics)  we print the note 
On Delivery Note (for client) we don't print the note

stock_picking_partner_note/i18n/fr.po
santostelmo: Changed

stock_picking_partner_note/models/stock_picking.py
santostelmo: The content of the note by default is taken from partner and recomputed if and only if partner change.

jbaudoux: For me it is fine as it is...

stock_picking_partner_note/models/stock_picking.py
jbaudoux: The purpose is to fill the note of the picking according to picking type and partner. We don't want another note field. Note field is used in report, edi, shopfloor and serves to provide notes to the operator

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    note = fields.Html(compute="_compute_note", store=True)
+
+    @api.depends("partner_id")
+    def _compute_note(self):
+        for picking in self:
+            pickig_type_note_type_ids = picking.mapped(
```
jbaudoux: no need to mapped

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+                "picking_type_id.partner_note_type_ids"
+            )
+            partner_picking_note_ids = picking.mapped(
```
jbaudoux: no need for mapped

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+                "partner_id.stock_picking_note_ids"
+            )
+            picking_notes = []
```
jbaudoux: Maybe enough like that:
```suggestion
            picking_notes = picking.partner_id.stock_picking_note_ids \
            .filtered(lambda n: n.note_type_id in picking_type_note_type_ids) \
            .mapped("name")

```

stock_picking_partner_note/models/stock_picking.py
jbaudoux: typo
```suggestion
            picking_type_note_type_ids = picking.mapped(
```

stock_picking_partner_note/views/res_partner.xml
```diff
@@ -0,0 +1,30 @@
...
+                        name="stock_picking_note_ids"
+                        widget="one2many_list"
+                        nolabel="1"
+                        colspan="2"
+                        mode="tree"
+                    >
+                        <tree>
+                            <field name="name" />
+                            <field name="note_type_id" />
+                            <field name="active" />
```
jbaudoux: I would not allow to set inactive a note from the partner as it can be affected on multiple partners
```suggestion
                            <field name="active" readonly="1" />
```

stock_picking_partner_note/views/stock_picking_note.xml
```diff
@@ -0,0 +1,16 @@
...
+    <record id="view_stock_picking_note_form" model="ir.ui.view">
+        <field name="name">stock.picking.note.form</field>
+        <field name="model">stock.picking.note</field>
+        <field name="arch" type="xml">
+            <form string="Picking Note">
+                <group>
+                    <field name="name" />
+                    <field name="note_type_id" />
+                    <field name="active" />
+                </group>
```
jbaudoux, line 12: Would be useful to see the list of partners having this note

stock_picking_partner_note/models/stock_picking_note.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import fields, models
+
+
+class StockPickingNote(models.Model):
+    _name = "stock.picking.note"
+    _order = "name"
```
jbaudoux: Make sequence related to note_type_id.sequence

```suggestion
    _order = "sequence,name"
```

jbaudoux (changes requested):
LG. Just a few small remarks

stock_picking_partner_note/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Stock Picking Partner Note",
+    "version": "16.0.1.0.1",
```
simahawk:
```suggestion
    "version": "16.0.1.0.0",
```

stock_picking_partner_note/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "development_status": "Beta",
+    "category": "Product",
+    "summary": "Add partner notes on picking",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "license": "AGPL-3",
+    "depends": ["stock", "sale_stock"],
```
simahawk:
```suggestion
    "depends": ["sale_stock"],
```

stock_picking_partner_note/models/stock_picking_note_type.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import fields, models
+
+
+class StockPickingNoteType(models.Model):
+    _name = "stock.picking.note.type"
```
simahawk, line 9:
```suggestion
    _name = "stock.picking.note.type"
    _description = ....
```

stock_picking_partner_note/models/stock_picking_note.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Camptocamp (<https://www.camptocamp.com>).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import fields, models
+
+
+class StockPickingNote(models.Model):
+    _name = "stock.picking.note"
```
simahawk, line 9:
```suggestion
    _name = "stock.picking.note"
    _description = ...
```

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -0,0 +1,27 @@
...
+
+            for note in partner_picking_note_ids:
+                if note.note_type_id in pickig_type_note_type_ids:
+                    picking_notes += [note.name]
+            picking.note = "<br>".join(picking_notes)
```
simahawk:
```suggestion
            picking.note = "<br />".join(picking_notes)
```

stock_picking_partner_note/models/stock_picking.py
simahawk: that's a list, you should append rather than combining lists and check if you have a real value
```suggestion
                    if note.description and note.description.strip():
                        picking_notes.append(note.description)
```

stock_picking_partner_note/models/stock_picking_type.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    partner_note_type_ids = fields.Many2many("stock.picking.note.type")
```
simahawk: help msg pls

stock_picking_partner_note/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+Allow to record a message for the person in charge of order preparation at the level of the customer, 
+then have it as a note on the picking transfer.
```
simahawk, line 2: a `CONFIGURE.rst` is needed too

stock_picking_partner_note/views/stock_picking_note.xml
simahawk: ok, but use a smart btn rather than a direct listing of m2m

simahawk (changes requested):

stock_picking_partner_note/models/stock_picking.py
santostelmo: okay for that. but using `note.name`. name is mandatory field

stock_picking_partner_note/views/stock_picking_note.xml
jbaudoux: nooooo :fearful: 
Keop it simple, just list them, there is nothing on that view. Easier to visualize and change

santostelmo: A smart button wouldn't work because `stock_picking_note_ids` because is a M2M and not O2M.
With a button you would be able to create a new note BUT detached from the partner.
I keep the simple version showing the notes directly in the form

santostelmo:
@jbaudoux @simahawk I updated the code . Can you please review again?

stock_picking_partner_note/models/stock_picking.py
mmequignon: yeah yeah, the description was just kind of an idea.
It would be easier to browse a tree view with explicit names, rather than block of content.

jbaudoux (approved):

stock_picking_partner_note/views/stock_picking_note.xml
simahawk: > A smart button wouldn't work because stock_picking_note_ids because is a M2M and not O2M.
> With a button you would be able to create a new note BUT detached from the partner.

This is wrong. You will have an action to handle the btn and you can search for the records you need. No matter if o2m or m2m. 
Regarding creation, that's wrong too: you simply have to set the default partner in the context.

> I keep the simple version showing the notes directly in the form

Ok, but that's about time and not about possibilities :wink:

stock_picking_partner_note/views/stock_picking_note_type.xml
```diff
@@ -0,0 +1,34 @@
...
+        </field>
+    </record>
+
+    <record id="view_stock_picking_note_type_form" model="ir.ui.view">
+        <field name="name">stock.picking.note.type.form</field>
+        <field name="model">stock.picking.note.type</field>
+        <field name="arch" type="xml">
+            <form string="Picking Note Type">
+                <group>
+                    <field name="sequence" />
```
simahawk:
```suggestion
```
not needed in a form

stock_picking_partner_note/views/stock_picking_partner_note_menus.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <menuitem
+        id="stock_picking_notes_menu"
```
simahawk:
```suggestion
        id="stock_picking_notes_root"
```

stock_picking_partner_note/views/stock_picking_partner_note_menus.xml
```diff
@@ -0,0 +1,17 @@
...
+        name="Picking Notes"
+        parent="stock.menu_stock_config_settings"
+        sequence="10"
+        groups="stock.group_stock_manager"
+    />
+    <menuitem
+        id="stock_picking_note_types_menu_menu"
```
simahawk:
```suggestion
        id="stock_picking_note_types_menu"
```

stock_picking_partner_note/views/stock_picking_note_type.xml
santostelmo: removed sequence from the form

stock_picking_partner_note/views/stock_picking_partner_note_menus.xml
santostelmo: done

stock_picking_partner_note/views/stock_picking_partner_note_menus.xml
santostelmo: done

simahawk (approved):

rousseldenis (approved):
Code review. IMHO, quite tough to manage for picking users if several per picking types. Fortunately, it's not one per customer :sweat_smile:

jbaudoux:
Note: Purpose is to remove from shopfloor the shopfloor.packing.info and replace it by this more generic module.

PR #1497, opened 2024-02-06
[16.0][ADD] stock_picking_to_batch_group_fields

EmilioPascual:
Currently this module [stock_picking_batch_extended](https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_picking_batch_extended) includes this functionality but to make it work it is necessary to activate the OCA batches, which forces to use other behaviours than Odoo's for the picking batches.

This module extends the functionality of the picking batchs and allows you to manually group picking by any field in the model.

@Shide @rafaelbn @fcvalgar @yajo please review, if you want.

@moduon MT-5022


fcvalgar (approved):
Functional review, 

LGTM, Thank you @EmilioPascual 

image

image

image

image

image

stock_picking_to_batch_group_fields/wizard/stock_picking_to_batch.py
```diff
@@ -0,0 +1,132 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class StockPickingToBatch(models.TransientModel):
+    """Create a stock.picking.batch from stock.picking"""
+
+    _inherit = "stock.picking.to.batch"
+    _group_field_param = "stock_batch_picking.group_field"
```
Shide: **suggestion (non-blocking):** Defining the param here is unnecessary

stock_picking_to_batch_group_fields/wizard/stock_picking_to_batch.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+
+    def load_store_fields(self):
+        group_field_ids = (
+            self.env["ir.config_parameter"].sudo().get_param(self._group_field_param)
+        )
+        group_fields = self.env["ir.model.fields"].browse(
+            group_field_ids and json.loads(group_field_ids)
+        )
+        return group_fields
```
Shide: **suggestion(non-blocking):** Store the parameters as strings. Eg.: ["user_id", "company_id"]

```suggestion
        group_field_names = (
            self.env["ir.config_parameter"].sudo().get_param("stock_batch_picking.group_field", None)
        )
        if not group_field_names:
            return self.env["ir.model.fields"].browse()
        
        return self.env["ir.model.fields"].search([
            ("model", "=", "stock.picking"), 
            ("name", "in", json.loads(group_fields)),
        ])
```

Shide (approved):
Nice job! 

https://www.loom.com/share/ab6cedb495ee4856ae38b7e446c097d6?sid=135bd39f-95cf-4c62-965c-dd0ab98a6cb6

stock_picking_to_batch_group_fields/wizard/stock_picking_to_batch.py
EmilioPascual: Yes, better names than ids

rafaelbn (approved):
I have been testing in staging and it work 👍🏼 

There is a UX point that we will improve later

rafaelbn (changes requested):
@EmilioPascual please, we should fix this

imagen

EmilioPascual:
> @EmilioPascual please, we should fix this
> 
> imagen

@rafaelbn I think you have reviewed other module. This image is an internal module from moduon to make **waves**, but this module is to make **batches**

rafaelbn (approved):
/ocabot merge patch

PR #1498, opened 2024-02-06
[16.0][IMP] stock_picking_group_by_partner_by_carrier: Control the original_group_id propagation.

lmignon:


lmignon:
cc @rousseldenis @jbaudoux

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -27,6 +27,12 @@ def _prepare_merge_moves_distinct_fields(self):
         # merged.
         return super()._prepare_merge_moves_distinct_fields() + ["original_group_id"]
 
+    def _keep_original_group(self):
+        """Keep the original group on the move when creating a joint group."""
+        for move in self:
+            if move.group_id and not move.original_group_id:
```
rousseldenis: Nothing more about printed pickings and state ?

rousseldenis (approved):
Code review

stock_picking_group_by_partner_by_carrier/models/stock_rule.py
```diff
@@ -2,12 +2,26 @@
...
+        help="Propagate the original group on the created moves. This allows to "
+        "prevent merging pulled moves. This allows to cancel a SO without "
+        "canceling pulled moves from other SO (as we ensure they are not "
+        "merged). If your warehouse is configured to make the delivery in two "
+        "steps, you should not propagate the original group on the picking if your"
+        "pick picking type is also configured to group pickings. In this way, "
+        "the generated moves for the same product but for different sales orders "
+        "will be merged into one unique move. If you propagate the original group "
+        "on the picking, the moves will not be merged and you will have one move "
+        "per sale order.",
```
jbaudoux: I would not talk about configuring merging on non outgoing moves as this should not be possible. The option is only visible on outgoing picking types. We should add a constrain to make it more explicit.
```suggestion
        "merged). You likely need this option if you don't propagate the merged procurement group to the pulled moves.",
```

stock_picking_group_by_partner_by_carrier/tests/test_grouping.py
```diff
@@ -267,6 +277,73 @@ def test_delivery_multi_step_group_pick(self):
         self.assertEqual(len(picks), 1)
         self.assertFalse(so1.picking_ids - so2.picking_ids)
 
+    def test_delivery_multi_step_group_out_pick_op_merged(self):
```
jbaudoux: This should not be possible

jbaudoux (changes requested):
This overall looks good to me

lmignon:
Back to draft. Compatibility issue with the stock_available_to_promise_release.

stock_picking_group_by_partner_by_carrier/tests/test_grouping.py
jbaudoux: I would drop this test that does not make sense as you should not be able to enable grouping on non outgoing picking type

jbaudoux:
@lmignon Can you drop changes to generated files?

lmignon:
@jbaudoux tests removed and rebase done

PR #1500, opened 2024-02-07
[15.0][ADD] stock_force_assign_by_type: New Module

pilarvargas-tecnativa:
cc @Tecnativa TT47597

@pedrobaeza @chienandalu please review


stock_force_assign_by_type/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
...
+
+    picking_type_force_reservation = fields.Boolean(
+        related="picking_type_id.force_reservation"
+    )
+
+
+class PickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    force_reservation = fields.Boolean()
```
chienandalu: Add a help description to the field

stock_force_assign_by_type/tests/test_stock_force_assign_by_type.py
```diff
@@ -0,0 +1,80 @@
...
+    def _create_purchase_order(self):
+        purchase_form = Form(self.env["purchase.order"])
+        purchase_form.partner_id = self.partner_supplier
+        with purchase_form.order_line.new() as purchase_line_form:
+            purchase_line_form.product_id = self.product1
+            purchase_line_form.product_qty = 10
+        purchase = purchase_form.save()
+        return purchase
+
+    def test_01_stock_force_assign_by_type(self):
```
chienandalu, line 35: It's interesting to check the regular behavior as well

stock_force_assign_by_type/views/stock_picking_views.xml
```diff
@@ -0,0 +1,31 @@
...
+            <button name="do_unreserve" position="attributes">
+                <attribute
+                    name="attrs"
+                >{'invisible': ['|', '|', '|', '&amp;', ('picking_type_force_reservation', '=', False), ('picking_type_code', '=', 'incoming'), ('immediate_transfer', '=', True), '&amp;', ('state', '!=', 'assigned'), ('move_type', '!=', 'one'), '&amp;', ('state', 'not in', ('assigned', 'confirmed')), ('move_type', '=', 'one')]}</attribute>
+            </button>
+            <field name="forecast_availability" position="attributes">
+                <attribute
+                    name="attrs"
+                >{'column_invisible': ['|', '|', ('parent.state', 'in', ['draft', 'done']), '&amp;', ('parent.picking_type_force_reservation', '=', False), ('parent.picking_type_code', '!=', 'outgoing'), ('parent.immediate_transfer', '=', True)]}</attribute>
+            </field>
```
chienandalu, line 23: Maybe it would be better to use: https://github.com/OCA/server-tools/tree/15.0/base_view_inheritance_extension

What do you think @pedrobaeza ?

pedrobaeza: AFAIK, the module doesn't support `attrs`. It's been partially added in https://github.com/OCA/server-tools/pull/2824 if I read it well.

chienandalu: Oh, ok. Let's put a good warning comment then

pilarvargas-tecnativa: done!

stock_force_assign_by_type/views/stock_picking_views.xml
```diff
@@ -0,0 +1,35 @@
...
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <field name="use_create_lots" position="after">
+                <field name="picking_type_force_reservation" />
+            </field>
+            <!-- WARNING: The 'attrs' attributes of the 'do_unreserve' button and the
+                'forecast_availability' field are overridden in this view. This change
+                may impact visibility and functionality in other modules. Ensure thorough
+                review of migrations and dependencies. -->
+            <button name="do_unreserve" position="attributes">
```
rousseldenis, line 14: @pilarvargas-tecnativa Why this action should be obfuscated depending on force_reservation parameter ? IMHO the _should_bypass_reservation() override is sufficent.

pilarvargas-tecnativa: Depending on the type of operation in which the stock reservation is being forced, the button that performs the unreserve action is not available. For example, in a purchase order in which the type of picking establishes a forced stock reservation, when we access the picking, this action is not available (domain: ('parent.picking_type_code', '!=', 'outgoing')) nor do we have visibility of the reserved units (forecast_availability). I think it is necessary to have this option available in any operation where a forced stock reservation is made.

chienandalu: Hi @rousseldenis the background context for this module is that we need to pay for something that goes out of our warehouse (the provider charges waste management fees), which is something unsupported by Odoo. So we need:

- To do a purchase order from which we can generate a vendor invoice.
- Only an incoming picking will complete the delivered (received) quantities in the PO.
- We need to reserve the right lots from our stock automatically as in a delivery picking (this is what this module allows us to do).
- We need to be able to re-assign/ unreserve as in a regular delivery. This is what this domain overrides allows us to do.

stock_force_assign_by_type/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _should_bypass_reservation(self, forced_location=False):
+        res = super()._should_bypass_reservation()
+        if self.picking_type_id.force_reservation:
+            return False
+        return res
```
pedrobaeza: No need of calling super in the case of forcing.

```suggestion
        if self.picking_type_id.force_reservation:
            return False
        return super()._should_bypass_reservation()
```

stock_force_assign_by_type/readme/CONTEXT.rst
```diff
@@ -0,0 +1,7 @@
+* Forced stock reservation: Allows you to configure picking operation types so that stock
```
pedrobaeza: This file is for putting the reasons to develop this module.

stock_force_assign_by_type/readme/USAGE.rst
```diff
@@ -0,0 +1,13 @@
+Picking operation type configuration:
```
pedrobaeza: This first section should go to CONFIGURE.rst.

pedrobaeza (changes requested):

stock_force_assign_by_type/readme/CONTEXT.rst
chienandalu: @pilarvargas-tecnativa something along these lines: https://github.com/OCA/stock-logistics-workflow/pull/1500#discussion_r1482610850

pilarvargas-tecnativa: thanks

chienandalu (approved):

stock_force_assign_by_type/models/stock_picking.py
```diff
@@ -0,0 +1,21 @@
...
+class Picking(models.Model):
+    _inherit = "stock.picking"
+
+    picking_type_force_reservation = fields.Boolean(
+        related="picking_type_id.force_reservation"
+    )
+
+
+class PickingType(models.Model):
+    _inherit = "stock.picking.type"
```
pedrobaeza: This is something that should go in its own file, as it's not a one2many model. And please name the classes with their complete name (`Picking` > `StockPicking`, `PickingType` > `StockPickingType`). If Odoo doesn't do this according guidelines doesn't mean that we shouldn't follow them.

stock_force_assign_by_type/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,10 @@
+This module provides additional functionality for stock allocation management in Odoo,
+allowing you to force stock reservation based on the type of picking operation:
+
+* Forced stock reservation: Allows you to configure picking operation types so that stock
```
pedrobaeza: I don't get the sense of these 3 points. You only enable a check, not 3 configuration options, and this, being a user oriented general description, shouldn't include these details. You may put them on USAGE.rst, but better redacted to indicate the "effects" of marking the check.

stock_force_assign_by_type/models/stock_picking.py
pilarvargas-tecnativa: done! thanks

pedrobaeza (approved):
/ocabot merge nobump

PR #1501, opened 2024-02-07
[FIX] [16.0] stock_move_line_expiration_date_required: Remove required from field definition

Shide:
Using the required on the field definition causes the database to set "not nullable" on expiry_date column on stock.move.line. 

Moved the required to the views.

MT-4224 @moduon @yajo @EmilioPascual @rafaelbn please review


stock_move_line_expiration_date_required/models/stock_move_line.py
```diff
@@ -9,11 +9,6 @@
 class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
 
-    expiration_date = fields.Datetime(
-        required="[('parent.use_expiration_date', '=', True),"
-        " ('parent.picking_code', '=', 'incoming')]"
```
yajo, line 14: You'll need a migration script to remove the constraint that was wrongly created by this.

yajo (changes requested):

stock_move_line_expiration_date_required/models/stock_move_line.py
Shide: Done. I've re-launched the `update_db` method on the `expiration_date` field.

Shide:
I think the ocabot merge will need to increment the verison of the module

stock_move_line_expiration_date_required/migrations/16.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2024 Moduon Team S.L. <info@moduon.team>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
+from odoo import SUPERUSER_ID, api, tools
+
+
+def migrate(cr, version):
```
rousseldenis: @Shide The current version of this module is 16.0.1.0.1 yet. So, you need both to change this directory name and manifest version

fcvalgar (approved):
Functional review.

LGTM, thank you @Shide.

https://www.loom.com/share/ec7071afa48644b69199d3d7bb8a537a?sid=22e04f0f-1975-4514-9126-db27f4847bac

stock_move_line_expiration_date_required/migrations/16.0.1.0.1/post-migration.py
Shide: Done, thanks! :)

rafaelbn (approved):
Functionaly 👍🏼 

I think we should wait for @yajo or @rousseldenis approval to merge, technical review, thanks

stock_move_line_expiration_date_required/__manifest__.py
```diff
@@ -3,12 +3,15 @@
 {
     "name": "Stock Move Line Expiration Date Required",
     "Summary": "Expiration date is required to enter manually on Move Lines.",
-    "version": "16.0.1.0.1",
+    "version": "16.0.1.0.2",
```
yajo, line 6: **suggestion(non-blocking):** usually, you shouldn't use migration scripts in patch versions. That should indicate that the version bump should be major.

According to [our guidelines](https://github.com/OCA/odoo-community.org/blob/8afdb5a5ebfd3314e2fc836c9b9f644468ee044e/website/Contribution/CONTRIBUTING.rst#L71-L72), it'd be a major or minor bump.

Future-proofing the code, [it's better to restrict migration scripts to major or minor versions too](https://github.com/orgs/OCA/discussions/117#discussion-5699815).

However, this one PR is a little bit crossing the line because it's really a bugfix. Thus, this is not required at this time IMHO. But at least there's some extra info for you to keep in mind for next times :)

yajo (approved):

PR #1226, opened 2023-02-15
[14.0][ADD] stock_picking_group_by_date_deadline: split picking per deadline date

petrus-v:
Not sure if it's in the proper repository nor proper module name !

I got the requirements to create one picking per day while using stock_request module with this PR: https://github.com/OCA/stock-logistics-warehouse/pull/1637

PR #1508, opened 2024-02-19
[ADD] stock_move_planned_consumed_date

petrus-v:
Base module to avoid mixing move line that is planned to consumed in different dates


rousseldenis:
@petrus-v Have you tried with date_deadline field ?

rousseldenis:
> @petrus-v Have you tried with date_deadline field ?

Which is propagated at sale order confirmation from the commitment_date on the order level

petrus-v:
> @petrus-v Have you tried with date_deadline field ?

I know about the date_deadline which propagate on stock_picking, but if you check the availability of those picking a bit earlier nothing prevent to reserved products that will be expired before the deadline (At the moment we haven't installed stock release modules mechanisms), also in complex industry with multiple warehouse where procurement group are not necessarily propagated we could get some transfer (stock.picking) with different moves where expected consumed date are different (to make this easier to reproduce we are using https://github.com/oca/stock-logistics-workflow/pull/1226 which basically round the deadline date time to date)

this is the base module for #1509 and #1511

rousseldenis:
> > @petrus-v Have you tried with date_deadline field ?
> 
> I know about the date_deadline which propagate on stock_picking, but if you check the availability of those picking a bit earlier nothing prevent to reserved products that will be expired before the deadline (At the moment we haven't installed stock release modules mechanisms), also in complex industry with multiple warehouse where procurement group are not necessarily propagated we could get some transfer (stock.picking) with different moves where expected consumed date are different (to make this easier to reproduce we are using #1226 which basically round the deadline date time to date)
> 
> this is the base module for #1509 and #1511

Ok, understand.

FYI, there is a change in v16 (and v15) merged yesterday that allows to ignore date_deadline when merging: https://github.com/odoo/odoo/commit/fdaba15c4f83f75fa2bc16c8e44d9c7093239ee3

stock_move_planned_consumed_date/readme/CONTRIBUTIRS.rst
alexandregaldeano, line 1: `CONTRIBUTIRS`? Are you sure? 😛

stock_move_planned_consumed_date/models/stock_move.py
```diff
@@ -0,0 +1,26 @@
...
+        "Planned consumed date",
+        readonly=True,
+        states={"draft": [("readonly", False)]},
+        help="This is the expected consumed/usage date of the product by the customer.",
+    )
+
+    def _prepare_procurement_values(self):
+        vals = super()._prepare_procurement_values()
+        vals["planned_consumed_date"] = self.planned_consumed_date
+        return vals
```
alexandregaldeano:
```suggestion
        return {
            **super()._prepare_procurement_values(),
            "planned_consumed_date": self.planned_consumed_date,
        }
```

WDYT?

stock_move_planned_consumed_date/models/stock_move.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @api.model
+    def _prepare_merge_moves_distinct_fields(self):
+        distinct_fields = super()._prepare_merge_moves_distinct_fields()
+        distinct_fields.append("planned_consumed_date")
+        return distinct_fields
```
alexandregaldeano:
```suggestion
        return [*super()._prepare_merge_moves_distinct_fields(), "planned_consumed_date"]
```

WDYT?

stock_move_planned_consumed_date/models/stock_rule.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    def _get_custom_move_fields(self):
+        fields = super()._get_custom_move_fields()
+        fields += ["planned_consumed_date"]
+        return fields
```
alexandregaldeano: Same here

alexandregaldeano (approved):
My comments are about code style preference, as such, you can of course discard them :)

stock_move_planned_consumed_date/models/stock_move.py
petrus-v: done

stock_move_planned_consumed_date/models/stock_move.py
petrus-v: done

stock_move_planned_consumed_date/models/stock_rule.py
petrus-v: yeah !! thanks

stock_move_planned_consumed_date/views/stock_picking_views.xml
```diff
@@ -0,0 +1,25 @@
...
+                expr="//page[@name='operations']/field[@name='move_ids_without_package']/tree/field[@name='reserved_availability']"
+                position="after"
+            >
+                <field name="planned_consumed_date" optional="hide" />
+            </xpath>
+            <xpath
+                expr="//page[@name='operations']/field[@name='move_ids_without_package']/form//field[@name='reserved_availability']"
+                position="after"
+            >
+                <field name="planned_consumed_date" optional="hide" />
```
petrus-v:
```suggestion
                <field name="planned_consumed_date" />
```

rousseldenis: @petrus-v I would say it should be optional="show" instead as tree view can be very overcharged.

PR #1512, opened 2024-02-20
[16.0] partner_delivery_windows: Add helper function to find next delivery w…

grindtildeath:
…indows

### DIFF
--- a/stock_partner_delivery_window/models/res_partner.py
+++ b/stock_partner_delivery_window/models/res_partner.py
@@ -7,6 +7,7 @@ from datetime import time
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools.misc import format_time
+from odoo.tools.date_utils import date_range
 
 from odoo.addons.partner_tz.tools import tz_utils
 
@@ -180,3 +181,76 @@ class ResPartner(models.Model):
         ]
         result["delivery_time_window_ids"] = [(0, 0, val) for val in values]
         return [result]
+
+    def next_delivery_window_start_datetime(self, from_date=None, timedelta_days=None):
+        """Get next starting datetime in a preferred delivery window.
+
+        If from_date is already in a delivery window, from_date is "the next"
+
+        :param from_date: Datetime object (Leave empty to use now())
+        :param timedelta_days: Number of days to use in the computation
+                               (Leave empty to use 7 days or 1 week)
+        :return: Datetime object
+        """
+        self.ensure_one()
+        if not from_date:
+            from_date = datetime.now()
+        if self.is_in_delivery_window(from_date):
+            return from_date
+        if timedelta_days is None:
+            timedelta_days = 7
+        if self.delivery_time_preference == "workdays":
+            datetime_windows = self.get_next_workdays_datetime(
+                from_date, from_date + timedelta(days=timedelta_days)
+            )
+        else:
+            datetime_windows = self.get_next_windows_start_datetime(
+                from_date, from_date + timedelta(days=timedelta_days)
+            )
+        for dwin_start in datetime_windows:
+            if dwin_start >= from_date:
+                return dwin_start
+        raise UserError(
+            _("Something went wrong trying to find next delivery window. Date: %s")
+            % str(from_date)
+        )
+
+    def get_next_workdays_datetime(self, from_datetime, to_datetime):
+        """Returns all the delivery windows in the provided date range.
+
+        :param from_datetime: Datetime object
+        :param to_datetime: Datetime object
+        :return: List of Datetime objects
+        """
+        dates = date_range(from_datetime, to_datetime, timedelta(days=1))
+        return [date for date in dates if date.weekday() < 5]
+
+    def get_next_windows_start_datetime(self, from_datetime, to_datetime):
+        """Get all delivery windows start time.
+
+        Range from from_datetime weekday to to_datetime weekday.
+
+        Note result can include a start datetime that is before from_datetime
+        on the from_datetime weekday
+
+        :param from_datetime: Datetime object
+        :param to_datetime: Datetime object
+        :return: List of Datetime objects
+        """
+        res = list()
+        for this_datetime in date_range(from_datetime, to_datetime, timedelta(days=1)):
+            this_weekday_number = this_datetime.weekday()
+            this_weekday = self.env["time.weekday"].search(
+                [("name", "=", this_weekday_number)], limit=1
+            )
+            # Sort by start time to ensure the window we'll find will be the first
+            # one for the weekday
+            this_weekday_windows = self.delivery_time_window_ids.filtered(
+                lambda w: this_weekday in w.time_window_weekday_ids
+            ).sorted("time_window_start")
+            for win in this_weekday_windows:
+                this_weekday_start_datetime = datetime.combine(
+                    this_datetime, win.get_time_window_start_time()
+                )
+                res.append(this_weekday_start_datetime)
+        return res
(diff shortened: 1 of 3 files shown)


stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,76 @@ def copy_data(self, default=None):
...
+        If from_date is already in a delivery window, from_date is "the next"
+
+        :param from_date: Datetime object (Leave empty to use now())
+        :param timedelta_days: Number of days to use in the computation
+                               (Leave empty to use 7 days or 1 week)
+        :return: Datetime object
+        """
+        self.ensure_one()
+        if not from_date:
+            from_date = datetime.now()
```
simahawk:
```suggestion
        from_date = from_date or datetime.now()
```

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,76 @@ def copy_data(self, default=None):
...
+        if self.is_in_delivery_window(from_date):
+            return fields.Datetime.to_datetime(from_date)
+        if timedelta_days is None:
+            timedelta_days = 7
```
simahawk, line 204:
```suggestion
        timedelta_days = timedelta_days or 7
```
unless you support `0`.

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,76 @@ def copy_data(self, default=None):
...
+        :param from_datetime: Datetime object
+        :param to_datetime: Datetime object
+        :return: List of Datetime objects
+        """
+        res = list()
+        for this_datetime in date_range(from_datetime, to_datetime, timedelta(days=1)):
+            this_weekday_number = this_datetime.weekday()
+            this_weekday = self.env["time.weekday"].search(
+                [("name", "=", this_weekday_number)], limit=1
+            )
```
simahawk: When the date_range is big you'll have many many searches.
I'd recommend doing the following:

```suggestion
        dts = date_range(from_datetime, to_datetime, timedelta(days=1))
        weekdays = {x.name: x for x in self.env["time.weekday"].search([("name", "in", dts)])}
        for this_datetime, this_weekday in weekdays.items():
            this_weekday_number = this_datetime.weekday()
```

simahawk (commented):
any test?

stock_partner_delivery_window/models/res_partner.py
grindtildeath: IMO it's usually safer to consider 0 when working on dates/times and deltas

grindtildeath:
@mmequignon since you're the author of the main commit here, did you write any tests?

mmequignon:
> @mmequignon since you're the author of the main commit here, did you write any tests?

be my guest https://github.com/OCA/sale-workflow/blob/4fd856f263a696f16814834e2ffd0e89bc230308/sale_delivery_date/tests/test_sale_partner_delivery_window.py

jbaudoux (approved):
LG

jbaudoux:
@grindtildeath pre-commit to fix

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,114 @@ def copy_data(self, default=None):
...
+
+        :param from_datetime: Datetime object
+        :param to_datetime: Datetime object
+        """
+        self.ensure_one()
+        if to_datetime is None:
+            to_datetime = from_datetime + timedelta(days=365)
+        for this_datetime in date_range(from_datetime, to_datetime, timedelta(days=1)):
+            if self.delivery_time_preference == "anytime":
+                yield this_datetime
```
jbaudoux: I haven't look your usage yet but I think your generator should allow a new from_datetime on each iteration.
Then you can ask multiple modules the next availability and send them the new start date to validate 
Something like:
```
this_datetime = from_datetime
while True:
  if self.delivery_time_preference == "anytime":
    from_datetime = (yield this_datetime)
  elif self.delivery_time_preference == "workdays" and this_weekday_number < 5:
    from_datetime = (yield this_datetime)
  else:
    ...

  if from_datetime is not None:
    this_datetime = from_datetime
  else:
    this_datetime += timedelta(days=1)
```

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,114 @@ def copy_data(self, default=None):
...
+            else:
+                this_weekday_number = this_datetime.weekday()
+                if self.delivery_time_preference == "workdays":
+                    if this_weekday_number < 5:
```
jbaudoux, line 286: else missing. You likely need to combine with above test

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,114 @@ def copy_data(self, default=None):
...
+                        yield this_datetime
+                else:
+                    this_weekday_windows = self.env[
+                        "partner.delivery.time.window"
+                    ].search(
```
jbaudoux, line 291: you should not call search on each iteration, use filtered

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,114 @@ def copy_data(self, default=None):
...
+                        [
+                            ("time_window_weekday_ids.name", "=", this_weekday_number),
+                            ("partner_id", "=", self.id),
+                        ],
+                        order="time_window_start ASC",
+                    )
+                    for win in this_weekday_windows:
```
jbaudoux, line 298: You should not loop on the time slots. If there is a slot, the day is valid. Next iteration should be on next day.

For now, we can keep time management as a nice to have. Ideally you should check that the time in this_datetime is before the end time of the slot and return the combine of the max(window_stat_time, this_datetime.time). In this case, you also need to reset time to 0 on next iteration (this_datetime += timedelta(days=1))

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,114 @@ def copy_data(self, default=None):
...
+            ).sorted("time_window_start")
+            for win in this_weekday_windows:
+                this_weekday_start_datetime = datetime.combine(
+                    this_datetime, win.get_time_window_start_time()
+                )
+                res.append(this_weekday_start_datetime)
+        return res
+
+    # TODO: Refactor with function above
+    def get_next_delivery_availability(self, from_datetime, to_datetime=None):
```
jbaudoux, line 262: I would have dropped the `to_datetime`. You can always manage this outside the generator with `myiter.close()`

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -180,3 +183,114 @@ def copy_data(self, default=None):
         ]
         result["delivery_time_window_ids"] = [(0, 0, val) for val in values]
         return [result]
+
+    def next_delivery_window_start_datetime(self, from_date=None, timedelta_days=None):
```
jbaudoux, line 189: Is this still useful or replaced by `get_next_delivery_availability` ?

jbaudoux (changes requested):
I dropped a few comments

cyrilmanuel:
@grindtildeath  little ping, could you made the modification ?

stock_partner_delivery_window/models/res_partner.py
grindtildeath: IMO we can drop this one

stock_partner_delivery_window/models/res_partner.py
grindtildeath: This is implemented in last commit for delivery windows preference

jbaudoux:
Being replaced by https://github.com/OCA/wms/pull/1026

PR #1513, opened 2024-02-21
[ADD] [16.0] stock_rule_reserve_max_quantity

Shide:
This module extends the functionality of stock reservation to support reserve max available quantity and to allow you to not double check quantity done in previous steps for Outgoing Shipments when is configured with 2 or 3 steps.

This operation will only apply when the available quantity exceeds the demanded quantity and the stock rule linked to the stock move has the field `Reserve Max Quantity` activated.

https://www.loom.com/share/7e9e0852b8c54fd1b17fa4d16bc80810?sid=79159165-a88a-4c9f-befb-f5a8b5987691

MT-5085 @moduon @rafaelbn @yajo @Gelojr @rousseldenis @legalsylvain please review if you want :)


rafaelbn (commented):
Great short video ! ❤️ Watching you 😄 happy video make me review faster and simpler!

- [x] After video I have tested in runboat and we should try to improve updating the "demand" with done quantity in the same way it makes when on-step is activated. Futhermore in other case Odoo creates a new stock.move with the weight over the demand.

Thank!

stock_rule_reserve_max_quantity/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "Stock Rule Reserve Max Quantity",
+    "summary": "Allows to reserve max available quantity when a move comes from an stock rule",
+    "version": "16.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Inventory/Inventory",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["Shide"],
```
yajo: issue:

```suggestion
    "maintainers": ["Shide", "rafaelbn"],
```

stock_rule_reserve_max_quantity/models/stock_rule.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import fields, models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    reserve_max_quantity = fields.Boolean(
+        help="If checked, the system will reserve the maximum quantity "
+        "available for the moves that has this rule linked and the quantity "
```
yajo: issue(typo): 

```suggestion
        "available for the moves that have this rule linked, when the quantity "
```

stock_rule_reserve_max_quantity/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+This module extends the functionality of stock reservation to support reserve max available quantity and to allow you to not double check quantity done in previous steps for Outgoing Shipments when is configured with 2 or 3 steps.
+
+This operation will only apply when the available quantity exceeds the demanded quantity and the stock rule linked to the stock move has the field `Reserve Max Quantity` activated.
```
yajo: issue: This isn't code:

```suggestion
This operation will only apply when the available quantity exceeds the demanded quantity and the stock rule linked to the stock move has the field *Reserve Max Quantity* activated.
```

stock_rule_reserve_max_quantity/tests/test_stock_rule_reservation.py
```diff
@@ -0,0 +1,143 @@
...
+                "type": "product",
+                "categ_id": cls.env.ref("product.product_category_all").id,
+            }
+        )
+
+    def _enable_pick_ship(self, create_backorder="never"):
+        """Enable pick_ship route and set the pick rule to reserve_max_quantity"""
+        self.wh = self.env["stock.warehouse"].search(
+            [("company_id", "=", self.env.user.id)], limit=1
+        )
```
yajo: issue: avoid polluting other tests:

```suggestion
        self.wh = self.env["stock.warehouse"].search(
            [("company_id", "=", self.env.user.id)], limit=1
        )
        self.addCleanup(delattr, self, "wh")
```

Please do the same with other attributes set in this method.

OTOH, you could just do all this stuff in `setUpClass` which would be more performant. After all, all tests need it.

yajo (changes requested):
Very good one! Some minor (but blocking) issues detected.

rafaelbn (approved):
Tested in runbot and recorded! 👍🏼 

Runnning safetly 1 week in prod

rafaelbn:
@yajo @EmilioPascual please review, take care of this! thanks!

rousseldenis:
@Shide @rafaelbn @EmilioPascual I'm wondering if this could be integrated or extend this : https://github.com/OCA/stock-logistics-warehouse/pull/1794

jbaudoux (changes requested):
Doesn't look right to do such. Looks quite dangerous to me.

Please redo the example shown in the video but process the 2 picks before processing a OUT. I bet you can have the first OUT with the full quantity and the 2nd OUT without any quantity.

jbaudoux:
> @Shide @rafaelbn @EmilioPascual I'm wondering if this could be integrated or extend this : [OCA/stock-logistics-warehouse#1794](https://github.com/OCA/stock-logistics-warehouse/pull/1794)

@rousseldenis Doesn't related at all to that module. The one you mention is related to MTS reservation rules. Here, they change MTO.

Shide:
> Doesn't look right to do such. Looks quite dangerous to me.
> 
> Please redo the example shown in the video but process the 2 picks before processing a OUT. I bet you can have the first OUT with the full quantity and the 2nd OUT without any quantity.

You can create those pickings on Runbot and try by yourself. If something doesn't work as expected, I'll review it and fix it as soon as possible.

We are using this PR un production with hundreds of pickings with 2 steps, with batches and waves, different UoM and tracking by lots in most of the products.

In the video I warn to not use this setting on a first step. 
Of course it's a dangerous check if you don't know how to use. Thanks the documentation and the field help is pretty well explained and you won't do that thing. In USAGE.md point 4, I tell you to not activate on the first step. If you need a further explanation, I can add it to create a more explicit warning.

This PR has been done after a lot of hours of investigation, deep code reading and try-harding debugging each line of code in the delivery process to find the best solution with less maintenance to avoid double or triple weighting quantities in 2 and 3 steps.

I'm sorry you don't like the module or the way that I do these things, but I cannot find another solution preserving the demand quantity and avoid the double/triple weighting without modifying Odoo core.

jbaudoux (approved):
Tested and working.

It relies on the fact that the call to the method in chain of moves is already limited to the brought quantity. In fact available_quantity is the brought quantity and not what is available on the location
https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_move.py#L1713

stock_rule_reserve_max_quantity/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+This module extends the functionality of stock reservation to support reserve max available quantity and to allow you to not double check quantity done in previous steps for Outgoing Shipments when is configured with 2 or 3 steps.
```
jbaudoux:
```suggestion
This module extends the functionality of chained (pull) stock reservation to support the reservation of the total quantity brought by the done chained move lines. By default, the reservation on each step of the chain is limited to the need. If you pick more than the need, with this module the chained moves will reserve the total picked quantity.
```

rafaelbn: Thank you @jbaudoux . This has been hard work. Great work @Shide !

EmilioPascual (approved):

stock_rule_reserve_max_quantity/readme/DESCRIPTION.md
Shide: Thanks for the suggestion. 
This description fits much better with what the module does. ❤️

rafaelbn (approved):
/ocabot merge patch

PR #1523, opened 2024-02-28
[ADD] [16.0] stock_picking_invoice_frequency

Shide:
Same module as [stock_picking_batch_invoice_frequency](https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_picking_batch_invoice_frequency) but pickings doesn't need to be validated from a batch.

MT-5246 @moduon @rafaelbn @EmilioPascual please review if you want


EmilioPascual (approved):
LGTM.
Functional and code review.

rousseldenis:
@Shide @EmilioPascual IMHO this is a duplicate as https://github.com/OCA/account-invoicing/tree/16.0/partner_invoicing_mode_at_shipping

rafaelbn:
Hello @rousseldenis ,

I's quite different, technically. We analysed it deeply before in https://github.com/OCA/sale-workflow/issues/2669 the dependencies make a big diferente, thousand of line of codes of difference.

"depends": ["account", "partner_invoicing_mode", "queue_job", "stock"],
partner_invoicing_mode:    "depends": ["account", "queue_job", "sale"],

- We are using 250 lines of code with at all.
- You proposal make us to use with all "set" near 5.000 lines of code for a really simple use case. We cannot maintain that.

Thank you!
Regards,
Rafael

PR #1527, opened 2024-03-05
[16.0][MIG] stock_valuation_fifo_lot

Saran440:


stock_valuation_fifo_lot/models/product.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _sort_by_all_candidates(self, all_candidates, sort_by):
+        """Hook function for other sort by"""
+        return all_candidates
+
+    def _get_all_candidates(self, company, sort_by=None):
```
AungKoKoLin1997, line 15: In odoo 16, we have this kind of function https://github.com/odoo/odoo/blob/4d154a9f08cb2204df8566ed5c1e4f5e59e5d212/addons/stock_account/models/product.py#L331.
Should we extend and passing sort_by as a context?

rousseldenis:
@Saran440 Do you mind if we switch to #1676 ?

Saran440:
No, go ahead.

PR #1537, opened 2024-03-11
[15.0][FIX] stock_picking_auto_create_lot: Lots repeated for different stock move lines with the same product

sergio-teruel:
For one stock move if I create two sml's and validate the picking two lots are created but the two sml have the same lot (The first lot created)

cc @Tecnativa TT48308

ping @carlosdauden @rousseldenis


rousseldenis (approved):
Code review. Seems fair

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to upper versions

PR #1543, opened 2024-03-15
[14.0][FIX] stock_picking_partner_note: make note field a text field

JuMiSanAr:
@santostelmo 

I'm opening this PR after noticing that the `note` field in `stock.picking`, which was created as an HTML, created some conflicts with what shopfloor expects.

Do you think we could make it a Text field (even more considering `note.name` is a Text field too) and then apply styling through CSS if needed?

ref: cos-4507

@simahawk

### DIFF
--- a/stock_picking_partner_note/models/stock_picking.py
+++ b/stock_picking_partner_note/models/stock_picking.py
@@ -8,7 +8,7 @@ from odoo import api, fields, models
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
-    note = fields.Html(compute="_compute_note", store=True)
+    note = fields.Text(compute="_compute_note", store=True)
 
     # TODO: ideally, we should add more detailed dependencies here;
     # however, there's a CacheMiss error that occurs when adding
@@ -27,4 +27,4 @@ class StockPicking(models.Model):
                 for note in picking_notes
                 if note.name and note.name.strip()
             ]
-            picking.note = "<br />".join(picking_notes)
+            picking.note = "\n".join(picking_notes)
--- a/stock_picking_partner_note/models/stock_picking.py
+++ b/stock_picking_partner_note/models/stock_picking.py
@@ -10,11 +10,6 @@ class StockPicking(models.Model):
 
     note = fields.Text(compute="_compute_note", store=True)
 
-    # TODO: ideally, we should add more detailed dependencies here;
-    # however, there's a CacheMiss error that occurs when adding
-    # "picking_type_id.partner_note_type_ids" and "partner_id.stock_picking_note_ids"
-    # so we're leaving it like this for now
-    # (the partner should be enough to trigger the compute).
     @api.depends("partner_id")
     def _compute_note(self):
         for picking in self:
--- a/stock_picking_partner_note/tests/test_stock_picking_partner_note.py
+++ b/stock_picking_partner_note/tests/test_stock_picking_partner_note.py
@@ -41,4 +41,4 @@ class StockPickingPartnerNote(common.SavepointCase):
             (6, 0, (self.note_type1 | self.note_type2).ids)
         ]
         self.order.action_confirm()
-        self.assertIn("<p>Note 1<br>Note 2</p>", self.order.picking_ids[0].note)
+        self.assertIn("Note 1\nNote 2", self.order.picking_ids[0].note)


santostelmo:
@JuMiSanAr do you have any screenshot of the final result in form ?
It's sounds okay for me

santostelmo:
I did some testing. With this change notes are showing in the form like that which is not a good presentation:
Partner Picking Notes
image

On stock picking aggregated notes are showing like that:
image

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -8,14 +8,11 @@
...
+    note = fields.Text(compute="_compute_note", store=True)
 
-    # TODO: ideally, we should add more detailed dependencies here;
-    # however, there's a CacheMiss error that occurs when adding
-    # "picking_type_id.partner_note_type_ids" and "partner_id.stock_picking_note_ids"
-    # so we're leaving it like this for now
-    # (the partner should be enough to trigger the compute).
-    @api.depends("partner_id")
+    @api.depends(
+        "partner_id.stock_picking_note_ids", "picking_type_id.partner_note_type_ids"
```
santostelmo: That will trigger a change in all picking notes if a single note is changed in the partner. We can't do that

santostelmo (changes requested):

JuMiSanAr:
Closing this PR as the shopfloor issue is resolved in customer proj

simahawk:
> I did some testing. With this change notes are showing in the form like that which is not a good presentation: Partner Picking Notes image
> 
> On stock picking aggregated notes are showing like that: image

@santostelmo if it's about that display we should simply add a computed html field. Any issue w/ that?

simahawk:
Another question: how does it look like on the delivery note report?

simahawk:
@JuMiSanAr as we discussed we can:
* add `note_html` field to be displayed only in view mode
* `note` will be displayed only in edit mode

This will keep the standard behavior *intact* everywhere, w/o side effects.

santostelmo:
> > I did some testing. With this change notes are showing in the form like that which is not a good presentation: Partner Picking Notes image
> > On stock picking aggregated notes are showing like that: image
> 
> @santostelmo if it's about that display we should simply add a computed html field. Any issue w/ that?

The html field is not only for display. It can also be edited by the user to customize notes

JuMiSanAr:
@simahawk so can we still follow the approach with two different computed fields? I don't see how we can keep the original behavior "intact" if the idea of this module is to give the user freedom to customize the notes as they please.

simahawk:
> The html field is not only for display. It can also be edited by the user to customize notes

Why do you need HTML for that? The notes are not HTML themselves. Why do you want to enforce this?
To me allowing customers to mess up w/ the markup is dangerous.

If you really want to keep it, you can do the contrary: convert the html field to txt by stripping markup and preserving only line breaks but before doing this I'd like to understand what's the real use case.

JuMiSanAr:
> > The html field is not only for display. It can also be edited by the user to customize notes
> 
> Why do you need HTML for that? The notes are not HTML themselves. Why do you want to enforce this? To me allowing customers to mess up w/ the markup is dangerous.
> 
> If you really want to keep it, you can do the contrary: convert the html field to txt by stripping markup and preserving only line breaks but before doing this I'd like to understand what's the real use case.

@santostelmo

JuMiSanAr:
@santostelmo could you please check Simone's last comment?

santostelmo:
> > The html field is not only for display. It can also be edited by the user to customize notes
> 
> Why do you need HTML for that? The notes are not HTML themselves. Why do you want to enforce this? To me allowing customers to mess up w/ the markup is dangerous.
> 
> If you really want to keep it, you can do the contrary: convert the html field to txt by stripping markup and preserving only line breaks but before doing this I'd like to understand what's the real use case.

Answering your points:
Why do you need HTML for that? it allows the user to customize the message layout of the message block
To me allowing customers to mess up w/ the markup is dangerous. => Why ? I see it used in multiple reports.
convert the html field to txt  => New field could be created indeed

JuMiSanAr:
@simahawk 

If in the end we keep this `note` field as an HTML and then we create an `note_text` field, we would still need a glue module to implement this in shopfloor right?

simahawk:
> Answering your points: Why do you need HTML for that? it allows the user to customize the message layout of the message block 

> To me allowing customers to mess up w/ the markup is dangerous. => Why ?

Because they might break the layout.


> I see it used in multiple reports. convert the html field to txt => New field could be created indeed

The point is that you are altering a default behavior and, to be honest, having to support a glue module just because a customer wants to edit the final html does not make much sense.
IMO this should go in a customer specific module. Not the contrary.

An alternative can be to make `note` a computed non stored field w/ an inverse. The inverse will store the value into `note_txt` and through a ctx key you can compute the html version. But I wouldn't do this now.
So, please, move this feature to the specific module.

> If in the end we keep this `note` field as an HTML and then we create an `note_text` field, we would still need a glue module to implement this in shopfloor right?

That's exactly what I would like to avoid, especially if this is the need only of 1 customer out there. I've never heard of such a requirement till now :wink:

simahawk:
> Why do you need HTML for that? it allows the user to customize the message layout of the message block

BTW I still miss this. The notes are not in HTML, only the note on the picking is (in last version) -> how do you enable them to edit the markup if the field is not editable? :thinking:

santostelmo:
I understand the constraints with HTML field on shopfloor. Let me know if you need anything on my side regarding this refactoring

simahawk (approved):
@santostelmo as for your issue, ping me and when can see together how to work around that.

jbaudoux (approved):
Because note field on picking is a Text field in v14

PR #1546, opened 2024-03-19
[15.0][IMP] stock_picking_batch_extended: batch relation in stock move

chienandalu:
Please check @carlosdauden @sergio-teruel 

Add the relation to the batch picking in the stock.move so we can filter / group by that criteria.

cc @Tecnativa TT47127

### DIFF
--- a/stock_picking_batch_extended/models/__init__.py
+++ b/stock_picking_batch_extended/models/__init__.py
@@ -3,5 +3,6 @@
 from . import product_product
 from . import res_config_settings
 from . import stock_batch_picking
+from . import stock_move
 from . import stock_move_line
 from . import stock_warehouse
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):

stock_picking_batch_extended/migrations/15.0.3.0.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2024 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+def _set_move_id_picking_batch(env):
+    """Pre-fill column"""
+    openupgrade.logged_query(
```
pedrobaeza: This script is not needed, as now related fields are pre-filled with a similar SQL to this one.

chienandalu: Ok, I just removed it :)

pedrobaeza (approved):
/ocabot merge minor

rousseldenis:
@chienandalu Maybe an index on it should have been great

PR #1552, opened 2024-03-27
[17.0] [MIG] stock_picking_line_sequence: Migration to V17

siemenv:


stock_picking_line_sequence/models/stock.py
```diff
@@ -0,0 +1,97 @@
...
+            move = move or move_line.move_id
+            uom = move.product_uom or move_line.product_uom_id
+            name = move.product_id.display_name
+            description = move.description_picking
+            if description == name or description == move.product_id.name:
+                description = False
+            product = move.product_id
+            line_key = f'{product.id}_{product.display_name}_{description or ""}_{uom.id}_{move.product_packaging_id or ""}'
+
+            return line_key
```
StefanRijnhart: Well spotted that the `line_key` definition had to change in order to be in sync with https://github.com/odoo/odoo/blob/49d25c5/addons/stock/models/stock_move_line.py#L778-L794, but is there any reason why we can't just call that method and use the line_key that that method returns in its result dictionary?

stock_picking_line_sequence/views/stock_view.xml
```diff
@@ -0,0 +1,38 @@
...
+                expr="//notebook/page/field[@name='move_ids_without_package']"
+                position="attributes"
+            >
+                    <attribute
+                    name="context"
+                >{'picking_type_code': picking_type_code, 'default_picking_id': id,
+                    'form_view_ref':'stock.view_move_form', 'address_in_id': partner_id,
+                    'default_picking_type_id': picking_type_id, 'default_location_id': location_id,
+                    'default_location_dest_id': location_dest_id, 'default_sequence': max_line_sequence}
+                    </attribute>
```
StefanRijnhart: Overwriting the context in the form will conflict (functionally) with other modules that need to insert a default value in the same place. Would you consider something like this to inject the additional context key on the code level?

```
    @api.model
    def get_view(self, view_id=None, view_type='form', **options):
        """Append the default sequence.

        The context of `move_ids_without_package` is already overloaded and replacing
        it in a view does not scale across other extension modules.
        """
        res = super().get_view(view_id=view_id, view_type=view_type, **options)
        if res.get('arch') and view_type == 'form':
            doc = etree.XML(res['arch'])
            elements = doc.xpath("//field[@name='move_ids_without_package']")
            if elements:
                element = elements[0]
                context = element.get('context', '{}')
                context = '%s%s, %s' % (context[0], "'default_sequence': max_line_sequence", context[1:])
                element.set('context', context)
            res['arch'] = etree.tostring(doc, encoding='unicode')
        return res
```

StefanRijnhart (changes requested):
I've implemented my own suggestions in https://github.com/siemenv/stock-logistics-workflow/pull/1

StefanRijnhart:
@siemenv Any comment on my suggestion to make the extension of the context attribute more scalable by injecting the additional key in python code?

StefanRijnhart (approved):
Thanks for taking my suggestions. I would have liked to retain attribution, maybe through Co-authored-by or maybe even keep the individual commits but I guess it can't be Christmas every day.

AurelijaNorvaisaite:
@rousseldenis @siemenv is this code ready to be merged?

siemenv:
> @rousseldenis @siemenv is this code ready to be merged?

I think it is ready.

StefanRijnhart:
@AurelijaNorvaisaite Please help to get this pull request merged by doing a technical or functional review (let me try to close and reopen this PR to get the pending runboat to build so we can see this code in action).

StefanRijnhart:
@siemenv Can you try a rebase to fix runboat? My attempt to revive it did not work.

stock_picking_line_sequence/models/stock.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2017 Camptocamp SA - Damien Crier, Alexandre Fayolle
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from lxml import etree
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
```
rousseldenis, line 10: Moving one class per file should be done

StefanRijnhart:
@rousseldenis thanks! @AurelijaNorvaisaite runboat is ready if you want to have a look.

stock_picking_line_sequence/models/stock.py
StefanRijnhart: I did this in a separate commit in the migration to 18.0 in https://github.com/OCA/stock-logistics-workflow/pull/1864. @siemenv you can pick that commit if you like.

AEstLo (approved):
Code LGTM

stock_picking_line_sequence/models/stock.py
AEstLo: I can create a new PR or apply the commit afterwards

LoisRForgeFlow (approved):
/ocabot merge nobump

stock_picking_line_sequence/models/stock.py
LoisRForgeFlow: @AEstLo It is merged now, you can go ahead and create the backports

AEstLo: thanks!

AEstLo: Done: https://github.com/OCA/stock-logistics-workflow/pull/1982

PR #1563, opened 2024-04-09
[FIX] stock_picking_group_by_partner_by_carrier: Don't get the assigned picking group

rousseldenis:
As Odoo changed its behavior to auto assign picking group to moves when picking_id field is filled in (backorder creation, assignation), we let the original group on move level.

Changed here:

https://github.com/odoo/odoo/commit/ca9aab977d6aa7d3565c2b69f1cddd3edcb7fc9f

Restored also inactive test.


AungKoKoLin1997 (approved):
LGTM.

EmilioPascual (approved):
LGTM

rousseldenis:
/ocabot merge patch

As tests are failing in 16.0 branch

PR #1565, opened 2024-04-10
[16.0][FIX] stock_picking_auto_create_lot: create new serial number for lots

dessanhemrayev:
This bug was discovered while working with this module #1567


rousseldenis:
@dessanhemrayev Thanks for this. Could you maybe add a test to avoid regression ?

rousseldenis:
@dessanhemrayev

dessanhemrayev:
> @dessanhemrayev

@rousseldenis  Thanks, I'll try to add a test for this
If you can tell me in which key to add the tests?

dessanhemrayev:
@rousseldenis  check please)

rousseldenis (approved):

PR #1569, opened 2024-04-12
[15.0][ADD] stock_picking_show_lot: New module

pilarvargas-tecnativa:
New module that adds a button to the sock picking to see the lots created when the "Auto create lots" option is active in the operation type.

cc @Tecnativa TT48693

@sergio-teruel @CarlosRoca13 please review


stock_picking_show_lot/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module adds a button to the "stock.picking" form if you have the option
+"Auto Create Lots" activated in the operation type that allows you to display the
```
rousseldenis: Why is it in this case only ? Maybe this should be moved to CONTEXT fragment. IMHO, this module should not depend on stock_picking_auto_create_lot module.

pilarvargas-tecnativa: It doesn't really depend on "stock_picking_auto_create_lot" and the description was not correct, now it is. The button should appear when the option to create lots is active in the operation type. Thanks

rousseldenis (approved):
Code review

stock_picking_show_lot/i18n/es.po
```diff
@@ -0,0 +1,29 @@
...
+
+#. module: stock_picking_show_lot
+#: model_terms:ir.ui.view,arch_db:stock_picking_show_lot.view_stock_picking_show_lot
+msgid "Show Lots"
+msgstr "Mostrar Lotes"
+
+#. module: stock_picking_show_lot
+#: code:addons/stock_picking_show_lot/models/stock_picking.py:0
+#, python-format
+msgid "View Lots"
```
sergio-teruel: I think just "Lots" is enough

stock_picking_show_lot/models/stock_picking.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "stock.picking"
+
+    def action_show_lots(self):
+        action = self.env["ir.actions.act_window"]._for_xml_id(
+            "stock.action_production_lot_form"
+        )
+        action["context"] = {}
+        action["domain"] = [("id", "in", self.move_line_ids.lot_id.ids)]
+        action["name"] = _("View Lots")
+        return action
```
sergio-teruel, line 25: Please display a tree view when All lots are more than one, if not, display the form view directly

stock_picking_show_lot/views/stock_picking_views.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='button_box']" position="inside">
+                <button
+                    string="Show Lots"
+                    type="object"
+                    name="action_show_lots"
+                    class="oe_stat_button"
+                    icon="fa-cubes"
+                    attrs="{'invisible': [('use_create_lots', '=', False)]}"
```
sergio-teruel: Revome, I think that we can show it for all pickings... 
Use show_lots_text field to determine the visibility..

sergio-teruel (changes requested):

stock_picking_show_lot/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module adds a button to the "stock.picking" form if you have the option
+"Create New Lots/Serial Numbers" activated in the operation type that allows you to display the
+created "stock.production.lots".
```
yostashiro: Just a minor suggestion.
```suggestion
created "stock.production.lot" records.
```

stock_picking_show_lot/views/stock_picking_views.xml
pilarvargas-tecnativa: Perhaps it is better to define a boolean calculated field to control visibility based on whether or not there are lots in the picking. If we use the show_lots_text field we limit the visibility if it has been validated, plus the "use_existing_lots" option has to be disabled.

https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_picking.py#L537

sergio-teruel: More easier!! display button if user has the security group...

stock_picking_show_lot/readme/DESCRIPTION.rst
pilarvargas-tecnativa: thanks!

stock_picking_show_lot/views/stock_picking_views.xml
pilarvargas-tecnativa: done!

stock_picking_show_lot/models/stock_picking.py
pilarvargas-tecnativa: done!

stock_picking_show_lot/i18n/es.po
pilarvargas-tecnativa: thanks!

stock_picking_show_lot/i18n/stock_picking_show_lot.pot
```diff
@@ -0,0 +1,27 @@
...
+"Last-Translator: \n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: stock_picking_show_lot
+#: model_terms:ir.ui.view,arch_db:stock_picking_show_lot.view_stock_picking_show_lot
+msgid "Show Lots"
```
sergio-teruel: Change by lots

stock_picking_show_lot/i18n/es.po
```diff
@@ -0,0 +1,29 @@
...
+"Language: es_ES\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Generator: Poedit 3.0.1\n"
+
+#. module: stock_picking_show_lot
+#: model_terms:ir.ui.view,arch_db:stock_picking_show_lot.view_stock_picking_show_lot
+msgid "Show Lots"
```
sergio-teruel: Change by lots

sergio-teruel (changes requested):

stock_picking_show_lot/i18n/es.po
pilarvargas-tecnativa: done!

sergio-teruel (approved):

PR #1572, opened 2024-04-17
[15.0][ADD] stock_weighing: New module

chienandalu:
A shopfloor assistant for weighing on stock operations.

Some bussiness must weigh their operations in order to process them. For example in fresh product factories where the demand could not fit the exact a amount of delivered product and that depends on the final operation weight.

Peek 17-04-2024 11-54

cc @Tecnativa TT47127

ping @sergio-teruel


pedrobaeza:
I'm thinking that we should create a new repo stock-weighing for hosting all these modules. What do you think?

It's created in https://github.com/OCA/repo-maintainer-conf

@etobella it's enough with that, isn't it?

etobella:
Yes, it is you can check the PR on the automation creation to see how I created one😄

chienandalu:
> I'm thinking that we should create a new repo stock-weighing for hosting all these modules. What do you think?

Ok for me :) I'll take a look

chienandalu:
Moving to its own repository: https://github.com/OCA/stock-weighing/pull/1

PR #1573, opened 2024-04-17
[15.0][ADD] stock_picking_batch_weighing: New module 

chienandalu:
We can go to the weighing operations from an action button in the batch
picking form.

Depends on:

- [ ] https://github.com/OCA/stock-logistics-workflow/pull/1572

cc @Tecnativa TT47127

Peek 17-04-2024 16-02


chienandalu:
Moved to:

- [ ] https://github.com/OCA/stock-weighing/pull/4

PR #1574, opened 2024-04-17
[15.0][ADD] stock_weighing_auto_create_lot: New module

chienandalu:
Auto-create lots when adding detailed operations from the weight kanban card.

Depends on:

- [ ] https://github.com/OCA/stock-logistics-workflow/pull/1572

cc @Tecnativa TT47127

auto-lot


chienandalu:
Moved to:

- [ ] https://github.com/OCA/stock-weighing/pull/6

PR #1578, opened 2024-04-18
[14.0][FIX] stock_picking_putinpack_restriction: Restriction check after action_done

mt-software-de:
If there is a move_line with a qty_done = 0
it will be removed by _action_done
therefor check the restriction after _action_done
otherwise lines will be checked which will be deleted

cc @jbaudoux @TDu


TDu (approved):

luisDIXMIT (approved):
Tested on runboat. LGTM!

rousseldenis (approved):

jbaudoux (approved):

PR #1583, opened 2024-04-29
[15.0][IMP] stock_picking_invoice_link: speed up initialization

chienandalu:
Follow up to https://github.com/OCA/stock-logistics-workflow/pull/1388

New compute stored fields on large tables like this can lead to high
initialation times when we rely on the ORM.

Also, a new stored computed field was introduced in v15 along the former
computed method. We need to pre-compute it to avoid large migration
times.

cc @Tecnativa TT46020

please review @pedrobaeza 

fyi @JordiBForgeFlow


rvalyi:
Hello, I understand that stock_picking_invoice_link already depends on the sale module.
But we find this not optimal... We would prefer to have it depend only on the stock module and have a new sale_stock_picking_invoice_link that would assume the dependency on the sale module.

One of the reason is that this module is a dependency of stock_picking_invoicing and it brings that dependency on the sale module. In Brazil we use stock_picking_invoicing to create invoices that are similar to intrastat/extrastat documents (with account move lines that are required legally) when goods are moved between federal states even when there is no sale operation. So this dependency on sale here sucks a bit for us and with this kind of optimization we would be increasing the dependency...

What do you think about such proposal? If you agree I could take care of making a PR with the sale_stock_picking_invoice_link module and that kind of optimization inside (notice we alteady proposed a new sale_stock_picking_invoicing module).

cc @mbcosta @renatonlima @marcelsavegnago @pedrobaeza

pedrobaeza:
@rvalyi please, that's a total off-topic of this PR. Open an issue about it.

stock_picking_invoice_link/init_hook.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _create_delivery_count(cr):
+    """Extracted to reuse in migration script. TODO: It can be merged into a sole
+    method in v16"""
+    _logger.info("Creating column: delivery_count for table account_move")
+    cr.execute(
+        """
+    ALTER TABLE account_move
+        ADD COLUMN IF NOT EXISTS delivery_count INTEGER;
```
pedrobaeza: Please take the occasion to rename it to `picking_count`, as it's totally incorrect (and more with `stock_picking_invoice_link_purchase`.

stock_picking_invoice_link/init_hook.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        UPDATE account_move am SET picking_count = t.picking_count
+        FROM (
+            SELECT
+                am.id AS account_move_id,
+                COUNT(DISTINCT sm.picking_id) AS picking_count
+            FROM
+                account_move am
+                JOIN account_move_line aml ON aml.move_id = am.id
+                JOIN stock_move_invoice_line_rel rel
```
pedrobaeza, line 39: Why not using here account_move_stock_picking_rel? Maybe you just need to swap order

pedrobaeza:
Superseded by #1617

PR #1585, opened 2024-05-03
[16.0][ADD] stock_customer_deposit_sale_margin

EmilioPascual:
Cost in sale order line when make sale order to deliver customer deposit should be zero because purchase price is in sale order that made customer deposit, otherwise the cost would be posted twice.

@yajo @Shide @Gelojr @fcvalgar @rafaelbn please review.

MT-5964


stock_customer_deposit_sale_margin/models/sale_order_line.py
```diff
@@ -0,0 +1,23 @@
...
+        "currency_id",
+        "product_uom",
+        "deposit_available_qty",
+    )
+    def _compute_purchase_price(self):
+        res = super()._compute_purchase_price()
+        for line in self.filtered_domain([("deposit_available_qty", ">", 0.0)]):
+            # Set purchase price to zero because purchase price is in sale order
+            # that made customer deposit, otherwise the cost would be posted twice.
+            line.purchase_price = 0.0
```
Shide: Maybe check if deposit available is < than demanded quantity?

```suggestion
        "deposit_available_qty",
        "product_uom_qty",
    )
    def _compute_purchase_price(self):
        res = super()._compute_purchase_price()
        for line in self.filtered_domain([("deposit_available_qty", ">", 0.0)]):
            # Set purchase price to zero because purchase price is in sale order
            # that made customer deposit, otherwise the cost would be posted twice.
            if line.deposit_available_qty < line.product_uom_qty:
                continue
            line.purchase_price = 0.0
```

Shide (changes requested):
Looks good!

question: How if deposit is < than demanded?

stock_customer_deposit_sale_margin/models/sale_order_line.py
EmilioPascual: I think this is not necessary, user can't confirm sale if customer deposit is lower than qty demanded. See https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_customer_deposit/models/sale_order.py#L61

yajo: Makes sense, indeed, because if the uom qty is higher, then it can't be borrowed from the deposit.

Beware, your suggestions has a bug IIUC: it should be `<=`, not just `<`. Also you can optimize it a bit more by looping only once:

```suggestion
        "deposit_available_qty",
        "product_uom_qty",
    )
    def _compute_purchase_price(self):
        res = super()._compute_purchase_price()
        for line in self:
            if 0 < line.deposit_available_qty <= line.product_uom_qty:
                continue
            # Set purchase price to zero because purchase price is in sale order
            # that made customer deposit, otherwise the cost would be posted twice.
            line.purchase_price = 0.0
```

stock_customer_deposit_sale_margin/readme/CONTEXT.md
```diff
@@ -0,0 +1 @@
+This module was developed because cost in sale order line when deliver customer deposit should be zero because purchase price is in sale order that made customer deposit, otherwise the cost would be posted twice.
```
yajo: suggestion: reword and simplify to make it easier to understand

```suggestion
This module was created to make sure that the cost on a sale order line is zero when delivering from customer deposits. This is needed because the purchase price already accounts for the customer deposit in the sale order. Without this module, the cost would be counted twice in the accounting.
```

stock_customer_deposit_sale_margin/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module extends the functionality of the consignment stock to support customer deposits and set cost in sale order line to zero when make sale order to deliver customer deposit
```
yajo:
```suggestion
This module extends the functionality of the consignment stock to support customer deposits and set cost in sale order line to zero when making another sale order to deliver some previously consigned customer deposit.
```

stock_customer_deposit_sale_margin/readme/USAGE.md
```diff
@@ -0,0 +1,29 @@
...
+1. Click on smart button with delivery.
+1. Set quantity done in operations or click on **Set Quantities**.
+1. Update location destination if it's necessary in operations detailed.
+1. Click on **Validate**.
+
+Deliver customer deposits:
+
+1. Go to Sales > Quotation.
+1. Create a new quotation.
+1. Select customer who has a deposit in your warehouse.
```
yajo:
```suggestion
1. Select a customer who has a deposit in your warehouse.
```

yajo (changes requested):
Some not-much-relevant suggestions, but pretty good overall.

stock_customer_deposit_sale_margin/tests/test_stock_customer_deposit_sale_margin.py
```diff
@@ -0,0 +1,47 @@
...
+        super().setUpClass()
+        (cls.productA | cls.productB | cls.productC).write({"standard_price": 10.0})
+        stock_dict = {
+            cls.productA: {cls.partner1: 100},
+            cls.productB: {cls.partner1: 120},
+            cls.productC: {False: 500},
+        }
+        cls.update_availiable_quantity(cls, stock_dict)
+
+    def test_something(self):
```
Shide: Change test name and let tests pass

Shide (changes requested):

Shide (approved):

yajo (approved):

Gelojr (approved):
Nice Job @EmilioPascual 
https://www.loom.com/share/c1ed426bcf334c33b2ed86d79321a1ec?sid=f34a684f-5bfb-4932-8ffc-48c090c181ad

rafaelbn (approved):
/ocabot merge patch

PR #1586, opened 2024-05-06
[16.0][IMP]add availibility on stock split picking

syera94:
This change add the possibility to split the picking between available and not available products.

@sebastienbeau 
I put it in V16 and i let you check the code and give me your return before to back port it in V14. 
(For my part it s more logical in this way but if you don t have time i can do it before, let me know )
I will add some test for this new feature.

### DIFF
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -12,34 +12,36 @@ class StockPicking(models.Model):
 
     _inherit = "stock.picking"
 
-    def _check_split_process(self):
+    def _check_split_process(self, split_move_type):
         # Check the picking state and condition before split
         if self.state == "draft":
             raise UserError(_("Mark as todo this picking please."))
-        if all([x.qty_done == 0.0 for x in self.move_line_ids]):
-            raise UserError(
-                _(
-                    "You must enter done quantity in order to split your "
-                    "picking in several ones."
-                )
-            )
 
-    def split_process(self):
+        if split_move_type == "quantity_done":
+            if all([x.qty_done == 0.0 for x in self.move_line_ids]):
+                raise UserError(
+                    _(
+                        "You must enter done quantity in order to split your "
+                        "picking in several ones."
+                    )
+                )
+
+    def split_process(self, split_move_type):
         """Use to trigger the wizard from button with correct context"""
         for picking in self:
-            picking._check_split_process()
+            picking._check_split_process(split_move_type)
 
             # Split moves considering the qty_done on moves
             new_moves = self.env["stock.move"]
             for move in picking.move_ids:
                 rounding = move.product_uom.rounding
-                qty_done = move.quantity_done
+                qty_compare_split = getattr(move, split_move_type)
                 qty_initial = move.product_uom_qty
                 qty_diff_compare = float_compare(
-                    qty_done, qty_initial, precision_rounding=rounding
+                    qty_compare_split, qty_initial, precision_rounding=rounding
                 )
                 if qty_diff_compare < 0:
-                    qty_split = qty_initial - qty_done
+                    qty_split = qty_initial - qty_compare_split
                     qty_uom_split = move.product_uom._compute_quantity(
                         qty_split, move.product_id.uom_id, rounding_method="HALF-UP"
                     )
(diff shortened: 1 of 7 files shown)


rousseldenis:
@syera94 Thanks for this. It could be great if you add tests that cover these changes.

syera94:
@rousseldenis 
@sebastienbeau 
Hi I add some test for this new feature.
Maybe it's a bit long cause i did not want to change all the previous tests and i add the picking creation directly in the method.
Let me know if a have to change it.

stock_split_picking/wizards/stock_split_picking.py
```diff
@@ -43,6 +52,23 @@ def _apply_move(self):
                 new_pickings += picking._split_off_moves(move)
         return self._picking_action(new_pickings)
 
+    def _apply_available_line(self):
+        """Create different pickings for available and not available move line"""
+        for picking in self.mapped("picking_ids"):
+            if picking.picking_type_code == "outgoing":
+                moves = picking.mapped("move_ids")
+                moves_available = moves.filtered(lambda move: move.state == "confirmed")
```
sebastienbeau: Wrong filter, available move are in the state "assigned" not "confirmed" 
```suggestion
                moves_available = moves.filtered(lambda move: move.state == "assigned")
```

stock_split_picking/wizards/stock_split_picking.py
```diff
@@ -43,6 +52,23 @@ def _apply_move(self):
...
+                new_picking = moves_available.mapped("picking_id")._split_off_moves(
+                    moves_available
+                )
+            else:
+                raise UserError(_("This type of mode is not available for the receipt"))
+        return self._picking_action(new_picking)
+
+    def _apply_available_product(self):
+        """Create different pickings for available and not available move line"""
+        return self.mapped("picking_ids").split_process("reserved_availability")
```
sebastienbeau: After splitting a picking based on available product, the picking that contain all the product available is still in the state confirm (user need to click on check availability), because all the line that was partially available are still partially available (the state is not recomputed automatically).
So after calling the split_process you need to call "_recompute_state" on the move
```suggestion
        self.picking_ids.split_process("reserved_availability")
        self.picking_ids.move_ids._recompute_state()
        return True
        
```

stock_split_picking/wizards/stock_split_picking.py
```diff
@@ -31,7 +40,7 @@ def action_apply(self):
         return getattr(self, "_apply_%s" % self[:1].mode)()
 
     def _apply_done(self):
-        return self.mapped("picking_ids").split_process()
+        return self.mapped("picking_ids").split_process("quantity_done")
```
sebastienbeau: mapped is useless. It's only need when you process field that do not containt an object
exemple 
- self.mapped("name") : mapped is used as name is a string
- self.picking_ids : no need to use mapped, odoo will do the magic (on old version it was necessary)


```suggestion
        return self.picking_ids.split_process("quantity_done")
```

stock_split_picking/wizards/stock_split_picking.py
```diff
@@ -43,6 +52,23 @@ def _apply_move(self):
                 new_pickings += picking._split_off_moves(move)
         return self._picking_action(new_pickings)
 
+    def _apply_available_line(self):
+        """Create different pickings for available and not available move line"""
+        for picking in self.mapped("picking_ids"):
+            if picking.picking_type_code == "outgoing":
```
sebastienbeau: do not filter on "outgoing". The split can be used in other context like internal move.
For sure for incoming it do not make sense as incoming are always available, but I do not think we need to add extra code to block that case.

sebastienbeau (changes requested):

sebastienbeau:
@syera94 thanks for your work, I have done some comment.
Can you also add the french translation for the new keyword added?

stock_split_picking/models/stock_picking.py
```diff
@@ -12,34 +12,38 @@ class StockPicking(models.Model):
...
+                raise UserError(
+                    _(
+                        "You must enter done quantity in order to split your "
+                        "picking in several ones."
+                    )
                 )
-            )
 
-    def split_process(self):
+    def split_process(self, split_move_type):
```
rousseldenis, line 29: As this parameter is in fact a field of stock move, maybe its name should reflect that ?

rousseldenis: A docstring should be great too

PR #1587, opened 2024-05-07
14.0 imp add available on stock split picking

syera94:
@sebastienbeau 
I back port it in V14.
This way you can check in V14 and in V16. 
I will also add test


rousseldenis:
@syera94 Could you point out where is the original PR ? The commit messages seem to be a little weird. They should mention the module name like `[IMP] stock_split_picking: <improvement>`. Thanks

PR #1490, opened 2024-01-24
[16.0][IMP] stock_picking_mass_action: Added the possibility to indicate to the method if we want other types of operation.

RodrigoBM:

PR #1599, opened 2024-05-23
[15.0][IMP] stock_picking_mass_action: allow to configure `check_assign_all` domain

LoisRForgeFlow:
It is useful for a easy tweaking of the configuration.

### DIFF
--- a/stock_picking_mass_action/models/stock_picking.py
+++ b/stock_picking_mass_action/models/stock_picking.py
@@ -10,9 +10,13 @@ class StockPicking(Model):
     _inherit = "stock.picking"
 
     @api.model
-    def check_assign_all(self):
+    def check_assign_all(self, domain=None):
         """Try to assign confirmed pickings"""
-        domain = [("picking_type_code", "=", "outgoing"), ("state", "=", "confirmed")]
+        if not domain:
+            domain = [
+                ("picking_type_code", "=", "outgoing"),
+                ("state", "=", "confirmed"),
+            ]
         records = self.search(domain, order="scheduled_date")
         records.action_assign()
 


stock_picking_mass_action/models/stock_picking.py
```diff
@@ -10,9 +10,13 @@ class StockPicking(Model):
     _inherit = "stock.picking"
 
     @api.model
-    def check_assign_all(self):
+    def check_assign_all(self, domain=None):
         """Try to assign confirmed pickings"""
-        domain = [("picking_type_code", "=", "outgoing"), ("state", "=", "confirmed")]
+        if not domain:
```
rousseldenis, line 15: @LoisRForgeFlow Should be the domain overridable or completely new one ?

LoisRForgeFlow: not sure what is the difference :sweat_smile: : If you set a domain in the cron code you will completely override the domain by a custom one.

rousseldenis: Yes. I wanted to say, you don't want to keep the 'outgoing' and 'confirmed' filters and extend the domain ?

LoisRForgeFlow: Yep, the intention is to give full control. And actually, my use case for this module is for a series of internal operations

rousseldenis (approved):

JoanSForgeFlow (approved):
LGTM!

LoisRForgeFlow:
Rebased to include fix for test in other module, ready to merge now.

/ocabot merge patch

LoisRForgeFlow:
A similar change was done in v16 (https://github.com/OCA/stock-logistics-workflow/pull/1490) so not really need to fwport.

PR #1600, opened 2024-05-24
[16.0][ADD] stock_picking_partner_block_out: init

AnizR:
### New module

This module allows you to block every deliveries to a partner. This is done by activating a boolean on a contact. It also adds a warning header on every delivery related to any blocked partner.

Here is an example where Deliveries to Abigail have been blocked:
Screenshot from 2024-05-24 13-49-52


stock_picking_partner_block_out/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    partner_picking_block_out = fields.Boolean(
+        related="partner_id.picking_block_out", store=True
+    )
+
+    @api.constrains("partner_picking_block_out", "state")
```
acsonefho, line 16: Can you check if you also have to add `picking_type_code` and `is_dropship` (who are behind `_is_to_external_location()`)

acsonefho (changes requested):
Just a question

stock_picking_partner_block_out/models/stock_picking.py
AnizR: I have modified my constraint because indeed, I should depend on `picking_type_code` but IMO, I shouldn't add 'is_dropship' since it is introduced by the module `stock_dropshipping`

AnizR: I changed my mind about adding `picking_type_code`:
* This field is a related -> I should store the field to be able to use add it in `@api.constrains`
* Changing the picking type is only possible in 'draft' state so in my case, it is not necessary to trigger the constraint when modifying the picking type

IT-Ideas (approved):
LGTM => code review

SamyBrahimi:
I made a functional test. The module works find. It will be also interesting to hide the button cancel.

Best

rousseldenis:
@AnizR Could you rebase ?

AnizR:
> @AnizR Could you rebase ?

Done :+1:

stock_picking_partner_block_out/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    partner_picking_block_out = fields.Boolean(
```
rousseldenis, line 12: @AnizR A value change on the partner side will recompute all the values for all the pickings (of the partner - but some DB's have great history). Do you really want to store it ?

rousseldenis:
@AnizR There is also this one (a little bit different): https://github.com/OCA/sale-workflow/tree/16.0/sale_stock_picking_blocking

AnizR:
Since another module exist, I'll close this PR

jbaudoux (commented):
Looks strange to me to allow to create a picking (or at least mark it as To Do) if you will never be able to process it. Seems a bit late in the process without knowing the real use case

PR #1561, opened 2024-04-08
[16.0][FIX] stock_picking_auto_create_lot: use standard method for lot val prep

AungKoKoLin1997:
This PR supersedes https://github.com/OCA/stock-logistics-workflow/pull/1297.

@qrtl

PR #1601, opened 2024-05-25
[15.0][IMP] stock_picking_auto_create_lot: Use Odoo core mechanism to create lots in pickings filling lot_name field instead of create all lots.

sergio-teruel:
[IMP] stock_picking_auto_create_lot: Use Odoo core mechanism to create lots in pickings filling lot_name field instead of create all lots.

Why this PR... well... I have noticed that if you install product_expiry, for example, you can set the expiration date in stock move lines directly, but stock_picking_auto_create_lot module break this date creating the lots with other dates (the default dates based on product time fields)...

Reviwing this module I think that is better let Odoo create the lots filling the lot_name field on stock move lines... 

After this PR, I will review future versions

ping @carlosdauden @CarlosRoca13 @rousseldenis


carlosdauden (approved):

stock_picking_auto_create_lot/models/stock_move_line.py
```diff
@@ -1,42 +1,18 @@
...
-            line.lot_id = lot
-            # If product has tracking we have to remove the lot to assign the next
-            # lot created. If not do it, different stock move lines will have the same
-            # lot and we will have orphans lots.
-            if lot.product_id.tracking != "none":
-                lots_by_product[line.product_id.id] -= lot
+        for line in self.filtered(
+            lambda sml: not sml.lot_name and sml.tracking != "none"
+        ):
+            line.lot_name = self.env["ir.sequence"].next_by_code("stock.lot.serial")
```
yostashiro: I think `filtered()` here is redundant as `_set_auto_lot()` in stock.picking has already done this filtering.  I guess we could simply remove this method and complete `lot_name` assignment in the method in the picking.

It would be good to put `self.env["ir.sequence"].next_by_code("stock.lot.serial")` part into a method for extensibility.  For example, a client of ours uses the `product_lot_sequence` module along with this module and they currently work nicely together (auto-created lots/serials follow the sequence defined in `product_lot_sequence`), but the suggested change (i.e. assign the `lot_name` instead of creating a lot) will "break" this design, although these two modules may have not been intended to work together in the first place.  Having a hook point will allow us to create a bridge module to restore the behavior as necessary.

We created https://github.com/OCA/stock-logistics-workflow/pull/1561 for 16.0 with a less radical fix for the addressed issue, by the way. But I like the approach suggested in this PR better as it probably lightens the maintenance burden for the future.

sergio-teruel:
@yostashiro Changes done!!!

stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py
```diff
@@ -36,8 +36,7 @@ def test_manual_lot(self):
         self.assertTrue(move.display_assign_serial)
 
         # Assign manual serials
-        for line in move.move_line_ids:
-            line.lot_id = self.lot_obj.create(line._prepare_auto_lot_values())
+        self._assign_manual_serials(move)
```
yostashiro, line 39:
```suggestion
        self._assign_manual_serials(move)
        self.picking.action_set_quantities_to_reservation()
```

stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py
```diff
@@ -62,6 +61,8 @@ def test_auto_create_lot(self):
             lambda m: m.product_id == self.product_serial_not_auto
         )
         self.assertTrue(move.display_assign_serial)
+        # Assign manual serials
+        self._assign_manual_serials(move)
```
yostashiro, line 66:
```suggestion
        self._assign_manual_serials(move)
        self.picking.action_set_quantities_to_reservation()
```

stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py
```diff
@@ -90,12 +91,10 @@ def test_auto_create_transfer_lot(self):
         moves = self.picking.move_lines.filtered(
             lambda m: m.product_id == self.product_serial_not_auto
         )
-
         # Assign manual serials
-        for line in moves.mapped("move_line_ids"):
-            line.lot_id = self.lot_obj.create(line._prepare_auto_lot_values())
+        self._assign_manual_serials(moves)
```
yostashiro, line 97:
```suggestion
        self._assign_manual_serials(moves)
        self.picking.action_set_quantities_to_reservation()
```

stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py
```diff
@@ -90,12 +91,10 @@ def test_auto_create_transfer_lot(self):
...
 
-        self.picking.button_validate()
+        self.picking.with_context(skip_immediate=True).button_validate()
```
yostashiro: I guess we don't need this adjustment.
```suggestion
        self.picking.button_validate()
```

stock_picking_auto_create_lot/tests/test_stock_picking_auto_create_lot.py
```diff
@@ -178,7 +168,15 @@ def test_multiple_sml_for_one_stock_move(self):
...
+        self.picking.move_line_ids.copy({"qty_done": 25.0})
         self.picking.button_validate()
         lots = self.picking.move_line_ids.lot_id
         self.assertEqual(len(lots), 2)
+
+    def _assign_manual_serials(self, moves):
+        # Assign manual serials
+        moves.picking_id._set_auto_lot()
+        moves.move_line_ids.qty_done = 1.0
+        moves.picking_id.move_lines._set_quantities_to_reservation()
```
yostashiro: IMO, this line should not be inside this method as it blurs the scope and could be confusing. The equivalent step should be inserted into the relevant parts of the test respectively (as suggested).
```suggestion
```

yostashiro (commented):
Thanks, @sergio-teruel ! Some comments on tests.

rousseldenis (approved):
LGTM. In fact, the implementation is better

sergio-teruel:
@yostashiro Changes done!!

yostashiro (approved):
LGTM. Thank you!

pedrobaeza:
Do you know why the error in the CI? Let's try anyway to merge:

/ocabot merge major

rousseldenis:
stock_picking_propagate_scheduled_date module tests compare dates but without using freezegun... @yankinmax

yankinmax:
> stock_picking_propagate_scheduled_date module tests compare dates but without using freezegun... @yankinmax

@vvrossem can you pls handle the needed changes to `stock_picking_propagate_scheduled_date` v15?

vvrossem:
@yankinmax cf. https://github.com/OCA/stock-logistics-workflow/pull/1604

PR #1602, opened 2024-05-28
[REF] stock_picking_batch_print_pickings: Upgrading for dynamic printing

EmilioPascual:
Module refactoring for upgrading for dynamic printing.

Add options to print:

- Only pickings.
- Only invoices.
- Invoices and pickings.
- Invoices or pickings. If picking has not invoice then print picking.

@Shide @Gelojr @rafaelbn @yajo @fcvalgar please review.

MT-6218 @moduon

### DIFF
--- a/stock_picking_batch_print_invoices/views/report_picking_batch_print_invoices.xml
+++ b/stock_picking_batch_print_invoices/views/report_picking_batch_print_invoices.xml
@@ -3,25 +3,46 @@
      License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0) -->
 <odoo>
     <template
-        id="report_picking_batch_print_invoices"
-        inherit_id="stock_picking_batch_print_pickings.report_picking_batch_print_pickings"
+        id="print_pickings_invoices"
+        inherit_id="stock_picking_batch_print_pickings.print_pickings"
     >
+        <xpath
+            expr="//t[@t-if=&quot;type_document == &apos;pickings&apos;&quot;]"
+            position="attributes"
+        >
+            <attribute
+                name="t-if"
+            >type_document in {'pickings', 'invoices_and_pickings'} or
+                type_document == 'invoices_or_pickings' and not doc.picking_ids.filtered(lambda p:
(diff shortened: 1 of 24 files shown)


EmilioPascual:
I have recorded a video with new behavior:

https://www.loom.com/share/bfe4c28a9a8449e1bfba58a5cf14415d?sid=3b7838e4-942d-4cb5-bca8-6843722d72b5

stock_picking_batch_print_invoices/views/report_picking_batch_print_invoices.xml
```diff
@@ -3,25 +3,44 @@
...
-                    </t>
-                </t>
+                <t t-call="stock_picking_batch_print_invoices.print_invoices" />
             </t>
         </xpath>
     </template>
+    <template id="print_invoices">
+        <t
+            name="print_invoices"
+            t-foreach="set(doc.picking_ids.filtered(lambda p: p.partner_id == partner).sale_id.invoice_ids)"
```
Shide: This will print all invoices even if it's an invoice from a picking invoiced before.

I think should be something like that:
```suggestion
            t-foreach="doc.picking_ids.filtered(lambda p: p.partner_id == partner).mapped("move_ids.sale_line_id.invoice_lines.move_id").filtered(lambda inv: inv.state == 'posted')"
```
With this approach you will only print recent invoices.

Shide (changes requested):

stock_picking_batch_print_invoices/views/report_picking_batch_print_invoices.xml
EmilioPascual: You're right. I have pushed changes. 
Thank you @Shide.

Shide: `set()` is not necessary

Shide (approved):
Code LGTM

fcvalgar (approved):
Functional review,

LGTM, thank you @EmilioPascual 

image

image

image

image

image

PR #1597, opened 2024-05-21
[16.0][FIX] stock_account_product_run_fifo_hook, rewrite to original method

ChrisOForgeFlow:
Update method to keep original algorithm in current version

@ForgeFlow
cc. @JordiBForgeFlow

PR #1605, opened 2024-05-29
[16.0][REM] stock_account_product_run_fifo_hook: Not usable right now

pedrobaeza:
After the refactoring done in https://github.com/odoo/odoo/pull/164295
    
As the changes are very deep and out of the PSC duties, it's disabled for now for greenifying the branch.

@Tecnativa


pedrobaeza:
@LoisRForgeFlow we need to put this into green quickly, as there's no easy patch for this, so I'm going to merge. You can re-enable the modules when you have the definitive patch. Anyway, you should rethink your strategy of doing this "hook" modules, as they are very fragile and not good at long term. Instead, you should try to do the needed things without overriding whole code, or ask for more hooks in upstream. Tell me if I can assist you in this process.

/ocabot merge nobump

JordiMForgeFlow:
@pedrobaeza we have the fix prepared in https://github.com/OCA/stock-logistics-workflow/pull/1597. We will revert this changes in the other PR.

CC @ChrisOForgeFlow

PR #1606, opened 2024-05-30
[14.0] update stock picking force availability

mathieudelva:


mathieudelva:
Modification of the module to allow selecting the quantity to be forced and the picking from which we want to take these quantities

mathieudelva (commented):
Modification of the module to allow selecting the quantity to be forced and the picking from which we want to take these quantities

PR #1617, opened 2024-06-10
[15.0][FIX] stock_picking_invoice_link: don't store auxiliar fields

chienandalu:
In c332869b09d99231f3c5f957089fddd257ee23ba a new computed field was introduced to show the count of a move's related pickings. But it was made stored, which is needless for this kind of field and adds a huge overhead when migrating from big DBs from previous versions.

With this commit:

- We drop the storing of this field which is meant to play a UI role.
- We rename the field to picking count, as a more precise nomenclature.

cc @Tecnativa TT46020

please review @pedrobaeza @sergio-teruel

### DIFF
--- a/stock_picking_invoice_link/models/account_move.py
+++ b/stock_picking_invoice_link/models/account_move.py
@@ -21,8 +21,8 @@ class AccountMove(models.Model):
         help="Related pickings (only when the invoice has been generated from a sale order).",
     )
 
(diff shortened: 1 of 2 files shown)


christian-ramos-tecnativa (approved):
LGTM

pedrobaeza (approved):

sergio-teruel (approved):

PR #1620, opened 2024-06-11
[16.0][FIX] stock_picking_invoice_link: don't store auxiliar fields

chienandalu:
fw of https://github.com/OCA/stock-logistics-workflow/pull/1617

In c332869b09d99231f3c5f957089fddd257ee23ba a new computed field was introduced to show the count of a move's related pickings. But it was made stored, which is needless for this kind of field and adds a huge overhead when migrating from big DBs from previous versions.

With this commit:

- We drop the storing of this field which is meant to play a UI role.
- We rename the field to picking count, as a more precise nomenclature.

cc @Tecnativa TT46020

please review @pedrobaeza @sergio-teruel

### DIFF
--- a/stock_picking_invoice_link/models/account_move.py
+++ b/stock_picking_invoice_link/models/account_move.py
@@ -19,8 +19,8 @@ class AccountMove(models.Model):
         help="Related pickings (only when the invoice has been generated from a sale order).",
     )
 
-    delivery_count = fields.Integer(
-        string="Delivery Orders", compute="_compute_picking_ids", store=True
+    picking_count = fields.Integer(
+        string="Pickings count", compute="_compute_picking_ids"
     )
 
     @api.depends("invoice_line_ids", "invoice_line_ids.move_line_ids")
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
Only 17.0 is pending

rousseldenis:
@pedrobaeza @chienandalu You forgot to watch warnings in logs: https://github.com/OCA/stock-logistics-workflow/actions/runs/9496218506/job/26170126787#step:8:89

Please fix it

rousseldenis:
> @pedrobaeza @chienandalu You forgot to watch warnings in logs: https://github.com/OCA/stock-logistics-workflow/actions/runs/9496218506/job/26170126787#step:8:89
> 
> Please fix it

You should use another compute method for the count

pedrobaeza:
Uhm, maybe other module is creating the same field? Both fields in this module don't have `compute_sudo`?

rousseldenis:
> Uhm, maybe other module is creating the same field? Both fields in this module don't have `compute_sudo`?

Nope. See https://github.com/odoo/odoo/blob/16.0/odoo/modules/registry.py#L322

rousseldenis:
Moreover you changed the field name.

It shouldn't as this was done at purpose to avoid name collisions. delivery_count is really the meaning of the field.

pedrobaeza:
No, that one is incorrect, as it doesn't reflect the real thing, and more with `purchase_stock_picking_invoice_link`. So this is finally a name collision with other module?

PR #1621, opened 2024-06-11
[16.0][FIX] stock_move_line_change_lot: fix tests

AungKoKoLin1997:
This PR aims to avoid test failures when there are modules that extend the behavior of lot record creation and to keep these modules working together. Additionally, creating `stock.lot` records with `tracking=none` is not a good practice.

@qrtl


rousseldenis (approved):

kanda999 (approved):
LGTM

MarwanBHL (approved):

PR #1629, opened 2024-06-14
[16.0][ADD] stock_quant_package_restrict_disposable_reuse

MarinaAForgeFlow:
This module restricts the reuse of disposable packages by not allowing to select them in stock
move lines after they have already been allocated in a delivery.

This restriction can be activated or deactivated at package type level. So the user can select
for each package type if the disposable packages can be reused or not.

Usage
=====

* Create a package and set the "package_use" to disposable.
* Create a package type with "restrict_reuse" to true and assign it to your package.
* Use the package in a delivery.
* Create another transfer and try to select the same package.
  You will see that it does not appear anymore on the list.


MarinaAForgeFlow:
I was wondering if when validating a delivery I should erase the package used from all the other non done transfers.  

Because you could have cases where a package has been selected for a move line on a delivery but at the same time in other moves. Since the delivery is not validated, you can select it and no error will occur.

An scenario where this could make sense is when there are two deliveries to the same customer, and they pack the products in one same package. So maybe a solution would be to erase it form all the other transfers that are not deliveries?

Any thoughts on this?

stock_quant_package_restrict_disposable_reuse/models/stock_move_line.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _check_disposable_package(self, package_id):
+        package = self.env["stock.quant.package"].browse(package_id)
+        if package.package_use == "disposable" and package.disposable_used:
+            raise ValidationError(
+                _(
+                    "You can't reuse a disposable package that has already been used"
+                    " in a validated delivery."
+                )
+            )
```
JordiMForgeFlow: you could have this as an api.constrain and you won't need to inherit the create and write

JordiMForgeFlow:
> I was wondering if when validating a delivery I should erase the package used from all the other non done transfers.
> 
> Because you could have cases where a package has been selected for a move line on a delivery but at the same time in other moves. Since the delivery is not validated, you can select it and no error will occur.
> 
> An scenario where this could make sense is when there are two deliveries to the same customer, and they pack the products in one same package. So maybe a solution would be to erase it form all the other transfers that are not deliveries?
> 
> Any thoughts on this?

@MarinaAForgeFlow what if you consider that it has been used the moment it has been assigned to a delivery? Not when the delivery has been done? If the package has already been allocated to a delivery, it should not be able to be allocated to another one.

MarinaAForgeFlow:
> @MarinaAForgeFlow what if you consider that it has been used the moment it has been assigned to a delivery? Not when the delivery has been done? If the package has already been allocated to a delivery, it should not be able to be allocated to another one.

@JordiMForgeFlow  I will do it like this. Thanks!

rousseldenis:
Great. I would have added an option to activate the restriction or not (to avoid breaking standard behavior):

- Either at company level (in Inventory settings)
- Either the mandatory use of a package type that is set as ('Restrict reuse') or something like that

navarrorico:
Good idea for a module!

> * Either at company level (in Inventory settings)

I think it should be by warehouse, rather than by company. Since the same company can have several warehouses with very different configurations.

MarinaAForgeFlow:
@rousseldenis @navarrorico 

I have finally made it at package type level. So you can define in the package type if you want to restrict the reuse of the disposable packages of that type or not. I believe this adds more flexibility, but let me know what you think :)

navarrorico (approved):
Functional review LGTM :+1:

stock_quant_package_restrict_disposable_reuse/models/stock_move_line.py
```diff
@@ -0,0 +1,66 @@
...
+            and package.package_type_id.restrict_reuse
+            and package.package_use == "disposable"
+            and package.disposable_used
+        ):
+            raise ValidationError(
+                _(
+                    "You can't reuse a disposable package that has already been used"
+                    " in a delivery."
+                )
+            )
```
JordiMForgeFlow: this method could check both package and result package. Then in the constrain you only need to make one call: self._check_disposable_package()

Also, in the error message it would be good to add the package number

JordiMForgeFlow (approved):
code review

stock_quant_package_restrict_disposable_reuse/models/stock_move_line.py
```diff
@@ -0,0 +1,80 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    result_package_id = fields.Many2one(
+        domain="['|', '|', ('location_id', '=', False), ('location_id', '=', location_dest_id),"
```
rousseldenis, line 12: We need to be conscient this is incompatible with another module doing the same.

stock_quant_package_restrict_disposable_reuse/models/stock_move_line.py
```diff
@@ -0,0 +1,80 @@
...
+    def _check_disposable_packages(self):
+        if (
+            self.package_id
+            and self.package_id.package_type_id.restrict_reuse
+            and self.package_id.package_use == "disposable"
+            and self.package_id.disposable_used
+        ):
+            raise ValidationError(
+                _(
+                    "You can't reuse package %s because it is a disposable package that has "
```
rousseldenis:
```suggestion
                    "You can't reuse package %(package_name)s because it is a disposable package that has "
                    "already been used in a delivery.", pacakge_name=self.package_id.name
```

stock_quant_package_restrict_disposable_reuse/models/stock_move_line.py
```diff
@@ -0,0 +1,80 @@
...
+            and self.result_package_id.package_type_id.restrict_reuse
+            and self.result_package_id.package_use == "disposable"
+            and self.result_package_id.disposable_used
+        ):
+            raise ValidationError(
+                _(
+                    "You can't reuse package %s because it is a disposable package that has "
+                    "already been used in a delivery."
+                )
+                % self.result_package_id.name
```
rousseldenis: Same as above

rousseldenis (commented):
Code review. Minor changes.

MarinaAForgeFlow:
Hi @rousseldenis , could you review the latest changes? Thanks!

JordiMForgeFlow:
Hi @rousseldenis 

could we get this reopened for merge?

PR #1632, opened 2024-06-14
[TESTS] Check warnings

rousseldenis:


rousseldenis:
@sbidoul @pedrobaeza @OCA/core-maintainers This is a proof of concept that shows how to trigger a failing check on PR's main page in order to warn people about current errors (showed also in tests) and warnings (the main goal of this).

I found that a lot of repositories have bunches of warnings that are never addressed, but should.

The idea is to warn earlier (before merge) people about those warnings. As github has no intermediate (just fail/success) status for checks we cannot do that in normal checks and has to be on separate one.

The idea is to collect logs across matrix steps and run checkog-odoo. The difficult part was to generate unique log file id (if someone has better idea).

Comments are welcome !

hbrunn:
yes please!

pedrobaeza:
Well, it sounds good to have it. Any drawback with the storage? Do we have a limit?

.github/workflows/test.yml
```diff
@@ -81,10 +87,53 @@ jobs:
...
+            echo 'FILELIST<<EOF'
+            find . -name '*.log' -print 
+            echo EOF
+          } >> "$GITHUB_ENV"
+      - name: Check logs
+        run: |
+          pip install checklog-odoo
+          echo $FILELIST
+          {
+          for x in $FILELIST; do
```
sbidoul, line 136:
```suggestion
          for x in $(ls *.log); do
```

You can probably do this in one step like this.

sbidoul:
@rousseldenis that's nice. Related: https://github.com/OCA/oca-addons-repo-template/issues/120

We could also enable this with an environment variable in oca-ci: https://github.com/OCA/oca-ci/issues/10

PR #1644, opened 2024-07-04
[FIX] stock_picking_propagate_scheduled_date: recursion error

yankinmax:


vvrossem (approved):

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -24,10 +24,14 @@ def _propagate_date(self, new_date):
...
+                move_dest.with_context(skip_date_propagation=True).write(
+                    {
+                        "date": move_dest.date - delta,
+                    }
+                )
 
     def write(self, vals):
         # propagate date changes in the stock move chain
-        if "date" in vals:
+        if "date" in vals and not self.env.context.get("skip_date_propagation", False):
```
gurneyalex: This fix is not correct. We want the date change to propagate beyond the 1st chained move.

gurneyalex (changes requested):

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -24,10 +24,14 @@ def _propagate_date(self, new_date):
                     continue
                 if move_dest.id in already_propagated_ids:
                     continue
-                move_dest.date -= delta
+                move_dest.with_context(skip_date_propagation=True).write(
+                    {
+                        "date": move_dest.date - delta,
+                    }
+                )
```
SilvioC2C: Since the context is already updated at line 16, I'd put this context key in there. This way, the context will be updated once for the whole recordset, instead of being updated singularly on each record.

However, a `RecursionError` looks strange.
Dates are propagated top-bottom using the `stock.move.move_dest_ids` field, which is a `Many2many` field.
I suppose that chaining the moves through this field may result (though it really shouldn't) in a circular structure, eg:
```
>>> self
stock.move(1,)
>>> self.move_dest_ids
stock.move(2,)
>>> self.move_dest_ids.move_dest_ids
stock.move(1,)
```
but when calling `self.date = <some_value>`, the propagation should stop at `stock.move(2,)`, since `stock.move(1,)`'s ID is already included in context's `date_propagation_ids` value (so condition at line 25 should be `True`, date on `stock.move(1,)` should not be updated, and the recursion ends).

Can you explain why this behavior didn't work, and what caused the `RecursionError`?

**EDIT**
I just noticed that `delta` is defined as `0` in case the origin move has no date; if this happens, then `move_dest.date - delta` will result in a `TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'int'`.
Please fix this as well.

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,65 @@
...
+            return chained_dest_moves
+
+        # Add the current move to the chained_dest_moves set
+        chained_dest_moves.add(self.id)
+
+        # Recursively find all dest moves in the down chain
+        for move_dest in self.move_dest_ids:
+            move_dest._get_moves_in_down_chain(chained_dest_moves)
+
+        return chained_dest_moves
```
gurneyalex: Could you try using a recursive SQL query here? 

Also the chained move which are Done or Canceled should not be updated

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,65 @@
...
-    def _propagate_date(self, new_date):
-        """Propagate the date of a move to its destination."""
-        already_propagated_ids = self.env.context.get(
-            "date_propagation_ids", set()
-        ) | set(self.ids)
-        self = self.with_context(date_propagation_ids=already_propagated_ids)
+    def write(self, vals):
+        previous_dates = {
+            move.id: move.date for move in self.with_context(prefetch_fields=False)
+        }
```
gurneyalex: we should compute this only if the condition on line 18 below is true (it is not needed if we are just updating the state, for instance)

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,65 @@
...
-                if move_dest.id in already_propagated_ids:
-                    continue
-                move_dest.date -= delta
-
-    def write(self, vals):
-        # propagate date changes in the stock move chain
-        if "date" in vals:
-            self._propagate_date(vals.get("date"))
-        return super().write(vals)
+                delta = timedelta(seconds=0)
```
gurneyalex: I would just `continue` in this case

stock_picking_propagate_scheduled_date/models/stock_move.py
gurneyalex: And possibly remove the ensure_one() to be able to work on a full recordset and perform a single SQL query.

stock_picking_propagate_scheduled_date/models/stock_move.py
yankinmax: We need previous dates to get the delta, because propagate is called after super call.

stock_picking_propagate_scheduled_date/models/stock_move.py
yankinmax: It was originally proposed by @vvrossem.

stock_picking_propagate_scheduled_date/models/stock_move.py
gurneyalex:
```
query = """
WITH RECURSIVE cte AS
(
SELECT
    r.move_orig_id, r.move_orig_id as from_id, r.move_dest_id as dest_id
FROM
    stock_move_move_rel r
WHERE r.move_orig_id in %s

UNION ALL

SELECT
    cte.move_orig_id,
    r.move_orig_id as from_id,
    r.move_dest_id as dest_id
FROM
    stock_move_move_rel r, cte
WHERE
    r.move_orig_id = cte.dest_id
)

SELECT 
   move_orig_id, 
   array_agg(dest_id)
FROM
   cte
group by move_orig_id
;"""
self.env.cr.execute(query, (tuple(self.ids),)
chained = dict(self.env.cr.fetchall()  # key: start move id, values: list of chained moves
```

stock_picking_propagate_scheduled_date/models/stock_move.py
yankinmax: I got the idea, sorry for the noise.

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
-                if move_dest.id in already_propagated_ids:
-                    continue
-                move_dest.date -= delta
-
     def write(self, vals):
-        # propagate date changes in the stock move chain
-        if "date" in vals:
-            self._propagate_date(vals.get("date"))
-        return super().write(vals)
+        """Override the write method to propagate the date to the destination moves."""
```
SilvioC2C: Do *not* use `"""` when overriding an existing method, it will override the original docstrings.
Use `#` comments instead.

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
+        )
+        if need_propagate:
+            # Get the previous date of the moves
+            previous_dates = {
+                move.id: move.date for move in self.with_context(prefetch_fields=False)
+            }
+        res = super().write(vals)
+        if need_propagate:
+            self._propagate_date_no_recursion(previous_dates, vals.get("date"))
+        return res
```
SilvioC2C:
```suggestion
        if "date" not in vals or self.env.context.get("_propagate_date_no_recursion"):
            return super().write(vals)
        dates = {d["id"]: d["date"] for d in self.read(["date"], load=None)}
        res = super().write(vals)
        self._propagate_date_no_recursion(dates, vals["date"])
        return res
```

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
+
+        self.env.cr.execute(query, (tuple(self.ids),))
+        # key: original move, values: list of destination moves
+        chained = dict(self.env.cr.fetchall())
+        return chained
+
+    def _propagate_date_no_recursion(self, previous_dates, new_date):
+        """Propagate the date of a move to its destination avoiding recursion."""
+        if not self:
+            return True
```
SilvioC2C: No need to return something:
```suggestion
            return
```

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
+                    continue
+                for move_dest in self.browse(dest_moves):
+                    if move_dest.state in ("done", "cancel"):
+                        continue
+                    move_dest.with_context(_propagate_date_no_recursion=True).write(
+                        {
+                            "date": move_dest.date - delta,
+                        }
+                    )
+        return True
```
SilvioC2C: No need to return something:
```suggestion
```

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
+                move_orig_id,
+                array_agg(dest_id)
+            FROM chained_moves
+            GROUP BY move_orig_id;
+        """
+
+        self.env.cr.execute(query, (tuple(self.ids),))
+        # key: original move, values: list of destination moves
+        chained = dict(self.env.cr.fetchall())
+        return chained
```
SilvioC2C:
```suggestion
        return dict(self.env.cr.fetchall())
```

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
+                else:
+                    continue
+                for move_dest in self.browse(dest_moves):
+                    if move_dest.state in ("done", "cancel"):
+                        continue
+                    move_dest.with_context(_propagate_date_no_recursion=True).write(
+                        {
+                            "date": move_dest.date - delta,
+                        }
+                    )
```
SilvioC2C:
```suggestion
                dest_move_ids = chained_dest_moves.get(move.id)
                if not dest_move_ids:
                    continue
                prev_date = previous_dates.get(move.id)
                if not prev_date:
                    continue
                delta = prev_date - fields.Datetime.to_datetime(new_date)
                dest_moves = self.browse(set(dest_move_ids))
                for dest_move in dest_moves.with_context(_propagate_date_no_recursion=1):
                    if dest_move.state not in ("done", "cancel"):
                        dest_move.date = move_dest.date - delta
```

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,80 @@
...
+                move.id: move.date for move in self.with_context(prefetch_fields=False)
+            }
+        res = super().write(vals)
+        if need_propagate:
+            self._propagate_date_no_recursion(previous_dates, vals.get("date"))
+        return res
+
+    def _get_moves_in_down_chain(self):
+        """Get all the destination moves in the down chain of a moves recordset recursively."""
+        query = """
```
SilvioC2C, line 25:
```suggestion
        if not self:
            return {}
        query = """
```

stock_picking_propagate_scheduled_date/models/stock_move.py
yankinmax: This way we don't start to update dates. We never come to `_propagate_date_no_recursion`

stock_picking_propagate_scheduled_date/models/stock_move.py
yankinmax: If we inject context with the recordset we update only first dest move in recordset and get into recursion. So, the correct way is to inject by record.

yankinmax:
Hello @SilvioC2C,
I've checked your suggestions and tried to debug different variants.
I've updated with the variant working without recursion.

stock_picking_propagate_scheduled_date/models/stock_move.py
SilvioC2C: The `if` statement I wrote was incorrect, I've updated it.
This way the code is cleaner and using `self.read()` should make it more performant.

stock_picking_propagate_scheduled_date/models/stock_move.py
SilvioC2C: Having
```
for rec in self:
    rec.with_context(flag=value).do_something()
```
is perfectly equivalent to
```
for rec in self.with_context(flag=value):
    rec.do_something()
```
The only difference is that, in the second case, you update the context on the whole recordset at once, instead of updating the context for each record singularly.

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -2,32 +2,78 @@
...
+                delta = prev_date - fields.Datetime.to_datetime(new_date)
+                dest_moves = self.browse(set(dest_move_ids))
+                for dest_move in dest_moves:
+                    if dest_move.state not in ("done", "cancel"):
+                        dest_move.with_context(_propagate_date_no_recursion=True).write(
+                            {
+                                "date": dest_move.date - delta,
+                            }
+                        )
+        return
```
SilvioC2C: You can remove the `return`

stock_picking_propagate_scheduled_date/models/stock_move.py
yankinmax: I understand that. I'll check again.

yankinmax:
Superseded by:
- [ ] https://github.com/OCA/stock-logistics-workflow/pull/1653

PR #1646, opened 2024-07-04
[15.0][ADD] stock_move_original_date

LoisRForgeFlow:
This module adds the *Original Date Scheduled* to stock moves. This is useful for
past analysis and warehouse performance reports.

In standard Odoo, once a transfer is confirmed with an scheduled date, this date
can still change to reflect changes in the planning (e.g. you receive a
notification from your supplier that goods will arrive a couple of days later).
In such situation you lost precious information of the originally planned date
that can be used to compute the real delay in the receipt from your supplier.


JordiMForgeFlow (approved):
code review LGTM, maybe a field showing the diff between the original and the current scheduled could be already added in the module

LoisRForgeFlow:
> code review LGTM, maybe a field showing the diff between the original and the current scheduled could be already added in the module

I was doubtful with this one. The reporting that can be driven from this field can be computed in different ways or focus in different metrics (delay compared with original date, real LT vs expected LT, you might want to tweak EE report with this or have a complete new report with CE...), therefore as it was not unequivocal I prefered to have the KPI/reporting in a different module and at least have a common column that can be used by anyone defined in this module.

JoanSForgeFlow (approved):

PR #1650, opened 2024-07-09
[14.0][FIX] bug on stock_picking_quick

mathieudelva:


rousseldenis:
@mathieudelva Do you have a test case for this (even the fix is trivial) ?

rousseldenis (approved):

rousseldenis:
@mathieudelva Could you also modify your commit message like `[FIX] stock_picking_quick: Don't pop nonexistent parameter`

PR #1653, opened 2024-07-10
[15.0] [FIX] stock_picking_propagate_scheduled_date: recursion error

ivantodorovich:
Depending on the installed modules, their resolution order (MRO), and the moves relationships through their `move_dest_ids` and `move_orig_ids` fields, we could get into a situation where the very same move that triggered the propagation gets propagated twice (or more) in the same transaction.

This happens because the modified context, which we use to keep track of the already propagated moves, is only available down the stack, and it's lost once we exit the method. However, other modules might also react to the changes and trigger other kinds of propagations, which might lead to our original record being propagated again.

This is the case with the core `mrp_subcontracting` module, which propagates the move's `date` to the related manufacturing order's `date_planned_start` and `date_planned_finished` fields. Which, in turn, triggers the propagation of these two fields to their related `move_raw_ids` and `move_finished_ids`.

It can just so happen that one of these related moves also points to our original move in their `move_dest_ids` or `move_orig_ids` fields chain.

The solution relies on the `_register_hook` method, which gets executed after the model has been initialized, and so it's able to patch the `write` method getting a top position in the stack, and inject the context key there.

---

Superseeds https://github.com/OCA/stock-logistics-workflow/pull/1644

@gurneyalex @SilvioC2C @vvrossem @yankinmax

### DIFF
--- a/stock_picking_propagate_scheduled_date/models/stock_move.py
+++ b/stock_picking_propagate_scheduled_date/models/stock_move.py
@@ -31,7 +31,7 @@ class StockMove(models.Model):
 
     def write(self, vals):
         # propagate date changes in the stock move chain
-        if "date" in vals:
+        if self and "date" in vals:
             self._propagate_date(vals.get("date"))
         return super(StockMove, self).write(vals)
 
--- a/stock_picking_propagate_scheduled_date/models/stock_move.py
+++ b/stock_picking_propagate_scheduled_date/models/stock_move.py
@@ -10,10 +10,13 @@ class StockMove(models.Model):
 
     def _propagate_date(self, new_date):
         """Propagate the date of a move to its destination."""
-        already_propagated_ids = self.env.context.get(
-            "date_propagation_ids", set()
-        ) | set(self.ids)
-        self = self.with_context(date_propagation_ids=already_propagated_ids)
+        # This should never happen (see :meth:`_register_hook`), but still..
+        if "date_propagation_ids" not in self.env.context:
+            self = self.with_context(date_propagation_ids=set())
+        # Update the already propagated ids, to avoid impacting the same move twice
+        already_propagated_ids = self.env.context.get("date_propagation_ids")
+        already_propagated_ids.update(self.ids)  # keep the same variable reference
+        # Propagate the date delta to destination moves
         for move in self:
             if move.date:
                 delta = move.date - fields.Datetime.to_datetime(new_date)
@@ -30,4 +33,20 @@ class StockMove(models.Model):
         # propagate date changes in the stock move chain
         if "date" in vals:
             self._propagate_date(vals.get("date"))
-        return super().write(vals)
+        return super(StockMove, self).write(vals)
+
+    def _register_hook(self):
+        # Patch `write` to initialize `date_propagated_ids` in the context, so that
+        # it's available on the whole stack chain, independent of the MRO.
+        res = super()._register_hook()
+
+        def make_write():
+            def write(self, vals, **kw):
+                if "date_propagation_ids" not in self.env.context:
+                    self = self.with_context(date_propagation_ids=set())
+                return write.origin(self, vals, **kw)
+
+            return write
+
+        self._patch_method("write", make_write())
+        return res
--- a/stock_picking_propagate_scheduled_date/models/__init__.py
+++ b/stock_picking_propagate_scheduled_date/models/__init__.py
@@ -1,1 +1,2 @@
 from . import stock_move
+from . import stock_picking
--- a/stock_picking_propagate_scheduled_date/models/stock_picking.py
+++ b/stock_picking_propagate_scheduled_date/models/stock_picking.py
@@ -1,0 +1,27 @@
+# Copyright 2024 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class StockPicking(models.Model):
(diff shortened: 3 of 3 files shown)


stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -30,4 +33,20 @@ def write(self, vals):
...
+        def make_write():
+            def write(self, vals, **kw):
+                if "date_propagation_ids" not in self.env.context:
+                    self = self.with_context(date_propagation_ids=set())
+                return write.origin(self, vals, **kw)
+
+            return write
+
+        self._patch_method("write", make_write())
+        return res
```
yankinmax, line 52: This is very specific to me.
LG, thanks

yankinmax (approved):

vvrossem (approved):
Next level fix :bow:

SilvioC2C (approved):
LGTM, thanks for the work done

stock_picking_propagate_scheduled_date/models/stock_move.py
```diff
@@ -30,4 +33,20 @@ def write(self, vals):
...
+
+    def _register_hook(self):
+        # Patch `write` to initialize `date_propagated_ids` in the context, so that
+        # it's available on the whole stack chain, independent of the MRO.
+        res = super()._register_hook()
+
+        def make_write():
+            def write(self, vals, **kw):
+                if "date_propagation_ids" not in self.env.context:
+                    self = self.with_context(date_propagation_ids=set())
```
rousseldenis, line 46: I'm still wondering why initializing context with void value ?

ivantodorovich: `_propagate_date` will take care of updating it:

```
        already_propagated_ids = self.env.context.get("date_propagation_ids")
        already_propagated_ids.update(self.ids)  # keep the same variable reference
```

We just need to make sure the variable exists, so that we can get its reference in `already_propagated_ids` by calling `self.env.context.get("date_propagation_ids")` instead of creating a new one

rousseldenis: And why not calling `already_propagated_ids = self.env.context.get("date_propagation_ids", set())`?

ivantodorovich: That's essentially what was done before, but here's the issue: we initialize the `set` on the first `_propagate_date` call, and we're sure it remains in the context and is reused downstream. However, it may happen that some upstream code keeps on making changes to other records, modifying other records dates, and that these changes will trigger yet another `_propagate_date` call, where the context key is no longer there.

This is the case with the `mrp` and `mrp_subcontracting` core modules installed.

The MRO would look something like this:

```
>>> type(self.env['stock.move']).__mro__
(
    ...
    <class 'odoo.addons.mrp_subcontracting.models.stock_move.StockMove'>,
    ...
    <class 'odoo.addons.stock_picking_propagate_scheduled_date.models.stock_move.StockMove'>,
    ...
    <class 'odoo.addons.stock.models.stock_move.StockMove'>,
    <class 'odoo.models.Model'>,
    ...
)
```

So, when the `write` method is executed with a new `date` in vals, the execution order will be the following (focusing only on the relevant modules)

```
mrp_subcontracting --> stock_picking_propagate_scheduled_date
```

1) `mrp_subcontracting` will call `super()`: https://github.com/odoo/odoo/blob/3a680893a1d61a43eb8569ae9a48b7abbd6a86f5/addons/mrp_subcontracting/models/stock_move.py#L72

2) `stock_picking_propagate_scheduled_date` will initialize the context key, and do the propagation ✅ 

3) `mrp_subcontracting` will continue its flow after `super()`. Since this is upstream code, the context key isn't here ⚠️ . It will then propagate the move's `date` to the originating move's MO `date_planned_start` and `date_planned_finished`: https://github.com/odoo/odoo/blob/3a680893a1d61a43eb8569ae9a48b7abbd6a86f5/addons/mrp_subcontracting/models/stock_move.py#L77-L80

4) `mrp` will react to this change, and propagate them to its related `move_raw_ids` and `move_finished_ids`'s `date`: https://github.com/odoo/odoo/blob/5caec324326a197bb219a4d428b0072b543079b9/addons/mrp/models/mrp_production.py#L757-L760

5) This will get us back to the `stock.move` `write` method, without the context key being set. Depending on how these moves are related, the `date` change delta will be propagated again to the same moves that were propagated in the first run. 💀 


The solution is, then, to ensure the context key is in the context upstream in the code.

1) Before calling any module's override, the patched write will be executed first, ensuring the set is initialized in the context key

2) Then, the process continues exactly like before, but this time with the empty `set` already initialized in the context. So the already propagated ids are tracked throughout the whole call stack

gurneyalex:
@ivantodorovich I updated your PR with 2 new commits:

* avoid calling _propagate_date when self is an empty recordset
* better fix in case a backorder is created and mrp_subcontracting is installed

rousseldenis (approved):

vvrossem:
Hello @rousseldenis, do you think this PR could be reopened and merge? :crossed_fingers:

rousseldenis:
@ivantodorovich Could you improve your last commits messages to make module name appear ? Thanks

ivantodorovich:
Done, thanks!

HviorForgeFlow (approved):

xaviedoanhduy (approved):

giarve (approved):
If this gets merged I can migrate the module to 16.0

PR #1656, opened 2024-07-11
[16.0][FIX] stock_picking_group_by_partner_by_carrier: add the same class style as in the base report

RodrigoBM:
This [PR](https://github.com/odoo/odoo/pull/172366) accepts a change that fixes how the information section looks like.

It is necessary to change the views that add changes in this section.


pedrobaeza (approved):
Fast-tracking this for having things aligned:

/ocabot merge minor

PR #1661, opened 2024-07-17
[15.0][ADD] stock_picking_operation_manual_lot_selection

StephaneMangin:
This module allows you to force the manual selection of a Serial Number/lot in the picking operations. This is useful when 
 you want to ensure that the right Serial Number/lot is selected for this given operation.


stock_picking_production_lot_manual_selector/models/stock_move_line.py
```diff
@@ -0,0 +1,38 @@
...
+    _inherit = "stock.move.line"
+
+    use_manual_lots = fields.Boolean(
+        related="picking_id.use_manual_lots",
+    )
+
+    manual_production_lot_id = fields.Many2one(
+        "stock.production.lot",
+        "Lot/Serial Number (manual)",
+        domain="[('product_id', '=', product_id), ('company_id', '=', company_id)]",
```
gurneyalex: Do we need to filter on lots which are in the source stock location  of the stock move related to the stock move line?

stock_picking_production_lot_manual_selector/models/stock_move_line.py
```diff
@@ -0,0 +1,38 @@
...
+        check_company=True,
+        copy=False,
+        help="This field is used to manually select a production lot.",
+    )
+
+    @api.onchange("manual_production_lot_id")
+    def onchange_manual_production_lot_id(self):
+        if self.manual_production_lot_id:
+            self.lot_id = self.manual_production_lot_id
```
gurneyalex: to check: consider a stock move src loc = Stock dest loc = Customers

Lot1 (reserved) is in Stock/ShelveA

Lot2 is in Stock/ShelveB


When switching the lot, is the source location of the stock.move.line updated?

stock_picking_production_lot_manual_selector/models/stock_move_line.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _action_done(self):
+        lines_to_block = self.filtered(
+            lambda line: line.use_manual_lots and not line.manual_production_lot_id
```
gurneyalex: and the done quantity is > 0 

Otherwise it should be allowed to not specify a lot (this will generate a backorder).

stock_picking_production_lot_manual_selector/tests/test_stock_picking_lot.py
```diff
@@ -0,0 +1,78 @@
...
+                "product_id": self.product_serial.id,
+                "product_uom_qty": 1,
+                "product_uom": self.product_serial.uom_id.id,
+            }
+        )
+        picking.action_confirm()
+        with Form(picking) as picking_form:
+            with picking_form.move_line_ids_without_package.new() as line_form:
+                line_form.product_id = move.product_id
+                line_form.qty_done = 1
```
gurneyalex: add a test with qty_done left to 0 -> this should not raise an exception but propose to create a back order.

stock_picking_production_lot_manual_selector/tests/test_stock_picking_lot.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
```
gurneyalex: are you sure?

stock_picking_production_lot_manual_selector/tests/test_stock_picking_lot.py
```diff
@@ -0,0 +1,78 @@
...
+            {
+                "name": "Serial2",
+                "product_id": self.product_serial.id,
+                "company_id": self.env.company.id,
+            }
+        )
+        picking_type = self.env["stock.picking.type"].create(
+            {
+                "name": "Manual Lot Selection",
+                "code": "incoming",
```
gurneyalex: for consistency with the locations, this should be outgoing.

stock_picking_production_lot_manual_selector/tests/test_stock_picking_lot.py
```diff
@@ -0,0 +1,78 @@
...
+                "sequence_code": "IN",
```
gurneyalex:
```suggestion
                "sequence_code": "OUT",
```

stock_picking_production_lot_manual_selector/tests/test_stock_picking_lot.py
```diff
@@ -0,0 +1,78 @@
...
+            with self.assertRaises(UserError):
+                picking._action_done()
+
+            with picking_form.move_line_ids_without_package.edit(0) as line_form:
+                line_form.manual_production_lot_id = serial2
+                self.assertEqual(line_form.lot_id, serial2)
+                self.assertEqual(line_form.manual_production_lot_id, serial2)
+                picking_form.save()
+
+            picking._action_done()
```
gurneyalex: please add another test with the sub locations to make sure that the stock does not get negative (see  [this comment](https://github.com/OCA/stock-logistics-workflow/pull/1661/files#diff-32b1a61db5271cbeec234d6f1cf2c099639a79c8028d289a0fa2f2510e4d22ebR27))

yostashiro:
The objective of the module seems to overlap with https://github.com/OCA/stock-logistics-warehouse/tree/15.0/stock_quant_manual_assign.  Is there something the existing module doesn't cover?

stock_picking_production_lot_manual_selector/models/stock_move_line.py
StephaneMangin: By default, there is a check on locations if there is an incompatibility at confirmation.
```
Serial number (0123456789) is not located in A, but is located in location(s): B.

Please correct this to prevent inconsistent data.
```

stock_picking_production_lot_manual_selector/models/stock_move_line.py
StephaneMangin: The lot_id onchange is preserved by using this onchange, standard behavior applies at this stage.

stock_picking_production_lot_manual_selector/models/stock_move_line.py
StephaneMangin: I presumed the check should be done on the product tracking type to keep standard behavior in that case:
`lambda line: line.use_manual_lots and not line.manual_production_lot_id and line.product_id.tracking != 'none')`

And IMO this is better because it keeps the quantity check later in the workflow:
`You cannot validate a transfer if no quantities are reserved nor done. To force the transfer, switch in edit mode and encode the done quantities.`

stock_picking_production_lot_manual_selector/tests/test_stock_picking_lot.py
StephaneMangin: No ^^

StephaneMangin:
Module renamed to `stock_picking_operation_manual_lot_selection`

StephaneMangin:
> The objective of the module seems to overlap with https://github.com/OCA/stock-logistics-warehouse/tree/15.0/stock_quant_manual_assign. Is there something the existing module doesn't cover?

This doesn't seem to be the same workflow. In this module, we want the user to manually select a lot in the operation view by keeping the same ergonomic workflow and block the picking validation if no selection occurs and should have been.

stock_picking_production_lot_manual_selector/models/stock_move_line.py
StephaneMangin: Adding both tracking type AND qty done

stock_picking_production_lot_manual_selector/models/stock_move_line.py
StephaneMangin: See https://github.com/OCA/stock-logistics-workflow/pull/1661#discussion_r1682245772

stock_picking_production_lot_manual_selector/models/stock_move_line.py
gurneyalex: To check with the customer, often they ask the SN to be filtered so as not to see SN which have already been delivered to a customer when working on a Stock -> Customer  picking.

gurneyalex (approved):

cyrilmanuel (approved):
Not an expert but LGTM

cyrilmanuel (approved):
Not an expert but LGTM

rousseldenis:
@StephaneMangin What is missing in Odoo stock module about this ? In fact, you can use the 'Use Existing Ones' option in stock.picking.type configuration. And then, Odoo won't validate the operation until you choose a lot.

Or am I missing something ?

StephaneMangin:
> @StephaneMangin What is missing in Odoo stock module about this ? In fact, you can use the 'Use Existing Ones' option in stock.picking.type configuration. And then, Odoo won't validate the operation until you choose a lot.
> 
> Or am I missing something ?

Hi @rousseldenis Thanks for your feedback. In fact, when using the option `Use Existing Ones`, Odoo almost blindly take a lot, in case we want the user to select the 'proper' lot number for the 'proper' product he has, we can't use this option efficiently without heavily change the standard behavior. That's why we choose to add a new field replacing the original one with the operation type condition to force the user to select the 'proper' lot manually.

PR #1666, opened 2024-08-02
[IMP] stock_grn: grn type & menu

jbaudoux:
Add menu entry to GRN types in settings
Allow to mark GRN type as inactive

Move "Goods Received Notes" menu under "Operations" instead of "Products" because it's not a master data

cc @rousseldenis


hparfr (approved):

jbaudoux:
FYI @phschmidt

gbrito (approved):

PR #1668, opened 2024-08-07
[14.0] [MIG] stock_picking_operation_quick_change

Ricardoalso:


stock_picking_operation_quick_change/wizards/stock_picking_wizard.py
```diff
@@ -0,0 +1,129 @@
...
+        active_model = self.env.context["active_model"]
+        active_ids = self.env.context["active_ids"] or []
+        picking = self.env[active_model].browse(active_ids)
+        res.update(self._prepare_default_values(picking))
+        return res
+
+    def _default_old_dest_location_id(self):
+        stock_picking_obj = self.env["stock.picking"]
+        pickings = stock_picking_obj.browse(self.env.context["active_ids"])
+        first_move_line = pickings.mapped("move_line_ids")[:1]
```
gbrito: No need to use mapped on relational fields, Odoo does it by default 
```suggestion
        first_move_line = pickings.move_line_ids[:1]
```

stock_picking_operation_quick_change/wizards/stock_picking_wizard.py
```diff
@@ -0,0 +1,129 @@
...
+    change_all = fields.Boolean(
+        string="Change All",
+        help="Check if you want change all operations without filter "
+        "by old location",
+    )
+
+    @api.depends("picking_id")
+    def _compute_allowed_product_ids(self):
+        for record in self:
+            record.allowed_product_ids = record.picking_id.move_lines.mapped(
```
gbrito, line 78: No need to use mapped on relational fields, Odoo does it by default
```suggestion
            record.allowed_product_ids = record.picking_id.move_lines.product_id
```

stock_picking_operation_quick_change/wizards/stock_picking_wizard.py
```diff
@@ -0,0 +1,129 @@
...
+        if forbidden_pickings:
+            raise UserError(
+                _(
+                    "You can not change operations destination location if "
+                    "picking state is not in %s"
+                )
+                % ",".join(self._get_allowed_picking_states())
+            )
+        pickings_with_chained_moves = pickings.filtered(
+            lambda x: x.move_lines.mapped("move_dest_ids")
```
gbrito: No need to use mapped on relational fields, Odoo does it by default
```suggestion
            lambda x: x.move_lines.move_dest_ids
```

stock_picking_operation_quick_change/wizards/stock_picking_wizard.py
```diff
@@ -0,0 +1,129 @@
...
+                    "You cannot change destination location if any of the moves "
+                    "has a destination move."
+                )
+            )
+
+    def action_apply(self):
+        stock_picking_obj = self.env["stock.picking"]
+        pickings = stock_picking_obj.browse(self.env.context["active_ids"])
+        self.check_allowed_pickings(pickings)
+        move_lines = pickings.mapped("move_line_ids")
```
gbrito: No need to use mapped on relational fields, Odoo does it by default
```suggestion
        move_lines = pickings.move_line_ids
```

stock_picking_operation_quick_change/wizards/stock_picking_wizard.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+        if pickings_with_chained_moves:
+            raise UserError(
+                _(
+                    "You cannot change destination location if any of the moves "
+                    "has a destination move."
+                )
+            )
+
+    def action_apply(self):
```
gbrito, line 105:
```suggestion
    def action_apply(self):
        self.ensure_one()
```

gbrito (changes requested):

Ricardoalso:
@gbrito Thanks for you review. Indeed I didn't pay attention to that during the migration, my bad.  New commit added :wink:

gbrito (approved):

FrancescoBallerini (approved):
Hello, I've tested this locally as the runbot is off. 
I think this can be reviewed as functionally ok but I do not understand why by applying the wizard with "CHANGE ALL" and setup a new `location_dest_id` , the value is correctly retrieved on wizard and business logic (e.g. when validating the picking the "done" quant is correctly moved to the location) but we still see the picking location dest ID on the stock move lines.. 

I think I can approve but I'd like to understand why this happens, if possible, thanks ! : )

florian-dacosta:
> Hello, I've tested this locally as the runbot is off. I think this can be reviewed as functionally ok but I do not understand why by applying the wizard with "CHANGE ALL" and setup a new `location_dest_id` , the value is correctly retrieved on wizard and business logic (e.g. when validating the picking the "done" quant is correctly moved to the location) but we still see the picking location dest ID on the stock move lines..
> 
> I think I can approve but I'd like to understand why this happens, if possible, thanks ! : )

@FrancescoBallerini There are 2 different notions actually : 
The stock moves (`stock.move`) which indicate a quantity of product to move from a zone to another => The same zones that are displayed at the picking level
The operations (`stock.move.line`). Depending on the configuration of your Odoo, you do not always see the operations displayed in a tab of the picking, but they exists anyway.
The goal of the operation is to indicate precisely what you move (which lot, which pack ? and which sublocation)

This module only change the destination location of the operations, that is why it does not change anything on the picking.

For instance, you have a picking with one line only with product quantity 10 and going from "supplier" to "WH/Stock".
You could fill the operations this way : 5 products are going to WH/stock/shelf1 and 5 products are going to WH/stock/shelf2
This way, only the destination location of both operation is changed, while the "zone", on the stock move is still "WH/stock".
This is how it happens, even without this module installed.
Hope it is clear

florian-dacosta (approved):

FrancescoBallerini:
@florian-dacosta I've missed your ping [here ](https://github.com/OCA/stock-logistics-workflow/pull/1232#issuecomment-2447560455) , glad it was merged aniway!

Thank you for the detailed explanation. I received similar insights about these models recently, and both were helpful.

Just a consideration: I think that if a user is going to change the "deepest" layer (`stock.move.line`) , also the highest level layer `stock.move` should reflect the change. Additionally I guess also the picking `location_dest_id` could change at this point if that's true. Does it make sense or there are some reason why it shouldn't change? 

I'm not really asking for implementation, it's just for my knowledge about this module since I'm considering to make additional customizations for a local deploy

Thank you again

PR #1669, opened 2024-08-08
[15.0][IMP] product_cost_price_avco_sync: Don't apply qty changes to landed costs

CarlosRoca13:
When a change is made to the quantities of a stock move that is associated with multiple stock valuation layers, including those related to landed costs, the quantity change is being applied to all the lines instead of only to the one related to the material entry.

With these changes, we restrict the quantity change solely to the stock valuation layer related to the material entry, preventing the change from being applied to the lines associated with landed costs.

cc @Tecnativa TT50028

ping @pedrobaeza @chienandalu


pedrobaeza (approved):
The TODO comment was referring to the landed cost thing.

chienandalu (approved):

PR #1557, opened 2024-03-29
[16.0][FIX] stock_picking_line_sequence

AungKoKoLin1997:
This PR is to follow up the updating behavior of the odoo standard done by this PR(https://github.com/odoo/odoo/pull/144276).

@qrtl

PR #1676, opened 2024-08-20
[16.0][MIG] stock_valuation_fifo_lot: Migration to 16.0

AungKoKoLin1997:
Superseding of https://github.com/OCA/stock-logistics-workflow/pull/1527

In this PR, I address the following points:

- Standard migration
- Improved code by using the Odoo standard function _get_fifo_candidates() and adjustments to _get_price_unit()
- Fixed the issue proposed in https://github.com/OCA/stock-logistics-workflow/pull/1350#issuecomment-1840407086
- Avoid completely overriding `_run_fifo()` and use the standard behavior.
- Update the product's standard price to the first available lot price.
- Improve `_get_price_unit()` to retrieve the value from the most recent incoming move line for the lot for the No PO (e.g. customer returns) stock moves.
- Add quantity and value-related fields in `stock_move_line` to cover all cases including multiple lots exist in an SVL record, and it is unclear which lot's remaining quantity is being delivered(Eg. Receive serials `001`, `002` and `003` for an incoming move. Deliver `002`, return `002` and deliver `002` again.)
- Add `force_fifo_lot_id` in `stock_move_line` to handle cases where a delivery needs to be made for an existing lot created before the installation of this module (e.g., lots `001` and `002` were received, lot `002` was delivered before the module's installation, and lot `001` is delivered after the module is installed).
- Add a `post_init_hook` to update the `lot_ids` in SVL for existing records and to update the field values in existing stock_move_line records.

Depends on https://github.com/odoo/odoo/pull/180245
@qrtl QT4650


AungKoKoLin1997:
Test failed from `stock_picking_line_sequence` is fixed at https://github.com/OCA/stock-logistics-workflow/pull/1557.

AungKoKoLin1997:
@matt454357 
Can you please review this PR and bring your ideas from your module https://github.com/OCA/stock-logistics-warehouse/pull/2139 as we talked?

matt454357:
I started a review, and will post some suggestions.

stock_valuation_fifo_lot/tests/test_stock_valuation_fifo_lot.py
```diff
@@ -0,0 +1,142 @@
...
+        return_picking_wizard = return_picking_wizard_form.save()
+        return_picking_wizard.product_return_moves.write({"quantity": 5})
+        return_picking_wizard_action = return_picking_wizard.create_returns()
+        return_picking = self.env["stock.picking"].browse(
+            return_picking_wizard_action["res_id"]
+        )
+        return_move = return_picking.move_ids
+        return_move.move_line_ids.qty_done = 5
+        return_picking.button_validate()
+        self.assertEqual(abs(return_move.stock_valuation_layer_ids.value), 1500)
```
matt454357:
```suggestion
        self.assertEqual(abs(return_move.stock_valuation_layer_ids.value), 1500)

    def test_customer_partial_return(self):
        """Verify partial customer return of delivery with multiple lots of the
        same product"""
        receipt_picking_1 = self.create_picking(
            self.supplier_location, self.stock_location, self.picking_type_in
        )
        move = self.create_stock_move(receipt_picking_1, self.product, 100)
        self.create_stock_move_line(move, "11111")

        receipt_picking_1.action_confirm()
        receipt_picking_1.action_assign()
        receipt_picking_1._action_done()

        receipt_picking_2 = self.create_picking(
            self.supplier_location, self.stock_location, self.picking_type_in
        )
        move = self.create_stock_move(receipt_picking_2, self.product, 200)
        self.create_stock_move_line(move, "22222")

        receipt_picking_2.action_confirm()
        receipt_picking_2.action_assign()
        receipt_picking_2._action_done()

        delivery_picking_2 = self.create_picking(
            self.stock_location, self.customer_location, self.picking_type_out
        )
        move = self.create_stock_move(delivery_picking_2, self.product)
        move_line = self.create_stock_move_line(move)
        lot_id = self.env["stock.lot"].search(
            [("name", "=", "11111"), ("product_id", "=", self.product.id)]
        )
        move_line.write({"lot_id": lot_id})
        move_line = self.create_stock_move_line(move)
        lot_id = self.env["stock.lot"].search(
            [("name", "=", "22222"), ("product_id", "=", self.product.id)]
        )
        move_line.write({"lot_id": lot_id})

        delivery_picking_2.action_confirm()
        delivery_picking_2.action_assign()
        delivery_picking_2._action_done()

        return_picking_wizard_form = Form(
            self.env["stock.return.picking"].with_context(
                active_ids=delivery_picking_2.ids,
                active_id=delivery_picking_2.id,
                active_model="stock.picking",
            )
        )
        return_picking_wizard = return_picking_wizard_form.save()
        return_picking_wizard.product_return_moves.write({"quantity": 5})
        return_picking_wizard_action = return_picking_wizard.create_returns()
        return_picking = self.env["stock.picking"].browse(
            return_picking_wizard_action["res_id"]
        )
        return_move = return_picking.move_ids
        return_move.move_line_ids.lot_id = lot_id
        return_move.move_line_ids.qty_done = 5
        return_picking.button_validate()
        self.assertEqual(abs(return_move.stock_valuation_layer_ids.value), 1000)

    def test_revaluation_layer(self):
        """Verify revaluation layers are included in unit price calculation"""
        receipt_picking_1 = self.create_picking(
            self.supplier_location, self.stock_location, self.picking_type_in
        )
        move = self.create_stock_move(receipt_picking_1, self.product, 1000)
        self.create_stock_move_line(move, "11111")

        receipt_picking_1.action_confirm()
        receipt_picking_1.action_assign()
        receipt_picking_1._action_done()

        receipt_picking_2 = self.create_picking(
            self.supplier_location, self.stock_location, self.picking_type_in
        )
        move = self.create_stock_move(receipt_picking_2, self.product, 2000)
        self.create_stock_move_line(move, "22222")

        receipt_picking_2.action_confirm()
        receipt_picking_2.action_assign()
        receipt_picking_2._action_done()

        wiz = (
            self.env["stock.valuation.layer.revaluation"]
            .sudo()
            .create(
                {
                    "company_id": self.company.id,
                    "product_id": self.product.id,
                    "added_value": -500.0,
                    "account_id": self.cost_account.id,
                    "reason": "test serial revaluation",
                }
            )
        )
        wiz.action_validate_revaluation()

        delivery_picking = self.create_picking(
            self.stock_location, self.customer_location, self.picking_type_out
        )
        move = self.create_stock_move(delivery_picking, self.product)
        move_line = self.create_stock_move_line(move)
        lot_id = self.env["stock.lot"].search(
            [("name", "=", "11111"), ("product_id", "=", self.product.id)]
        )
        move_line.write({"lot_id": lot_id})

        delivery_picking.action_confirm()
        delivery_picking.action_assign()
        delivery_picking._action_done()

        self.assertEqual(abs(move.stock_valuation_layer_ids.value), 4500)
```

AungKoKoLin1997:
@matt454357 
What I am thinking is that I will handle adding the option for a single lot/serial per SVL record or multiple lots/serials per SVL record in this PR (since some users may not need a single lot/serial per SVL).

After you review and the PR is merged, you can create a new PR to follow up on your idea.
Or would you prefer to include your idea as a new commit in this PR by creating a PR in our repository?

What do you think?

AungKoKoLin1997:
> I will handle adding the option for a single lot/serial per SVL record or multiple lots/serials per SVL record in this PR (since some users may not need a single lot/serial per SVL).

@matt454357 I will not add this enhancement in the current PR because it is not related to migration, and there is no use case with the current design. So, could you please handle this when you work on revaluation?

stock_valuation_fifo_lot/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Stock Valuation Fifo Lot",
+    "version": "16.0.1.0.0",
+    "category": "Warehouse Management",
+    "development_status": "Alpha",
+    "license": "AGPL-3",
+    "author": "Ecosoft, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "depends": ["stock_account", "stock_no_negative", "mrp_landed_costs"],
```
matt454357: I'm not sure we should depend on `mrp_landed_costs`.  Could that be handled in a separate module?

stock_valuation_fifo_lot/models/stock_landed_cost.py
matt454357, line 1: Maybe move to a separate module?

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _get_price_unit(self):
+        """
+        No PO, Get price unit from lot price
+        """
+        self.ensure_one()
+        if not self.company_id.use_lot_get_price_unit_fifo:
+            return super()._get_price_unit()
+        if (
+            not self.purchase_line_id
```
matt454357: Depends on `purchase` module, which I guess is okay.

matt454357:
I think it's a good idea to finish this migration PR, and then do a follow up PR.  And, I'm happy to do the follow up PR.  However, there a couple of problems with allowing multiple lots/serials per SVL.

Maybe this initial migration should prevent revaluation of tracked products?

Maybe also raise an error when `_get_price_unit()` is called with multiple `lot_ids`?

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,97 @@
...
+            and self.product_id.cost_method == "fifo"
+            and len(self.lot_ids) == 1
```
matt454357: Should we raise an error when `len(self.lot_ids) > 1`?

stock_valuation_fifo_lot/tests/test_stock_valuation_fifo_lot.py
matt454357: These suggested tests fail because we are not enforcing single lot/serial per SVL.

matt454357:
The believe PR https://github.com/OCA/stock-logistics-workflow/pull/1677 supports the idea of enforcing single lot/serial per SVL.

matt454357 (commented):
I'm just commenting on this PR, not suggesting that we make these changes immediately.

AungKoKoLin1997:
> I think it's a good idea to finish this migration PR, and then do a follow up PR. And, I'm happy to do the follow up PR. However, there a couple of problems with allowing multiple lots/serials per SVL.

Thanks for the pointer and your interest in following up. What do you think of my idea of adding an option to choose a single or multiple lot/serial per SVL? As I mentioned, I don't want to make this adjustment in the migration PR because it would make the module bigger, and the community might be less interested in reviewing it. So, my idea is to get this PR merged quickly and then follow up with the changes.

stock_valuation_fifo_lot/__manifest__.py
AungKoKoLin1997: It was exist since old version. So, I think we don't need to handle in this PR.

stock_valuation_fifo_lot/models/stock_landed_cost.py
AungKoKoLin1997: Same as above. May be later.

AungKoKoLin1997: Yes. I will handle it.

matt454357:
Yes, keep this PR small.

Regarding the option to choose between multiple or single lot per SVL: I'm not sure why someone would want multiple, but it sounds like you have a use case in mind. So, I suppose it's fine, as long as we raise errors for the issues described above.

stock_valuation_fifo_lot/models/stock_move.py
AungKoKoLin1997: I added `hasattr(self, "purchase_line_id")` in the conditon. So, there is no issue I guess.

stock_valuation_fifo_lot/models/stock_move.py
AungKoKoLin1997: Do you have any concerns or inconveniences if we use Odoo's standard behavior for this case?

AungKoKoLin1997:
@TheerayutEncoder 
I have separated the module for passing `lot_producing_id` of `mrp_production_ids` to the SVL when adding the landed cost for these MOs, as you proposed in this [commit](https://github.com/OCA/stock-logistics-workflow/pull/1350/commits/acf76aa557da10452affad47f6f40e4489467a7f).
What do you think about it?
Additionally, I am not entirely sure of the purpose of this, since the `remaining_qty` of the SVL for landed costs is 0.0, and I believe these SVLs will not be considered when `_run_fifo(`) is called, if I understand correctly. Is this intended for user experience, to see the related `lot_ids` in the SVL? Could you please explain the initial purpose?

AungKoKoLin1997:
> What do you think of my idea of adding an option to choose a single or multiple lot/serial per SVL?

@matt454357 Please let me handle this part and include it in this PR because we urgently need this module, and I don't know when this PR will be merged.

matt454357:
@AungKoKoLin1997 any contribution you make is very much appreciated,  Do it the way you feel is best, and I will work with it.

matt454357 (approved):
Looks good to me

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,98 @@
...
+                .sudo()
+                .search(
+                    [
+                        ("product_id", "=", self.product_id.id),
+                        (
+                            "lot_ids",
+                            "in",
+                            self.lot_ids.ids,
+                        ),
+                        ("remaining_qty", ">", 0),
```
yostashiro: I think we misunderstood the usage of this logic. I now think this doesn't work. The original logic isn't optimal either. I'm thinking of a different implementation...

stock_valuation_fifo_lot/models/stock_move.py
yostashiro: > Should we raise an error when `len(self.lot_ids) > 1`?

I think we could make it configurable if the community requires it. I don't have a strong need for that at the moment.

matt454357: Okay, single lot per SVL would not work when migrating/installing with existing data.  So, I don't think we can do that.

matt454357:
We will also need https://github.com/odoo/odoo/pull/179447, in order to do revaluation (later).

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,122 @@
...
+        if forced_quantity and correction_ml:
+            correction_ml.qty_base += forced_quantity
+            return super()._create_in_svl(forced_quantity=forced_quantity)
+        layers = self.env["stock.valuation.layer"]
+        for move in self:
+            layer = super(StockMove, move)._create_in_svl(
+                forced_quantity=forced_quantity
+            )
+            layers |= layer
+            # Calculate standard price (sorted by lot created date)
```
yostashiro: This comment is in the wrong position.
```suggestion
```

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,122 @@
...
+            product = move.product_id
+            if product.cost_method != "fifo" or product.tracking == "none":
+                continue
+            for ml in layer.stock_move_id.move_line_ids:
+                ml.qty_base = ml.qty_done
+            # Change product standard price to the first available lot price.
```
yostashiro:
```suggestion
            # Update product standard price to the first available lot price (by sorting by lot create date).
```

stock_valuation_fifo_lot/models/stock_move.py
```diff
@@ -0,0 +1,122 @@
...
+            product = product.with_context(sort_by="lot_create_date")
+            candidate = product._get_fifo_candidates(move.company_id)[:1]
+            if not candidate:
+                continue
+            product = product.with_company(move.company_id.id)
+            product = product.with_context(disable_auto_svl=True)
+            product.sudo().standard_price = candidate.unit_cost
```
yostashiro:
```suggestion
            # `candidate.unit_cost` is not totally accurate in 16.0 because of https://github.com/odoo/odoo/issues/171464
            product.sudo().standard_price = candidate.remaining_value / candidate.remaining_qty
```

yostashiro: I'm actually not clear about the business rationale why we need this part.

@ps-tubtim @Saran440 Could you please explain what business needs you had to have this logic? In my view, product standard price doesn't mean much in FIFO (it doesn't affect the transaction except for the case where inventory is created with a new lot with no purchase order) and this logic can be removed.

stock_valuation_fifo_lot/tests/test_stock_valuation_fifo_lot.py
```diff
@@ -0,0 +1,474 @@
...
+        )
+
+        return_move = self.transfer_return(receipt_picking_3, 5.0)
+        self.assertEqual(
+            abs(return_move.stock_valuation_layer_ids.value),
+            1500.0,
+            "Stock valuation after return should be 1500.0",
+        )
+
+    def test_receive_deliver_return_lot(self):
```
yostashiro: I didn't quite get the point of this test. Does it prove anything when all lots have the same cost?

stock_valuation_fifo_lot/tests/test_stock_valuation_fifo_lot.py
```diff
@@ -0,0 +1,474 @@
...
+            is_receipt=False,
+        )
+        self.assertEqual(
+            abs(move_out_2.stock_valuation_layer_ids.value),
+            500.0,
+            "Stock valuation for second delivery of lot 002 should be 500.0",
+        )
+        self.assertEqual(move_in_2.stock_valuation_layer_ids.remaining_qty, 0.0)
+
+    def test_cost_tracking_by_lot(self):
```
yostashiro: This could be combined with the first test? It looks redundant...

stock_valuation_fifo_lot/models/stock_move.py
kittiu: @yostashiro https://github.com/OCA/stock-logistics-workflow/blob/15.0/stock_valuation_fifo_lot/models/stock_move.py
As I check the same file in version 15, I don't see that part of code.

yostashiro: @kittiu Sorry, we made it hard to track. It's this part: https://github.com/OCA/stock-logistics-workflow/blob/a87e986/stock_valuation_fifo_lot/models/stock_move.py#L47-L62

stock_valuation_fifo_lot/tests/test_stock_valuation_fifo_lot.py
AungKoKoLin1997: This was for the case like we receive serials `001`, `002` and `003` for an incoming move.Then, Deliver `002`, return `002` and deliver `002` again. In this case, we want to check the `remaining_qty` was taken from return move's SVL for latest delivery instead of first incoming move's SVL because there remaining qty of first incoming move's SVL is only for both `001` and `002` serials now.

stock_valuation_fifo_lot/models/stock_move.py
kittiu: I remember we have a case where it will be incorrect without this fix. I can't remember the case. @newtratip do you remember the case or any where issue is open?

newtratip: @yostashiro
As far as I remember, this part will affect the case where you return items to lot in the warehouse.

**Example**

1. Receive goods (WH/IN) with lot001 10 unit, price/unit = $10 => Standard price = $10 (Price from lot001)
2. Receive goods (WH/IN) with lot002 5 unit, price/unit = $15 => Standard price = $10 (Price from lot001)
3. Issue goods (WH/OUT) 12 unit => Standard price = $15 (Price from lot002)
4. Return goods (WH/RET) lot001 3 unit => Standard price = $10 (Price from lot001)

This part of code will check when incoming goods to warehouse, the system will pull price from first lot always.

AungKoKoLin1997: @newtratip Thanks for the explanation. I understand your logic, but I don't fully understand what happens if we don't update the standard price in this scenario. I assume that we don't care about the standard price for FIFO products. Could you please explain the problem if we don't update the standard price in this case?

AungKoKoLin1997: @kittiu @newtratip I removed this part as I don’t see any business need for it, and I don't think we need to be concerned about the standard price for FIFO products. Please let me know if I’m missing something regarding the business or operations.

matt454357: @AungKoKoLin1997 I believe the standard price is used in at least 2 places, even for FIFO products:

- Sales order line cost and margin
- Inventory adjustment / new quant

https://github.com/odoo/odoo/blob/bc7b2154cbaf49c33af4452f78300cb791e2e3db/addons/sale_margin/models/sale_order_line.py#L27

AungKoKoLin1997: @matt454357 Thanks for the pointer. I think this part is useful in cases where the first lot of a product is delivered in full and then some quantity is returned. In this situation, the standard price is updated to match the price of the first lot.

So, I think using the standard price of the first lot for the Sales Order line cost doesn't have an impact. Am I misunderstanding?

For inventory adjustments or new quants, adjustments for existing inventory will use the lot price of that inventory with the current design. When creating a new quant, which means a new lot, the standard price of the first or any other lot isn’t necessarily relevant, as I wouldn’t expect it to use the same standard price. So, user should check the standard price before they create new quant.

ValentinVinagre:
@AungKoKoLin1997 @yostashiro Is this PR dead? 🦴

yostashiro (commented):
@AungKoKoLin1997 Can you please squash your refactoring commits and assign a (big) commit message?

yostashiro:
> @AungKoKoLin1997 @yostashiro Is this PR dead? 🦴

@ValentinVinagre It's alive and you are more than welcome to review it. 😁

ValentinVinagre:
> > @AungKoKoLin1997 @yostashiro Is this PR dead? 🦴
> 
> @ValentinVinagre It's alive and you are more than welcome to review it. 😁

Perfect, I'll sign up to review it 👍🏻  . I thought it was at a standstill, so I wanted to know the status.

stock_valuation_fifo_lot/tests/test_stock_valuation_fifo_lot.py
yostashiro: True. I think we should mention this limitation in the ROADMAP file for now and follow up on this if the need arises. The workaround is to deduct the stock of the target lot and re-create the stock with inventory adjustment upon adjusting the product cost.

@AungKoKoLin1997 Can you please follow up on this point?

stock_valuation_fifo_lot/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,56 @@
...
+    - 'Qty Remaining' [*]: Remaining quantity (the total by product should match that
+      of the inventory valuation). In product UoM.
+    - 'Value Remaining' [*]: Remaining value (the total by product should match that
+      of the inventory valuation).
+    - 'Force FIFO Lot/Serial': Used when you are stuck by not being able to find a FIFO
+      balance for the lot in an outgoing move line.
+ 
+ .. [*] Updated only for products with FIFO costing method only, for valued incoming
+        moves, and outgoing moves where the qty_done has been reduced in the completed
+        state.
```
yostashiro:
```suggestion
 .. [*] Updated only for products with FIFO costing method, for valued incoming
        moves, and for outgoing moves where the qty_done has been reduced in the completed
        state.
```

yostashiro (commented):
@AungKoKoLin1997 Can you please add ROADMAP.rst with the following content:

The module currently doesn't support product revaluation for specific lots.
If such a need arises, a workaround is to deduct the stock of the target lot and recreate it using an inventory adjustment after modifying the product cost.

AungKoKoLin1997:
I squashed the commits.

yostashiro (approved):
Used in production now.

kanda999 (approved):
Functional and code review: this module works in production DB

kanda999 (commented):
Sorry I didn't review latest code. I couldn't approve this yet.

kanda999 (approved):
Functional review

stock_valuation_fifo_lot/models/product.py
```diff
@@ -0,0 +1,111 @@
...
+                if candidate.lot_ids:
+                    return min(candidate.lot_ids.mapped("create_date"))
+                return candidate.create_date
+
+            all_candidates = all_candidates.sorted(key=sorting_key)
+        elif sort_by is not None:
+            all_candidates = self._sort_by_all_candidates(all_candidates, sort_by)
+        return all_candidates
+
+    # Depends on https://github.com/odoo/odoo/pull/180245
```
yostashiro: Let's remove this comment.

stock_valuation_fifo_lot/models/stock_valuation_layer.py
```diff
@@ -11,9 +11,41 @@ class StockValuationLayer(models.Model):
         comodel_name="stock.lot",
         string="Lots/Serials",
     )
+    is_lot_revaluation = fields.Boolean()
```
yostashiro: Please add help text.

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+                )
+            else:
+                reval.new_value_by_qty = 0.0
+        return super(
+            StockValuationLayerRevaluation, self - revaluations
+        )._compute_new_value()
+
+    def action_validate_revaluation(self):
+        self.ensure_one()
+        if self.lot_id:
```
yostashiro: Better to flip the condition to reduce nesting, given the added logic is quite long.

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+            # Overrides standard behavior to store the remaining value in a new SVL rather than
+            #  modifying the existing one.
```
yostashiro:
```suggestion
            # modifying the existing one.
```

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+            if self.currency_id.is_zero(self.added_value):
+                raise UserError(
+                    _("The added value doesn't have any impact on the stock valuation")
+                )
+            product_id = self.product_id.with_company(self.company_id)
```
yostashiro:
```suggestion
            product = self.product_id.with_company(self.company_id)
```

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+    def _search_svl_move_lines(self, lot_id):
+        domain = [
+            ("product_id", "=", self.product_id.id),
+            ("company_id", "=", self.env.company.id),
+            ("lot_ids", "in", lot_id.ids),
+        ]
+        to_date = self.env.context.get("to_date")
+        if to_date:
+            to_date = fields.Datetime.to_datetime(to_date)
+            domain.append(("create_date", "<=", to_date))
```
yostashiro: What do we need this part for?

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+    current_quantity_lot_svl = fields.Float(
+        "Current Quantity", compute="_compute_current_value_lot_svl"
+    )
+    lot_id = fields.Many2one(
+        "stock.lot",
+        "Lot/Serial Number",
+    )
+    display_lot_id = fields.Boolean(compute="_compute_display_lot_id")
+
+    def _search_svl_move_lines(self, lot_id):
```
yostashiro: Wouldn't it be more efficient to directly search `stock.move.line` instead of getting records via SVL?

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+
+from odoo import Command, _, api, fields, models
+from odoo.exceptions import UserError
+from odoo.tools import float_compare, float_is_zero
+
+
+class StockValuationLayerRevaluation(models.TransientModel):
+    _inherit = "stock.valuation.layer.revaluation"
+
+    current_value_lot_svl = fields.Float(
```
yostashiro: I think we can remove the `_svl` part as we are not looking at SVL records.
```suggestion
    current_value_lot = fields.Float(
```

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+                description = _("Manual Stock Valuation: %s.", self.reason)
+            else:
+                description = _("Manual Stock Valuation: No Reason Given.")
+            if product_id.categ_id.property_cost_method == "average":
+                description += _(
+                    " Product cost updated from %(previous)s to %(new_cost)s.",
+                    previous=product_id.standard_price,
+                    new_cost=product_id.standard_price
+                    + self.added_value / self.current_quantity_lot_svl,
+                )
```
yostashiro: This case should be out of scope.
```suggestion
```

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+                "remaining_value": self.added_value,
+                "quantity": 0,
+            }
+            remaining_qty = sum(remaining_svls.mapped("remaining_qty"))
+            remaining_value = self.added_value
+            remaining_value_unit_cost = self.currency_id.round(
+                remaining_value / remaining_qty
+            )
+            svls = self.env["stock.valuation.layer"]
+            for svl in remaining_svls:
```
yostashiro: Just to improve the readability.
```suggestion
            rounding = product.uom_id.rounding
            for svl in remaining_svls:
```

yostashiro (commented):
Partial review on the revaluation feature.

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+        return super(
+            StockValuationLayerRevaluation, self - revaluations
+        )._compute_new_value()
+
+    def action_validate_revaluation(self):
+        self.ensure_one()
+        if self.lot_id:
+            # Overrides standard behavior to store the remaining value in a new SVL rather than
+            #  modifying the existing one.
+            if self.currency_id.is_zero(self.added_value):
```
yostashiro: To add a shortcut.
```suggestion
            svl_model = self.env["stock.valuation.layer"]
            if self.currency_id.is_zero(self.added_value):
```

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+                [
+                    ("product_id", "=", product_id.id),
+                    ("remaining_qty", ">", 0),
+                    ("company_id", "=", self.company_id.id),
+                    ("lot_ids", "in", self.lot_id.id),
+                ]
+            )
+            remaining_svls = remaining_svls.filtered(
+                lambda l: l._get_unconsumed_in_move_line(self.lot_id)
+            )
```
yostashiro: I think we need to switch to using the `stock.move.line` records rather than SVLs when it comes to identifying the remaining qty and value. The current implementation doesn't block you from devaluing more than the selected lot contains.

stock_valuation_fifo_lot/wizard/stock_valuation_layer_revaluation.py
```diff
@@ -0,0 +1,244 @@
...
+            previous_value_svl = self.current_value_lot_svl
+            revaluation_svl = self.env["stock.valuation.layer"].create(
+                revaluation_svl_vals
+            )
+            # Assign value after revaluation layer is created to trigger compute method
+            # from stock move line
+            svls.write({"has_lot_revaluation": True})
+            if self.property_valuation != "real_time":
+                return True
+            accounts = product_id.product_tmpl_id.get_product_accounts()
```
yostashiro: Please cut the following part into a separate method for better readability.

yostashiro (changes requested):

yostashiro (approved):
The module now handles revaluation for specific lots.

kanda999 (approved):
Functional review: it works as expected

PR #1678, opened 2024-08-21
[16.0][IMP] stock_split_picking: assign created moves only if reservation_method is 'at_confirm'

AnizR:
This is a small improvement:

When splitting a picking that have a picking type with a reservation method different than 'at confirm', I except my reservation to not be done.

### DIFF
--- a/stock_split_picking/models/stock_picking.py
+++ b/stock_split_picking/models/stock_picking.py
@@ -61,7 +61,6 @@ class StockPicking(models.Model):
                         new_move = self.env["stock.move"].create(new_move_vals)
                     else:
                         new_move = move
-                    new_move._action_confirm(merge=False)
                     new_moves |= new_move
 
             # If we have new moves to move, create the backorder picking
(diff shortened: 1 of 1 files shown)


IT-Ideas (approved):
LGTM, Thanks!

stock_split_picking/models/stock_picking.py
```diff
@@ -71,7 +71,8 @@ def split_process(self):
                 new_moves.mapped("move_line_ids").write(
                     {"picking_id": backorder_picking.id}
                 )
-                new_moves._action_assign()
+                if backorder_picking.picking_type_id.reservation_method == "at_confirm":
```
rousseldenis: @AnizR I'm not sure either it's sufficent either it's not the good approach.

Looking at this, I think there should be additionnal filters:

https://github.com/odoo/odoo/blob/16.0/addons/stock/models/stock_move.py#L1409

But, shouldn't `_action_confirm()`be called at this moment (and not above - line 64) ? So, the base code is called and you don't bother about reservation method or...

AnizR: In deed, moving the call to `_action_confirm` will do the trick! Thanks

rousseldenis (approved):
Code review

jbaudoux (approved):

PR #1680, opened 2024-08-26
[16.0][ADD] stock_picking_status_notification: Add new module to notify user after change picking states

GabbasovDinar:
This module allows to send notifications to selected backend users when the state of stock picking changes . For this configurable notification rules are used. Rules are checked using flexible patterns. The first one rule matching the pattern is triggered.

(This module depends on PR: [OCA/web#2923](https://github.com/OCA/web/pull/2923))

Task: 3874


rvalyi:
Just an observation: some modules similar modules with mail templates like edi_notification or account_bank_payment_notification use the notification suffix instead of notify. In the OCA the notify suffix has been used in web_notify for server push. So I wonder if it might not be better to user notification here for consistency. I don't have a strong opinion about this; it's just an observation.

cc @pedrobaeza @sbidoul

pedrobaeza:
About the formal aspect, the tag of the commit should be `[ADD]`, not `[NEW]`. About the module name and feature, `stock_picking_state_tracking` can be more appropriated if the implementation is with a mail subtype and directly subscribers of the thread receive the notifications using standard options. You can put that subtype inheritable in picking types.

ivs-cetmix:
> About the formal aspect, the tag of the commit should be `[ADD]`, not `[NEW]`. About the module name and feature, `stock_picking_state_tracking` can be more appropriated if the implementation is with a mail subtype and directly subscribers of the thread receive the notifications using standard options. You can put that subtype inheritable in picking types.

Hello @pedrobaeza thank you for your feedback!
Actually this module has nothing to do with the standard tracking mechanism. Moreover people who will get notified can be even not the followers of the document they are notified about.
I have put the general idea into the [CONTEXT](https://github.com/cetmix/oca-stock-logistics-workflow/blob/16.0-t3874-stock_notify_picking-new-module/stock_picking_notify/readme/CONTEXT.md) section, here is some further elaboration:
We have a warehouse with around 30 people working there. Odoo is installed on the laptop which is located on a table in the middle. So once a new SO is completed in the shop someone should take care of it. 
Actually we are going to use sound notification because people are not just sitting at the table and waiting. 
Here is a PR for that: https://github.com/OCA/web/pull/2923

pedrobaeza:
Well, I think that should be a pro-active action, checking a list of "completed orders" more than a reactive one, having a notification per order (even with sound), but anyone takes the approach they consider...

`stock_picking_status_notification` then for the current approach of the module.

ivs-cetmix:
> Well, I think that should be a pro-active action, checking a list of "completed orders" more than a reactive one, having a notification per order (even with sound), but anyone takes the approach they consider...
> 
> `stock_picking_status_notification` then for the current approach of the module.

This is not the case here. This will require someone stop his work, go to the laptop and check the orders any eg 10 minutes.
Which will lead to the excessive time loss. Because sometimes there can be one order in 2-3 minutes, and sometimes there can be no orders for an hour.

pedrobaeza:
I see something weird in that flows, but I'm not blocking anyway. I don't see normal that someone has to stop their work to do "something" on that case. Automate that "something" or batch it somehow.

ivs-cetmix:
> I see something weird in that flows, but I'm not blocking anyway. I don't see normal that someone has to stop their work to do "something" on that case. Automate that "something" or batch it somehow.

Well this happens. Let's say you are packing the website orders which are queued and planned. But when there is an offline sale and the customer is here you to serve him right now. And then continue with your website orders.

pedrobaeza:
Uhm, OK, that seems a real need, although missing the notification/sound seems a real possibility, so you should consider a fallback.

ivs-cetmix (approved):
Functional review LGTM

hoangtrann (approved):
Code LGTM

ivs-cetmix:
Hey @rousseldenis @pedrobaeza sorry for bothering you, however this module could be a great new member of the OCA module family. We are already using it in several projects, and it really rocks)

pedrobaeza:
This is depending of another pull request that must be merged before as I see in the CI (but no reference in the initial comment)

ivs-cetmix:
> This is depending if another pull request that must be merged before as I see in the CI (but no reference in the initial comment)

@pedrobaeza indeed! Thank you for noticing that! @GabbasovDinar could you please add a link to that PR in this PR description?

ivs-cetmix:
> This is depending of another pull request that must be merged before as I see in the CI (but no reference in the initial comment)

@pedrobaeza looks like the PR this one depends on is already merged. Could you please have a look at this so we could merge this one too?

stock_picking_status_notification/models/stock_picking.py
```diff
@@ -0,0 +1,53 @@
...
+        "which is used to retrieve all records for this attribute "
+        "and to send a notification.",
+    )
+
+    def _write(self, vals):
+        # We use the _write method instead of write because _write is a low-level implementation
+        # that bypasses certain restrictions related to computed fields. In this case, it is
+        # crucial to ensure that notifications are triggered correctly when the state of the
+        # picking changes, even if this field (state) was computed by dependent fields
+        if "state" in vals:
```
rousseldenis, line 22: @GabbasovDinar @ivs-cetmix IMHO, this is too harsh as it involves every picking even those not concerned by notification.

We should filter the pickings to notify instead.

As this leads to DB locks (when there are a lot of movements - and only few are concerned by this flow).

ivs-cetmix: @rousseldenis I see, however this PR is already merged and we are not managing this project any longer.

PR #1682, opened 2024-08-28
[17.0][MIG] stock_delivery_note: Migration to 17.0

SodexisTeam:


MarwanBHL (approved):
Code review

stock_delivery_note/reports/report_picking.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="report_delivery_document" inherit_id="stock.report_delivery_document">
+        <xpath expr="//div[@name='signature']" position="before">
+            <p t-if="o.delivery_note">
+                <div t-field="o.delivery_note" />
+            </p>
+        </xpath>
```
christian-ramos-tecnativa: While testing it in runboat: 
If I print the delivery slip without signature and without delivery note I get a traceback. 
If both are set, then the signature is not displayed.
There must be something off with the xml on this version because if I change the position to be after everything seems to work, but it looks odd having the signature before the notes.

christian-ramos-tecnativa (changes requested):
Thanks for contributing, can you take a look to the comments? Thanks in advance

rousseldenis (approved):
Code review

stock_delivery_note/reports/report_picking.xml
christian-ramos-tecnativa: I just figure it out, with this xpath you are breaking the next if else [statement](https://github.com/OCA/OCB/blob/5d6192807491f41d572f5d89595c7d3d322c2df8/addons/stock/report/report_deliveryslip.xml#L197-L198)

SodexisTeam: Changed the xpath for the delivery note before the signature.

christian-ramos-tecnativa (approved):

PR #1688, opened 2024-08-30
[16.0][FIX] stock_picking_invoice_link, ensure field product_uom_qty value was changed to raise exception

ChrisOForgeFlow:
I found cases where product_uom_qty is raised as a change in field, but is for decimal errors, for example this one 50.0 != 50.00000000000001

@ForgeFlow

### DIFF
--- a/stock_picking_invoice_link/models/stock_move.py
+++ b/stock_picking_invoice_link/models/stock_move.py
@@ -5,6 +5,7 @@
 
 from odoo import _, fields, models
 from odoo.exceptions import UserError
+from odoo.tools import float_compare
 
 
 class StockMove(models.Model):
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
I think iit makes sense

MiquelRForgeFlow (approved):

stock_picking_invoice_link/models/stock_move.py
```diff
@@ -30,7 +31,14 @@ def write(self, vals):
         if "product_uom_qty" in vals and not self.env.context.get(
             "bypass_stock_move_update_restriction"
         ):
-            for move in self:
+            for move in self.filtered(
```
LoisRForgeFlow, line 34: I would do it inside the loop to not iterate twice the recordset (filtered + for).

stock_picking_invoice_link/models/stock_move.py
```diff
@@ -30,7 +31,14 @@ def write(self, vals):
...
+                lambda x: float_compare(
+                    x.product_uom_qty,
+                    vals.get("product_uom_qty"),
```
LoisRForgeFlow: Can you put a fallback value of 0.0. Just in case, as it it possible that you have the key in vals but with None as value.

LoisRForgeFlow (commented):
Two minor suggestions

stock_picking_invoice_link/models/stock_move.py
ChrisOForgeFlow: As far as I understand "filtered" function limit records in "for" function iteration, Please can you explain me a bit why iterate twice?

stock_picking_invoice_link/models/stock_move.py
ChrisOForgeFlow: Done!

rvalyi (approved):

astirpe:
Backported to V15 in https://github.com/OCA/stock-logistics-workflow/pull/1891

PR #1694, opened 2024-09-04
[14.0][IMP] sale_stock_mto_as_mts_orderpoint: return executed op

TDu:


rousseldenis:
@TDu Prefer explicit naming `sale_stock` in commit message:

image

jbaudoux (commented):
I find it strange to return ids and not recordset

TDu:
> I find it strange to return ids and not recordset

True, I did not want to browse for the records when not needed.

TDu:
> @TDu Prefer explicit naming `sale_stock` in commit message:
> 
> image

I updated the commit message. Always a juggle :juggling_person:  between the length and the meaning on the title of the commit message.

jbaudoux:
> I updated the commit message. Always a juggle 🤹 between the length and the meaning on the title of the commit message.

But the module name is with underscores and you put hyphens ...

TDu:
> > I updated the commit message. Always a juggle 🤹 between the length and the meaning on the title of the commit message.
> 
> But the module name is with underscores and you put hyphens ...

:zombie:

rousseldenis (approved):

jbaudoux:
> > I find it strange to return ids and not recordset
> 
> True, I did not want to browse for the records when not needed.

I would post the msg in the chatter in this module

rousseldenis:
@TDu

PR #1696, opened 2024-09-06
[14.0][FIX] sale_stock_mto_as_mts_orderpoint: fix orderpoint creation

mmequignon:
When generation the orderpoint from the sale order line, it might happen that the sale_order_line's product is a BOM, the moves that are generated are based on the components that composes it.

For this reason, when creating the purchase order, we have to get the product from moves, not sale order lines.

ping @sebalix @jbaudoux @TDu


jbaudoux (approved):

jbaudoux:
Obvious fix

/ocabot merge patch

PR #1697, opened 2024-09-09
[15.0][REF] product_cost_price_avco_sync: Refactoring for handling all the changes at once

pedrobaeza:
The current problem is that when there are crossed changes between products (like in MRP, having changes in components that affect manufactured products), a lot of writes are done, as each product is processed separately, and thus, there can be continuous resync between them.

With this new method, we can process the whole sync in one step, blocking it till pending data is processed, avoiding continuos sync triggerings.

Main changes:

- Extract the processing code corresponding to one SVL to the method `_process_avco_sync_one`.
- Include all the state variables - previous_unit_cost, previous_qty, inventory_processed and unit_cost_processed - into a super-structure (called `svls_dic`, which previously holds the SVL values for the product being synchronized). This super-structure is a dictionary having as keys the pair (product, company), and the value another dictionary with such state variables and storing the SVL values in "svls" key.
- 2 main loops for iterating in the elements to sync, one for the pair (product, company), and other inside for the SVLs to sync. First `while` loop is driven by these control variables:

  * index: which position - pair (product, company) - to process
  * reloop: activated when something is blocking
  * any_processed: to control if there's no progress in a whole loop

  The second one uses the value inside "to_sync" key of the super-structure to know which is the SVL to sync currently, and when processed, we dynamically look for the next one.
- `_preprocess_main_svl_line` and `_preprocess_rest_svl_to_sync` dissapears in favor of one control method: `_is_avco_sync_processable`, which can return a falsy value if we can't sync that SVL.
- We add extra things to sync modifying the super-structure in `_process_avco_sync_one` - add new (product, company) keys or change "to_sync" value -.
- All the AVCO sync methods have been converted to private.
- The processing of the SVLs previous to the one to sync is done in a light way, but using the same code of `_process_avco_sync_one`. After obtaining the state variables, the SVLs dict is discarded for saving memory.

@Tecnativa TT50028

### DIFF
--- a/product_cost_price_avco_sync/models/__init__.py
+++ b/product_cost_price_avco_sync/models/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import product_product
 from . import stock_move_line
 from . import stock_picking
 from . import stock_valuation_layer
--- a/product_cost_price_avco_sync/models/product_product.py
+++ b/product_cost_price_avco_sync/models/product_product.py
@@ -1,0 +1,15 @@
+# Copyright 2024 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import models
+from odoo.tools import config
+
+
+class ProductProduct(models.Model):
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Closing as merged in later versions.

PR #1699, opened 2024-09-09
[14.0][FIX] sale_stock_mto_as_mts_orderpoint: fix orderpoint activation condition

mmequignon:
When running orderpoints for mto products, we should ensure that move has been issued by an MTO product and that move's product is MTO.


jbaudoux (approved):
indeed as delivery moves are processed

mt-software-de (approved):
LGTM

PR #1701, opened 2024-09-10
[14.0][ADD] sale_stock_mto_as_mts_orderpoint_mrp

mmequignon:


sale_stock_mto_as_mts_orderpoint_bom/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": True,
+    "depends": [
+        "sale_stock_mto_as_mts_orderpoint",
+        "mrp",
+        "mrp_helper",
```
jbaudoux: Module was renamed https://github.com/OCA/manufacture/pull/1350
```suggestion
        "mrp_bom_find_ignore",
```
You need to add it in the test requirement

sale_stock_mto_as_mts_orderpoint_bom/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+        "Integration module between sale_stock_mto_as_mts_orderpoint and mrp"
+    ),
+    "version": "14.0.1.0.0",
+    "category": "Operations/Inventory/Delivery",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["mmequignon"],
+    "license": "AGPL-3",
+    "installable": True,
+    "auto_install": True,
```
mt-software-de: auto install is not really working here, since the mrp_bom_find_ignore is a helper addon which is not auto_installed.

mt-software-de (changes requested):
Could this addon named 
`sale_stock_mto_as_mts_orderpoint_mrp`

sale_stock_mto_as_mts_orderpoint_bom/__manifest__.py
mmequignon: <s>Well, that's a glue module, right ?
Shouldn't this module only depend on 2 modules ?</s>

I was thinking this module should integrate mrp_bom_find_ignore, but it doesn't make any sense.

sale_stock_mto_as_mts_orderpoint_bom/models/sale_order_line.py
```diff
@@ -0,0 +1,59 @@
...
+                **{
+                    "active_model": "stock.warehouse.orderpoint",
+                    "active_ids": orderpoints_to_procure_ids,
+                    "ignore_bom_find": True,
+                }
+            )
+            .create({})
+        )
+        wiz.make_procurement()
+        return super(SaleOrderLine, self - sale_bom_lines)._run_orderpoints_for_mto_products()
```
mt-software-de:
```suggestion
# Copyright 2024 Camptocamp SA
# Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
from odoo import api, models
from odoo.tools import groupby


class SaleOrderLine(models.Model):
    _inherit = "sale.order.line"

    @api.model
    def _all_moves_can_be_purchased(self, moves):
        for move in moves:
            if not move.product_id.purchase_ok:
                return False
        return True

    def _run_orderpoints_for_mto_products(self):
        orderpoints_to_procure_ids = []
        sale_bom_lines = self.browse()

        for line in self:
            delivery_moves = line.move_ids.filtered(
                lambda m: m.picking_id.picking_type_code == "outgoing"
                and m.state not in ("done", "cancel")
                and m.is_from_mto_route
                and m.bom_line_id
            )
            if not delivery_moves:
                continue
            if line._all_moves_can_be_purchased(delivery_moves):
                # if all moves are buyable, they should be bought individually
                continue
            for bom, move_list in groupby(
                delivery_moves, lambda m: m.bom_line_id.bom_id
            ):
                bom_moves = self.env["stock.move"].union(*move_list)
                bom_lines = bom.bom_line_ids
                if bom_lines != bom_moves.bom_line_id:
                    continue
                orderpoint = line._get_mto_orderpoint(bom.product_id)
                if orderpoint.procure_recommended_qty:
                    orderpoints_to_procure_ids.append(orderpoint.id)
            sale_bom_lines |= line
        if orderpoints_to_procure_ids:
            wiz = ( 
                self.env["make.procurement.orderpoint"]
                .with_context(
                    **{
                        "active_model": "stock.warehouse.orderpoint",
                        "active_ids": orderpoints_to_procure_ids,
                        "ignore_bom_find": True,
                    }
                )
                .create({})
            )
            wiz.make_procurement()
        return super(
            SaleOrderLine, self - sale_bom_lines
        )._run_orderpoints_for_mto_products()
```

mt-software-de: I updated the code, it was not working because also lines were used without any bom and which were not coming from mto.

mmequignon:
> Could this addon named `sale_stock_mto_as_mts_orderpoint_mrp`

done

sale_stock_mto_as_mts_orderpoint_bom/models/sale_order_line.py
mmequignon: It's complicated to find the differences here....

mmequignon: @mt-software-de I updated this method. Can you please redo this without selecting the whole block ? That'd be helpful.
Thanks !

mmequignon: Also, I kind of disagree with this `_all_moves_can_be_purchased` method.
This has nothing to do on the sale.order.line model.
I could move that on stock.move if you really want me to extract that in a method.
But I'm quite ok with the inner function.

mt-software-de: yes i don't really care if it is inner or not. i just fixed the method in our testing
the main difference is the filter on delivery_moves, we don't want to do it for moves which do not have a bom_line_id set or not is_from_mto_route
```
        for line in self:
            delivery_moves = line.move_ids.filtered(
                lambda m: m.picking_id.picking_type_code == "outgoing"
                and m.state not in ("done", "cancel")
                and m.is_from_mto_route
                and m.bom_line_id
            )
            if not delivery_moves:
                continue
```
also i changed the groupby
instead of itertools the default from odoo.tools
this does not require a sorting
```
from odoo.tools import groupby
```

jbaudoux: @mt-software-de you cannot exclude with `and m.is_from_mto_route` as the component might not be mto while the parent is. That's all the point of this module

mmequignon:
@jbaudoux @mt-software-de 
Thanks for your reviews guys, I updated the PR, included a « refactoring » of `sale_stock_mto_as_mts_orderpoint`, adding a few hooks to ease extension and overrides.
Can you please drop another review please ?

sale_stock_mto_as_mts_orderpoint/models/stock_move.py
```diff
@@ -16,3 +16,57 @@ def _compute_is_from_mto_route(self):
...
+                        "product_min_qty": 0.0,
+                        "product_max_qty": 0.0,
+                    }
+                )
+            )
+        return orderpoint
+
+    def _whatever(self, warehouse):
+        orderpoint_to_procure_ids = []
+        if not warehouse.mto_as_mts:
```
jbaudoux: I would have placed this before building delivery_moves, so that you just continue to next item of the group by loop. But it's a matter of taste. Would prevent to make this test also in the overrides

sale_stock_mto_as_mts_orderpoint/models/stock_move.py
```diff
@@ -16,3 +16,57 @@ def _compute_is_from_mto_route(self):
...
+                            warehouse._get_locations_for_mto_orderpoints().id
+                        ),
+                        "product_min_qty": 0.0,
+                        "product_max_qty": 0.0,
+                    }
+                )
+            )
+        return orderpoint
+
+    def _whatever(self, warehouse):
```
jbaudoux: What about
```suggestion
    def _get_mto_as_mts_orderpoints(self, warehouse):
```

sale_stock_mto_as_mts_orderpoint/models/stock_move.py
```diff
@@ -16,3 +16,57 @@ def _compute_is_from_mto_route(self):
...
+        for record in self:
+            if (
+                not record.is_from_mto_route
+                and mto_route not in record.product_id.route_ids
+            ):
+                continue
+            orderpoint = record._get_mto_orderpoint(warehouse, record.product_id)
+            if orderpoint.procure_recommended_qty:
+                orderpoint_to_procure_ids.append(orderpoint.id)
+        return orderpoint_to_procure_ids
```
jbaudoux: I prefer to return recordset than ids

sale_stock_mto_as_mts_orderpoint/models/stock_move.py
```diff
@@ -16,3 +16,57 @@ def _compute_is_from_mto_route(self):
         else:
             for move in self:
                 move.is_from_mto_route = move.rule_id.route_id == mto_route
+
+    @api.model
+    def _get_mto_orderpoint(self, warehouse, product):
```
jbaudoux, line 21: Moved it to `product.product` and try to be close to v16 implementation https://github.com/OCA/stock-logistics-orderpoint/blob/16.0/stock_orderpoint_mto_as_mts/models/product_product.py#L26
You could backport the content of product_product.py from v16.

The main difference in v16 is that the trigger is automatic and you don't need to call the wizard

sale_stock_mto_as_mts_orderpoint/models/stock_move.py
```diff
@@ -16,3 +16,57 @@ def _compute_is_from_mto_route(self):
...
+        return orderpoint
+
+    def _whatever(self, warehouse):
+        orderpoint_to_procure_ids = []
+        if not warehouse.mto_as_mts:
+            return orderpoint_to_procure_ids
+        mto_route = self.env.ref("stock.route_warehouse0_mto", raise_if_not_found=False)
+        if not mto_route:
+            return
+        for record in self:
```
jbaudoux:
```suggestion
        for move in self:
```

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def all_moves_can_be_bought(self):
+        for move in self:
+            if not move.product_id.purchase_ok:
+                return False
+        return True
```
jbaudoux, line 15:
```suggestion
        return all(move.product_id.mapped("purchase_ok"))
```

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,63 @@
...
+
+            bom_product = bom.product_id
+            bom_lines = bom.bom_line_ids
+            move_bom_lines = bom_moves.bom_line_id
+            if bom_lines != move_bom_lines:
+                continue  # FIXME
+
+            if not bom_moves.all_moves_can_be_bought() and self.product_is_mto(
+                bom_product
+            ):
```
jbaudoux: group all your refutations
```suggestion
            if (
                bom_lines != move_bom_lines
                or bom_moves.all_moves_can_be_bought()
                or not self.product_is_mto(bom_product)
            ):
                continue
```

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,63 @@
...
+                seen_moves |= bom_moves
+
+            # otherwise, create orderpoints for remaining mto moves
+            for move in bom_moves:
+                product = move.product_id
+                if self.product_is_mto(product):
+                    orderpoint = self._get_mto_orderpoint(warehouse, product)
+                    if orderpoint.procure_recommended_qty:
+                        orderpoints_to_procure_ids.append(orderpoint.id)
+                    seen_moves |= move
```
jbaudoux: Looks like what super is doing, so could be dropped

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,63 @@
...
+    def all_moves_can_be_bought(self):
+        for move in self:
+            if not move.product_id.purchase_ok:
+                return False
+        return True
+
+    @api.model
+    def product_is_mto(self, product):
+        mto_route = self.env.ref("stock.route_warehouse0_mto")
+        return mto_route in product.route_ids
```
jbaudoux: Move to base module and put on `product.product`. In v16, this is provided by odoo standard

sale_stock_mto_as_mts_orderpoint/models/stock_move.py
```diff
@@ -16,3 +16,57 @@ def _compute_is_from_mto_route(self):
...
+            if (
+                not record.is_from_mto_route
+                and mto_route not in record.product_id.route_ids
```
jbaudoux: Use `product_is_mto` as you defined it

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,63 @@
...
+                # In case a move doesn't have a bom, do nothing
+                continue
+
+            bom_product = bom.product_id
+            bom_lines = bom.bom_line_ids
+            move_bom_lines = bom_moves.bom_line_id
+            if bom_lines != move_bom_lines:
+                continue  # FIXME
+
+            if not bom_moves.all_moves_can_be_bought() and self.product_is_mto(
```
jbaudoux: Add also support for MTO route on SO line like base module
`product_is_mto or all(bom_moves.is_from_mto_route)`

sale_stock_mto_as_mts_orderpoint_mrp/tests/common.py
```diff
@@ -0,0 +1,239 @@
...
+                    ),
+                    (
+                        0,
+                        0,
+                        {"product_id": cls.product_office_desk.id, "product_qty": 1.0},
+                    ),
+                    (
+                        0,
+                        0,
+                        {"product_id": cls.product_office_bin.id, "product_qty": 1.0},
```
jbaudoux: safer if you test with qty > 1 to ensure the right qty is used

sale_stock_mto_as_mts_orderpoint_mrp/tests/test_bom.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from .common import Common
+
+
+class TestMtsAsMtoBom(Common):
+    def test_bom_1(self):
```
jbaudoux:
```suggestion
    def test_purchase_components(self):
```

sale_stock_mto_as_mts_orderpoint_mrp/tests/test_bom.py
```diff
@@ -0,0 +1,82 @@
...
+        purchase_orders_after = self.purchase_order_model.search(
+            [("id", "not in", purchase_orders_before.ids)]
+        )
+        self.assertTrue(purchase_orders_after)
+        # As only the structure is MTO, generated PO should only contain the bed structure
+        purchase_lines = purchase_orders_after.order_line
+        self.assertEqual(purchase_lines.product_qty, 1.0)
+        self.assertEqual(purchase_lines.product_id, self.product_bed_structure)
+
+    def test_components_cannot_be_bought(self):
```
jbaudoux:
```suggestion
    def test_purchase_parent_kit_single_level(self):
```

sale_stock_mto_as_mts_orderpoint_mrp/tests/test_bom.py
```diff
@@ -0,0 +1,82 @@
...
+            [("id", "not in", purchase_orders_before.ids)]
+        )
+        self.assertTrue(purchase_orders_after)
+        # As none of the components are purchaseable, only the BOM product should
+        # be in the PO
+        purchase_lines = purchase_orders_after.order_line
+        self.assertEqual(purchase_lines.product_qty, 1.0)
+        self.assertEqual(purchase_lines.product_id, self.product_office_furniture)
+
+    def test_bom_3(self):
```
jbaudoux:
```suggestion
    def test_purchase_parent_kit_multi_level(self):
```

sale_stock_mto_as_mts_orderpoint_mrp/tests/test_bom.py
```diff
@@ -0,0 +1,82 @@
...
+        order.action_confirm()
+        purchase_orders_after = self.purchase_order_model.search(
+            [("id", "not in", purchase_orders_before.ids)]
+        )
+        self.assertTrue(purchase_orders_after)
+        # As none of the components are purchaseable, only the BOM product should
+        # be in the PO
+        purchase_lines = purchase_orders_after.order_line
+        self.assertEqual(purchase_lines.product_qty, 1.0)
+        self.assertEqual(purchase_lines.product_id, self.product_office_furniture)
```
jbaudoux: I would invert the 2 assert to first check there is only a line for the office furniture.
And also make one of the component purchasable to ensure it is not purchased as not all are purchasable

sale_stock_mto_as_mts_orderpoint_mrp/tests/test_bom.py
```diff
@@ -0,0 +1,82 @@
...
+            lambda l: l.product_id == self.product_garden_table
+        )
+        self.assertTrue(table_line)
+        self.assertEqual(table_line.product_qty, 1.0)
+        # Chair is MTO, we should have a line for it, with qty = 4.0
+        chair_line = purchase_lines.filtered(
+            lambda l: l.product_id == self.product_garden_chair
+        )
+        self.assertTrue(chair_line)
+        self.assertEqual(chair_line.product_qty, 4.0)
```
jbaudoux: Add a check that there are no other purchase_lines than table and chair.

You should have the same result if the table leg is purchasable or if the garden furniture is purchasable. The last one should be tested as it is important. In this test or in a separate one.

jbaudoux (changes requested):
Algorithm seems now to be properly there.
Needs some final cleanup.
Have a look at how it has been cleaned up in v16. Note that in v16, it's still lacking the support of the mto route on the SO line (but it's also strange to create a permanent orderpoint for this case). The orderpoint is created when the route is set on the product. And orderpoints are triggered automatically in v16.

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
mmequignon: No. Doing this loop is more efficient in term of resources.

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
mmequignon: I don't like those 4 lines one-liners…

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
mmequignon: nice catch

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+    _inherit = "stock.move"
+
+    def all_moves_can_be_bought(self):
+        for move in self:
+            if not move.product_id.purchase_ok:
+                return False
+        return True
+
+    def _get_mto_as_mts_orderpoints(self, warehouse):
+        orderpoints_to_procure = self.env["stock.warehouse.orderpoint"]
```
jbaudoux, line 18: You need to repeat here
`        if not warehouse.mto_as_mts:`
that's why you better move it before the function call

jbaudoux (approved):
Lgtm

jbaudoux:
@mmequignon can you rebuild the generated files as you renamed the module. You may also add BCIM as co author in the manifest. Thanks

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
```diff
@@ -0,0 +1,48 @@
...
+                # In case a move doesn't have a bom, do nothing
+                continue
+
+            bom_product = bom.product_id
+            bom_lines = bom.bom_line_ids
+            move_bom_lines = bom_moves.bom_line_id
+            if bom_lines != move_bom_lines:
+                continue
+
+            is_mto = bom_product._is_mto() or all(bom_moves.is_from_mto_route)
```
mt-software-de, line 36:
```suggestion
            is_mto = bom_product._is_mto() or all(bom_move.is_from_mto_route for bom_move in bom_moves)
```

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
mt-software-de, line 1: I think we should point out, that this can only work for simple bom kits. (Where no component is a kit it self)
A more complex bom structure like. 
Product 1
- > Component 1.1
- > Component 1.2 (kit)
-- Component 1.2.1
-- Component 1.2.2

This can never order all of the necessary products.
Since there will be 1 move (1.1) linked to bom 1
and 2 moves (1.2.1, 1.2.2) linked to bom 1.2.

The move from bom 1 can not be reordered by mto because of this `if bom_lines != move_bom_lines:`
it would be just possible to reorder Component 1.2 .

mt-software-de (changes requested):

sale_stock_mto_as_mts_orderpoint_mrp/models/stock_move.py
mt-software-de: @mmequignon could you change this? @jbaudoux then we should be good to merge it.

PR #1702, opened 2024-09-10
[14.0][IMP] stock_picking_auto_create_package: Without packaging assume qty of 1

mt-software-de:
This will put the move line per piece in a separate package.
Also it is now possible to package an amount which is not a multiple of a its packaging amount.
For example: a move over 5 pieces and with a packaging amount of 2, will create 3 packages with following amounts 2 2 1

### DIFF
--- a/stock_picking_auto_create_package/models/stock_picking.py
+++ b/stock_picking_auto_create_package/models/stock_picking.py
@@ -3,9 +3,7 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import _, api, models
-from odoo.exceptions import UserError
-from odoo.tools.float_utils import float_round
+from odoo import api, models
 
 
 class StockPicking(models.Model):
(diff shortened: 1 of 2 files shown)


mt-software-de:
cc @jbaudoux @TDu

TDu (approved):
LG

stock_picking_auto_create_package/models/stock_picking.py
```diff
@@ -38,35 +36,16 @@ def _auto_create_delivery_package_per_smallest_packaging(self) -> None:
...
-                            "a multiple of product packaging"
-                        )
-                        % move_line.product_id.display_name
-                    )
-                for _i in range(int(qty)):
-                    move_line.qty_done = smallest_packaging.qty
+                smallest_qty = 1
+                packagings = move_line.product_id.packaging_ids
+                if packagings:
+                    smallest_packaging = packagings.sorted("qty")[0]
```
jbaudoux: For safety, I would exclude packagings with qty <= 0

stock_picking_auto_create_package/models/stock_picking.py
```diff
@@ -38,35 +36,16 @@ def _auto_create_delivery_package_per_smallest_packaging(self) -> None:
...
+                    smallest_qty = smallest_packaging.qty
+                while qty_done:
+                    qty_todo = min(qty_done, smallest_qty)
+                    qty_done -= qty_todo
+                    move_line.qty_done = qty_todo
```
jbaudoux: I find the variable names misleading. I would have inverted todo and done or used different names

vuwnevska (approved):

jbaudoux (approved):

PR #1721, opened 2024-09-18
[FIX] stock_picking_return_lot: group return line by move and lot

sbejaoui:
In case a picking out contains multiple moves for the same product/lot, it's essential to split the return lines by lot move to ensure the correct linking between the return move and the delivery move.

### DIFF
--- a/stock_picking_return_lot/tests/test_stock_picking_return_lot.py
+++ b/stock_picking_return_lot/tests/test_stock_picking_return_lot.py
@@ -62,7 +62,9 @@ class StockPickingReturnLotTest(TransactionCase):
             .create({})
         )
 
-    def _create_validate_picking(self):
+    def _create_picking(self, product=None):
+        if not product:
+            product = self.product
         picking = self.picking_obj.create(
             {
                 "partner_id": self.partner.id,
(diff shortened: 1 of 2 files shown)


petrus-v (approved):
Code looks good

I've create a module on version 14.0 that when you are using lot restrict returns on that lot: cf https://github.com/OCA/stock-logistics-workflow/pull/1441 (The PR is staled but will re-open it soon)

stock_picking_return_lot/tests/test_stock_picking_return_lot.py
```diff
@@ -189,15 +195,47 @@ def test_multiple_move_same_product_same_lot(self):
...
+        moves = picking_returned.move_ids.filtered(
             lambda m, lot=self.lot_1: m.restrict_lot_id == lot
         )
-        self.assertEqual(move_1.move_line_ids.lot_id, self.lot_1)
-        self.assertEqual(move_1.product_qty, 2)
+        self.assertEqual(moves.move_line_ids.lot_id, self.lot_1)
+        self.assertEqual(moves[0].product_qty, 1)
+        self.assertEqual(moves[1].product_qty, 1)
+
+    def test_(self):
```
rousseldenis:
```suggestion
    def test_<something>(self):
```

stock_picking_return_lot/wizards/stock_return_picking_line.py
```diff
@@ -13,6 +13,7 @@ class StockReturnPickingLine(models.TransientModel):
         string="Lot/Serial Number",
         domain="[('product_id', '=', product_id)]",
     )
+    origin = fields.Char(related="move_id.origin", readonly=True)
```
rousseldenis: Shouldn't it declared elsewhere (another module) as not really linked directly to lots ?

sbejaoui: In this particular case, if a picking contains more than one move for the same product and lot, it's helpful for users to see the origin of each return line in the return wizard. This helps them, for example, to identify which sale order to return.

Before this change, only a single line per product was shown in the return wizard, so the origin of the return didn't matter and wasn't necessary.

rousseldenis (approved):

PR #1730, opened 2024-09-26
[IMP] sale_global_stock_route: prevent to update route_id on confirme…

bealdav:
…d sales

@cvinh @carlosdauden @CarlosRoca13

### DIFF
--- a/sale_order_global_stock_route/views/sale_order_views.xml
+++ b/sale_order_global_stock_route/views/sale_order_views.xml
@@ -11,6 +11,11 @@
                      field in sale order lines.
                  -->
                 <field name="route_id" groups="stock.group_adv_location" />
+                <field
+                    name="route_id"
+                    attrs="{'readonly': [('state', 'in', ('sale', 'done', 'cancel'))]}"
+                    groups="stock.group_adv_location"
+                />
             </field>
         </field>
     </record>


cvinh (approved):
Even if it's still draft, I understand what you did and of course I'm approving

carlosdauden (approved):

CarlosRoca13 (approved):

rousseldenis:
Mmmh, I understand the problem you have but:

- In Odoo core, the route_id field on sale order line is readonly in "done" and "cancel"

That's why I think the best is to mimic the same states.

bealdav:
@rousseldenis Thanks for your suggestion.
But readonly is only in the view then method in current module can change the field isn't it ?

bealdav:
And on the user side, it can let think we can update the flow whereas we should not. Then it's error prone or at least a source of misunderstood isn't it ?

cvinh:
> Mmmh, I understand the problem you have but:
> 
> * In Odoo core, the route_id field on sale order line is readonly in "done" and "cancel"
> 
> That's why I think the best is to mimic the same states.

Well IMHO sometimes, Odoo's way is not the best... but sometimes it is... 
I prefer sale, done... if it's cancelled then why would it be forbidden to change it...
I vote for @bealdav way

bealdav:
@rousseldenis maybe I misunderstood your message but my commit also makes field readonly when state in done/cancel, then that's the odoo way ? I missed something ?

cvinh:
> @rousseldenis maybe I misunderstood your message but my commit also makes field readonly when state in done/cancel, then that's the odoo way ? I missed something ?

@bealdav you did
attrs="{'readonly': [('state', 'in', ('sale', 'done'))
and I vote for you

rousseldenis:
> @rousseldenis maybe I misunderstood your message but my commit also makes field readonly when state in done/cancel, then that's the odoo way ? I missed something ?

In fact, that's the behaviour for the lines in `sale` state, you can still add lines with a route (e.g. : for MTO).

Here, the field is global, so, it can be good to protect `sale`state too.

But for the `cancel`state, you should not be able to change something too. So, I would have added that state

bealdav:
Sorry I misread states, really sorry  :-(

Everybody is OK for  `'readonly': [('state', 'in', ('sale', 'done', 'cancel'))]` ?

rousseldenis (approved):

PR #1735, opened 2024-10-01
[14.0][ADD] stock_quant_serial_partner_location

christian-ramos-tecnativa:


stock_quant_serial_partner_location/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module remove the check on serial number quants when the location is type customer or vendor.
+It is usefull to specific situation, when you sell products to one company and then the same one provide
+dropship service. When you try to send to the final customer the same serial number the error will pop up.
+This module block the raise and allow you to send it on that situation.
```
thomaspaulb: Corrected text:

This module removes the check on serial number quants when the location is type customer or vendor.
It is useful to specific situation, when you sell products to one company and then the same one provides a
dropship service. When you try to send to the final customer the same serial number the error will pop up.
This module blocks the raise and allows you to send it in that situation.

stock_quant_serial_partner_location/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+This module remove the check on serial number quants when the location is type customer or vendor.
+It is usefull to specific situation, when you sell products to one company and then the same one provide
```
thomaspaulb: Can you elaborate on the use case for this module? I tested and it works as intended, but it seems to me a bit dangerous that when installing this module, you can send out the same serial number twice.

Why would this be needed? Because a serial number normally can reside only in one place.

Maybe you can give a specific example of a dropship flow in which this happens, in steps.

thomaspaulb (changes requested):

stock_quant_serial_partner_location/readme/DESCRIPTION.rst
christian-ramos-tecnativa: Done, Thanks

stock_quant_serial_partner_location/readme/DESCRIPTION.rst
christian-ramos-tecnativa: Hello @thomaspaulb, thanks for the review. Here are the steps to reproduce the issue:

1. **Setup**:
   - Company A and Company B are separate entities.
   - Company A initially stores products in its own warehouse and regularly sells them.

2. **Scenario**:
   - Company A decides to stop storing products in its own warehouse and instead subcontracts Company B to handle dropshipping for them.
   - As part of this change, Company A sells all of its stock to Company B (using the `purchase_sale_inter_company` module in Odoo to facilitate this process).
   - Company A delivers the products to Company B with serialized numbers, which Company B retains.

3. **Issue**:
   - When Company A sells a product via dropshipping, Company B fulfills the delivery to the end customer.
   - An error is triggered because Company A is trying to register a delivery to the customer using the same serial numbers already used in the transfer to Company B.

This is indeed a specific situation, but it could potentially affect others in a similar setup. If it’s a one-time scenario, performing an inventory adjustment on the customer location might temporarily resolve it, but I’m uncertain about the long-term correctness of that approach.

thomaspaulb: Hi, thanks for explaining the scenario!

I wonder why you get an error in that scenario, because we have exactly the same setup as you describe for a customer, but we don't get that error. It's probably because in our case, we maintain the multi-company security rule on serial numbers, so we dont share serial numbers between companies - they have to be copied. (OCA intercompany module copies them automatically, not sure about Odoo Enterprise version).

Because they are copied you wont get the error because the serial numbers are different records with different ids

PR #1740, opened 2024-10-09
[ADD] stock_picking_putaway_recompute

rousseldenis:


rousseldenis:
@lmignon @jbaudoux @phschmidt

stock_picking_putaway_recompute/models/stock_move_line.py
```diff
@@ -0,0 +1,58 @@
...
+class StockMoveLine(models.Model):
+
+    _inherit = "stock.move.line"
+
+    can_recompute_putaways = fields.Boolean(
+        compute="_compute_can_recompute_putaways",
+    )
+
+    @api.depends(
+        "picking_type_id.allow_to_recompute_putaways", "picking_id.printed", "qty_done"
```
simahawk: `allow_to_recompute_putaways` is not stored, should not stay here

stock_picking_putaway_recompute/tests/test_recompute_putaway.py
```diff
@@ -0,0 +1,325 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.fields import Command, first
+
+from odoo.addons.base.tests.common import BaseCommon
```
simahawk, line 5: `BaseCommon` does too many things and it slows down setup. If you don't need company, currency, user, etc, use `TransactionCase` + `tracking_disable` or `DISABLED_MAIL_CONTEXT`.

simahawk (commented):
LG, minor remarks

stock_picking_putaway_recompute/models/stock_move_line.py
grindtildeath: Why? AFAIK you can use non stored computed fields as dependencies for other non stored computed fields :shrug: 

EDIT: BTW seems we need more dependencies, cf my comment below
EDIT 2: Actually it's a stored field on picking type :wink:

stock_picking_putaway_recompute/models/stock_move_line.py
```diff
@@ -0,0 +1,58 @@
...
+    def _filtered_for_putaway_recompute(self) -> Self:
+        """
+        Recompute putaways on operations that:
+
+            - have their picking type configured for that
+            - have their picking not printed (started)
+            - have their qty_done field != 0
+        """
+        return self.filtered(
+            lambda line: line.picking_type_id.allow_to_recompute_putaways
```
grindtildeath: Wouldn't it make more sense to have the filtering function defined at line level instead of one that filters out without the ability to override?

stock_picking_putaway_recompute/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+class StockPicking(models.Model):
+
+    _inherit = "stock.picking"
+
+    can_recompute_putaways = fields.Boolean(
+        compute="_compute_can_recompute_putaways",
+        help="Technical field in order to display the Recompute Putaways button.",
+    )
+
+    def action_recompute_putaways(self) -> dict:
```
grindtildeath, line 19: I don't know much about annotations but AFAIU this function must return a dict but doesn't return anything

stock_picking_putaway_recompute/models/stock_picking.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        for picking in self:
+            picking.move_line_ids._recompute_putaways()
+
+    def _filtered_can_recompute_putaways(self) -> Self:
+        """
+        Filter current recordset in order to get the pickings that can
+        """
+        return self.filtered(
+            lambda picking: picking.state == "assigned" and not picking.printed
```
grindtildeath: Same as above about filtering function. Moreover, this does not check if recompute putaway is allowed at picking type level :confused: 

I feel like the filtering function from stock move should somehow call the filtering function from picking for consistency.

I would suggest to do as follows:
```
    def _can_recompute_putaway(self):
        self.ensure_one()
        return self.picking_id._can_recompute_putaway() and not self.result_package_id and not self.qty_done

    def _filtered_can_recompute_putaways(self)
        return self.filtered(lambda li: li._can_recompute_putaway())


    def _can_recompute_putaway(self):
        self.ensure_one()
        return self.picking_type_id.allow_to_recompute_putaways and self.state == "assigned" and not self.printed

    def _filtered_can_recompute_putaways(self)
        return self.filtered(lambda pick: pick._can_recompute_putaway())
```
What do you think?

grindtildeath (changes requested):

grindtildeath:
@rousseldenis I did the suggested changes here to speed up the merge of that PR: https://github.com/acsone/stock-logistics-workflow/pull/2

Can you please check and merge if that's good for you? :pray:

lmignon:
> @rousseldenis I did the suggested changes here to speed up the merge of that PR: [acsone#2](https://github.com/acsone/stock-logistics-workflow/pull/2)
> 
> Can you please check and merge if that's good for you? 🙏

done :smirk:

jbaudoux (approved):

jbaudoux:
Module already merged in v18

/ocabot merge nobump

PR #1741, opened 2024-10-09
[16.0][ADD] stock_split_picking_dimension: Split picking based on dimension

lmignon:
This module extends the functionality of the stock_split_picking module to allow the user to split the picking by dimensions to ensure that the contents of the picking to process will not exceed a certain volume and/or weight and/or number of lines.


lmignon:
ping @rousseldenis @jbaudoux @sbejaoui

stock_split_picking_dimension/readme/CREDITS.md
```diff
@@ -0,0 +1,3 @@
+The development of this module has been financially supported by:
+
+- ALCYON Benelux
```
jbaudoux:
```suggestion
- ALCYON Belux
```

jbaudoux (commented):
What if no moves are kept ?

Shouldn't you always keep the first move as a start and then check if you can keep other moves according to the constrains?

rousseldenis:
Depends on:

- https://github.com/OCA/stock-logistics-workflow/pull/1788

stock_split_picking_dimension/readme/CONTEXT.md
```diff
@@ -0,0 +1,5 @@
+When dealing with large quantities of products, it could be useful to be able to split
+the picking into multiple pickings to simplify the picking process and improve the
+efficiency of the warehouse. Some times, the picking you would like to process is too
+large to be picked in one go or too heavy to be carried by one person or contains too
+many different products to be picked in one go.
```
jbaudoux, line 5:
```suggestion
When dealing with large quantities of products, it is useful to split
the picking into multiple pickings to reduce the time spent on a picking and eventually parallelize the work to improve the
efficiency of the warehouse. Sometimes the picking you would like to process is too
large to be picked in one go or too heavy to be carried by one person or contains too
many different products to be picked in one go.
```

stock_split_picking_dimension/wizards/stock_split_picking.py
```diff
@@ -0,0 +1,223 @@
...
+                        precision_rounding=weight_uom.rounding,
+                    )
+                    > 0
+                ):
+                    moves_kept |= move
+                    nbr_lines += 1
+                    volume += move_volume
+                    weight += move_weight
+            if moves_kept != picking.move_ids:
+                moves_to_split = picking.move_ids - moves_kept
```
jbaudoux, line 221: Should be the other way round to keep the remaining moves in the current transfer.
Otherwise, the remaining moves are put in a backorder and will be processed as last due to the picking sorting on id

jbaudoux:
@ivantodorovich is migrating this to v18. He will tackle comments in separate commits that could be backported

jbaudoux (approved):

jbaudoux:
Can you rebase and drop do not merge commit?

lmignon:
> Can you rebase and drop do not merge commit?

done :smirk:

PR #1749, opened 2024-10-24
[FIX] [16.0] stock_customer_deposit_sale_margin: Fix case when deposit is equal to product_uom_qty

Shide:
When quantities on deposit are equal to the product_uom_qty, doesn't set purchase_price to 0.0.

Fixed + refactor to view clearly all the cases

MT-6706 @moduon @Gelojr @fcvalgar @yajo @rafaelbn please review if you want.

### DIFF
--- a/stock_customer_deposit_sale_margin/models/sale_order_line.py
+++ b/stock_customer_deposit_sale_margin/models/sale_order_line.py
@@ -2,6 +2,7 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
 
 from odoo import api, models
+from odoo.tools import float_compare, float_is_zero
 
 
 class SaleOrderLine(models.Model):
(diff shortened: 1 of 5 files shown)


stock_customer_deposit_sale_margin/models/sale_order_line.py
```diff
@@ -18,9 +18,13 @@ class SaleOrderLine(models.Model):
     def _compute_purchase_price(self):
         res = super()._compute_purchase_price()
         for line in self:
-            if 0 <= line.deposit_available_qty <= line.product_uom_qty:
+            if line.deposit_available_qty <= 0.0:  # No deposit available
```
rousseldenis: You should use `float_compare`

stock_customer_deposit_sale_margin/models/sale_order_line.py
```diff
@@ -18,9 +18,13 @@ class SaleOrderLine(models.Model):
...
                 continue
-            # Set purchase price to zero because purchase price is in sale order
-            # that made customer deposit, otherwise the cost would be posted twice.
-            line.purchase_price = 0.0
+            elif not line.product_uom_qty:  # No quantity to deliver
```
rousseldenis: You should use `float_is_zero()`

stock_customer_deposit_sale_margin/models/sale_order_line.py
```diff
@@ -18,9 +18,13 @@ class SaleOrderLine(models.Model):
...
+                continue
+            elif line.deposit_available_qty >= line.product_uom_qty:
```
rousseldenis: You should use `float_compare()`

stock_customer_deposit_sale_margin/tests/test_stock_customer_deposit_sale_margin.py
```diff
@@ -28,7 +28,7 @@ def test_stock_customer_deposit_sale_margin(self):
         so_form.partner_id = self.partner1
         so_form.warehouse_id = self.warehouse
         product_qty_dict = {
-            self.productA: 50.0,
+            self.productA: 100.0,
```
rousseldenis, line 31: Was the original test incorrect?

Shide: The test was not covering the case when deposit is 100 and you want to deliver 100. In the original code this case was failing.

EmilioPascual (approved):
Code review

rafaelbn (approved):
/ocabot merge patch

PR #1750, opened 2024-10-24
[FIX] [16.0] stock_customer_deposit: Create valuation layers when performing a deposit

Shide:
Create a valuation layer when performing a deposit to take into account that the goods are not part of your valuated inventory.

MT-7678 @moduon @rafaelbn @Gelojr @fcvalgar @EmilioPascual @yajo please review if you want :)

### DIFF
--- a/stock_customer_deposit/models/stock_move.py
+++ b/stock_customer_deposit/models/stock_move.py
@@ -39,6 +39,17 @@ class StockMove(models.Model):
             StockMove, (moves_customer_deposits - moves_owner).with_context(owner=False)
         )._action_assign(force_qty=force_qty)
 
+    def _get_out_move_lines(self):
+        """Also consider as "out" move lines to those that belongs to
+        a picking type that assigns owners
+        and are in a location that should be valued."""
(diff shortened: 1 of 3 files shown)


Gelojr (approved):
LGTM
Nice job @Shide 
https://www.loom.com/share/5a255632a0cb4095beef6f3b54f5a0e9?sid=750c3f88-c82b-428f-99c7-1746c5183005

EmilioPascual (approved):
Code review

rousseldenis:
@Shide What about deposit with owner you put in stock back ?

Shide:
> @Shide What about deposit with owner you put in stock back ?

For the valuation perspective, it counts as a normal incoming transfer, so valuation should be created as usually without any modification.

Also, deposit returns are not being properly done with the return button because odoo cannot transfer one location with owner to the same location without owner. We are doing this with inventory adjustments and reevaluating manually.

This is an issue and will be addressed soon (i hope).

rafaelbn (approved):
/ocabot merge patch

PR #1752, opened 2024-10-31
[MIG] stock_picking_invoice_link: Migration to 18.0

carlosdynapps:


stock_picking_invoice_link/models/account_move.py
```diff
@@ -0,0 +1,86 @@
...
+    def copy_data(self, default=None):
+        """Copy the move_line_ids in case of refund invoice creating new invoices
+        (refund_method="modify") for multiple records."""
+        vals_list = super().copy_data(default)
+
+        for record, vals in zip(self, vals_list, strict=False):
+            if (
+                self.env.context.get("force_copy_stock_moves")
+                and "move_line_ids" not in vals
+            ):
```
NachoAlesLopez: This condition can be placed outside the `for` loop, which can save a bit of time if there are many records.

```suggestion
        if self.env.context.get("force_copy_stock_moves"):
            for record, vals in zip(self, vals_list, strict=False):
                if "move_line_ids" not in vals:
```

stock_picking_invoice_link/models/account_move.py
```diff
@@ -0,0 +1,86 @@
...
+                vals["move_line_ids"] = [(6, 0, record.move_line_ids.ids)]
```
NachoAlesLopez: Can you replace this set operation with a `Command`?

```suggestion
                vals["move_line_ids"] = Command.set(record.move_line_ids.ids)
```

NachoAlesLopez (commented):
Functionally it works perfectly fine. I've left some notes regarding small details in code

luc-adhoc (approved):
functional review

PR #1759, opened 2024-11-07
[14.0][FIX] stock_picking_invoice_link: inv duplication: copy false

DiegoParadeda:
When copying an invoice, the related stock move relations are currently being kept. I believe this is not the intended behavior


DiegoParadeda:
The failing tests are suggesting otherwise, did I miss something?

rvalyi (approved):

PR #1760, opened 2024-11-08
[16.0][IMP] stock_move_line_lock_qty_done: prevent test cases from other modules from failing

AungKoKoLin1997:
Before this PR, if we had other test cases that updated the done quantity after the transfer was done in our module, they would be blocked by `stock_move_line_lock_qty_done`. This PR addresses that issue.

@qrtl QT4650

### DIFF
--- a/stock_move_line_lock_qty_done/models/res_company.py
+++ b/stock_move_line_lock_qty_done/models/res_company.py
@@ -1,0 +1,14 @@
+# Copyright 2024 Quartile
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    lock_qty_done = fields.Boolean(
+        string="Limit Updates to Done Quantity After Validation",
+        help="Only users in the 'Can Edit Done Quantity for Done Stock Moves' group"
+        " are allowed to edit the 'done' quantity for validated transfer.",
+    )
--- a/stock_move_line_lock_qty_done/models/res_config_settings.py
+++ b/stock_move_line_lock_qty_done/models/res_config_settings.py
@@ -1,0 +1,16 @@
+# Copyright 2024 Quartile
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    lock_qty_done = fields.Boolean(
+        related="company_id.lock_qty_done",
+        readonly=False,
+        string="Limit Updates to Done Quantity After Validation",
+        help="Only users in the 'Can Edit Done Quantity for Done Stock Moves' group"
+        " are allowed to edit the 'done' quantity for validated transfer.",
+    )
--- a/stock_move_line_lock_qty_done/models/stock_move_line.py
+++ b/stock_move_line_lock_qty_done/models/stock_move_line.py
@@ -16,7 +16,7 @@ class StockMoveLine(models.Model):
             "stock_move_line_lock_qty_done.group_stock_move_can_edit_done_qty"
         )
         for rec in self:
-            if rec.state != "done":
+            if rec.state != "done" or not rec.company_id.lock_qty_done:
                 continue
             if rec.is_locked:
                 raise UserError(
--- a/stock_move_line_lock_qty_done/readme/CONFIGURE.rst
+++ b/stock_move_line_lock_qty_done/readme/CONFIGURE.rst
@@ -1,2 +1,4 @@
-To configure this module, you need to add the users allowed to edit the done quntity
-for done moves to the group "Can edit done quantity for done stock moves"
+Update the following settings:
+
+* Go to *Inventory > Settings* and select the 'Limit Updates to Done Quantity After Validation' option to enable the restriction. This should be done for each company.
+* Add users who are allowed to edit the done quantity for done moves to the group 'Can edit done quantity for done stock moves'.
--- a/stock_move_line_lock_qty_done/README.rst
+++ b/stock_move_line_lock_qty_done/README.rst
@@ -28,8 +28,8 @@ Lock Done Quantity Changes in Stock Moves
 
 |badge1| |badge2| |badge3| |badge4| |badge5|
 
-This module limits the capability to modify the done quantity for stock moves
-once they have been validated.
+This module can limit the ability to modify the done quantity of stock moves once they are validated,
+configurable per company through the settings.
 
 **Table of contents**
 
@@ -39,8 +39,10 @@ once they have been validated.
 Configuration
 =============
 
-To configure this module, you need to add the users allowed to edit the done quntity
-for done moves to the group "Can edit done quantity for done stock moves"
+Update the following settings:
+
+* Go to *Inventory > Settings* and select the 'Limit Updates to Done Quantity After Validation' option to enable the restriction. This should be done for each company.
+* Add users who are allowed to edit the done quantity for done moves to the group 'Can edit done quantity for done stock moves'.
 
 Bug Tracker
 ===========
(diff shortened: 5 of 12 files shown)


AungKoKoLin1997:
I assume the test cases failed from `stock_move_line_auto_fill` in this PR is not related with the changes because I see the other PR is facing the same issue.

stock_move_line_lock_qty_done/models/stock_move_line.py
```diff
@@ -3,13 +3,20 @@
...
 
     _inherit = "stock.move.line"
 
     def _check_qty_done_change_allowed(self, vals):
+        check_qty_done_change_allowed = (
+            config["test_enable"]
+            and self.env.context.get("test_stock_move_line_lock_qty_done")
+        ) or not config["test_enable"]
+        if not check_qty_done_change_allowed:
+            return True
```
yostashiro: Shouldn't this suffice?
```suggestion
        if config["test_enable"] and not self.env.context.get("test_stock_move_line_lock_qty_done"):
            return True
```

yostashiro (approved):
LGTM.

stock_move_line_lock_qty_done/models/stock_move_line.py
```diff
@@ -3,13 +3,18 @@
...
 from odoo.exceptions import UserError
+from odoo.tools import config
 
 
 class StockMoveLine(models.Model):
 
     _inherit = "stock.move.line"
 
     def _check_qty_done_change_allowed(self, vals):
+        if config["test_enable"] and not self.env.context.get(
```
rousseldenis: I know this is an easy workaround that flourishes everywhere but it's not a good practice.

Introduce a configuration setting that allows to enable or disable the feature instead.

Thanks !

rousseldenis (changes requested):
See comment

rousseldenis:
@sbejaoui

stock_move_line_lock_qty_done/models/stock_move_line.py
yostashiro: @rousseldenis I thought you might say so. ;)

AungKoKoLin1997:
@sbejaoui Can you please review?

stock_move_line_lock_qty_done/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <!-- System Parameter for Enabling Move Line Lock Qty Done Check -->
+    <record id="param_move_line_lock_qty_done_enable" model="ir.config_parameter">
+        <field
+            name="key"
+        >stock_move_line_lock_qty_done.move_line_lock_qty_done_enable</field>
```
yostashiro: Non-blocking. Just thought the key could be shortened.
```suggestion
        >stock_move_line_lock_qty_done.lock_qty_done</field>
```

yostashiro (approved):
LGTM. 👍

stock_move_line_lock_qty_done/models/res_config_settings.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2024 Quartile
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    lock_qty_done = fields.Boolean(
```
rousseldenis, line 10: I'd prefer a related on a company field.

That will allow you to avoid `get/set_values` calls.

AungKoKoLin1997: If we are related on a company field instead of system parameter and enable this feature by default, do you want to set `lock_qty_done` to `false` at the end of the test cases of the module to prevent this feature from other test cases? May be I don't get your intention.

rousseldenis: No.

From a functional point of view, enabling or disabling the feature per company is a good practice.
From a technical point of view (that's a matter of taste), I don't like such writings. Moreover, in v16, you have the field attribute `config_parameter`that allows to avoid such `get/set`.

stock_move_line_lock_qty_done/demo/demo_company.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="base.main_company" model="res.company">
+        <field name="lock_qty_done">False</field>
+    </record>
```
yostashiro: To disable the feature for all companies.
```suggestion
    <function model="res.company" name="write">
        <value model="res.company" search="[]" />
        <value eval="{'lock_qty_done': False}" />
    </function>
```

stock_move_line_lock_qty_done/models/stock_move_line.py
```diff
@@ -10,6 +10,8 @@ class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
 
     def _check_qty_done_change_allowed(self, vals):
+        if not self.company_id.lock_qty_done:
+            return True
```
yostashiro:
```suggestion
```
`self` is a recordset. Apply the check inside the `for` loop.

yostashiro (changes requested):

stock_move_line_lock_qty_done/models/res_company.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    lock_qty_done = fields.Boolean(
+        default=True,
```
rousseldenis: But last question, why enabling it by default?

IMHO, that should be done on your project side.

AungKoKoLin1997: My thought is that users install this module because they want to use its features. So, requiring them to change the settings after installing the module feels a bit strange to me.

stock_move_line_lock_qty_done/demo/demo_company.xml
AungKoKoLin1997: I updated the code.

stock_move_line_lock_qty_done/models/stock_move_line.py
AungKoKoLin1997: Updated the code.

stock_move_line_lock_qty_done/models/res_company.py
rousseldenis: That has been always a debate. But that changes Odoo standard behavior. Which is not always wanted. That's why enabling the feature on deployment is better.

And you avoid disabling it in demo data

AungKoKoLin1997: I updated the code.

AungKoKoLin1997:
@yostashiro Please review the PR with the latest changes.

rousseldenis (approved):

stock_move_line_lock_qty_done/readme/CONFIGURE.rst
```diff
@@ -1,2 +1,4 @@
 To configure this module, you need to add the users allowed to edit the done quntity
 for done moves to the group "Can edit done quantity for done stock moves"
+
+Enable the 'Limit Updates to Done Quantity After Validation' option under 'Inventory > Settings' to enable the restriction of modifying the done quantity for validated stock moves.
```
yostashiro:
```suggestion
Update the following settings:

* Go to *Inventory > Settings* and select 'Limit Updates to Done Quantity After Validation' option to enable the restriction. This should be done for each company.
* Add users allowed to edit the done quantity for done moves to the group 'Can edit done quantity for done stock moves'.
```

yostashiro (approved):
LGTM.

PR #1766, opened 2024-11-15
[FIX] allow_negative_stock field visibility in product form view

moaadbourhim:
In Odoo v18, the handling of trackable (storable) products has been updated. This PR modifies the condition for displaying the allow_negative_stock field in the product.template form view.

Before Update:
The form view of product.template before applying the update:
Screenshot from 2024-11-15 14-56-40

After Update:
The form view of product.template after applying the update:
Screenshot from 2024-11-15 14-55-27

### DIFF
--- a/stock_no_negative/views/product_product_views.xml
+++ b/stock_no_negative/views/product_product_views.xml
@@ -11,7 +11,10 @@
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
             <field name="categ_id" position="after">
-                <field name="allow_negative_stock" invisible="type != 'product'" />
+                <field
+                    name="allow_negative_stock"
+                    invisible="type != 'consu' or is_storable == False"
+                />
             </field>
         </field>
     </record>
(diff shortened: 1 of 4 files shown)


rousseldenis:
@moaadbourhim Thanks for this.

In fact, this is not a migration but a fix.

Could you reword your commit to `[FIX] ...` and PR's title accordingly ?

moaadbourhim:
@rousseldenis Done

stock_no_negative/views/product_product_views.xml
```diff
@@ -11,7 +11,10 @@
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
             <field name="categ_id" position="after">
-                <field name="allow_negative_stock" invisible="type != 'product'" />
+                <field
+                    name="allow_negative_stock"
+                    invisible="type != 'consu' or is_storable == False"
```
Vang-NguyenPhu, line 16: I think like this `invisible="not is_storable"` is enough, right?
In the compute func of `is_storable` already filtered `t.type != 'consu'`.

Vang-NguyenPhu (changes requested):

StefanRijnhart:
Same fix merged in https://github.com/OCA/stock-logistics-workflow/pull/1830. Thank you for your contribution though!

PR #1776, opened 2024-11-21
[16.0][IMP] stock_account_product_run_fifo_hook: add hooks that allows to change unit cost layer precision

AaronHForgeFlow:
Set precision Product Price 4 digits and change unit_cost field in layers to that precision (that is standard Odoo change in v17)

This is causing many rounding issues when using higher Product Price precision. 

Example 1:

Product price 8.2833 and qty 10
Remove one unit: it is removed @ 8.2833 unit_cost --> it is removed @ 8.2800 Add oen unit back: it is added @ 8.2833 unit_cost

If you do this many times you are adding value incorrectly

Example 2:

Product Price 10.0033 and qty 10
Change product price to 8.2833
The system is not adding any value to your valuation

Change product price to 8.2844
The system is adding 0.01 to your valuation

In my opinion when you use the unit cost like this: https://github.com/odoo/odoo/blob/16.0/addons/stock_account/models/product.py#L354

when the reamining value is rounded already to 2, you are creating rounding issues in your valuation. It makes no sense to have separate unit of measure when you are using this value in the calculation

Ok, value makes sense to be Monetary, but then do use the full number in the calculation otherwise it will generate many unit cost rounding issues.

Ok, if I don't change the precision of the unit cost in the layer (remember that is standard change in v17) this issue is not happening as the unit cost is already rounded and the rounding issues are less than 0.01. However, the hooks make sense, don't they?

Ok, you won't merge this, but at least let me argue a bit.

### DIFF
--- a/stock_account_product_run_fifo_hook/hooks.py
+++ b/stock_account_product_run_fifo_hook/hooks.py
@@ -4,7 +4,8 @@ from collections import defaultdict
 from datetime import datetime
 
 from odoo import _
-from odoo.tools import float_is_zero
+from odoo.exceptions import UserError
+from odoo.tools import clean_context, float_compare, float_is_zero, float_round
 
 from odoo.addons.stock_account.models.product import ProductProduct
 from odoo.addons.stock_account.models.stock_move import StockMove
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
Example in runbot v16:

- Define a product with unit cost 8.2833:

image


- Add 40 units in stock with an inventory adjustment. 40*82833 = 331.332 **Odoo is ok**:

image

- Define 3 consecutive inventory adjustments, removing 1 unit each. the result is 37*8.2833 = 306.4821. **Odoo is wrong**

image

- Repeat the process several times to leave 1 unit in stock at 8.2833. **Odoo is wrong again**:
image


This is due incorrect rounding using the currency digital precision and we can inherit all those hooks in the custom module and round to the precision of the Product Price and not the currency. It is really worth it? maybe not, but this case is not so uncommon, and the errors can be more than $0.01

AaronHForgeFlow:
BTW this is not necessary in v17. Odoo fixed that in code. In the unlikely event of merging this, the forward port is not needed.

PR #1649, opened 2024-07-08
[16.0][ADD] stock_account_fifo_return_origin

AungKoKoLin1997:
This module makes adjustment to the standard behavior with regards to candidate SVLs that are used in purchase returns; if the SVL linked to the origin move (receipt) has remaining quantity, that SVL should be consumed first.

@qrtl QT4657

PR #1777, opened 2024-11-22
[16.0][ADD] stock_move_value

AungKoKoLin1997:
This module adds value fields to the stock move model, to add visibility of how the move has affected the stock valuation.

@qrtl QT4685


yostashiro (commented):
@AungKoKoLin1997 Can you please add a CONTEXT.md as the intention is probably not clear to most people.

"""
Odoo calculates the cost of outgoing stock moves based on the product's costing method, no matter what the cause of the operation is. This can be problematic especially when a user just reverts the purchase receipt operation which was done by mistake (e.g., processing the wrong receipt picking) for FIFO and AVCO products, as the returned value is not the same as the value of the original receipt. Some auditors might see such inconsistencies as the result of intended manipulation on the financial performance.

For FIFO products, you can use the OCA module stock_account_fifo_return_origin (yet to be added with https://github.com/OCA/stock-logistics-workflow/issues/1649) to return the same value as the receipt.  However, there is no complete solution for AVCO products.

This module aims to provide users with a means to identify discrepancies resulting from their purchase return operations, allowing them to take necessary actions (e.g., inventory revaluation) as appropriate.
"""

AungKoKoLin1997:
@yostashiro I added the `CONTEXT.md`.

yostashiro (approved):
Code review and functional test.

kanda999 (approved):
Code and functional review

yostashiro (approved):

kanda999 (approved):

PR #1780, opened 2024-11-27
[14.0] [ADD] stock_putaway_last_location

paradoxxxzero:
This module allows you to use the last putaway location of a product as the default putaway location.


sebalix:
@paradoxxxzero I pushed a commit here to extract the move line domain in a separate method

PR #1789, opened 2024-12-11
[16.0][FIX] stock_picking_group_by_partner_by_carrier: merged procurement group should not be linked to the original sales order

sbejaoui:
When a merged procurement group is created, it should not be linked to the original sales order. Linking it would cause any moves manually added to the delivery picking to be associated with that sales order.

The first commit reproduce the bug and show how the manual add of a move to the picking impact the sale order.

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
@@ -111,7 +111,7 @@ class StockPicking(models.Model):
     def _prepare_merge_procurement_group_values(self, move_groups):
         """Build a new procurement group that is the merge of given procurement
         group."""
-        sales = move_groups.sale_id
+        sales = move_groups.sale_id + move_groups.sale_ids
         partners = move_groups.sale_id.partner_id
         name = _("Merged procurement")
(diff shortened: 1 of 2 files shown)


sbejaoui:
cc/ @rousseldenis , @lmignon

lmignon (approved):
Thank you for this FIX. Indeed, without this a merged procurement could introduce hard to understand unexpected behaviour. LGTM (Code review)

rousseldenis (approved):

lauradiaz22 (approved):
Good morning,
I have reviewed this FIX at a functional level and this LGTM👍

jbaudoux (changes requested):
You cannot empty the `sale_id` field as it is used by many module (enterprise) and few OCA in delivery.
I logged what is I think the real issue https://github.com/OCA/stock-logistics-workflow/issues/1833

jbaudoux:
@lmignon @rousseldenis This needs to be reworked

lmignon:
> @lmignon @rousseldenis This needs to be reworked

@jbaudoux Ticket created into our issue tracker to don't forget it..

PR #1796, opened 2024-12-19
[16.0][IMP] stock_picking_mass_action: imp reservation logic from odoo base

alvaro-gmz:
The check_assign_all() method in the stock.picking model of the stock_picking_mass_action module has been enhanced to align with Odoo's stock reservation logic.

Changes implemented:
- Assignement at stock.move level:
    - Availability checks now respect the domain defined by Odoo's scheduler logic.
- Custom domain
    - The planned action accepts a configurable domain(at stock.move level)t o adapt the movements to be processed
- Batch processing
    - Added support for processing moves in configurable batches, with a default size of 1000

### DIFF
--- a/stock_picking_mass_action/models/stock_picking.py
+++ b/stock_picking_mass_action/models/stock_picking.py
@@ -2,23 +2,35 @@
 # Copyright 2018 Tecnativa - Vicent Cubells
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+
 from odoo import _, api
 from odoo.models import Model
+from odoo.osv import expression
 
 
 class StockPicking(Model):
     _inherit = "stock.picking"
 
     @api.model
-    def check_assign_all(self, domain=None):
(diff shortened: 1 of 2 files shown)


RodrigoBM (approved):

stock_picking_mass_action/models/stock_picking.py
```diff
@@ -2,23 +2,35 @@
...
+        move_assign_domain = self.env["procurement.group"]._get_moves_to_assign_domain(
+            company
+        )
         if domain:
-            search_domain += domain
-        else:
-            search_domain += [("picking_type_code", "=", "outgoing")]
-        records = self.search(search_domain, order="scheduled_date")
-        records.action_assign()
+            move_assign_domain = expression.AND([domain, move_assign_domain])
```
jbaudoux, line 23: I don't know what it calling this method but I guess the domain is a picking domain and not a move domain. So you cannot do such

jbaudoux (commented):
I don't think you can do such change

cc @wouitmil

PR #1797, opened 2024-12-19
 [17.0][MIG] stock_auto_move 

ferran-S73:
Standard migration. Deleted the `_get_auto_moves_by_pickings` method because it was no longer used in this module


stock_auto_move/tests/test_stock_auto_move.py
```diff
@@ -69,8 +69,11 @@ def test_10_auto_move(self):
             }
         )
         move3._action_confirm()
+        self.env["stock.quant"]._update_available_quantity(
```
rousseldenis, line 72: @ferran-S73 Why is this necessary ?

ferran-S73: @rousseldenis The move was not reserving stock because there was no available stock of the product. Maybe in older versions some other module added stock for the product?

ferran-S73: @rousseldenis can you review please?

ferran-S73:
@ThomasBinsfeld @hbrunn could you review please?

stock_auto_move/models/stock_move.py
```diff
@@ -15,7 +15,7 @@ class StockMove(models.Model):
 
     def _auto_assign_quantities(self):
         for move in self:
-            move.quantity_done = move.reserved_availability
+            move.picked = True
```
hbrunn, line 18: please explain this change, I'd have expected a call to `_set_quantity_done`

ferran-S73: @hbrunn Why would I use that function? The move should already be reserved thus having quantity. We need to set the picked field so that `_action_done` can validate the move after this

hbrunn: yes right, that was v16 thinking

hbrunn (approved):

hildickethan (approved):

PR #1799, opened 2024-12-19
[16.0][FIX] stock_picking_return_restricted_qty: remove _onchange_quantity method

ced-adhoc:
This fix replaces the UserError in _onchange_quantity with a warning message. Now, users receive a non-blocking notification when attempting to return more products that delivered, but allows them to modify the quantity before proceeding.

The restriction logic remains unchanged for _create_returns, ensuring hard enforcement at the validation stage.

- Improves UX by preventing unnecessary hard errors on simple mistakes.
- Maintains the business rule enforcement while allowing users to correct their input easily.

### DIFF
--- a/stock_picking_return_restricted_qty/wizard/stock_picking_return.py
+++ b/stock_picking_return_restricted_qty/wizard/stock_picking_return.py
@@ -49,7 +49,11 @@ class ReturnPickingLine(models.TransientModel):
         qty = self.get_returned_restricted_quantity(self.move_id)
 
         if restrict_return_qty and self.quantity > qty:
-            raise UserError(_("Return more quantities than delivered is not allowed."))
+            warning = {
+                "title": ("Warning!"),
+                "message": ("Return more quantities than delivered is not allowed."),
+            }
+            return {"warning": warning}
 
     def get_returned_restricted_quantity(self, stock_move):
         """This function is created to know how many products
--- a/stock_picking_return_restricted_qty/tests/test_stock_picking_return_restricted_qty.py
+++ b/stock_picking_return_restricted_qty/tests/test_stock_picking_return_restricted_qty.py
@@ -102,7 +102,7 @@ class StockPickingReturnRestrictedQtyTest(common.TransactionCase):
 
         wiz.product_return_moves.quantity = 80
         with self.assertRaises(UserError):
-            wiz.product_return_moves._onchange_quantity()
+            wiz._create_returns()
 
     def test_multiple_return_without_restriction(self):
         """On this test we are going to follow a sequence that a client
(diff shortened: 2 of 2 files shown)


luc-adhoc (approved):
functional review

mav-adhoc (approved):

stock_picking_return_restricted_qty/wizard/stock_picking_return.py
```diff
@@ -40,17 +40,6 @@ def _create_returns(self):
...
-    @api.onchange("quantity")
-    def _onchange_quantity(self):
-        restrict_return_qty = (
-            self.move_id.picking_id.picking_type_id.restrict_return_qty
-        )
-
-        qty = self.get_returned_restricted_quantity(self.move_id)
-
-        if restrict_return_qty and self.quantity > qty:
-            raise UserError(_("Return more quantities than delivered is not allowed."))
```
rousseldenis, line 52: @ced-adhoc You remove here a fucntionality. Maybe in order to get user warned before validating the return should be to add an alert on the wizard form (if it is possible) or by other mean.

rousseldenis (changes requested):

mav-adhoc:
@rousseldenis What's the point to put the UserError on an onchange_quantity method if either way the quantity restriction will be validated when clicking on the return button? It looks redundant. It makes any difference? Sorry to ask! I was just wondering because it doesn't make too much sense to me.

ced-adhoc:
@rousseldenis Hi! I'm taking up this task based on you suggestion. I replaced the UserError with a warning in the onchange quantity method, and it seems to work as expected. Could you please review it again and let me know your thoughts? Thanks!

mav-adhoc (approved):

stock_picking_return_restricted_qty/wizard/stock_picking_return.py
ced-adhoc: @rousseldenis Hi! I'm taking up this task based on you suggestion. I replaced the UserError with a warning in the onchange quantity method, and it seems to work as expected. Could you please review it again and let me know your thoughts? Thanks!

florenciafrigieri2 (approved):
Test working ok.
When the restriction is set in the operation type, Odoo is not allowing to return bigger quantities. If you have a mistake, it is possible to correct it.

ced-adhoc:
Hi @rousseldenis could you please check this fix? Thanks in advance!

ced-adhoc:
Hi @rousseldenis could you please check this? thanks!

pedrobaeza:
The idea to raise a `UserError` is to rollback the quantity to the previous one. With this change, it won't be rollbacked, and then, it's possible to validate a picking returning more quantity, which is not correct.

PR #1801, opened 2024-12-20
[16.0][ADD] stock_move_negative_stock_location

santostelmo:
### Context
When making an stock move with a negative quantity, odoo convert it to a positive return move and the locations are inverted.
For stock moves with negative quantities, this module sets the source location with the return 
picking type default destination location. Like that destination location is properly set on the return move.

### Usage
In `Returns` picking type set the default destination location as `Stock`
Create a stock move from Stock to Customer with a negative quantity and validate it. 
That move will be converted into a stock move from Customer to Stock with a positive quantity.


rousseldenis:
@santostelmo @jbaudoux Thanks for this. But, shouldn't it be a "bug" in stock module behavior ?

santostelmo:
> @santostelmo @jbaudoux Thanks for this. But, shouldn't it be a "bug" in stock module behavior ?

You're absolutely right ! I am opening a issue to odoo. In the meantime we have this module to workaround.

stock_move_negative_stock_location/models/stock_move.py
```diff
@@ -0,0 +1,43 @@
...
+    def _check_company(self, fnames=None):
+        # Set the location_id of negative return moves to the default location_dest_id
+        # of the return picking type
+        # We hook into this method used in stock.move._action_confirm
+        # (see https://github.com/odoo/odoo/blob/
+        # 3a63c90fff615b70881131e11d7375af2ae082a6/addons/stock/models/stock_move.py
+        # #L1381C14-L1381C27)
+        # to be able to invert location_id and location_dest_id
+        if self.env.context.get("neg_r_moves"):
+            self.set_negative_return_moves_location()
```
sebalix, line 42: And handle the context only where it is needed (if in later version Odoo adds a hook for instance, we could trash easily this context handling without touching the business method `set_negative_return_moves_location`:
```suggestion
            neg_moves = self.browse(self.env.context.get("neg_r_moves")).exists()
            neg_moves.set_negative_return_moves_location()
```

stock_move_negative_stock_location/models/stock_move.py
```diff
@@ -0,0 +1,43 @@
...
+            < 0
+        )
+        self = self.with_context(neg_r_moves=neg_r_moves.ids)
+        return super()._action_confirm(merge=merge, merge_into=merge_into)
+
+    def set_negative_return_moves_location(self):
+        """Set the location_id of negative return moves to the default location_dest_id
+        of the return picking type."""
+        neg_r_moves = self.browse(self.env.context["neg_r_moves"])
+        for move in neg_r_moves:
```
sebalix, line 27: Better to get methods not depending on context but on current recordset, easier to test and understand:
```suggestion
        for move in self:
```

stock_move_negative_stock_location/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+When making an stock move with a negative quantity, odoo convert it to a positive return move and the locations are inverted.
```
sebalix, line 1:
```suggestion
When making a stock move with a negative quantity, Odoo converts it to a positive return move and the locations are inverted.
```

sebalix (commented):
Few comments.
For sure, having this fixed in core would be better :)
If not possible in `16.0`, maybe in `master`.

stock_move_negative_stock_location/tests/test_stock_move_negative_stock_location.py
```diff
@@ -0,0 +1,54 @@
...
+        # WH: Stock → Customers
+        rule = self.env["stock.rule"].search(
+            [
+                ("action", "=", "pull"),
+                ("picking_type_id", "=", self.env.ref("stock.picking_type_out").id),
+                ("location_src_id", "=", location_source.id),
+                ("location_dest_id", "=", location_dest.id),
+                ("procure_method", "=", "make_to_stock"),
+            ]
+        )
```
sebalix, line 39: These config data can be retrieved in `setUpClass`, to keep transactional data and asserts only in the test

cyrilmanuel:
@santostelmo hi, could you check the status from the odoo ticket and here to define a solution please ?

PR #1815, opened 2024-12-31
[18.0][MIG] product_expiry_simple

alexis-via:
1 thing left to do: inherit the JS widget to generate lots from serial ranges to add support for expiry date. In v16, it was a wizard ; but in v18, it's in javascript, cf

https://github.com/odoo/odoo/blob/18.0/addons/stock/static/src/widgets/lots_dialog.xml
https://github.com/odoo/odoo/blob/18.0/addons/stock/static/src/widgets/generate_serial.js


rousseldenis:
@alexis-via Any news on this ?

francesco-ooops:
@alexis-via please please please add a USAGE file to this module, it's annoying for functionals to have to ask devs "how do I use this?"

PR #1817, opened 2025-01-03
[16.0][FIX] stock_picking_restrict_cancel_printed: fix no backorder

jbaudoux:
Allow to validate a partial picking when the backorder policy is set to never


Ricardoalso (approved):

jbaudoux:
cc @sebalix @sbejaoui

PR #1461, opened 2023-12-22
[14.0][IMP/ADD] stock_product_variant_mto, sale_stock_mto_as_mts_orderpoint_product_variant: Handle mto route at variant level

mmequignon:
I have to move commits in dedicated PRs, but pushed everything here for the time being

PR #1819, opened 2025-01-06
[18.0][MIG] stock_product_variant_mto: Migration to 18.0

chaule97:
- I format-patch from https://github.com/OCA/stock-logistics-workflow/pull/1461
- I updated `route_ids` `stored=True` because [this code](https://github.com/odoo/odoo/blob/bfab889e4d1578092e03e50a893773d549c78648/addons/stock/models/product.py#L667C9-L667C98) made error `Non-stored field product.product.route_ids cannot be searched. `


stock_product_variant_mto/models/product_product.py
```diff
@@ -23,30 +23,31 @@ class ProductProduct(models.Model):
     )
 
     route_ids = fields.Many2many(
-        "stock.location.route",
+        "stock.route",
         compute="_compute_route_ids",
         domain="[('product_selectable', '=', True)]",
-        store=False,
+        store=True,
```
grindtildeath: Can't we define a search function instead?

stock_product_variant_mto/models/product_product.py
```diff
@@ -23,30 +23,31 @@ class ProductProduct(models.Model):
...
-            if mto_route and mto_route in product.product_tmpl_id.route_ids:
-                if not product.is_mto:
-                    product.route_ids = product.product_tmpl_id.route_ids - mto_route
-                    continue
-            else:
-                if mto_route and product.is_mto:
-                    product.route_ids = product.product_tmpl_id.route_ids + mto_route
-                    continue
-            product.route_ids = product.product_tmpl_id.route_ids
+            new_route_ids = product.product_tmpl_id.route_ids
```
grindtildeath: Nitpicking a little, but this is a recordset, not a list of ids. Also these ain't new routes but the routes from the template. I would call this variable something like `template_routes` or `product_routes`.

However, all this rewriting does not bring anything in the context of module migration as the conditions are the same as before... IMO it only adds more work for the reviewer...

stock_product_variant_mto/models/product_template.py
```diff
@@ -1,58 +1,72 @@
...
-        (
+        templates_mto_added = templates_not_mto_before & templates_mto_after
+        templates_mto_removed = self - templates_mto_after - templates_not_mto_before
+
+        affected_product_variants = (
             templates_mto_added | templates_mto_removed
-        ).product_variant_ids._compute_is_mto()
+        ).product_variant_ids
+        if affected_product_variants:
+            affected_product_variants._compute_is_mto()
```
grindtildeath: Basically these are pure cosmetic change that will only obfuscate the original contribution when blaming the code, I don't think it must be part of a migration.

grindtildeath (commented):
Thanks for the mig.

However, IMO we should keep migration commit to a minimum in order to highlight the required changes and avoid rewriting for cosmetic changes or fixing typos

stock_product_variant_mto/models/product_product.py
chaule97: I think not, because compute can track when `is_mto` was changed

rousseldenis:
@chaule97 @grindtildeath @jbaudoux Maybe it should have been great to merge this one with that one : https://github.com/OCA/product-attribute/tree/16.0/product_route_mto

stock_product_variant_mto/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,2 @@
+This module allows to define if a product variant can use the Make To
+Order route without any dependency on its template routes settings.
```
jbaudoux, line 2: Add a big warning that the routes are moved from product template to product product !

jbaudoux:
> @chaule97 @grindtildeath @jbaudoux Maybe it should have been great to merge this one with that one : https://github.com/OCA/product-attribute/tree/16.0/product_route_mto

Certainly NOT. However, this module should be in product-attribute repo and I would rename it `product_variant_route_mto`

chaule97:
> Certainly NOT. However, this module should be in product-attribute repo and I would rename it `product_variant_route_mto`

Do you or I move it?

jbaudoux:
Close this PR and open a new one on product attribute

chaule97:
Hello everyone, I move it to https://github.com/OCA/product-attribute/pull/1879, please review it in new PR, thanks

PR #1825, opened 2025-01-08
[16.0][IMP] stock_move_line_change_lot: add hooks before/after moves reservation

sebalix:
With the extraction of `shopfloor` "change lot" action into this `stock_move_line_change_lot` module in 16.0 (see https://github.com/OCA/wms/commit/51d5f4ae25), we lost one feature (which was still not merged in 14.0 at that time, but now it is): addition of hooks before/after the reassignment of moves.

This PR ports these hooks that were introduced in 14.0 in `shofploor` here:
- https://github.com/OCA/wms/pull/712/commits/c296499726f0f95082bfe40bfb0c137dfd2fb690
- https://github.com/OCA/wms/pull/945/commits/28d4a0116e9c891cd1f37e794e3a7643798260ac

### DIFF
--- a/stock_move_line_change_lot/models/stock_move_line.py
+++ b/stock_move_line_change_lot/models/stock_move_line.py
@@ -187,7 +187,9 @@ class StockMoveLine(models.Model):
                 move_line.move_id._recompute_state()
                 already_processed |= move_line
 
-        res &= super(StockMoveLine, self - already_processed).write(vals)
+        still_todo = self - already_processed
+        if still_todo:
+            res &= super(StockMoveLine, still_todo).write(vals)
         if to_reassign_moves:
             self._handle_change_lot_reassign(
                 lot, to_reassign_moves, moves_by_previous_lot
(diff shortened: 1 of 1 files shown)


stock_move_line_change_lot/models/stock_move_line.py
```diff
@@ -161,6 +165,25 @@ def write(self, vals):
         res &= super(StockMoveLine, self - already_processed).write(vals)
 
         if to_reassign_moves:
-            to_reassign_moves._action_assign()
+            self._handle_change_lot_reassign(
```
sebalix, line 194: Put in a separate method to avoid flake8 issue:
`stock_move_line_change_lot/models/stock_move_line.py:64:5: C901 'StockMoveLine.write' is too complex (17)`

simahawk:
@jbaudoux ping

simahawk:
Let's give precedence to https://github.com/OCA/stock-logistics-workflow/pull/1970

then here you can find the adaptation of this PR to that one https://github.com/camptocamp/stock-logistics-workflow/pull/new/16-change-lot-hooks

simahawk:
Rewritten on top of #1970  + added check on lines still todo before calling write.

sebalix:
LG

jbaudoux (approved):

PR #1828, opened 2025-01-10
[18.0][MIG] sale_stock_mto_as_mts_orderpoint: Migration to 18.0

chaule97:
-  because `stock_orderpoint_manual_procurement` is [deprecated](https://github.com/OCA/stock-logistics-warehouse/pull/1641#issuecomment-1445054077), so I remove it in depends and write more code about `stock_warehouse_orderpoint`


rousseldenis:
@chaule97 Thanks for this. Orderpoints related module have migrated to : https://github.com/OCA/stock-logistics-orderpoint

Thanks to do the PR there.

PR #1830, opened 2025-01-14
[18.0][FIX] stock_no_negative: fix field not visible in view

woutswennen:
Oca module didnt work as we where not able to sellect the allow_negative_stock field on product

### DIFF
--- a/stock_no_negative/views/product_product_views.xml
+++ b/stock_no_negative/views/product_product_views.xml
@@ -10,9 +10,9 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
-            <field name="categ_id" position="after">
-                <field name="allow_negative_stock" invisible="type != 'product'" />
-            </field>
+            <xpath expr="//field[@name='is_storable']/parent::div" position="after">
+                <field name="allow_negative_stock" invisible="not is_storable" />
+            </xpath>
         </field>
     </record>
     <record id="product_category_form_view_inherit" model="ir.ui.view">


stock_no_negative/views/product_product_views.xml
```diff
@@ -11,7 +11,7 @@
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
             <field name="categ_id" position="after">
-                <field name="allow_negative_stock" invisible="type != 'product'" />
+                <field name="allow_negative_stock" invisible="type == 'service'" />
```
StefanRijnhart: `not is_storable`, I think. See https://github.com/odoo/odoo/commit/728d9f83f6d14

StefanRijnhart (changes requested):

stock_no_negative/views/product_product_views.xml
StefanRijnhart: Also, might be a nice improvement to move the field until after `is_storable` itself.
image

rousseldenis (changes requested):
@woutswennen You should follow migration guide here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#technical-method-to-migrate-a-module-from-170-to-180-branch

Especially, you should keep commits history. 

Thanks

StefanRijnhart:
Ouch, that history was there before but it was squashed into oblivion.

StefanRijnhart:
Wait, this is not the migration, it's a fix of the migration. @woutswennen Please update commit message and PR name accordingly.

StefanRijnhart:
Also, duplicate of https://github.com/OCA/stock-logistics-workflow/pull/1766

woutswennen:
@StefanRijnhart Looks like there the previous MR is a bit dead. I changed the commit message, anything else that i should change?

StefanRijnhart:
@woutswennen We can go with this one. About the commit message: please include the module name, and drop the version number. Something like `[FIX] stock_no_negative: fix field not visible in view`. Please apply pre-commit as well.

StefanRijnhart (approved):

StefanRijnhart:
Looking good now!
image
image

BT-crodriguez (approved):
LGTM!

rousseldenis (approved):

PR #1832, opened 2025-01-16
[16.0][FIX] stock_lock_lot fix: fixed permissions issue when creating product categories

alvaro-gmz:
**Fix permissions validation when changing 'lot_default_locked' field**

The permission checks for modifying the lot_default_locked field has been improved. Now, the system compares the current value of lot_default_locked with the new value being set and ensures the user has the necessary permissions (stock_lock_lot.group_lock_lot) before allowing the change. This prevents unauthorized users from blocking or unblocking serial numbers/lots.

### DIFF
--- a/stock_lock_lot/models/product_category.py
+++ b/stock_lock_lot/models/product_category.py
@@ -2,7 +2,7 @@
 # Copyright 2015 AvanzOsc (http://www.avanzosc.es)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import _, api, exceptions, fields, models
+from odoo import _, exceptions, fields, models
 
 
 class ProductCategory(models.Model):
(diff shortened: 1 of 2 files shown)


stock_lock_lot/models/product_category.py
```diff
@@ -14,9 +14,15 @@ class ProductCategory(models.Model):
         "by default",
     )
 
-    @api.constrains("lot_default_locked")
-    def _check_category_lock_unlock(self):
-        if not self.user_has_groups("stock_lock_lot.group_lock_lot"):
+    def _check_category_lock_unlock(self, vals):
+        if self.lot_default_locked != vals and not self.user_has_groups(
```
rousseldenis: @FLAlvaroGomez My question here: why is the constrains() trigger if the value hasn't changed ?

alvaro-gmz: @rousseldenis Updated, now the write method checks if the attribute changes to avoid executing the _check_category_lock_unlock method unnecessarily

RodrigoBM: @rousseldenis can see the error in the video made in runbot.

[Runboat-builds.webm](https://github.com/user-attachments/assets/5293d83e-b1f7-4298-b84b-a4294458d247)

RodrigoBM: Did @rousseldenis see the video?

PR #1834, opened 2025-01-17
[16.0][FIX] stock_picking_restrict_cancel_printed: allow moves merge

jbaudoux:
Support validation of picking with quantity in excess. Odoo creates an extra move that gets merged and then canceled. No need to allow this cancellation.

@santostelmo @rousseldenis @mt-software-de

### DIFF
--- a/stock_picking_restrict_cancel_printed/tests/test_stock_picking_restrict_cancel_printed.py
+++ b/stock_picking_restrict_cancel_printed/tests/test_stock_picking_restrict_cancel_printed.py
@@ -65,3 +65,13 @@ class TestPickingRestrictCancel(TransactionCase):
         self.picking.move_ids.quantity_done = 1
         self.picking_type.create_backorder = "never"
         self.picking.button_validate()
+
+    def test_stock_move_restrict_cancel_printed_enabled_move_merge(self):
+        """Check a picking processed in excess can be validated
+
+        This will trigger the creation of an extra move.
+        That extra gets merged and then canceled.
+        """
+        self.picking.printed = True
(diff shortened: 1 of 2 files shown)


stock_picking_restrict_cancel_printed/tests/test_stock_picking_restrict_cancel_printed.py
```diff
@@ -65,3 +65,13 @@ def test_stock_move_restrict_cancel_printed_enabled_nobackorder(self):
...
         self.picking_type.create_backorder = "never"
         self.picking.button_validate()
+
+    def test_stock_move_restrict_cancel_printed_enabled_move_merge(self):
+        """Check a picking processed in excess can be validated
+
+        This will trigger the creation of an extra move.
+        That extra gets merged and then canceled.
+        """
+        self.picking.printed = True
```
mt-software-de, line 75: Shouldn't be `restrict_cancel_if_printed` set to true.
Otherwise it will always call super._action_cancel?
```suggestion
        self.picking_type.restrict_cancel_if_printed = True 
        self.picking.printed = True
```

mt-software-de (changes requested):
I think the test needs to be changed?

stock_picking_restrict_cancel_printed/tests/test_stock_picking_restrict_cancel_printed.py
jbaudoux: there is a default=True, so should be the case

mt-software-de (approved):

santostelmo (approved):

PR #1835, opened 2025-01-17
[15.0][MIG] stock_move_forced_lot

DavidJForgeFlow:
Standard migration to 15.0


LoisRForgeFlow (commented):
Could you fix the commit message? (17 -> 15)

rousseldenis:
@DavidJForgeFlow @LoisRForgeFlow Have you check this one ? https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_restrict_lot

Which is largely used through various versions

LoisRForgeFlow:
> @DavidJForgeFlow @LoisRForgeFlow Have you check this one ? https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_restrict_lot
> 
> Which is largely used through various versions

@DavidJForgeFlow it looks like it does the same thing, could you check if there is any problem for our use case? If no, we could migrate the dependencies to stock_restrict_lot

DavidJForgeFlow:
> @DavidJForgeFlow @LoisRForgeFlow Have you check this one ? https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_restrict_lot
> 
> Which is largely used through various versions

Seems stock_restrict_lot implements the same features as stock_move_forced_lot but also includes more tests and little details, so I will deprecate stock_move_forced_lot and migrate stock_restrict_lot to 15.0 and forward (if needed)

@LoisRForgeFlow @rousseldenis

PR #1839, opened 2025-01-20
[18.0][MIG] stock_picking_partner_note: Migration to 18.0

BhaveshHeliconia:


stock_picking_partner_note/readme/CONTRIBUTORS.md
```diff
@@ -2,3 +2,4 @@
 - Jacques-Etienne Baudoux \<<je@bcim.be>\>
 - [360ERP](https://www.360erp.com):
   - Andrea Stirpe
+- Heliconia Solutions Pvt. Ltd. \<<https://www.heliconia.io>\>
```
rousseldenis: @HeliconiaSolutions You can put your company here but as a title for an individual contributor (see here above)

BhaveshHeliconia: @rousseldenis Thank you for the feedback!
I’ve made the changes to include the company as a title for an individual contributor as per your suggestion.

jbaudoux (approved):

stock_picking_partner_note/models/stock_picking_note.py
```diff
@@ -0,0 +1,39 @@
...
+    _description = "Picking Note"
+    _order = "sequence,name"
+
+    name = fields.Text(required=True)
+    active = fields.Boolean(default=True)
+    note_type_id = fields.Many2one("stock.picking.note.type", required=True)
+    sequence = fields.Integer(related="note_type_id.sequence", store=True)
+
+    def write(self, vals):
+        self.check_note_already_in_use()
```
ivantodorovich: Shouldn't we check which fields are updated here?

For example, shouldn't we allow archiving a note (writing `{'active': False}`), even if it's already in use? 🤔 

Since this method is not covered tested, could we take the opportunity to write some?

ivantodorovich (approved):

stock_picking_partner_note/models/stock_picking_note.py
simahawk: @HeliconiaSolutions ping

BhaveshHeliconia: @ivantodorovich , @simahawk , Thanks for the suggestions! The issue is fixed now. Could you review it?

ivantodorovich: Hi @HeliconiaSolutions , I think there was a misunderstanding here

My point was that it should be **allowed** to archive a note, even if it's in use.
Seems you actually did the opposite (which was already the case in fact with the write override)

BhaveshHeliconia: @ivantodorovich, Thanks for pointing that out! I’ve updated the code to allow archiving a note even if it's in use, as intended. Let me know if there’s anything else!

simahawk (approved):

PR #1840, opened 2025-01-21
[14.0][REF] stock_restrict_lot: change AGPL-3 to LGPL-3

DavidJForgeFlow:
Hi @florian-dacosta,
I was migrating the `stock_move_forced_lot` module to 15.0 (https://github.com/OCA/stock-logistics-workflow/pull/1835), and @rousseldenis pointed out that your module `stock_restrict_lot` not only performs the same functionality but also offers more features and is used by more modules. As a result, I began updating the dependencies in some of our modules, such as RMA: https://github.com/ForgeFlow/stock-rma/pull/573.
However, I noticed a key difference between this module and `stock_move_forced_lot`: the license. I think that a module widely used as a base may need to change its license from AGPL to LGPL to encourage broader adoption by allowing its use in proprietary software without imposing the AGPL's strict copyleft requirements.
What do you think?

@LoisRForgeFlow

### DIFF
--- a/stock_restrict_lot/README.rst
+++ b/stock_restrict_lot/README.rst
@@ -13,9 +13,9 @@ Stock Restrict Lot
 .. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
     :target: https://odoo-community.org/page/development-status
     :alt: Beta
-.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
-    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
-    :alt: License: AGPL-3
+.. |badge2| image:: https://img.shields.io/badge/licence-LGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+    :alt: License: LGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fstock--logistics--workflow-lightgray.png?logo=github
     :target: https://github.com/OCA/stock-logistics-workflow/tree/14.0/stock_restrict_lot
     :alt: OCA/stock-logistics-workflow
--- a/stock_restrict_lot/__manifest__.py
+++ b/stock_restrict_lot/__manifest__.py
@@ -1,4 +1,4 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     "name": "Stock Restrict Lot",
@@ -8,7 +8,7 @@
     "website": "https://github.com/OCA/stock-logistics-workflow",
     "author": "Akretion, Odoo Community Association (OCA)",
     "maintainers": ["florian-dacosta"],
-    "license": "AGPL-3",
+    "license": "LGPL-3",
     "installable": True,
     "depends": ["stock"],
 }
--- a/stock_restrict_lot/models/product_product.py
+++ b/stock_restrict_lot/models/product_product.py
@@ -1,5 +1,5 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import models
 
 
--- a/stock_restrict_lot/models/stock_move.py
+++ b/stock_restrict_lot/models/stock_move.py
@@ -1,4 +1,4 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, exceptions, fields, models
 from odoo.exceptions import UserError
--- a/stock_restrict_lot/models/stock_rule.py
+++ b/stock_restrict_lot/models/stock_rule.py
@@ -1,4 +1,4 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import models
 
--- a/stock_restrict_lot/static/description/index.html
+++ b/stock_restrict_lot/static/description/index.html
@@ -8,11 +8,10 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
+:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
-Despite the name, some widely supported CSS2 features are used.
 
 See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
 customize this style sheet.
@@ -275,7 +274,7 @@ pre.literal-block, pre.doctest-block, pre.math, pre.code {
   margin-left: 2em ;
   margin-right: 2em }
 
-pre.code .ln { color: gray; } /* line numbers */
+pre.code .ln { color: grey; } /* line numbers */
 pre.code, code { background-color: #eeeeee }
 pre.code .comment, code .comment { color: #5C6576 }
 pre.code .keyword, code .keyword { color: #3B0D06; font-weight: bold }
@@ -301,7 +300,7 @@ span.option {
 span.pre {
   white-space: pre }
 
-span.problematic, pre.problematic {
+span.problematic {
   color: red }
 
 span.section-subtitle {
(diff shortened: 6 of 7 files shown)


pedrobaeza:
I don't agree on starting to relicense AGPL-3 to LGPL-3. That's restricting the opening of the source code. Why are you ForgeFlow now switching to that?

LoisRForgeFlow:
Hi @pedrobaeza 

You are right in the sense that we should encourage the use of AGPL to force to contribute back. 

However, I see AGPL more useful for modules that are completely functional by themselves, when you have such functionalities you prefer that people contribute back to opensource, instead of potencialy extend your work and do not share it back. E.g: cycle count app, tier validation and all submodules, production serial matrix, etc.

In the other hand, when you contribute a module that is more a technical piece that does not provide functionality by itself but instead it represents a useful mixin, a good framework, small helpers, etc. I see that the benefit comes from the broadest adoption possible, in other words the more people using your module for different end usages, the best. The more usage, also the more chances to have a more robust module as it has been tested in more situation or cases. E.g.: queue_job, edi_oca, stock_helper...

I think this module falls in this second category. Don't you think?

pedrobaeza:
I don't see the difference. The only exception we do is for e-commerce modules, as cosmetics changes made in custom theme modules shouldn't be opened.

LoisRForgeFlow:
@pedrobaeza There must be a difference when queue job was initially AGPL and then relicenced to LGPL. Just to mention a big framework as an example...

pedrobaeza:
For me that move was a mistake and promoted closed modules depending on it.

LoisRForgeFlow:
But that's the point! you want the framework to be used as much as possible to make it the standard in industry and only alternative for (following the example of queue job) asynchronous jobs. If you force people into AGPL to implement their core applications, they might prefer to do not rely on that big framework and keep their core applications private. In some cases you can even force the birth of an alternative/competitor for your framework.

pedrobaeza:
Yes, that's fine for me. Closed people shouldn't use open resources.

rousseldenis:
> Closed people shouldn't use open resources.

IMHO, that could be counter-productive in some cases.

@LoisRForgeFlow I personally think that that kind of moves should be done punctually. Nevertheless, @florian-dacosta has the last word as maintainer.

pedrobaeza:
Yes, I don't have word in this specific module, but I'm just saying that in this greedy enterprise ecosystem, I don't want OCA to be friendlier with them to continue cannibalizing things.

rvalyi:
Hello, I don't have a particular opinion about this specific module and I think this kind of important relicensing should be authorized by the module authors themselves if we don't want to see people loosing trust in the OCA.

However, I do understand the rationale behind it and for some rare modules I'm actually in favor of such changes. I once have read an important FSF article about it (but cannot remember exactly which).

Basically the rationale is: if a software (or an Odoo module) is simple enough to replicate under a permissive license, then it is better to relicense it under LGPL and see it with a larger adoption than seeing it copied under a permissive license and die slowly because less adopted.

the "simple enough to replicate" can be for various reasons like: it's technically very little code to re implement, it can be because it's a very simple set of fields and trivial methods that one cannot claim it could be copyrighted (like say a math formula). It can also be because Odoo will pass over it in some sort of way as they did when they changed their license in 2015. It can be that these proprietary modules will each reimplement 40 lines of code they need without fear somebody will sue them...

So IMHO the debate may happen for some very specific modules and I think module authors should adjust their strategy to ensure that the vast majority of the OCA stays copyleft, while a small set modules may somewhat fall into what can be considered a commodity and better be licensed under a permissive license than be replicated elsewhere. In fact I defend the switch for some 1% of very specific modules we wrote at Akretion (but certainly respecting the wish of the developpers who wrote it).

florian-dacosta:
Hello,

To be honest, I am not the big expert on licenses.
Is it not an option to change the Forgeflow RMA modules toward AGPL ? @LoisRForgeFlow @DavidJForgeFlow 

While I agree that we should encourage the use of AGPL, in this particular case, I guess it is worth the change if it is the only option to use it for the forgeflow RMA modules.
If we do not change the license, the other option left for Forgeflow will be to maintain the other oca concurrent module (stock_move_forced_lot) which make no sense IMHO

So I prefer (and so agree with) the relicensing of this module so it can be used as a dependency of the RMA modules than the use of another conccurent module.

pedrobaeza:
Starting with, I don't like such "RMA fork" done outside OCA, that has also been promoted inside OCA, and having such license. But again, on this particular module, I can't decide not being the author/maintainer.

About what Raphaël stays, AGPL variant is not going to die, as in this ecosystem, enterprise/closed source actors contribute little to nothing back to code foundations.

LoisRForgeFlow:
> Starting with, I don't like such "RMA fork" done outside OCA, that has also been promoted inside OCA, and having such license. But again, on this particular module, I can't decide not being the author/maintainer.

That "fork" was created long ago, even before than the current RMA version in the OCA. We proposed it to the OCA, but at that time in v7, the authors of the OCA/rma didn't accept our proposal so we were forced to maintain it in our own repo, we always wanted to contribute it to OCA. The old OCA/rma implementation died with the time (it was not a good design, that was the reason for us to propose a complete refactor), but later a new RMA was proposed in OCA this time without a "young" active version and so it was easier to take that vacant.

I just want to make it clear that ForgeFlow/stock-rma was always intended to be at OCA. Somebody would say it was just timing, some others think that it was just politics, but at this point I guess it is not relevant for most of the people. The thing that is important it is that it is not fork, it has been an stable and useful RMA aplication that is even older than the current RMA app at OCA repo.

> About what Raphaël stays, AGPL variant is not going to die, as in this ecosystem, enterprise/closed source actors contribute little to nothing back to code foundations.

I don't think such a small technical module as this one could stop contributions. We already have an LGPL module that does what we need (stock_move_forced_lot) but we are glad to change to a more robust and used module. If we are forced to change all our 20 RMA modules to AGPL, we might prefer to stick to our LGPL stock_move_forced_lot.

This is a perfect example on how LGPL helps contributions, if we start to use it in RMA and other apps, we might find bugs and be sure that we'll contribute back. If we start to use the stock_move_forced_lot LGPL version, we are not going to be very concerned about stock_restrict_lot and as a consequence the community efforts are split.

rousseldenis:
> If we are forced to change all our 20 RMA modules to AGPL

@LoisRForgeFlow Doing the devil's advocate, OCA RMA modules have been improved in v16. Haven't you think about switching back to them ? In order to merge efforts on that so unexploited functional flows (it is normal as it is usually envisaged when all other flows are running fine).

pedrobaeza:
Yes, and having a LGPL RMA version on OCA for me is a no-go.

rousseldenis:
> Yes, and having a LGPL RMA version on OCA for me is a no-go.

That was my next question :sweat_smile: 

@LoisRForgeFlow Why having a full stack of functional modules branded to LGPL ? It seems weird to me.

LoisRForgeFlow:
> > If we are forced to change all our 20 RMA modules to AGPL
> 
> @LoisRForgeFlow Doing the devil's advocate, OCA RMA modules have been improved in v16. Haven't you think about switching back to them ? In order to merge efforts on that so unexploited functional flows (it is normal as it is usually envisaged when all other flows are running fine).

I would be a huge effort for us, it will impact a big portion of our customers, most of them would probably see that OCA/rma does not cover their case and will require analysis, coding and training from our side. So it would be a huge effort of re-implementation with no real value added to be honest.

Don't you think this questions goes further away from "check this other similar module"?

LoisRForgeFlow:
> Yes, and having a LGPL RMA version on OCA for me is a no-go.

If that is the only thing preventing to put our RMA apps under OCA/rma, we would switch for sure. But I'm afraid that, Pedro, you are not going to make the re-implementation effort to migrate your installations, are you?

rousseldenis:
> Don't you think this questions goes further away from "check this other similar module"?

Of course. IMHO, we can go further and switch this one to LGPL.

@LoisRForgeFlow IMHO, you could keep that in mind for further migrations and avoid - take it with a touch of irony and humour -  schyzophrenic speech

> I just want to make it clear that ForgeFlow/stock-rma was always intended to be at OCA

rousseldenis (approved):

pedrobaeza:
Well, you are the OCA outsiders, not me. We already invested a lot to create (and later improve) current OCA rma modules.

Anyway, several "competing" solutions can be put inside OCA at the same time (look `contract` and `subscription_oca`). The pity is that, the same as is happening with the example put, both finish being very similar (people are "improving" the subscription one to be more and more complex and similar to `contract` one), while playing with the modularity is better IMO.

If your customer habits/problem is to have several products grouped, the solution is to build on top of the current ones other modules like `rma_batch` or similar, but take into account that we came when designing current RMA from a proprietary RMA module that performs such groupings, and states, actions, control flows were a nightmare. That's why we designed it this way, that sounds weird at the beginning, but it makes sense once you are on it, and having into account that you can still perform batch actions.

LoisRForgeFlow:
> Well, you are the OCA outsiders

Pedro, everybody that has participated in this conversation is among the top 10 contributors of the OCA. There are no "OCA outsiders" here.

I clarified what happened when we proposed our RMA application to OCA. We invested a lot of effort, you invested a lot of effort. The result: 2 valid and good RMA applications, each with pros and cons. Sometimes there are alternatives, and all of them are valid so the cost of the change is not worth it.

florian-dacosta (approved):
Well, having a set of module inside the OCA does not make it better than a set of module outside.
Also, it is some times just a question of timing and/or politics like explained by Lois.

I doubt we can all agree on the best set of RMA modules nor if we can replace one by another here and now. It is also not the place do do so...

So on my side, like I said before, ok for the re-licensing even though I'd prefer having the submodule as AGPL, I understand that it is more easy to say so than to change it overnight. Also I have no doubt about Forgeflow's involvement into the OCA...

pedrobaeza:
I'm not doubting either about their involvement. It's just the duplication effort. The cost of maintaining them may be similar to switching, but I don't insist. The license thing is the weird and non explained here. OK about the history of the alternative version, but why is LGPL? And why forcing OCA things to be LGPL then?

rousseldenis:
> I clarified what happened when we proposed our RMA application to OCA.

@LoisRForgeFlow In fact, not to come back on past, but to talk on same things, you've decided to go your side (in v9):

https://github.com/OCA/rma/pull/110#issuecomment-318412153

That's why I suggested you come back on the good side of the force in a further version :smile: to merge efforts of top contributors we all are

LoisRForgeFlow:
About the licence in RMA, I have checked and it has been like this from the very beginning so I would need to review/ask if there is any reason for that. That's why I do not want to rush that change in RMA, plus we also use stock_move_forced_lot in other applications/modules and it felt logical that a technical module like this could be used anywhere opensource or not.

pedrobaeza:
Lois, let me know if you want to have a meeting to analyze both solutions and see the gaps.

LoisRForgeFlow:
> > I clarified what happened when we proposed our RMA application to OCA.
> 
> @LoisRForgeFlow In fact, not to come back on past, but to talk on same things, you've decided to go your side (in v9):
> 
> [OCA/rma#110 (comment)](https://github.com/OCA/rma/pull/110#issuecomment-318412153)
> 
> That's why I suggested you come back on the good side of the force in a further version 😄 to merge efforts of top contributors we all are

I might have lost that part of the RMA history :thinking: and that's one more reason for me to not rush a license change in our RMA. I'm sure that in the initial conception we proposed it to OCA because I was involved in the call discussing it with the original authors at that time.

LoisRForgeFlow:
> Lois, let me know if you want to have a meeting to analyze both solutions and see the gaps.

That can be a good topic for a code sprint, maybe they are not that different.

pedrobaeza:
Write down it for the next Spanish OCA days. They are from 2 to 5 of June in Cartagena. I count with you all!

rvalyi:
> I'm not doubting either about their involvement. It's just the duplication effort. The cost of maintaining them may be similar to switching, but I don't insist. The license thing is the weird and non explained here. OK about the history of the alternative version, but why is LGPL? And why forcing OCA things to be LGPL then?

If >80% of the new modules contributed to the OCA keep coming as AGPL, even if we relicense 1% of the simplest commodity modules as LGPL every year, it is peanut in the big picture and the OCA keeps consolidating, both as the serious Odoo modules and as a copyleft model overall.

I know Odoo EE is also growing a lot, but IMHO we are yet to see the backfires coming. Odoo largely cannibalized its serious partner network, that's kind of the last asset they could have burned in exchange of some temporary market boost. It's likely the market will come back to open source after some delusion phase...

LoisRForgeFlow:
Thanks everybody to give your opinion on this discussion, and also thanks to @florian-dacosta for the flexibility accepting the change. We certainly have some pending discussion and review of the RMA applications, but for the time being let's merge this PR and meet in June to discuss RMA and for sure more things :rocket: 

/ocabot merge minor

PR #1842, opened 2025-01-21
[17.0] [MIG] stock_putaway_hook: Migration to 17.0

alexandregaldeano:
* replaced `fields_view_get` to `_get_view` and within it, `attrs` updates have been replaced by direct attributes changes
* added `odoo_test_helper` as test dependency


rousseldenis:
@alexandregaldeano Thanks for this. A test with `Form` could be great as `_get_view()` override is always touchy. Thanks

alexandregaldeano:
@rousseldenis that was a great suggestion, I understand more how this code works. I have added tests as you suggested. But now I'm questioning whether the _get_view override is useful at all. I have added the following test to check what was the behavior without the module: https://github.com/OCA/stock-logistics-workflow/pull/1842/commits/95a9784efc3cee014b82e6e660690650d7f5a2f9#diff-0c5186a1fcedfbbd1cd6b4d951d6a372b6c527330313759f55edb1f9e4a4f0a7R74
I feel like the changes in how Odoo 17 works make this part of the change irrelevant, what do you think?

sebalix (commented):
Thanks @alexandregaldeano for this migration

About the `_get_view` override if it is still relevant: it is if we still want to migrate `stock_putaway_by_route` that was adding a `route_id` field with the `exclusive_selection` option set.

However @jbaudoux thinks this was a bad design to configure route on put-away rules, and therefore we don't plan to use `stock_putaway_by_route` in the future. That means the current override of `_get_view` is not useful anymore as it was done for this purpose initially.

My take: we should drop all the `_get_view` plumbery (less code to maintain).

cc @jbaudoux

jbaudoux:
> However @jbaudoux thinks this was a bad design to configure route on put-away rules, and therefore we don't plan to use `stock_putaway_by_route` in the future. That means the current override of `_get_view` is not useful anymore as it was done for this purpose initially.

`stock_putaway_by_route` wasn't a bad design but a simple solution I did in v10.0 and that was migrated up to v14.0. Now, odoo integrated the package type concept for put-away and there is the `stock_storage_type` module for advanced routing. So there is no reason to continue to support `stock_putaway_by_route`.

It is dropped in the v18.0 migration. You could also cherry-pick https://github.com/OCA/stock-logistics-workflow/pull/1931/commits/8c3ca0151f4a1f339cbd6cde476141fcb91735d5 here

alexandregaldeano:
@jbaudoux I integrated changes from https://github.com/OCA/stock-logistics-workflow/commit/8c3ca0151f4a1f339cbd6cde476141fcb91735d5 into the migration commit if that's okay to you?
If not I could add an additional commit just before the [pre-commit auto fixes](https://github.com/OCA/stock-logistics-workflow/pull/1842/commits/3c84227210151f5636fa7b75316f8d858ebc417c) one :)

jbaudoux:
> @jbaudoux I integrated changes from [8c3ca01](https://github.com/OCA/stock-logistics-workflow/commit/8c3ca0151f4a1f339cbd6cde476141fcb91735d5) into the migration commit if that's okay to you? If not I could add an additional commit just before the [pre-commit auto fixes](https://github.com/OCA/stock-logistics-workflow/pull/1842/commits/3c84227210151f5636fa7b75316f8d858ebc417c) one :)

Thanks

jbaudoux (approved):

sebalix:
Thanks @alexandregaldeano , I'll update https://github.com/OCA/stock-logistics-workflow/pull/1931 accordingly.

sebalix (approved):

PR #1843, opened 2025-01-21
[15.0][FIX] stock_return_request: Cache miss error when access to qty_returnable computed field on stock move

sergio-teruel:
cc @Tecnativa TT54590

ping  @chienandalu @carlosdauden

### DIFF
--- a/stock_return_request/models/stock_move.py
+++ b/stock_return_request/models/stock_move.py
@@ -16,7 +16,8 @@ class StockMove(models.Model):
     def _compute_qty_returnable(self):
         """Looks for chained returned moves to compute how much quantity
         from the original can be returned"""
-        for move in self.filtered(lambda x: x.state not in ["draft", "cancel"]):
+        self.update({"qty_returnable": 0.0})
+        for move in self:
             if not move.returned_move_ids:
                 if move.state == "done":
                     move.qty_returnable = move.quantity_done


stock_return_request/models/stock_move.py
```diff
@@ -16,7 +16,10 @@ class StockMove(models.Model):
     def _compute_qty_returnable(self):
         """Looks for chained returned moves to compute how much quantity
         from the original can be returned"""
-        for move in self.filtered(lambda x: x.state not in ["draft", "cancel"]):
```
rousseldenis, line 19: @sergio-teruel 
To avoid multiple `__set__` calls and n condition evaluations, do this:

`self.update({"qty_returnable": 0.0})`

chienandalu, line 19: Why not simply setting it to `False` at the beggining?

sergio-teruel:
changes done!

marcelsavegnago (approved):

sergio-teruel:
ping @chienandalu @rousseldenis

stock_return_request/models/stock_move.py
pedrobaeza: AFAIK, doing `self.qty_returnable = 0` has the same performance against `update`, as there's only one value to set.

chienandalu (approved):

PR #1844, opened 2025-01-22
[18.0] [MIG] stock_picking_backorder_strategy: Migration to 18.0

bizzappdev:
Module renamed to: stock_picking_backorder_strategy_cancel


jbaudoux (approved):

jbaudoux:
@rousseldenis Would you rename this module to stock_picking_backorder_strategy_cancel as now the backorder strategy is odoo standard and this only adds a cancel strategy ?

bizzappdev:
@jbaudoux Indeed, it makes more sense.
@rousseldenis, could you please confirm this? If you agree as well, we can take care of renaming.

rousseldenis:
Yes, no problem for that.

In fact, this is still a module for backorder strategy management :-)

jbaudoux (approved):
Thanks. Can you add in the mig commit message that you renamed the module?
And I would drop the old changelog that is obsolete.

bizzappdev:
> Thanks. Can you add in the mig commit message that you renamed the module? And I would drop the old changelog that is obsolete.

The commit-msg changed. also added a new changelog. (not sure if I should remove the old changelog or not)

stock_picking_backorder_strategy_cancel/models/stock_picking_type.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    create_backorder = fields.Selection(
```
simahawk, line 10: don't we need any migration of data after the renaming?

jbaudoux: Indeed, the old field was `backorder_strategy`

PR #1847, opened 2025-01-23
[18.0][MIG] stock_picking_product_availability_inline: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT54341

ping @sergio-teruel @carlosdauden 

The name_get method no longer exists, so the changes would need to be applied to compute_display_name. However, if we did this, the value would also be computed when it's not related to the dropdown for selecting a product. To solve this issue, the name_search method has been adapted, as it is the one that returns the name to be used in the dropdown, as shown in this JavaScript code: https://github.com/odoo/odoo/blob/26a992ea187c09ac2b39a14d0c64ed5dd658cfaf/addons/web/static/src/views/fields/relational_utils.js#L306-L337


jbaudoux (commented):
This module would be better in https://github.com/OCA/stock-logistics-availability

stock_picking_product_availability_inline/__manifest__.py
```diff
@@ -2,9 +2,9 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Stock Picking Product Availability Inline",
-    "summary": "Show product availability in product drop-down in stock picking form view.",
-    "version": "15.0.1.0.1",
-    "development_status": "Beta",
+    "summary": "Show product availability in product drop-down of picking form view.",
+    "version": "18.0.1.0.0",
+    "development_status": "Production/Stable",
```
sergio-teruel, line 7: Plase add maintainer key

sergio-teruel (approved):
Minor comment.
About change module to other repositore... 
@pedrobaeza What do you think?? To me is OK

CarlosRoca13:
Closing this PR because this new one is opened: https://github.com/OCA/stock-logistics-availability/pull/44

PR #1852, opened 2025-01-26
[18.0][MIG] stock_picking_sale_order_link: Migration to 18.0

rjaraspearhead:
In this PR, the migration to v17 has been carried out. If there are any issues, I will be available to resolve them. If there are no problems, I would appreciate your approval.


StefanRijnhart:
Module is already available in 17.0 and 18.0

PR #1853, opened 2025-01-27
[16.0][ADD] stock_move_group_by_release_blocked

santostelmo:
Group stock moves in picking depending on the release blocked status of the move.


stock_move_group_by_release_blocked/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "Summary": "Group stock moves in 1 picking based on release blocked",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "category": "Warehouse",
+    "depends": [
+        "stock_picking_group_by_base",
+        "stock_available_to_promise_release_block",
+    ],
+    "installable": True,
```
sebalix, line 16: I would add `auto_install` too

stock_move_group_by_release_blocked/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+
+{
+    "name": "Stock Move: group by release blocked",
+    "Summary": "Group stock moves in 1 picking based on release blocked",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "category": "Warehouse",
+    "depends": [
+        "stock_picking_group_by_base",
```
sebalix, line 12: It's encouraged to add some code on `stock.picking` as shown in the Usage section of `stock_picking_group_by_base`: https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_picking_group_by_base#usage

stock_move_group_by_release_blocked/models/stock_move.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _search_picking_for_assignation_domain(self):
+        return super()._search_picking_for_assignation_domain() + [
+            ("release_blocked", "=", self.release_blocked)
+        ]
```
sebalix: Better to use `odoo.osv.expression.AND`

sebalix: You also have to override method `_key_assign_picking` on `stock.move` to add the `self.release_blocked`

stock_move_group_by_release_blocked/__manifest__.py
santostelmo: release_blocked is a boolean => no need to add index in this case no ?

santostelmo:
@sebalix I added your changes. All green

stock_move_group_by_release_blocked/__manifest__.py
sebalix: Yep probably not needed. I don't know if it's useful in a multi-column index. Let's not add it for now.

sebalix (approved):
LG, ping @jbaudoux your review is also welcome

jbaudoux (commented):
What happens when you unblock a move that was blocked? Then you have a picking containing blocked and non blocked moves. Releasing that move will be processed separately from the any other moves that were never blocked, so you end-up with different packages.

santostelmo:
> unblock a move that was blocked?
What do you do you propose in this case ?
If user unblock a move it should split the picking if needed no ?

PR #1854, opened 2025-01-28
[16.0][IMP] stock_owner_restriction

AungKoKoLin1997:
Supersede https://github.com/OCA/stock-logistics-workflow/pull/1385.
This PR does the following:

- Extend _update_available_quantity() and _update_reserved_quantity() in stock.quant
  to handle scenarios where owner_restriction should be enforced.
- Create a separate _get_owner_restriction() method to improve clarity and extensibility for other modules.
- Fix read_group() in stock.quant, which was incorrectly assigning the owner record
  instead of the owner ID in the domain.
- Add compute method for restrict_partner_id to dynamically control owner assignment
  based on picking type restrictions.
- Unlink stock.move.line records when owner_id is updated and owner_restriction is set.
- Update test cases.

@qrtl QT4207

### DIFF
--- a/stock_owner_restriction/models/stock_move.py
+++ b/stock_owner_restriction/models/stock_move.py
@@ -1,13 +1,29 @@
 # Copyright 2020 Carlos Dauden - Tecnativa
 # Copyright 2020 Sergio Teruel - Tecnativa
+# Copyright 2023-2024 Quartile
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from collections import defaultdict
 
-from odoo import models
+from odoo import api, fields, models
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
+
+    # Add compute method to the standard field
+    restrict_partner_id = fields.Many2one(
+        compute="_compute_restrict_partner_id",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("picking_type_id", "picking_id.owner_id", "move_dest_ids")
+    def _compute_restrict_partner_id(self):
+        for move in self:
+            if move.picking_type_id.owner_restriction == "picking_partner":
+                move.restrict_partner_id = move._get_owner_for_assign()
+            else:
+                move.restrict_partner_id = False
 
     def _get_moves_to_assign_with_standard_behavior(self):
         """This method is expected to be extended as necessary. e.g. you may not want to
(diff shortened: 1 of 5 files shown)


kanda999 (approved):
LGTM: This update has been used in production for a year.

stock_owner_restriction/models/stock_picking.py
```diff
@@ -8,3 +8,11 @@ class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     owner_restriction = fields.Selection(related="picking_type_id.owner_restriction")
+
+    def write(self, vals):
+        if "owner_id" in vals:
+            for pick in self:
+                owner_restriction = pick.picking_type_id.owner_restriction
+                if owner_restriction in ("unassigned_owner", "picking_partner"):
```
yostashiro:
```suggestion
                if pick.owner_restriction in ("unassigned_owner", "picking_partner"):
```

yostashiro (commented):
Could we perhaps make the intention of adding _compute_restrict_partner_id() clearer?

stock_owner_restriction/models/stock_move.py
```diff
@@ -1,14 +1,32 @@
...
+from odoo import api, fields, models
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    restrict_partner_id = fields.Many2one(
+        # Add compute to override the standard 'restrict_partner_id' behavior:
+        # assign the owner only when the picking type's restriction is 'picking_partner',
+        # and clear it for other restriction types to avoid unintended ownership.
```
yostashiro: I think this should do. No need to explain the content of the compute method.
```suggestion
    # Add compute method to the standard field
    restrict_partner_id = fields.Many2one(
```

stock_owner_restriction/models/stock_move.py
```diff
@@ -1,14 +1,30 @@
...
+    restrict_partner_id = fields.Many2one(
+        compute="_compute_restrict_partner_id",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("picking_type_id", "picking_id.owner_id", "move_dest_ids")
+    def _compute_restrict_partner_id(self):
+        for move in self:
+            if move.picking_type_id.owner_restriction == "picking_partner":
```
yostashiro, line 23: Please check the validity of this suggestion.
```suggestion
            if move.picking_type_id.owner_restriction in ("picking_partner", "partner_or_unassigned"):
```

AungKoKoLin1997: I think this is not valid. Even if we have move lines from an unassigned owner together with move lines from an assigned owner, the move will still be assigned to the partner. Therefore, we should continue allowing False for this restriction type.

PR #1856, opened 2025-01-29
[18.0][MIG] stock_picking_mass_action: Migration to 18.0

yankinmax:


jbaudoux (commented):
Would be better to split into 2 modules:
- stock_picking_mass_confirm
- stock_picking_mass_validate

rrebollo (approved):
Code Review. LGTM! Would U be so kind to do a review on mine #1916?

pedrobaeza (approved):
Let's continue for now with the module as is for having the functionalities. We can split it on next version depending on what Odoo incorporates in the core.

/ocabot merge nobump

PR #1861, opened 2025-01-31
[16.0][FIX] stock_picking_line_sequence model_create_multi

andreampiovesana:
fix warning @api.model_create_multi

### DIFF
--- a/stock_picking_line_sequence/models/stock.py
+++ b/stock_picking_line_sequence/models/stock.py
@@ -23,7 +23,7 @@ class StockMove(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        moves = super(StockMove, self).create(val_list)
+        moves = super(StockMove, self).create(vals_list)
         for move in moves:
             # We do not reset the sequence if we are copying a complete picking
             # or creating a backorder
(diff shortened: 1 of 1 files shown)


StefanRijnhart (changes requested):
Please return the `moves` at the end of the `create` method. Also, squash commits and apply a conventionally formatted commit message (such as the title of the PR).

stock_picking_line_sequence/models/stock.py
```diff
@@ -21,14 +21,14 @@ class StockMove(models.Model):
...
-            move.picking_id._reset_sequence()
-        return move
+    @api.model_create_multi
+    def create(self, vals_list):
+        moves = super(StockMove, self).create(vals_list)
+        for move in moves:
+            # We do not reset the sequence if we are copying a complete picking
+            # or creating a backorder
+            if not self.env.context.get("keep_line_sequence", False):
+                move.picking_id._reset_sequence()
```
rousseldenis: IMHO, this should be done once per picking. This should be grouped per picking.

StefanRijnhart:
Picked an improved version of this in the migration of this module to 18.0 in https://github.com/OCA/stock-logistics-workflow/pull/1864

TheMule71:
@andreampiovesana you should squash all commits together leaving just one

PR #1862, opened 2025-01-31
[FIX] stock_no_negative: use display_name instead of name_get because it is deprecated

micheledic:
name_get is deprecated the error 
`AttributeError: 'stock.lot' object has no attribute 'name_get' `
appear

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -39,7 +39,7 @@ class StockQuant(models.Model):
             ):
                 msg_add = ""
                 if quant.lot_id:
-                    msg_add = _(" lot {}").format(quant.lot_id.name_get()[0][1])
+                    msg_add = _(" lot %(name)s", name=quant.lot_id.display_name)
                 raise ValidationError(
                     _(
                         "You cannot validate this stock operation because the "
--- a/stock_no_negative/tests/test_stock_no_negative.py
+++ b/stock_no_negative/tests/test_stock_no_negative.py
@@ -134,6 +134,26 @@ class TestStockNoNegative(TransactionCase):
         with self.assertRaises(ValidationError):
             self.stock_picking.button_validate()
 
+    def test_check_constrains_with_lot(self):
+        """Assert that constraint is raised when user
+        tries to validate the stock operation which would
+        make the stock level of the product negative with
+        a product with lot"""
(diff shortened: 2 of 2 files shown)


stock_no_negative/models/stock_quant.py
```diff
@@ -39,7 +39,7 @@ def check_negative_qty(self):
             ):
                 msg_add = ""
                 if quant.lot_id:
-                    msg_add = _(" lot {}").format(quant.lot_id.name_get()[0][1])
+                    msg_add = _(" lot {}").format(quant.lot_id.display_name)
```
rousseldenis: Could you remove also `format()` method and use translation function arguments instead ?

```suggestion
                    msg_add = _(" lot %(name)s", name=quant.lot_id.display_name)
```

And add a test to avoid regression ? Thanks

rousseldenis (changes requested):

stock_no_negative/models/stock_quant.py
micheledic: Hello, i changed the format function.
How should i create the test ? Should i assert the FULL ValidationError message content ?

the msg_add is used in a dynamic string passed to a validation error

micheledic:
@rousseldenis can you check? it is a bug not an enhancement

rousseldenis (approved):

rousseldenis:
@micheledic Could you add a test to avoid regression ? Thanks

micheledic:
> @micheledic Could you add a test to avoid regression ? Thanks

Sorry can you explain ? the commit only modify the dynamic test inside a validationerror. Can you give me an example about how should i create the test?

rousseldenis:
> > @micheledic Could you add a test to avoid regression ? Thanks
> 
> Sorry can you explain ? the commit only modify the dynamic test inside a validationerror. Can you give me an example about how should i create the test?

You should add a test with a product with a lot

micheledic:
> > > @micheledic Could you add a test to avoid regression ? Thanks
> > 
> > 
> > Sorry can you explain ? the commit only modify the dynamic test inside a validationerror. Can you give me an example about how should i create the test?
> 
> You should add a test with a product with a lot

done @rousseldenis

rousseldenis (approved):
Thanks

robyf70 (approved):
LGTM

PR #1866, opened 2025-02-04
[MIG][16.0] stock_picking_tier_validation

em230418:


em230418:
Hello, @ps-tubtim!
Could please you explain why do we have this method in TierValidationTester, that is not even covered by tests in 14.0
https://app.codecov.io/gh/OCA/stock-logistics-workflow/blob/14.0/stock_picking_tier_validation%2Ftests%2Ftier_validation_tester.py#L24

rousseldenis:
> Hello, @ps-tubtim! Could please you explain why do we have this method in TierValidationTester, that is not even covered by tests in 14.0 https://app.codecov.io/gh/OCA/stock-logistics-workflow/blob/14.0/stock_picking_tier_validation%2Ftests%2Ftier_validation_tester.py#L24

@em230418 When writing modules, the best situation is the one where all lines are covered. But, sometimes, it's not the case.

Do you plan to add that test ? :smiley:

PR #1870, opened 2025-02-10
[FIX] stock_picking_batch_validate_confirm: avoid propagating context with defaults

yajo:
When this module was installed along with `sale_elaboration`, elaboration sale order lines could be created automatically when confirming a picking.

The `sale.order.line` model also has a `move_ids` field. Thus, those lines were created associated to the `stock.move` records found from within the batch.

To make it even more fun, a batch can contain moves associated to unrelated sales. Thus, the elaboration lines were getting stock moves (which is wrong because they are services) and those moves came from unrelated sale orders.

This mismatch was resulting in unexpected behaviors here and there.

Now, we just produce defaults based on the standard context that all wizards get by default. This is safe and doesn't pollute any sub-calls.

@moduon MT-9033 MT-7490

### DIFF
--- a/stock_picking_batch_validate_confirm/models/stock_picking_batch.py
+++ b/stock_picking_batch_validate_confirm/models/stock_picking_batch.py
@@ -15,24 +15,14 @@ class StockPickingBatch(models.Model):
             [("state", "not in", ("cancel", "done"))]
         )
         if pending_orig_moves:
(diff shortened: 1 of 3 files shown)


Shide (approved):
Code LGTM

EmilioPascual (approved):
LGTM. Code review

PR #1874, opened 2025-02-11
[16.0][ADD] stock_move_product_expired_alert

rousseldenis:
Depends on:

- https://github.com/OCA/product-attribute/pull/1882


stock_move_product_expired_alert/models/stock_move.py
```diff
@@ -0,0 +1,59 @@
...
+                lambda line: line.has_expired_product
+            ):
+                # Building the alert message
+                note += _(
+                    "\n - Product: %(product_name)s (Lot: %(lot_name)s)",
+                    product_name=line.product_id.display_name,
+                    lot_name=line.lot_id.display_name,
+                )
+            for user in group.users:
+                picking.activity_schedule(
```
lmignon: In my opinion, creating the same activity for multiple users is a sure way to make the system unusable and erode user trust in Odoo's activity management system. Once one user processes or manages the activity, others are not notified of its resolution, leaving them wondering why they received the notification

rousseldenis: Done

PR #1875, opened 2025-02-13
[18.0][MIG] stock_picking_tier_validation: Migration to version 18.0

Ricardo-MC:


stock_picking_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,64 @@
...
+                "definition_domain": "[('test_field', '>', 1.0)]",
+            }
+        )
+
+        cls.test_record = cls.test_model.create({"test_field": 2.5})
+
+    @classmethod
+    def tearDownClass(cls):
+        teardown_test_model(cls.env, [TierValidationTester])
+        super().tearDownClass()
```
celm1990: I think you can add a real test for the `stock.picking` model instead of creating a fake model, similar to:
https://github.com/OCA/sale-workflow/blob/17.0/sale_tier_validation/tests/test_tier_validation.py

celm1990 (changes requested):
Tested and working. Just a technical comment to make the test as realistic as possible. What do you think?

CC @pedrobaeza @bosd @StefanRijnhart

stock_picking_tier_validation/tests/test_tier_validation.py
StefanRijnhart: Agreed! The tests here are clearly copied from base_tier_validation, which makes no sense, and it would be good to have them replaced by a proper validation flow as is the case in for instance https://github.com/OCA/stock-logistics-workflow/blob/17.0/stock_scrap_tier_validation/tests/test_tier_validation.py#L50-L66.

Saran440 (approved):
Functional Test :+1:

stock_picking_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,102 @@
...
+        self.assertEqual(picking_in.state, "done")
+
+    def test_tier_validation_picking_out(self):
+        picking_out = self._create_picking(
+            self.picking_type_out, self.location_stock, self.location_customer
+        )
+        picking_out.action_confirm()
+        picking_out.action_assign()
+        self.assertEqual(picking_out.state, "assigned", "The picking is not assigned.")
+        picking_out.request_validation()
```
StefanRijnhart: Thanks for improving the tests! Right now, it is not actually tested that the picking cannot be processed before the tier validation is completed. Can you add this here?

Ricardo-MC: @StefanRijnhart, Hi Thanks for the suggestion! I’ve added an assertion to ensure the picking cannot be processed before the tier validation is completed.

marcelsavegnago (approved):
LGTM

StefanRijnhart (approved):
Thanks for the final test tweaks! @celm1990 is it alright for you as well?

nikul-serpentcs:
@rousseldenis @StefanRijnhart, can we move forward?

PR #1878, opened 2025-02-18
[13.0][FIX] stock_picking_filter_lot: if quantity changes location ids are not computed

dalonsod:
We had quantities in a location and weren't able to select it and we traced the problem to be this missing dependency.

Supersedes #1049 for 13.0

### DIFF
--- a/stock_picking_filter_lot/models/stock_production_lot.py
+++ b/stock_picking_filter_lot/models/stock_production_lot.py
@@ -11,7 +11,7 @@ class StockProductionLot(models.Model):
         comodel_name="stock.location", compute="_compute_location_ids", store=True
     )
 
-    @api.depends("quant_ids", "quant_ids.location_id")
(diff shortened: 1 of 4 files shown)


dalonsod:
@lk-eska @BernatPForgeFlow could you review?

If it's merged, I'll port it later versions up to v17. Thanks!

rousseldenis:
@dalonsod Could you reflect module name also in commit message ? Thanks

dalonsod:
> @dalonsod Could you reflect module name also in commit message ? Thanks

Done!

rousseldenis:
> @dalonsod Could you reflect module name also in commit message ? Thanks

Ok done

rousseldenis (approved):
Seems coherent.

Could you maybe add a little test for that?

dalonsod:
> Could you maybe add a little test for that?

Done! I've made them in a separate commit, if two commits should be squashed, I'll do it.

dalonsod:
@newtratip @fd-oerp could you review? Thanks!

PR #1885, opened 2025-02-21
[18.0][MIG] stock_picking_return_lot

StefanRijnhart:
Full refactoring to adapt to Odoo's refactored stock.picking.return wizard which has separate options for returning all quantities or manually entered quantities.` This module now logically integrates with stock_picking_return_restricted_qty because the restricted quantities have to be split up per lot so that the modules would be incompatible and have the duplicate functionality of setting the default quantity otherwise.


sbejaoui (approved):
code review + functional test

stock_picking_return_lot/tests/test_stock_picking_return_lot.py
```diff
@@ -55,11 +64,9 @@ def setUpClass(cls):
 
     @classmethod
     def create_return_wiz(cls, picking):
-        return (
-            cls.env["stock.return.picking"]
-            .with_context(active_id=picking.id, active_model="stock.picking")
-            .create({})
-        )
+        wizard = cls.env["stock.return.picking"].create({"picking_id": picking.id})
+        # wizard.set_quantities()
```
marielejeune: Could you remove this commented line?

stock_picking_return_lot/wizards/stock_return_picking_line.py
```diff
@@ -7,14 +7,49 @@
...
+        return vals
+
+    @api.model
+    def get_returned_restricted_quantity(self, stock_move):
+        """Get the restricted quantity by lot (if any).
+
+        Overwrite of this method from stock_picking_return_restricted_quantity.
+        """
+        if self:
+            self.ensure_one()
```
marielejeune, line 44: Isn't this strange to have a `self.ensure_one()` on an `api.model` method?

stock_picking_return_lot/tests/test_stock_picking_return_lot.py
StefanRijnhart: Thanks, done.

stock_picking_return_lot/wizards/stock_return_picking_line.py
StefanRijnhart: I can see why you'd think that! But it is simply the case that `stock_picking_return_restricted_qty` calls this method on both empty recordsets and single records. For our purposes (filtering the quantities by lot) it works either way but not if someone decides to call it on two records or more so I'm just making things explicit here.

The decorator might just as well be removed from a technical point of view, because this method is only going to be called from code and not from UI, cron or RPC so the module would still be working fine without it.

marielejeune: Thanks for the explanation!

marielejeune (approved):
LGTM

PR #1892, opened 2025-03-04
[18.0][MIG] stock_lock_lot: Migration to version 18.0

henrybackman:
## Issues:
I tried to follow the configuration guide in functional testing, but it didn't match what I was seeing and I couldn't set the user rights:
```
To allow a user to block or unblock a Lot:

1.  Open the user settings (menu "Settings \> Users & Companies \>
    Users")
2.  In the "Warehouse" section, check the box "Allow to block/unblock
    Serial Numbers/Lots"

To allow move locked lots to a location: \#. Open the locations (menu
"Inventory \> Configuration \> Warehouse Management \> Locations") \#.
check the box "Allow Locked"
```
There is no "Warehouse" section in user settings after installing this module.

Not sure if this is an issue to be fixed in migration or not.


sebalix:
@henrybackman in fact the section is now called `Inventory`, and the group/check box will appear only with debug mode enabled (or if the module `base_technical_features` is installed):
image

So we should probably update the config instructions to enable debug mode first.

stock_lock_lot/models/stock_lot.py
```diff
@@ -55,24 +55,25 @@ def _check_lock_unlock(self):
...
-                    "product_id", self.env.context.get("default_product_id", False)
-                ),
+        for vals in vals_list:
+            product = self.env["product.product"].browse(
+                vals.get(
+                    "product_id",
+                    # Web quick-create provide in context
+                    self.env.context.get(
+                        "product_id", self.env.context.get("default_product_id", False)
+                    ),
```
sebalix: Now we are looping, we could move this context getter outside the `for` loop:
```python
        default_product_id = self.env.context.get(
            "product_id", self.env.context.get("default_product_id", False)
        )
```

Then in the loop:
```python
            product = self.env["product.product"].browse(
                vals.get("product_id", default_product_id)
            )
```

henrybackman: Good point, now changed to have the `default_product_id` outside of the loop

henrybackman:
> @henrybackman in fact the section is now called `Inventory`, and the group/check box will appear only with debug mode enabled (or if the module `base_technical_features` is installed): image
> 
> So we should probably update the config instructions to enable debug mode first.

Did updates to the CONFIGURE and USAGE to match the current steps

TDu (approved):
LG, eventually test coverage could be improved

sebalix (approved):

henrybackman:
> LG, eventually test coverage could be improved

Added additional tests

hparfr (approved):
/ocabot merge nobump

sebalix (approved):

PR #1895, opened 2025-03-06
[14.0][FIX] delivery_procurement_group_carrier: Fix carrier update

mmequignon:
In odoo standard, it's possible to update carrier on SO even when confirmed, and picking is updated at the same time.
This PR aims to align carrier on the procurement group when this happens.

ping @mt-software-de @jbaudoux

### DIFF
--- a/delivery_procurement_group_carrier/models/stock_picking.py
+++ b/delivery_procurement_group_carrier/models/stock_picking.py
@@ -1,0 +1,47 @@
+# Copyright 2025 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _align_group_carrier(self):
+        for picking in self:
+            picking_carrier = picking.carrier_id
+            picking_group = picking.group_id
+            if picking_group and picking_carrier != picking_group.carrier_id:
+                picking_group.carrier_id = picking_carrier
+                need_align_pickings = self.search(
+                    [
+                        ("group_id", "=", picking_group.id),
+                        (
+                            "state",
+                            "not in",
+                            (
+                                "done",
+                                "cancel",
+                            ),
+                        ),
+                        ("carrier_id", "!=", picking_carrier.id),
+                        ("carrier_id", "!=", False),
+                    ]
+                )
+                need_align_pickings.carrier_id = picking_carrier
+
+    def write(self, values):
+        if "carrier_id" not in values or self.env.context.get(
+            "skip_align_group_carrier"
+        ):
+            # We only track when carrier changes. Avoid useless computation when
+            # carrier_id isn't in values
+            return super().write(values)
+        carrier_mapping = {record.id: record.carrier_id for record in self}
+        res = super().write(values)
+        # Align group on pickings where carrier was updated
+        updated_pickings = self.filtered(
+            lambda p: p.carrier_id != carrier_mapping.get(p.id)
+        )
+        updated_pickings._align_group_carrier()
+        return res
--- a/delivery_procurement_group_carrier/models/__init__.py
+++ b/delivery_procurement_group_carrier/models/__init__.py
@@ -1,3 +1,4 @@
 from . import procurement_group
 from . import sale_order
 from . import stock_move
+from . import stock_picking
--- a/delivery_procurement_group_carrier/tests/test_carrier.py
+++ b/delivery_procurement_group_carrier/tests/test_carrier.py
@@ -2,44 +2,65 @@
 # Copyright 2020 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo.tests import tagged
-
-from odoo.addons.sale.tests.common import TestSaleCommonBase
+from odoo.tests import SavepointCase, tagged
+from odoo.tests.common import Form
 
 
 @tagged("post_install", "-at_install")
-class TestProcurementGroupCarrier(TestSaleCommonBase):
-    # FIXME: TestSale is very heavy and create tons of records w/ no tracking disable
-    # for every test. Move to SavepointCase!
-    def setUp(self):
-        super().setUp()
-        self.carrier1 = self.env["delivery.carrier"].create(
+class TestProcurementGroupCarrier(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.product = cls.env["product.product"].create(
+            {"type": "product", "name": "Test Product"}
+        )
+        cls.carrier = cls.env["delivery.carrier"].create(
             {
                 "name": "My Test Carrier",
-                "product_id": self.env.ref("delivery.product_product_delivery").id,
+                "product_id": cls.env.ref("delivery.product_product_delivery").id,
             }
         )
-        self.partner = self.env["res.partner"].create({"name": "Test Partner"})
+        cls.partner = cls.env["res.partner"].create({"name": "Test Partner"})
+
+    @classmethod
+    def _add_carrier_to_order(cls, order, carrier):
+        order.set_delivery_line(carrier, 1)
+
+    @classmethod
+    def _create_sale_order(cls, product_qty, carrier=None):
+        with Form(cls.env["sale.order"]) as order_form:
+            order_form.partner_id = cls.partner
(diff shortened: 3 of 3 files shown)


delivery_procurement_group_carrier/models/sale_order.py
```diff
@@ -13,6 +13,19 @@ def action_draft(self):
...
+    def set_delivery_line(self, carrier, amount):
+        # Updating procurement group carrier before calling super, because
+        # some compute methods triggered by picking.carrier_id change
+        # might rely on group_id.carrier_id.
+        # Both needs to be aligned if this happens.
+        for order in self:
+            pending_deliveries = order.picking_ids.filtered(
+                lambda p: p.state not in ("done", "cancel")
+                and not any(m.origin_returned_move_id for m in p.move_lines)
+            )
```
jbaudoux: This is already done in standard odoo, isn't it?

delivery_procurement_group_carrier/models/sale_order.py
```diff
@@ -13,6 +13,19 @@ def action_draft(self):
...
+            pending_deliveries.group_id.carrier_id = carrier
```
jbaudoux:
```suggestion
            order.group_id.carrier_id = carrier
```

jbaudoux (changes requested):

delivery_procurement_group_carrier/models/sale_order.py
jbaudoux: a better solution would be to do that only if the order is in state done/cancel and only for pickings having the same proc group that the one from the order. You then need to create a new copy of the proc group, put it on the SO, update the carrier of the proc group, update the pickings having the old SO proc group. And call `order._create_delivery_line(carrier, amount)`. And finally not call super on those orders

delivery_procurement_group_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _align_group_carrier(self):
+        for picking in self:
```
jbaudoux, line 11: Here you should check if the carrier needs to be updated on the group. Then you can just write the new carrier on the group and update all open pickings of the group having the old carrier with the new carrier. (to be compatible with new versions where the carrier can be propagated on the chain)

That remark about creating a new group was only in case you change the carrier on the SO but it is not relevant anymore as you dropped that part.

jbaudoux: What I wrote isn't right. It will conflict with https://github.com/OCA/wms/blob/14.0/delivery_carrier_preference/models/stock_move.py#L71

delivery_procurement_group_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,36 @@
...
+            picking_carrier = picking.carrier_id
+            picking_group = picking.group_id
+            if picking_carrier != picking_group.carrier_id:
+                picking_group.carrier_id = picking_carrier
+                need_align_pickings = self.search(
+                    [
+                        ("group_id", "=", picking_group.id),
+                        ("carrier_id", "!=", picking_carrier.id),
+                    ]
+                )
```
jbaudoux:
```suggestion
                need_align_pickings = picking_group.picking_ids.filtered(
                    lambda p: p.state not in ("cancel", "done") and p.carrier_id and p.carrier_id != picking_carrier
                )
```

delivery_procurement_group_carrier/models/stock_picking.py
jbaudoux: A solution would be to add a context variable to skip this in the caller write.
`        if "carrier_id" not in values or self.env.context.get("skip_align_group_carrier"):`

Then here https://github.com/OCA/wms/blob/14.0/delivery_carrier_preference/models/stock_picking.py#L42 we could enable that skip

mmequignon: This is fixed here https://github.com/OCA/wms/pull/996

delivery_procurement_group_carrier/models/stock_picking.py
mmequignon: This group.picking_ids comes from `delivery_carrier_preference`. I'm not gonna add a dependency for this.
I updated the domain though.

delivery_procurement_group_carrier/models/stock_picking.py
mt-software-de: What i don't understand is, why you want to align all open pickings with the same carrier.
Case 1: Carrier is changed via Sale order -> by default all open pickings will get the new carrier set
Case 2: Carrier is changed via Picking and there is more than one transfer with the same group -> the carrier should not change for all open transfers
In both cases i think we should check the related pickings and the carriers and if there is not all carriers of the transfers are the same, we should create a new group

mt-software-de (changes requested):

delivery_procurement_group_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,46 @@
...
+                        ("group_id", "=", picking_group.id),
+                        (
+                            "state",
+                            "not in",
+                            (
+                                "done",
+                                "cancel",
+                            ),
+                        ),
+                        ("carrier_id", "!=", picking_carrier.id),
```
jbaudoux, line 27:
```suggestion
                        ("carrier_id", "!=", picking_carrier.id),
                        ("carrier_id", "!=", False),
```

delivery_procurement_group_carrier/models/stock_picking.py
mmequignon: here is the idea, stock_warehouse_flow uses `move.group_id.carrier_id` to determine what is the best flow for a given picking.
If you do not align group's carrier, the flow will never be updated after a carrier change (whether it's from the picking itself or the sale order).
Now, if you change the carrier of the picking, we update the carrier on the procurement group. We have to align the other pickings referencing this procurement group.
We cannot have a discrepancy between group.carrier_id / picking.carrier_id, it doesn't make any sense.

mt-software-de: Yes i understand this. But what wasn't clear why all of those other transfers in the group have to change the carrier. For my understanding this is not necessary right now. It is mimicking https://github.com/odoo/odoo/blob/1bd7ed14a6d70a570e2924c13129ca6a594bca80/addons/delivery/models/stock_move.py#L38 to align the carriers also on none outgoing transfers. My concern was also more about what if there is a second pending outgoing transfer in the group. But this should not happen.

delivery_procurement_group_carrier/models/stock_picking.py
mt-software-de: Wouldn't this mean, that there will be never a carrier updated on a none outgoing transfer, because in v14 the carrier is only set for outgoing transfers. `vals['carrier_id'] = self.mapped('sale_line_id.order_id.carrier_id').id`

delivery_procurement_group_carrier/models/stock_picking.py
```diff
@@ -0,0 +1,47 @@
...
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _align_group_carrier(self):
+        for picking in self:
+            picking_carrier = picking.carrier_id
+            picking_group = picking.group_id
```
mt-software-de, line 13: You should check if there is a group set, if not ignore it. 
```suggestion
            picking_group = picking.group_id
            if not picking_group:
                continue
```

mt-software-de (changes requested):

delivery_procurement_group_carrier/models/stock_picking.py
jbaudoux: that's the purpose, to update only pickings of the group where a carrier was set (normally only outgoing in v14) but not set a carrier where it's not supposed to have one

mmequignon: done

delivery_procurement_group_carrier/models/stock_picking.py
mmequignon: done

delivery_procurement_group_carrier/models/stock_picking.py
mt-software-de: OK but when you will have more than Just one outgoing where the Carrier is set. Is it because this should Work also for v16 where the Carrier can be propagated?

jbaudoux (approved):

mt-software-de (approved):

PR #1898, opened 2025-03-09
[MIG] stock_picking_auto_revert: Migration to 17.0

dreispt:
- [ADD]stock_picking_auto_revert
- [IMP] stock_picking_auto_revert: black, isort, prettier
- [MIG] stock_picking_auto_revert: Migration to v15
- [IMP] stock_picking_auto_revert: pre-commit execution
- [MIG] stock_picking_auto_revert: Migration to 17.0


stock_picking_auto_revert/models/stock_picking.py
```diff
@@ -54,10 +43,15 @@ def action_revert_recreate(self):
...
+            "are not available in the destination location"
+        )
         for rm in return_wiz.product_return_moves:
-            sm = pick.move_lines.filtered(
-                lambda x: x.product_id.id == rm.product_id.id and x.state == "done"
+            sm = pick.move_ids.filtered(
+                lambda x, rm=rm: x.product_id.id == rm.product_id.id
+                and x.state == "done"
             )
             if rm.quantity < sum(sm.mapped("product_uom_qty")):
```
rousseldenis, line 56: Shouldn't we use `float_compare()`?

stock_picking_auto_revert/tests/test_stock_picking.py
```diff
@@ -49,76 +48,58 @@ def setUpClass(cls):
...
-        picking = self.so.picking_ids
-        picking.move_lines.write({"quantity_done": 5.0})
-        picking.move_lines.filtered(
-            lambda x: x.product_id == self.product2
-        )._action_cancel()
-        picking.button_validate()
+        # confirm our standard so and deliver
+        picking = self._do_confirm_and_deliver(self.so)
         # Create return picking
-        return_form = Form(self.env["stock.return.picking"])
```
rousseldenis: Why removing this way of testing ?

stock_picking_auto_revert/models/stock_picking.py
dreispt: This is the preexisting logic, the change is only to fix a lint error.
And this is an ID comparison, so I don't think float_compare() is helpful here.

stock_picking_auto_revert/tests/test_stock_picking.py
dreispt: It was not needed.

PR #1906, opened 2025-03-29
[14.0][IMP] delivery_procurement_group_carrier: Align Carrier once per group

mt-software-de:
I checked the tests and figured out that the test never has hit https://github.com/OCA/stock-logistics-workflow/blob/fc42bed4eebbf2b937c6c2ef3b89ddde4e37d7e5/delivery_procurement_group_carrier/models/stock_picking.py#L31 with a picking. 

That's because by default `set_delivery_line` sets the new carrier on all pickings of the proc group of the sale.order.

As result i extended the test to ensure the carrier is just set on the other transfers of the group if they have a carrier set. I changed the align carrier method to ensure it gets only once called per group. 

@jbaudoux @mmequignon

### DIFF
--- a/delivery_procurement_group_carrier/models/stock_picking.py
+++ b/delivery_procurement_group_carrier/models/stock_picking.py
@@ -1,34 +1,39 @@
 # Copyright 2025 Camptocamp SA
+# Copyright 2025 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import models
+from odoo.tools import groupby
 
 
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     def _align_group_carrier(self):
-        for picking in self:
-            picking_carrier = picking.carrier_id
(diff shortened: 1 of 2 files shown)


mt-software-de:
ping @jbaudoux @mmequignon

delivery_procurement_group_carrier/models/stock_picking.py
```diff
@@ -1,34 +1,38 @@
...
-            if picking_group and picking_carrier != picking_group.carrier_id:
-                picking_group.carrier_id = picking_carrier
-                need_align_pickings = self.search(
-                    [
-                        ("group_id", "=", picking_group.id),
+        for group, pickings in groupby(self, lambda pick: pick.group_id):
+            if not group:
+                continue
+            pickings = self.browse().union(*pickings)
+            carrier = pickings.carrier_id
```
jbaudoux, line 17: just to be sure
```suggestion
            carrier = pickings.carrier_id
            carrier.ensure_one()
```

jbaudoux (approved):

PR #1909, opened 2025-04-01
[18.0][MIG] stock_move_line_auto_fill: Migration to 18.0

MS-OSI:
Stock Move Line Auto Fill
===================

This module alows to auto fill quantities in picking operations and
autoassignment lots quantities.

In Odoo, if you schedule to transfer 50 products and only receive 49,
you have to change the quantity directly on the picking. As the quantity
by default is 0 for each line, you have to write the received quantity
on 49 lines.

In this module we have added a button that helps users to fill
automatically the scheduled quantities. Then, the user can just change
back the quantities for the product that hasn't been received yet.


stock_move_line_auto_fill/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Stock Move Line Auto Fill",
     "summary": "Stock Move Line auto fill",
-    "version": "17.0.1.0.0",
+    "version": "18.0.1.0.0",
```
rousseldenis, line 8: @MS-OSI Please put migration changes in migration commit

MS-OSI: @rousseldenis Done

matteotognini:
Hi @MS-OSI, Could you please rebase it? So I can test and leave a review! Thanks a lot

pedrobaeza:
Is this module still needed in 18? Seeing the default behavior, I doubt it.

matteotognini:
Thanks @pedrobaeza, I think is a default behavior so it's not needed in v18.

metaminux:
Hi @pedrobaeza 

Do we need to make an [OU-ADD] PR against OpenUpgrade to make this explicit ?

pedrobaeza:
Yes, to include it in `merge_modules` section in apriori.py.

metaminux:
OK, new PR submitted

pedrobaeza:
Closing as not needed.

PR #1783, opened 2024-12-04
[18.0][MIG] stock_partner_delivery_window: Migration to 18.0

thienvh332:

PR #1910, opened 2025-04-03
[18.0][MIG] stock_partner_delivery_window: Migration to 18.0

ArnauCForgeFlow:
Migration of **stock_partner_delivery_window** to version 18.0

Supersedes: https://github.com/OCA/stock-logistics-workflow/pull/1783


AaronHForgeFlow (approved):
It works ok, but I still see a "False" in the error message:

image

image

ArnauCForgeFlow:
Hey @AaronHForgeFlow that's because the partner doesn't have the timezone set, that's why its False.

AaronHForgeFlow:
Ok! so I assume the user must inform this value in order to use this feature.

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -20,17 +21,16 @@ def _onchange_scheduled_date(self):
         if not partner or anytime_delivery or outgoing_picking:
             return
         if not partner.is_in_delivery_window(self._planned_delivery_date()):
-            return {"warning": self._scheduled_date_no_delivery_window_match_msg()}
+            raise UserError(self._scheduled_date_no_delivery_window_match_msg())
```
sebalix: We are switching from a warning (non-blocking) to a blocking flow here. Is it intended?

sebalix: I think this change relates to this comment in previous PR: https://github.com/OCA/stock-logistics-workflow/pull/1783#discussion_r1910076100
But by "alert" it wasn't meant a blocking UserError, but a simple text displayed on top of the form with
```xml
<div class="alert alert-warning" role="alert">
...
</div>
```

ArnauCForgeFlow: @sebalix you are right, I switched it from UserError to alert now.

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+    def _planned_delivery_date(self):
+        return self.scheduled_date
+
+    @api.onchange("partner_id", "scheduled_date")
+    def _compute_delivery_date_warning_message(self):
+        self.ensure_one()
+        partner = self.partner_id
+        if not partner:
+            self.delivery_date_warning_message = False
+            return
```
sebalix: And set the default value of the computed field to False at the beginning (to not set it in `if/else` statements afterwards):
```suggestion
            picking.delivery_date_warning_message = False
            if not partner:
                continue
```

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+    delivery_date_warning_message = fields.Char(
+        compute="_compute_delivery_date_warning_message"
+    )
+
+    def _planned_delivery_date(self):
+        return self.scheduled_date
+
+    @api.onchange("partner_id", "scheduled_date")
```
sebalix:
```suggestion
    @api.depends("partner_id", "scheduled_date")
```

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+    def _compute_delivery_date_warning_message(self):
+        self.ensure_one()
```
sebalix: You cannot use `ensure_one()` in a computed field (it could be triggered on a recordset of size > 1), so we have to loop here:
```suggestion
        for picking in self:
```

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+        outgoing_picking = self.picking_type_id.code == "outgoing_picking"
+
+        if anytime_delivery or outgoing_picking:
+            self.delivery_date_warning_message = False
+        elif not partner.is_in_delivery_window(self._planned_delivery_date()):
+            self.delivery_date_warning_message = (
+                self._scheduled_date_no_delivery_window_match_msg()
+            )
+        else:
+            self.delivery_date_warning_message = False
```
sebalix:
```suggestion
        if anytime_delivery or outgoing_picking:
            continue
        elif not partner.is_in_delivery_window(picking._planned_delivery_date()):
            picking.delivery_date_warning_message = (
                picking._scheduled_date_no_delivery_window_match_msg()
            )
```

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, fields, models
+from odoo.tools.misc import format_datetime
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+    delivery_date_warning_message = fields.Char(
+        compute="_compute_delivery_date_warning_message"
+    )
```
sebalix: Field name is maybe too common, could conflict with other modules installed in the DB, I propose:
```suggestion
    partner_delivery_window_warning = fields.Char(
        compute="_compute_partner_delivery_window_warning"
    )
```

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,61 @@
...
+        self.ensure_one()
+        partner = self.partner_id
+        if not partner:
+            self.delivery_date_warning_message = False
+            return
+
+        anytime_delivery = partner and partner.delivery_time_preference == "anytime"
+        outgoing_picking = self.picking_type_id.code == "outgoing_picking"
+
+        if anytime_delivery or outgoing_picking:
```
sebalix: Not related to this migration, but these lines of code are wrong: there is no `outgoing_picking` picking type code (it's `outgoing`), so currently this check always return `False`, and I guess it's a good thing, as I don't see the point to have it returning `True` on delivery orders, making all this warning message useless.

`outgoing_picking` check should be replaced by `picking.picking_type_id.code != "outgoing"`, as it was before this commit: https://github.com/OCA/stock-logistics-workflow/commit/8464cfab1cd036e5dedcec68811d6b23bb996e57
(to fix in a separate commit so it can be backported easily)

ArnauCForgeFlow:
@sebalix Thanks for your comments! could you review it again?

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -22,7 +22,7 @@ def _compute_partner_delivery_window_warning(self):
                 continue
 
             anytime_delivery = partner and partner.delivery_time_preference == "anytime"
-            outgoing_picking = self.picking_type_id.code == "outgoing_picking"
+            outgoing_picking = picking.picking_type_id.code != "outgoing"
```
sebalix: Still two things to fix here:
- the `s/self/picking/` fix should land in your IMP commit that transform the onchange into a computed warning msg
- now we are doing `code != "outgoing"`, the variable name should become `not_outgoing_picking` to be meaningful

ArnauCForgeFlow:
@sebalix Done! let me know if it is okay now.

Thanks!

stock_partner_delivery_window/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+            anytime_delivery = partner and partner.delivery_time_preference == "anytime"
+            not_outgoing_picking = picking.picking_type_id.code != "outgoing"
+
+            if anytime_delivery or not_outgoing_picking:
+                continue
+
+            elif not partner.is_in_delivery_window(self._planned_delivery_date()):
+                self.partner_delivery_window_warning = (
+                    self._scheduled_date_no_delivery_window_match_msg()
+                )
```
sebalix, line 33: I tested on runboat, and with multiple delivery windows I get the following rendering:
image

We should transform the `\n` carriage returns into HTML new lines, maybe with that helper (didn't try):
```suggestion
                picking.partner_delivery_window_warning = self.env["ir.qweb.field.text"].value_to_html(
                    picking._scheduled_date_no_delivery_window_match_msg(), {}
                )
```

We should replace `self` by `picking` too.

sebalix: Or simply by changing the `fields.Char` into a `fields.Text` for the warning msg?

ArnauCForgeFlow: Done, thanks!

sebalix: Thank you, working fine now :)

sebalix (approved):

PR #1911, opened 2025-04-08
[16.0][IMP] stock_picking_partner_note propagate from partner to picking

santostelmo:
Partner notes are computed when picking is created. 
When partner notes change we update open pickings notes

### DIFF
--- a/stock_picking_partner_note/models/res_partner.py
+++ b/stock_picking_partner_note/models/res_partner.py
@@ -12,3 +12,15 @@ class ResPartner(models.Model):
         comodel_name="stock.picking.note",
         string="Picking Notes",
     )
+
+    def write(self, vals):
+        # Update notes on open pickings
+        res = super().write(vals)
+        if "stock_picking_note_ids" in vals:
+            self.env["stock.picking"].search(
+                [
+                    ("partner_id", "child_of", self.ids),
+                    ("state", "in", ("draft", "waiting", "confirmed", "assigned")),
+                ]
+            )._compute_note()
+        return res
--- a/stock_picking_partner_note/tests/test_stock_picking_partner_note.py
+++ b/stock_picking_partner_note/tests/test_stock_picking_partner_note.py
@@ -66,3 +66,27 @@ class StockPickingPartnerNote(common.TransactionCase):
         # We cannot delete a note that is already in use
         with self.assertRaises(UserError):
             partner_b.stock_picking_note_ids.unlink()
+
+    def test_update_partner_note_propagated_to_open_picking(self):
+        """Test that update of partner note is propagated to open pickings."""
+        with Form(self.env["sale.order"]) as order_form:
+            order_form.partner_id = self.partner_a
+            with order_form.order_line.new() as line_form:
+                line_form.product_id = self.product_a
+                line_form.product_uom_qty = 1
+
+        self.order = order_form.save()
+        self.order.warehouse_id.out_type_id.partner_note_type_ids = [
+            (6, 0, (self.note_type1 | self.note_type2).ids)
+        ]
+        self.order.action_confirm()
+        self.partner_a.write(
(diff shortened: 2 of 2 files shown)


cyrilmanuel (approved):

stock_picking_partner_note/models/stock_picking.py
```diff
@@ -10,7 +10,7 @@ class StockPicking(models.Model):
 
     note = fields.Html(compute="_compute_note", store=True)
 
-    @api.depends("partner_id")
+    @api.depends("partner_id", "origin")
```
jbaudoux: origin is not used in the compute, no reason to add it. It's purely based on partner and picking type.
```suggestion
    @api.depends("partner_id", "picking_type_id")
```

jbaudoux (changes requested):

stock_picking_partner_note/models/stock_picking.py
santostelmo: The idea behind this depends is to update note when picking is merged (origin is the concatenation of all orders)

jbaudoux: I got it but the partner is not changing, so the note will remain the same no matter how many SOs you have in the delivery

santostelmo: That's the point notes can change in the partner.
Use case:
Partner1 note1 added: Take care with the cat
Sale1 confirmed and note 1 propagated to picking1
Partner1 note2 added: Take care with the dog
Sale2 confirmed and note 1 + note2 should be propagated to picking1

jbaudoux: Then it should update even without the SO2.
What about a hook on partner for any change of `stock_picking_note_ids`, find all not done/cancel and not printed stock.picking and call `_compute_note` to update the note.

santostelmo: > Then it should update even without the SO2. What about a hook on partner for any change of `stock_picking_note_ids`, find all not done/cancel and not printed stock.picking and call `_compute_note` to update the note.

And when do you trigger `_compute_note` in this case ? Onchange at partner level ?

jbaudoux: something like this
```
_inherit = "res.partner"
def write(self, vals):
  if `stock_picking_note_ids` in vals:
     open_partner_picking._compute_note()
```

santostelmo: That way is an alternative. I'll implement like that

santostelmo: @jbaudoux propagating notes from partner to open pickings

stock_picking_partner_note/models/res_partner.py
```diff
@@ -12,3 +12,15 @@ class ResPartner(models.Model):
...
         string="Picking Notes",
     )
+
+    def write(self, vals):
+        # Update notes on open pickings
+        res = super().write(vals)
+        if "stock_picking_note_ids" in vals:
+            self.env["stock.picking"].search(
+                [
+                    ("partner_id", "child_of", self.ids),
```
jbaudoux, line 22: the note is coming from the partner, I don't see the reason for the child_of
```suggestion
                    ("partner_id", "in", self.ids),
```

stock_picking_partner_note/models/res_partner.py
```diff
@@ -12,3 +12,15 @@ class ResPartner(models.Model):
...
+                    ("state", "in", ("draft", "waiting", "confirmed", "assigned")),
+                ]
```
jbaudoux, line 24: never modify a printed picking
```suggestion
                    ("printed", "=", False),
                ]
```

jbaudoux (changes requested):

PR #1920, opened 2025-04-17
[16.0][ADD] stock_picking_return_empty_package

jbaudoux:
Ensure only package content is put in stock when doing a picking return using standard wizard.

Requires to enable feature on the return operation type

cc @rousseldenis @sbejaoui


rousseldenis (approved):
LGTM

PR #1929, opened 2025-04-28
[16.0][FIX] stock_move_line_change_lot: Use id with browse

mmequignon:
Default value in vals.get() might be used by browse(), and should be an ID, not a record

### DIFF
--- a/stock_move_line_change_lot/models/stock_move_line.py
+++ b/stock_move_line_change_lot/models/stock_move_line.py
@@ -85,7 +85,7 @@ class StockMoveLine(models.Model):
                 vals.get("location_id", move_line.location_id.id)
             )
             package = move_line.package_id.browse(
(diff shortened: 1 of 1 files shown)


sebalix (approved):

jbaudoux (approved):

stock_move_line_change_lot/models/stock_move_line.py
```diff
@@ -85,7 +85,7 @@ def write(self, vals):
                 vals.get("location_id", move_line.location_id.id)
             )
             package = move_line.package_id.browse(
-                vals.get("package_id", move_line.package_id)
+                vals.get("package_id", move_line.package_id.id)
             )
```
sebalix, line 89: I would even add a call to `exists()` to filter out wrong ID in user's input
```suggestion
            ).exists()
```

PR #1930, opened 2025-04-30
[WIP] [14.0] [IMP] stock_picking_invoice_link: add security group to restrict canceling stock moves linked to invoices/bills

rrebollo:
This PR introduces a restriction to prevent the cancellation of stock moves that are already linked to invoices or vendor bills. Only users who are members of the "Allow to cancel stock moves linked to invoices/bills" group will be authorized to perform this action.

## Why Restrict Cancelling Stock Moves Linked to Invoices/Bills

- **Data Integrity**: Prevents inconsistencies between inventory and accounting.
- **Controlled Access**: Cancellation is still allowed, but only for users in the designated security group.
- **Legal and Financial Compliance**: Avoids issues with invoiced or billed records.
- **Auditability**: Maintains a clear and traceable record of exceptional actions.
- **Best Practice**: Encourages using proper workflows like credit notes or returns.

## ROADMAP

- [x] create security group
- [x] restrict cancellation at the stock move model level
- [x] update translations
- [ ] create a test

@BinhexTeam

### DIFF
--- a/stock_picking_invoice_link/README.rst
+++ b/stock_picking_invoice_link/README.rst
@@ -36,6 +36,10 @@ In standard, if you make a partial delivery and invoice it, then make remaining
 delivery and invoice it, it is impossible to known to what delivery the
 invoices relate to. You only have the quantity.
 
+Additionally, the cancellation of stock moves already linked to invoices or
+bills is restricted to users who belong to the "Allow to cancel stock moves
+linked to invoices/bills" group.
+
 This module is also useful if you want to present data on the invoice report
 grouped by deliveries.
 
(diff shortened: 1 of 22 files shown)


rrebollo:
After careful consideration during the design of the tests, we have concluded that this feature would fit better as part of the `stock_picking_invoicing` addon from `OCA/account-invoicing`.

Although the code base related to the issue this feature aims to address is indeed in the current addon (responsible for the picking-journal items link), the conditions that lead to the issue itself (the abuse of cancelling pickings that are not done but already linked to invoices/bills) cannot actually occur solely with this addon.

As far as this addon is concerned, the only pickings linked to invoices or bills are those that are already in the `done` state.

Therefore, we are closing this PR in favor of proposing it directly to the `OCA/account-invoicing` repository.

PR #1921, opened 2025-04-18
[18.0][MIG] stock_move_backdating

bosd:

PR #1594, opened 2024-05-15
[16.0][MIG] stock_move_backdating

Saran440:
This PR migrated from version15 and enhance to supported backdate scrap

PR #1935, opened 2025-05-07
[17.0][MIG] stock_move_backdating: Migration to 17.0

map-240221:
Using the same base as the migration to 16.0 and 18.0 in #1594 and #1921, respectively. All modifications belong to those two requests.

Errors on tests are caused by `stock_valuation_layer_usage` and `product_cost_price_avco_sync` (I believe the same error also happened on the other requests as well). Might look at them later once this is done.

The previous merge request (#1932) can be ignored as it's pretty "dirty" with all trials and errors. All further changes will be made on this request instead.


stock_move_backdating/models/stock_quant.py
```diff
@@ -0,0 +1,112 @@
...
+                new_move_line_ids = []
+                for line_command in res["move_line_ids"]:
+                    if (
+                        isinstance(line_command, tuple | list)
+                        and len(line_command) == 3
+                        and line_command[0] == 0
+                    ):
+                        line_values = line_command[2]
+                        # Add the backdating date to the move line values
+                        line_values["date"] = date_to_use
```
kafai-lam:
```suggestion
                        line_values["date_backdating"] = date_to_use
```

I think we should use `date_backdating` instead of `date` in move line vals in order to trigger `_backdating_action_done` and `_backdating_stock_valuation_layers` call on stock move done.

map-240221: Fine by me, seems to be in-line with the migration to 16.0 as well.
What do you think about it @bosd ?

stock_move_backdating/models/stock_quant.py
```diff
@@ -0,0 +1,112 @@
...
+            location_dest_id,
+            package_id=package_id,
+            package_dest_id=package_dest_id,
+        )
+
+        date_backdating_ctx = self.env.context.get("date_backdating", False)
+
+        date_to_use = date_backdating_ctx
+
+        if date_to_use:
```
FernandoRomera: For consistency in the module (line 33), rename date variable.
date_backdating = self.env.context.get("date_backdating", False)
if date_backdating:

and in
line_values["date"] = date_to_use
replace by:
line_values["date_backdating"] = date_backdating

map-240221:
Updated with the suggestions

map-240221:
Added fix for the unit test problem from https://github.com/OCA/stock-logistics-workflow/pull/1594

Can you check again, @kafai-lam @FernandoRomera ?

FernandoRomera (approved):
LGTM

bosd (approved):

kafai-lam:
qqq

PR #1939, opened 2025-05-12
[17.0][MIG] stock_move_quantity_product_uom

DavidJForgeFlow:
Standard migration to v17.0

Backport from 18.0


stock_move_quantity_product_uom/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Stock Move Quantity Product UOM",
-    "version": "18.0.1.0.1",
```
LoisRForgeFlow: 17.0.1.0.0

jbaudoux:
it's on the wrong repo but was the same mistake on v18. Should have been on https://github.com/OCA/stock-logistics-warehouse

DavidJForgeFlow:
The last commit should be forward ported to v18.0!

LoisRForgeFlow:
> it's on the wrong repo but was the same mistake on v18. Should have been on https://github.com/OCA/stock-logistics-warehouse

@jbaudoux  That would have been more accurate indeed, what would you do? Maybe we can put a note in v18 in order to point that it should moved during migration to 19?

LoisRForgeFlow:
@jbaudoux @DavidJForgeFlow I think we can merge this one now as the module is already in this repo in v18. I would move it in v18, but I do not want to find the hard way that there is a problem with any tool in the OCA stack because I've never seen a repo change of a merged module.

We'll add a note in v18 so that we don't forget to move in v19.

stock_move_quantity_product_uom/hook.py
```diff
@@ -0,0 +1,31 @@
...
+        UPDATE stock_move sm
+        SET quantity_product_uom = sm.quantity
+        FROM product_product pp
+        JOIN product_template pt ON pt.id = pp.product_tmpl_id
+        WHERE sm.product_id = pp.id
+          AND pt.uom_id = sm.product_uom
+        """
+    )
+    env.cr.commit()
+    # env.cr.execute(
```
rousseldenis: @DavidJForgeFlow Could you remove unused code ?

stock_move_quantity_product_uom/hook.py
```diff
@@ -0,0 +1,22 @@
...
+
+def pre_init_quantity_product_uom(env):
+    if not column_exists(env.cr, "stock_move", "quantity_product_uom"):
+        create_column(env.cr, "stock_move", "quantity_product_uom", "float8")
+    # The installation of this module in a large database could take some time
+    # (and maybe exceed the CPU time limit). That's why we are pre-filling some
+    # of the values for the new column with a simple SQL sentence and the rest of
+    # the values will be filled out normally by Odoo.
+    # This reduces the installation time.
+    env.cr.execute(
```
LoisRForgeFlow: This is not true, Odoo only computes the values when it creates the column, if you are doing it before, you need to provide values for all the rows. 

I actually did a test locally, this is the result:

```
stock_move_quantity_product_uom_v17_1=# select id, name, reference, state, product_uom, quantity, quantity_product_uom, product_uom_qty, product_qty from stock_move where product_id = 37;
 id |    name     |  reference   |  state   | product_uom | quantity | quantity_product_uom | product_uom_qty | product_qty 
----+-------------+--------------+----------+-------------+----------+----------------------+-----------------+-------------
 52 | TEST widget | WH/OUT/00008 | assigned |           2 |     1.00 |                      |            1.00 |        12.0
 53 | TEST widget | WH/IN/00010  | assigned |          28 |     2.00 |                      |            2.00 |        20.0
 42 | TEST widget | WH/IN/00006  | done     |          27 |    10.00 |                      |           10.00 |         1.0
 44 | TEST widget | WH/IN/00007  | done     |          28 |     2.00 |                      |            1.00 |        10.0
 46 | TEST widget | WH/IN/00008  | done     |           2 |     3.00 |                      |            2.00 |        24.0
 48 | TEST widget | WH/IN/00009  | done     |          27 |    30.00 |                      |           20.00 |         2.0
 43 | TEST widget | WH/IN/00007  | done     |           1 |     3.00 |                    3 |            3.00 |         3.0
 49 | TEST widget | WH/IN/00009  | done     |           1 |     5.00 |                    5 |            0.00 |         0.0

```

Only the ones with uom ID 1 (unit) were filled.

LoisRForgeFlow (changes requested):

stock_move_quantity_product_uom/hook.py
DavidJForgeFlow: You are right, this was a mistake on my part. I just added a post script to cover this cases.

LoisRForgeFlow: Works good now :+1:

LoisRForgeFlow (approved):
All good now

/ocabot merge nobump

LoisRForgeFlow:
> @jbaudoux @DavidJForgeFlow I think we can merge this one now as the module is already in this repo in v18. I would move it in v18, but I do not want to find the hard way that there is a problem with any tool in the OCA stack because I've never seen a repo change of a merged module.
> 
> We'll add a note in v18 so that we don't forget to move in v19.

@jbaudoux readme note added in https://github.com/OCA/stock-logistics-workflow/pull/1949

PR #1946, opened 2025-05-18
[14.0][FIX] delivery_procurement_group_carrier: Do not align carrier again, allow to set carrier to False

mt-software-de:
Right now it not is not possible to set the carrier on a picking to False, since there is the `ensure_one` check. 
Also the method `_align_group_carrier` gets called multiple times, if the carrier gets set in `_align_group_carrier`, with using the already existing context key this will not happen. 

cc @jbaudoux @mmequignon @i-vyshnevska @raumschmiede-joshuaL


mt-software-de:
cc @raumschmiede-joshuaL

rousseldenis (approved):
LGTM

jbaudoux (approved):

PR #1947, opened 2025-05-19
[18.0][MIG] stock_picking_group_by_partner_by_carrier

theangryangel:
I needed to make some assumptions and adjustments to this module based on how 18.0 changes the standard / out-of-the-box multi-step routes from pull to push routes.

Notable changes:
* The module now supports both push and pull rule configuration. When in the new-style push rules, the behaviour is slightly different.
* Now depends on `stock_warehouse_out_pull` to easily/quickly accommodate the fact that historical database may be configured with pull rules and new database will have push rules.
* `group_pickings` is now visible based on the configuration from stock_warehouse_out_pull - if push rules then it's shown as per previous behaviour. If it's new pull rules then it's shown at all times.
* `StockMove._domain_search_picking_handle_move_type` required an additional if statement added to ensure that `TestMergeWizard` passed all tests
* `StockPicking._merge_procurement_groups` required a change in the from `picking_type_id.code != "outgoing"` to checking the `picking_type_id.group_pickings`
* Switched from itertools.groupby to odoo.tools.groupby
* New tests were added to handle the push rules
* Old test file handling pull rules was renamed for clarity
* `sale_stock` was added as a dependency and the delivery slip report inherit_id changed as it was depending on a field added through `sale_stock`


stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -114,8 +111,11 @@ def _domain_search_picking_handle_move_type(self):
         By default the move type is taken from the procurement group.
         Override to customize this behavior.
         """
-        # avoid mixing picking policies
-        return [("move_type", "=", self.group_id.move_type)]
+        if self.group_id.move_type:
+            # avoid mixing picking policies
+            return [("move_type", "=", self.group_id.move_type)]
```
jbaudoux: if the move type is missing on the procurement group, it now defaults to the one configured on the picking type. Otherwise, we can assume direct as default.
```suggestion
        group = self.group_id.move_type or self.picking_type_id.move_type or "direct"
        return [("move_type", "=", self.group_id.move_type)]
```

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
```diff
@@ -111,21 +111,21 @@ def _update_merged_origin(self):
...
+        sales = move_groups.sale_id + move_groups.sale_ids
         partners = move_groups.sale_id.partner_id
         name = _("Merged procurement")
         if partners:
             name = _(
                 "Merged procurement for partners: %(partners_name)s",
                 partners_name=", ".join(partners.mapped("display_name")),
             )
-        return {"sale_ids": [(6, 0, sales.ids)], "name": name}
+        return {"sale_ids": [(6, 0, sales.ids)], "name": name, "sale_id": False}
```
jbaudoux: Do not include https://github.com/OCA/stock-logistics-workflow/pull/1789 in your migration commit.
And as I explained, this is a wrong approach as it creates incompatibility with all modules relying on this standard `sale_id` value.

stock_picking_group_by_partner_by_carrier/report/report_delivery_slip.xml
```diff
@@ -167,13 +168,13 @@
 
         <!-- Replace the customer refs with the one based on the move line displayed -->
         <xpath
-            expr='//p[@t-field="o.sudo().sale_id.client_order_ref"]'
+            expr='//div[@t-field="o.sudo().sale_id.client_order_ref"]'
```
jbaudoux, line 175: Not linked to the migration but shouldn't the template inherit `sale_stock.report_delivery_document_inherit_sale_stock` instead ?
If you do this change, do it in a separate fix commit

stock_picking_group_by_partner_by_carrier/views/stock_picking_type.xml
```diff
@@ -5,13 +5,9 @@
...
         <field name="arch" type="xml">
-            <field name="show_reserved" position="after">
-                <field
-                    name="group_pickings"
-                    groups="base.group_no_one"
-                    attrs="{'invisible': [('code', '!=', 'outgoing')]}"
-                />
-            </field>
+            <xpath expr="//group[@name='first']/group[2]" position="inside">
+                <field name="group_pickings" />
```
jbaudoux: This change needs a big warning in the module readme.
You can configure grouping of picking only in case of push flow. Otherwise, only the outgoing can be enabled

jbaudoux (changes requested):
Thanks for the work.

It's nice you added tests for the new push flow of odoo but we need to keep the tests for the case you continue using pull flow to ensure non regression.
You can duplicate the impacted tests. And you can rely on https://github.com/OCA/stock-logistics-warehouse/pull/2353 for applying pull flow setup.

jbaudoux:
@theangryangel your changes to support the new push flow are legit and I don't see any change breaking the previous behavior. Sounds good to me. Keeping also the previous tests as they were should show there are no regressions.
If you want to use https://github.com/OCA/stock-logistics-workflow/pull/1789, put it in a separate PR as it creates incompatibilities and won't be merged as is.

theangryangel:
>  I don't see any change breaking the previous behavior. Sounds good to me. Keeping also the previous tests as they were should show there are no regressions.

I had significant issues trying to support both workflows. It might've been that I was having a blurry-eyed halucination from a 2am wake up from a toddler 😅 

I'll take a look at some point today or tomorrow and re-add those tests :)

Thank you for the feedback <3

stock_picking_group_by_partner_by_carrier/views/stock_picking_type.xml
theangryangel: I've added some additional logic to show/hide based on stock_warehouse_out_pull and I've added some extra information to the documentation describing the difference in configuration. Not sure if it's a strong enough message though.

stock_picking_group_by_partner_by_carrier/models/stock_picking.py
theangryangel: Reverted.

theangryangel:
I believe that I have made the relevant changes. I've incorporated a fix from bizappdev and the suggestions from jbaudoux. I've updated the PR description with the current state of things.

I'm not sure if the USAGE and CONFIGURATION changes are clear enough. Input would be appreciated.

stock_picking_group_by_partner_by_carrier/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,21 @@
...
+When the delivery slip is printed, the list of pending quantities to
+deliver is shown at the end, grouped by order.
+
+The grouping can also be applied in case of external resupply. Moves at
+the destination of the same delivery address defined on the resupply
+stock rule will be grouped in a same delivery order.
+
+Note: The grouping is currently not propagated to the pulled internal
+moves if you don't use the module stock_available_to_promise_release in
+the wms repository.
```
jbaudoux: it's being moved
```suggestion
the stock-logistics-workflow repository.
```

jbaudoux (approved):
Thanks for the updates

can you rewrite your fix commit to add the module name?
`[FIX] stock_available_to_promise_release: report inherit_id`

stock_picking_group_by_partner_by_carrier/__manifest__.py
```diff
@@ -0,0 +1,35 @@
...
+{
+    "name": "Stock Picking: group by partner and carrier",
+    "Summary": "Group sales deliveries moves in 1 picking per partner and carrier",
+    "version": "18.0.1.0.0",
+    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "category": "Warehouse",
+    "depends": [
+        "delivery_procurement_group_carrier",
+        "sale_stock",
```
bizzappdev, line 13: IMO, the stock_delivery module https://github.com/odoo/odoo/blob/18.0/addons/stock_delivery/__manifest__.py#L14 is already adding the dependency, explicit dependency is unnecessary.

bizzappdev (changes requested):
@theangryangel Thank you for the contribution,
Please squash administrative commits (if any) with the previous commit for reducing commit noise as per the [migration checklist](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0) requirements. Ref: [Merge/squash the commits generated by bots or Weblate](https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate)

stock_picking_group_by_partner_by_carrier/__manifest__.py
simahawk: It depends. If we have a direct dependency on stock_delivery (a field, an override, a view, whatever), it's better to keep it explicit.

theangryangel: I believe that I have followed the rules in terms of squashing and reordering. Thank you for pointing it out bizzappdev. I'd completely forgotten that step. It's been a moment since I last contributed :/

theangryangel:
I believe I've made all the requested changes. Anything else, please do let me know :)

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -0,0 +1,128 @@
...
+                render_values={"self": self.picking_id, "origin": sales, "edit": True},
+                subtype_xmlid="mail.mt_note",
+            )
+
+    def _search_picking_for_assignation_domain(self):
+        domain = super()._search_picking_for_assignation_domain()
+        if (
+            not self.picking_type_id.group_pickings
+            or self.partner_id.disable_picking_grouping
+            or self.group_id.sale_id.picking_policy == "one"
```
jbaudoux, line 76: Not grouping when `one` should be configurable

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # TODO: this part and everything related to generic grouping
+    # should be split into `stock_picking_group_by` module.
+    def _assign_picking_group_domain(self):
+        domain = [
+            # same partner
+            ("partner_id", "=", self.group_id.partner_id.id),
+            # don't search on the procurement.group
+        ]
+        domain += self._domain_search_picking_handle_move_type()
```
jbaudoux, line 96: When the move type is `one`, the deadline should be respected when there is a `sale_line_id.order_id.commitment_date`

stock_picking_group_by_partner_by_carrier/models/stock_move.py
```diff
@@ -0,0 +1,128 @@
...
+            domain += [
+                ("carrier_id", "=", self.group_id.carrier_id.id),
+            ]
+        if self.env.context.get("picking_no_copy_if_can_group"):
+            # we are in the context of the creation of a backorder:
+            # don't consider the current move's picking
+            domain.append(("id", "!=", self.picking_id.id))
+        return domain
+
+    def _domain_search_picking_handle_move_type(self):
```
jbaudoux, line 108: What we want to hook is the retrieval of the `move_type`, not the domain

jbaudoux (commented):
Not linked to the migration, but would be nice to allow merging also when the policy is `one` unless there is a commitment date

stock_picking_group_by_partner_by_carrier/models/stock_move.py
theangryangel: So I'll counter that one - The customer we're porting this for actually specifically ignores the commitment date if they can consolidate, risking upsetting the customer by delaying one order by a day or two, but proactively inform the customer that they're gonna get order A, B and at once, for example.

The customer doesn't really get a choice, the margins of the business are "low".

I'd look to make this configurable, but I'm hard pressed to devote time to add additional features on this PR right now

bizzappdev (approved):
Code and Funtional test LGTM

stock_picking_group_by_partner_by_carrier/models/stock_move.py
jbaudoux: improved in https://github.com/OCA/stock-logistics-workflow/pull/1987

stock_picking_group_by_partner_by_carrier/models/stock_move.py
jbaudoux: @theangryangel See https://github.com/OCA/stock-logistics-workflow/pull/1987

PR #1952, opened 2025-05-22
[18.0][MIG] stock_dynamic_routing: Migration to 18.0

sebalix:
Was hosted in OCA/wms.

`<stock.move>._split` method changed with "quantity pocalypse" Odoo commit: https://github.com/odoo/odoo/commit/7dda6bb92715e , code adapted accordingly.


stock_dynamic_routing/models/stock_routing.py
```diff
@@ -256,6 +256,7 @@ def _routing_rule_for_moves(self, moves):
     # time we clear it. We only need a local cache used for the duration of the
     # execution of
     @lru_cache
+    @staticmethod
```
simahawk, line 259: static method w/ self?

stock_dynamic_routing/models/stock_routing.py
```diff
@@ -0,0 +1,285 @@
...
+    def _routing_message_template(self):
+        pull_rules = self.rule_ids.filtered(lambda r: r.method == "pull")
+        push_rules = self.rule_ids.filtered(lambda r: r.method == "push")
+        pull_message = ""
+        push_message = ""
+
+        if pull_rules:
+            rule_messages = []
+            for rule in pull_rules:
+                msg = _(
```
simahawk:
```suggestion
                msg = self.env._(
```
same for all gettext calls

simahawk (commented):
LG. Minor remarks

stock_dynamic_routing/models/stock_routing.py
sebalix: oups, forgot to remove this parameter, thanks

sebalix: Changed the caller from `self` to `StockRouting` as well

jbaudoux (approved):
maybe add in readme:

Compared to new odoo dynamic push rules, here all steps are planned in advance. The move to the final destination always exists and and the put-away has already been executed so that the final bin is already known. The intermediate step before reaching this final destination will also have it's put-away computed.
As each step is planned in advance, you can also properly respect the operation shipping policy.

sebalix:
@jbaudoux pushed in DESCRIPTION.rst, can you check? As I do not master really well what you wrote ;)

rousseldenis (approved):
LGTM. Functional review

jbaudoux:
We should investigate those 2 PRs in v14.0 that could not make it in v16.0
https://github.com/OCA/wms/pull/809

sebalix:
@jbaudoux @rousseldenis I pushed a small change in `stock_dynamic_routing/tests/test_routing_pull.py` regarding the reservation method of WH picking types:
```python
        cls.wh.pick_type_id.reservation_method = "manual"
        cls.wh.out_type_id.reservation_method = "manual"
```

It was making tests in https://github.com/OCA/stock-logistics-workflow/pull/1989 failing. By doing that, we align the behavior as it was in v14/v16, and we avoid side-effects in tests when confirming moves.

sebalix:
@jbaudoux I forward ported missing commits from 14.0 directly there, tests are :green_circle:

PR #1957, opened 2025-05-26
[18.0] [MIG] delivery_package_default_shipping_weight: Migration to 18.0

bizzappdev:


delivery_package_default_shipping_weight/models/stock_package_type.py
```diff
@@ -0,0 +1,34 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.tools import float_compare
+
+
+class PackageType(models.Model):
+    _inherit = "stock.package.type"
+
+    package_default_shipping_weight = fields.Float(
```
jbaudoux, line 11: There is already a weight field on the package type. I would not an additional one but add a boolean 
```suggestion
    is_base_weight_shipping_weight = fields.Boolean(
    string="Is shipping weight",
    help="When flagged, the package type weight will be the package total weight. The product weight won't be added"
    )
```

delivery_package_default_shipping_weight/models/stock_quant_package.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import api, models
+
+
+class StockQuantPackage(models.Model):
```
jbaudoux, line 6: instead of extending this, you need to adapt `_compute_shipping_weight` in `choose.delivery.package`

delivery_package_default_shipping_weight/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+This module allows to define a Default shipping weight on Delivery
+packages (product.packaging) in order to set automatically the Shipping
+weight on the packages (stock.quant.package) using this Package type.
```
jbaudoux, line 3:
```suggestion
This module allows to use the delivery package weight as the default total shipping weight
for a package using this delivery package. The weight of the products is then not anymore added to the package type weight.
```

jbaudoux (changes requested):
This needs to be refactored.
Move to deliver-carrier repo.
Rename to delivery_package_type_shipping_weight

delivery_package_default_shipping_weight/models/stock_quant_package.py
```diff
@@ -0,0 +1,34 @@
...
+    _inherit = "stock.quant.package"
```
jbaudoux, line 7: you also need to extend `_compute_weight` on `stock.quant.package`

jbaudoux (changes requested):
it would be more accurate to also adapt
* https://github.com/odoo/odoo/blob/f2421a625a7179787fbbe34af80642ffa60aada6/addons/stock_delivery/models/delivery_carrier.py#L109
* https://github.com/odoo/odoo/blob/f2421a625a7179787fbbe34af80642ffa60aada6/addons/stock_delivery/models/delivery_carrier.py#L147

maybe set a value in context of order/picking to that `_get_estimated_weight` returns 0

jbaudoux:
replaced by https://github.com/OCA/delivery-carrier/pull/1022

PR #1960, opened 2025-05-26
[17.0][ADD] stock_picking_backorder_strategy_restrict

bizzappdev:
This module restricts the creation of backorders if configured at the level of operation type.


stock_picking_backorder_strategy_restrict/readme/INSTALL.md
```diff
@@ -0,0 +1,5 @@
+Your preferred way to install addons will work with Stock Picking Backorder Strategy: Restrict.
+
+An easy way to install it with all its dependencies is using pip:
+
+pip install --pre odoo11-addon-stock-picking-backorder-strategy-restrict then restart Odoo, update the addons list in your database, and install the Stock Picking Backorder Strategy Restrict application.
```
rousseldenis: IMHO this has no sense (because of version). Was it generated through IA ?

bizzappdev: @rousseldenis, We added it from a similar module. Now, we have removed the install file, which is reflected in the readme.

PR #1966, opened 2025-05-27
[18.0] [MIG] stock_move_source_relocate: Migration to 18.0

bizzappdev:
Moved module from [wms](https://github.com/OCA/wms)


stock_move_source_relocate/views/stock_source_relocate_views.xml
```diff
@@ -0,0 +1,95 @@
...
+                    <div class="oe_button_box" name="button_box" />
+                    <widget
+                        name="web_ribbon"
+                        title="Archived"
+                        bg_color="bg-danger"
+                        invisible="active"
+                    />
+                    <group string="Applicability" name="configuration">
+                        <group>
+                            <field name="active" invisible="1" />
```
twalter-c2c: If I am not mistaken, it is not needed to have such invisible fields in the view, they are added automatically now.

From the OCA guidelines:
> Missing fields used in domains and other attributes are added to the view as invisible automatically. If you encounter fields in views with invisible="True", they can probably be removed. Reference: https://github.com/odoo/odoo/pull/137031

stock_move_source_relocate/readme/CREDITS.md
```diff
@@ -0,0 +1,2 @@
+The migration of this module from 13.0 to 14.0 was financially supported
+by Camptocamp
```
twalter-c2c: Past credits should be removed, I guess.

From the OCA guidelines:
> Remove references in CREDITS.rst of past financed migrations by other companies.

twalter-c2c (commented):
Technical review only. Looks good to me, a couple of tiny comments.

stock_move_source_relocate/views/stock_source_relocate_views.xml
```diff
@@ -60,24 +60,24 @@
             </search>
         </field>
     </record>
-    <record id="view_stock_source_relocate_tree" model="ir.ui.view">
```
twalter-c2c: I am not sure if you need to rename it, in the Odoo core, an `id` of views remains unchanged.

bizzappdev: @twalter-c2c 
Multiple MR reviewers are asking to replace the tree with a list; technically, it is right. that is why we are changing it.

bizzappdev: Reverting it now.
Just check the guidelines.
Replace tree view type by list. This means everywhere in Python, JS or XML where there's a mentioning of this type, but keep XML-IDs that contain tree in its name for avoiding subsequent changes in dependent modules.

bizzappdev: @twalter-c2c 
Proper changes are done.

jbaudoux (approved):

stock_move_source_relocate/views/stock_source_relocate_views.xml
twalter-c2c: Thank you @bizzappdev The naming convention in this case is confusing now, I have the impression that it is simply an in-between state now.

twalter-c2c (approved):

PR #1970, opened 2025-05-29
[16.0] stock_move_line_change_lot: pre-commit fixes from v18 + split write

simahawk:


stock_move_line_change_lot/models/stock_move_line.py
```diff
@@ -61,11 +61,38 @@ def _change_lot_free_other_lines(self, need, location, product, lot, package):
...
         if self.env.context.get("bypass_reservation_update"):
-            return super(StockMoveLine, self).write(vals)
+            return super().write(vals)
 
         if not vals.get("lot_id"):
-            return super(StockMoveLine, self).write(vals)
+            return super().write(vals)
 
+        res, already_processed, to_reassign_moves = self._do_change_lot(vals)
+        res &= super(StockMoveLine, self - already_processed).write(vals)
```
jbaudoux, line 70: Skip this if recordset is empty

simahawk: if the recordset is empty it won't do anything

sebalix (approved):

stock_move_line_change_lot/models/stock_move_line.py
jbaudoux: it will still go through all the steps and checks inside write for nothing

simahawk: I'll merge this one and take care of this on #1825

PR #1979, opened 2025-06-05
[18.0][MIG] stock_checkout_sync

sebalix:
Migrated from `OCA/wms#14.0`.

### Issue about updating locations directly on moves and on backorders

Starting from Odoo 18.0, source/dest locations on moves are coming directly from the related `stock.picking`. Fields are `stored=True / readonly=False`, but when Odoo is putting a backorder move in its own transfer, this is resetting the source/dest locations that were set on it (see https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_picking.py#L1585).

Should we split these moves in their own picking (with the expected dest location) when syncing them?

```
M1 (assigned) A -> B
M2 (assigned) A -> B
```
Sync both moves to reach C:
```
M1 (assigned) A -> C
M2 (assigned) A -> C
```
Validate M1, so we get M2 in its own backorder transfer:
```
M1 (done)     A -> C
M2 (assigned) A -> B    # /!\ Reset to 'stock.picking' dest location
```
An idea would be to set the synced dest location not only on the moves themselves, but on their `stock.picking`, so all moves in this transfert will be synced to reach `C`.

The issue by doing that is it doesn't fulfill an existing use-case, which is having one of the moves having a different `stock.picking` as a destination than the others, so not impacted by the sync of other moves:

```
M1 (assigned) A -> B from PICK1, targetting M3 part of PACK1
M2 (assigned) A -> B from PICK1, targetting M4 part of PACK2
```
Sync M1 to C:
```
M1 (assigned) A -> C from PICK1, targetting M3 part of PACK1
M2 (assigned) A -> B from PICK1, targetting M4 part of PACK2
```
In 14.0 we were able to get different moves (M1 + M2) part of the same transfert (PICK1) but having different dest locations (as soon as they respect the parent location defined by the `stock.picking`).
Keeping this logic in 18.0 seems dangerous:

- technically that could still work, excepted when dealing with backorders as seen above (dest location is reset)
- updating directly location fields on the move seems fragile in 18.0, the system is thought to update locations on the related `stock.picking`, not directly on moves independently (I see it that way, could be wrong).

**EDIT:** after discussion with @jbaudoux , the use-case having moves within the same PICK transfer targetting different PACK transfers should not happen / be supported => removing/adapting tests accordingly + sync the location both on `stock.picking` + moves, no discrepancy allowed.


ivantodorovich (approved):

twalter-c2c (approved):
Technically, LGTM.

stock_checkout_sync/wizards/stock_move_checkout_sync.py
```diff
@@ -0,0 +1,129 @@
...
+        help="Destination operation for these moves.",
+    )
+    # in case we have several steps to sync all the moves,
+    # this field will contain them
+    remaining_help = fields.Html(readonly=True)
+    done_dest_picking_ids = fields.Many2many(
+        comodel_name="stock.picking",
+        relation="stock_move_checkout_sync_stock_picking_done_dest_rel",
+    )
+    show_skip_button = fields.Boolean(default=False)
```
jbaudoux: i guess this should be enough
```suggestion
    show_skip_button = fields.Boolean()
```

jbaudoux (approved):

marcelsavegnago (approved):
LGTM

jbaudoux (commented):
As discussed, I'm not sure it's still right to change the destination of the moves and not of the picking as now the destination of the move is computed.

jbaudoux (changes requested):
Because it doesn't work anymore on backorder due to the compute. There should be a failing test.

stock_checkout_sync/models/stock_move.py
```diff
@@ -17,12 +17,16 @@ def sync_checkout_destination(self, location):
         # destination move can change, so handle this case too. (there is a
         # glue module stock_dynamic_routing_checkout_sync).
         moves_to_update = self.filtered(lambda m: m.location_dest_id != location)
-        moves_to_update.write({"location_dest_id": location.id})
+        moves_to_update.picking_id.location_dest_id = location
+        moves_to_update.location_dest_id = location
         # Sync the source of the destination move too, if it's still waiting.
-        moves_to_update.move_dest_ids.filtered(
+        moves_dest = moves_to_update.move_dest_ids.filtered(
+            # FIXME add partially_available?
```
jbaudoux, line 26: yes, and `confirmed` as well

jbaudoux (approved):
@sebalix you squash the fixups?

jbaudoux (commented):
You still have to replace readme/*.rst by .md

simahawk:
@sebalix ping

PR #1987, opened 2025-06-16
[18.0][IMP] stock_picking_group_by_partner_by_carrier: allow to group when policy is one

jbaudoux:
Allow to also group when policy is "When all are ready".

Mixed policies are currently not merged. Not on purpose, just to make it easy. Could be improved in the future if someone needs that feature.

@theangryangel @mt-software-de @sebalix @rousseldenis

PR #1992, opened 2025-06-19
[18.0][FIX] stock_picking_group_by_partner_by_carrier: improperly ported method _create_backorder on stock.picking

theangryangel:
Without this fix raising backorders from a `stock.picking` can fail under certain circumstances.

I'm sorry I should've picked this up during the initial port.

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
@@ -85,12 +85,19 @@ class StockPicking(models.Model):
         else:
             return super().action_cancel()
 
-    def _create_backorder(self):
+    def _create_backorder(self, backorder_moves=None):
         backorders = self.browse()
         for picking in self:
             if not picking._is_grouping_disabled():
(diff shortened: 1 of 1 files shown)


theangryangel:
FYI @jbaudoux, @bizzappdev as you were involved in the original PR. Figured you might need to be aware if you've got this deployed already.

jbaudoux (approved):

jbaudoux:
@theangryangel I just spotted that existing deliveries are not shown anymore on SO because sale_ids is not filled in from sale_id on the procurement group when the module is installed. Did we lost a migration script? I haven't checked yet

jbaudoux:
@theangryangel can you also review https://github.com/OCA/stock-logistics-workflow/pull/1987 regarding this module

jbaudoux:
> @theangryangel I just spotted that existing deliveries are not shown anymore on SO because sale_ids is not filled in from sale_id on the procurement group when the module is installed. Did we lost a migration script? I haven't checked yet

i logged the issue here https://github.com/OCA/stock-logistics-workflow/issues/1993

PR #2016, opened 2025-07-04
[IMP] stock_picking_grn_mandatory: enforce GRN check in _action_done

sbejaoui:
some processes bypass the button_validate method and call _action_done directly to validate pickings, which skips the GRN requirement check

this PR ensures consistency by applying the same GRN check in _action_done as in button_validate


sbejaoui:
cc/ @rousseldenis

rousseldenis (approved):

rousseldenis:
@jbaudoux @lmignon

jbaudoux (approved):

PR #2017, opened 2025-07-04
[17.0] [FIX] stock_picking_invoice_link: add colspan attribute to invoice_line_ids field in stock.move form

rrebollo:
This pull request includes a minor update to the `stock_view.xml` file. The change adds a `colspan` attribute with a value of `2` to the `invoice_line_ids` field, improving the layout of the field in the user interface. Previously:

image

### DIFF
--- a/stock_picking_invoice_link/views/stock_view.xml
+++ b/stock_picking_invoice_link/views/stock_view.xml
@@ -26,22 +26,22 @@
         <field name="model">stock.move</field>
         <field name="inherit_id" ref="stock.view_move_form" />
         <field name="arch" type="xml">
-            <group name="linked_group" position="after">
+            <xpath expr="//group" position="after">
                 <field name="invoice_line_ids" invisible="1" />
                 <group
                     name="invoice_line"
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Isn't the double nested group the problem?

rrebollo:
> Isn't the double nested group the problem?

:thinking: almost all elements in that form are inside a `group` tag.

image

So, this is happening on a nested group (not double If a got the "double" right): there is the parent group an then this. For me is like the colspan and the nolabel caused this. I'm sure there is plenty of ways of make this right. Maybe I should remove the colspan from the group instead of add it to the field but ...

pedrobaeza:
Yeah, I meant 2 nested levels. I always try to simplify the architecture when possible. That's why I propose to remove one nested group.

pedrobaeza (approved):

rrebollo:
@pedrobaeza all this is also happening in 18.0. I'm doing a PR with both commits
image

rrebollo:
ping @rvalyi

PR #2022, opened 2025-07-09
[NEW][REF][16.0] stock_picking_invoice_link split to sale_stock_picking_invoice_link

renatonlima:
## Current Scenario

The stock_picking_invoice_link module, which adds the relationship between the invoice and stock picking, depends on the sale_stock module, but the purchase_stock_picking_invoice_link module depends on stock_picking_invoice_link, making it mandatory to install both the purchase and sale modules, thus losing modularity.

## Proposed Solution

The sale dependency on stock_picking_invoice_link was removed, and a new sale_stock_picking_invoice_link module was created. In this way, the stock_picking_invoice_link module is a base module for the sale_stock_picking_invoice_link and purchase_stock_picking_invoice_link modules, preserving modularity.


pedrobaeza:
Although the idea is good, I'm afraid we can't do this in a stable version, as it supposes to break some installations due to the lack of the new module (not all deployment methods auto-download and auto-install new modules). This should be done in a new version.

rvalyi:
> Although the idea is good, I'm afraid we can't do this in a stable version, as it supposes to break some installations due to the lack of the new module (not all deployment methods auto-download and auto-install new modules). This should be done in a new version.

@pedrobaeza that's legit and that's why he also made a similar PR in v18 #2023

rvalyi:
so I reverted the PR to draft to avoid merging by mistake as indeed this an impacting dependency change and we will focus on getting the merge in 18.0 instead to avoid impacting stable instances all around.

PR #2023, opened 2025-07-10
[NEW][REF][18.0] stock_picking_invoice_link split to sale_stock_picking_invoice_link

renatonlima:
Port PR #2022 

## Current Scenario

The stock_picking_invoice_link module, which adds the relationship between the invoice and stock picking, depends on the sale_stock module, but the purchase_stock_picking_invoice_link module depends on stock_picking_invoice_link, making it mandatory to install both the purchase and sale modules, thus losing modularity.

## Proposed Solution

The sale dependency on stock_picking_invoice_link was removed, and a new sale_stock_picking_invoice_link module was created. In this way, the stock_picking_invoice_link module is a base module for the sale_stock_picking_invoice_link and purchase_stock_picking_invoice_link modules, preserving modularity.


sale_stock_picking_invoice_link/models/stock_move.py
```diff
@@ -0,0 +1,38 @@
...
+                    .sudo()
+                    .search(
+                        [
+                            ("sale_line_ids", "=", stock_move.sale_line_id.id),
+                            ("move_id.move_type", "=", inv_type),
+                        ]
+                    )
+                )
+                if inv_line:
+                    stock_move.invoice_line_ids = [(4, m.id) for m in inv_line]
```
rvalyi: I would say you could use = [Command.link(aml.id) for aml in inv_line]
Also eventually use in_lines in plural.

pedrobaeza (approved):

sale_stock_picking_invoice_link/models/stock_move.py
renatonlima: Done !

rvalyi: sorry, in fact it would have been cleaner to use = [Command.set(inv_lines.ids)] directly and do that also in sale_order.py

rvalyi (approved):

PR #2030, opened 2025-07-18
[18.0][MIG] stock_picking_auto_create_package

JordiMForgeFlow:
In <=V16 Odoo used to automatically split the move lines when packing based on the quantity done on the move line (see: https://github.com/odoo/odoo/blob/fd5fe84f70dabc4b7d045f5181508658b6596c9c/addons/stock/models/stock_picking.py#L1457).

Since V17 the qty done field is not used anymore and Odoo is no longer splitting (see: https://github.com/odoo/odoo/blob/c5222a82887b6c5c781fe823865e2573c14e8b7b/addons/stock/models/stock_picking.py#L1782). Therefore, we manage the split ourselves.


AaronHForgeFlow (approved):
:+1:  Code review LGTM

JasminSForgeFlow (approved):

stock_picking_auto_create_package/models/stock_picking.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @api.model
+    def _auto_create_delivery_package_filter(self, move_lines):
+        """
+        Filter move lines that can be put in pack
+        """
+        return move_lines.filtered(
+            lambda ml: not ml.result_package_id
+            and ml.state not in ("cancel", "done")
+            and ml.quantity
```
jbaudoux, line 27: When you have multiple move lines for a move, only some can be picked. Non picked move lines will be deleted by call to action_done occuring afterwards.
When the move is not picked, all move lines will be considered as picked on action_done
```suggestion
            and ml.quantity
            and (not ml.move_id.picked or ml.picked)
```

jbaudoux (changes requested):

jbaudoux:
To notify contributors about this migration @raumschmiede-joshuaL @mt-software-de @rousseldenis

stock_picking_auto_create_package/models/stock_picking.py
```diff
@@ -24,31 +25,63 @@ def _auto_create_delivery_package_filter(self, move_lines):
         return move_lines.filtered(
             lambda ml: not ml.result_package_id
             and ml.state not in ("cancel", "done")
-            and ml.qty_done
+            and ml.quantity
         )
 
+    def _split_move_line_package_qty(self, ml, package_qty):
```
mt-software-de: Can this method be moved into `stock.move.line` model, please? Since it's always executed for a line, not for a transfer.

stock_picking_auto_create_package/models/stock_picking.py
```diff
@@ -24,31 +25,63 @@ def _auto_create_delivery_package_filter(self, move_lines):
...
+            rounding_method="HALF-UP",
+        )
+        new_move_line = ml.copy(default={"quantity": quantity_left_todo})
+        vals = {}
+        if self.picking_type_id.code == "incoming":
+            if ml.lot_id:
+                vals["lot_id"] = False
+            if ml.lot_name:
+                vals["lot_name"] = False
+        new_move_line.write(vals)
```
mt-software-de: Is there are a reason why not checking for incoming and lot before hand and then passing it as default value to copy?

stock_picking_auto_create_package/models/stock_picking.py
```diff
@@ -24,31 +25,63 @@ def _auto_create_delivery_package_filter(self, move_lines):
...
+            precision_rounding=ml.product_uom_id.rounding,
+            rounding_method="HALF-UP",
+        )
+        new_move_line = ml.copy(default={"quantity": quantity_left_todo})
+        vals = {}
+        if self.picking_type_id.code == "incoming":
+            if ml.lot_id:
+                vals["lot_id"] = False
+            if ml.lot_name:
+                vals["lot_name"] = False
```
mt-software-de: In v16 it removed the lot, but is this really necessary, and a good approach?

If there is a `stock.move.line` with a lot for a quantity, it will then be depending on the qty_to_pack, how much you will receive for the scanned lot.

For example: You picked 10 times the same lot and then want to create a automatic package for the qty 1. This will end up with 10 move lines, with a qty of 1. But only the first move.line is containing the lot. The rest will not have any lot.

mt-software-de: ping @jbaudoux

mt-software-de (changes requested):
Thx, for the migration.

stock_picking_auto_create_package/models/stock_picking.py
JordiMForgeFlow: @mt-software-de I was simply keeping same odoo standard logic used in v16, I can adapt

stock_picking_auto_create_package/models/stock_picking.py
JordiMForgeFlow: @mt-software-de mm yes, it does not make much sense to remove the lot if we are just splitting

mt-software-de (approved):
LG

rousseldenis (approved):

jbaudoux (commented):
As the package creation occurs before action done, to keep same behavior as before, we should rely on a picked quantity and not on the reserved quantity.
So depend on https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_move_line_qty_picked and replace quantity by picked_qty
WDYT @mt-software-de

mt-software-de:
> As the package creation occurs before action done, to keep same behavior as before, we should rely on a picked quantity and not on the reserved quantity. So depend on https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_move_line_qty_picked and replace quantity by picked_qty WDYT @mt-software-de

If i am getting it correct `stock_move_line_qty_picked` is there, for not losing the reservation until the transfers is set to done and keeping the picked qty in a separate field. 
Why is it done like that and not just created a new move line with the remaining qty, when something is picked? 
Also i have seen that this addon is already be used as a dependency for shopfloor. Was it considered to create new move lines? But this doesn't really matter, here. 

When you suggesting to rely on `qty_picked` we need `stock_move_line_qty_picked` as dependency here, but also on all other addons which are using `quantity` of `stock.move.line`. 

IMO there a two possible options here
- add the new dependency and rely on qty_picked
- rely on `quantity` and ensure that `qty_picked` is in sync with quantity before action_done is called, for example by using `_register_hook` in `stock_move_line_qty_picked` instead of directly overwriting it https://github.com/OCA/stock-logistics-workflow/blob/5e864299d5ab5a44b8275ac0b45ba0a051d9e49b/stock_move_line_qty_picked/models/stock_move.py#L28 this means that we do not need a dependency here or on any other addon which is using `quantity` of `stock.move.line`.

mt-software-de:
> As the package creation occurs before action done, to keep same behavior as before, we should rely on a picked quantity and not on the reserved quantity. So depend on https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_move_line_qty_picked and replace quantity by picked_qty WDYT @mt-software-de

After checking everything for a response of your initial question, i maybe found a general problem with `stock_picking_auto_create_package`?

For example: If you end up with:
- a move with a demand of 2
- two move lines with a quantity of 1
- picking type configured `automatic_package_creation_mode == "packaging"
- smallest packaging on product with a quantity of 2

Then the addon will never create package for a qty of 2 instead i will create two packages with a quantity of 1. 
Because https://github.com/ForgeFlow/stock-logistics-workflow/blob/73fc56d39ff52a78e2241c4bf7b32efe873f96fc/stock_picking_auto_create_package/models/stock_picking.py#L41 relies on the quantity of a move.line and not on the quantity of the move. 

What do you think @jbaudoux

mt-software-de:
> > As the package creation occurs before action done, to keep same behavior as before, we should rely on a picked quantity and not on the reserved quantity. So depend on https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_move_line_qty_picked and replace quantity by picked_qty WDYT @mt-software-de
> 
> If i am getting it correct `stock_move_line_qty_picked` is there, for not losing the reservation until the transfers is set to done and keeping the picked qty in a separate field. Why is it done like that and not just created a new move line with the remaining qty, when something is picked? Also i have seen that this addon is already be used as a dependency for shopfloor. Was it considered to create new move lines? But this doesn't really matter, here.
> 
> When you suggesting to rely on `qty_picked` we need `stock_move_line_qty_picked` as dependency here, but also on all other addons which are using `quantity` of `stock.move.line`.
> 
> IMO there a two possible options here
> 
>     * add the new dependency and rely on qty_picked
> 
>     * rely on `quantity` and ensure that `qty_picked` is in sync with quantity before action_done is called, for example by using `_register_hook` in `stock_move_line_qty_picked` instead of directly overwriting it https://github.com/OCA/stock-logistics-workflow/blob/5e864299d5ab5a44b8275ac0b45ba0a051d9e49b/stock_move_line_qty_picked/models/stock_move.py#L28
>        this means that we do not need a dependency here or on any other addon which is using `quantity` of `stock.move.line`.

https://github.com/OCA/stock-logistics-workflow/pull/2091

jbaudoux:
> > As the package creation occurs before action done, to keep same behavior as before, we should rely on a picked quantity and not on the reserved quantity. So depend on https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_move_line_qty_picked and replace quantity by picked_qty WDYT @mt-software-de
> 
> If i am getting it correct `stock_move_line_qty_picked` is there, for not losing the reservation until the transfers is set to done and keeping the picked qty in a separate field. 

Yes exactly 

> Why is it done like that and not just created a new move line with the remaining qty, when something is picked? Also i have seen that this addon is already be used as a dependency for shopfloor. Was it considered to create new move lines? But this doesn't really matter, here.

Because it's convenient

> When you suggesting to rely on `qty_picked` we need `stock_move_line_qty_picked` as dependency here, but also on all other addons which are using `quantity` of `stock.move.line`.

Beside shopfloor, this addon is the only one I found so far having to care for this. Other addons hook after action done and not before or don't care about the quantity.

> IMO there a two possible options here
> 
> * add the new dependency and rely on qty_picked
> * rely on `quantity` and ensure that `qty_picked` is in sync with quantity before action_done is called, for example by using `_register_hook` in `stock_move_line_qty_picked` instead of directly overwriting it https://github.com/OCA/stock-logistics-workflow/blob/5e864299d5ab5a44b8275ac0b45ba0a051d9e49b/stock_move_line_qty_picked/models/stock_move.py#L28
>    this means that we do not need a dependency here or on any other addon which is using `quantity` of `stock.move.line`.

JordiMForgeFlow:
@jbaudoux Odoo standard already gives us the "picked" boolean in order to check the lines that have been picked, not really sure if we should need to force the dependency or if it can simply be a glue module in case you are using both.

mt-software-de:
> @jbaudoux Odoo standard already gives us the "picked" boolean in order to check the lines that have been picked, not really sure if we should need to force the dependency or if it can simply be a glue module in case you are using both.

With this PR https://github.com/OCA/stock-logistics-workflow/pull/2091 a glue module wouldn't be needed.

JordiMForgeFlow:
@rousseldenis could we reopen this one?

jbaudoux (approved):

mt-software-de (approved):

PR #2032, opened 2025-07-18
[18.0][IMP] delivery_total_weight_from_packaging: Add hooks

mmequignon:
Make use of stock.move.quantity which already aggregates stock.move.line.quantity, and add hooks allowing to use a different quantity I.E. stock_available_to_promise_release.


mmequignon:
<s>Tests are :red_circle: 
I got the same result locally with only stock_move_actual_date installed.
https://github.com/OCA/stock-logistics-workflow/actions/runs/16381601281/job/46294090796?pr=2032#step:8:968
@AungKoKoLin1997 any idea what's wrong here?</s>
tests are :green_circle: now

simahawk (approved):

PR #2034, opened 2025-07-22
[16.0][FIX] stock_owner_restriction : fix wrong data type in domain

AdrianaSaiz:


stock_owner_restriction/models/stock_quant.py
```diff
@@ -44,10 +44,14 @@ def _gather(
...
+        restricted_owner_id = self.env.context.get(
+            "force_restricted_owner_id", None
+        )
         if restricted_owner_id is not None:
-            domain = expression.AND([domain, [("owner_id", "=", restricted_owner_id)]])
-        return super(StockQuant, self).read_group(
+            domain = expression.AND(
+                [domain, [("owner_id", "=", restricted_owner_id)]]
+            )
+        return super()._read_group(
```
rousseldenis, line 54: @AdrianaSaiz What do you try to fix here?

rousseldenis (changes requested):

PR #2042, opened 2025-07-28
[15.0][MIG] stock_picking_group_by_partner_by_carrier: migration to 15.0

PolComas:
Migrate the stock_picking_group_by_partner_by_carrier addon to be fully compatible with Odoo 15, minimizing external dependencies and avoiding the need to port stock_picking_group_by_base.

Key Changes:
- Removed the dependency on stock_picking_group_by_base. Its picking‑indexing and grouping logic is now implemented directly on the stock.picking model.

Field renames to match Odoo 15 conventions:
- move_ids renamed to move_lines
- reserved_uom_qty renamed to product_qty
- Removed Odoo 16’s invalidate_recordset() calls in tests.
- All tests  have been updated for Odoo 15 and now pass successfully.


ivilata (approved):
Looks good to me. Thanks for your contribution!

etobella (changes requested):
Follow migration guidelines:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

You have lost all history of commits. Also, the migration should have just one commit.

PR #2049, opened 2025-07-31
[17.0][ADD] stock_landed_costs_revert

edescalona:
@BinhexTeam

Create new landed costs from another.

Currently, if we need to undo some landed costs, the system doesn't allow it. The only way is to create another shipping cost and set the value to negative. When there are too many products, this becomes complicated; this module attempts to provide the user with a shortcut to generate a reversal of the original landed cost.


stock_landed_costs_revert/models/stock_landed_cost.py
```diff
@@ -0,0 +1,63 @@
...
+                                    "price_unit": -line.price_unit,
+                                    "split_method": line.split_method,
+                                }
+                            )
+                            for line in landed_cost.cost_lines
+                        ],
+                    ],
+                }
+            )
+            landed_cost.revert_landed_cost_ids = [Command.link(landed_cost_id.id)]
```
christian-ramos-tecnativa: I think this name may be more appropiate `revert_landed_cost` instead of `landed_cost_id` and open the new landed cost directly after creation:
```suggestion
            revert_landed_cost = landed_cost.copy(
                {
                    "state": "draft",
                    "picking_ids": landed_cost.picking_ids.ids,
                    "vendor_bill_id": landed_cost.vendor_bill_id.id,
                    "cost_lines": [
                        Command.clear(),
                        *[
                            Command.create(
                                {
                                    "product_id": line.product_id.id,
                                    "name": line.name,
                                    "account_id": line.account_id.id,
                                    "price_unit": -line.price_unit,
                                    "split_method": line.split_method,
                                }
                            )
                            for line in landed_cost.cost_lines
                        ],
                    ],
                }
            )
            landed_cost.revert_landed_cost_ids = [Command.link(revert_landed_cost.id)]
            return {
                "name": "Revert Landed Costs",
                "view_type": "form",
                "view_mode": "tree,form,kanban",
                "res_model": "stock.landed.cost",
                "type": "ir.actions.act_window",
                "domain": [("id", "=", revert_landed_cost.id)],
            }
```

christian-ramos-tecnativa (changes requested):
Just small comments

edescalona:
Hello @Christian-RB  ready for review, thank you for your comments.

edescalona:
Hi @christian-ramos-tecnativa @rrebollo @jbaudoux @lmignon, please review, thanks

stock_landed_costs_revert/models/stock_landed_cost.py
```diff
@@ -0,0 +1,69 @@
...
+                            )
+                            for line in landed_cost.cost_lines
+                        ],
+                    ],
+                }
+            )
+            landed_cost.revert_landed_cost_ids = [Command.link(revert_landed_cost.id)]
+            return self.with_context(
+                revert_landed_cost_id=revert_landed_cost.id
+            ).action_view_revert_landed_cost()
```
rrebollo, line 56: You're looping over `self` but returning at every step of the loop, so only the first record in the sequence actually gets processed. Are you sure you meant to do this only for a single record? If that's the case why the loop then?

stock_landed_costs_revert/models/stock_landed_cost.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def action_view_revert_landed_cost(self):
+        domain = [("id", "in", self.revert_landed_cost_ids.ids)]
```
rrebollo, line 59:
```suggestion
        self.ensure_one()
        domain = [("id", "in", self.revert_landed_cost_ids.ids)]
```

rrebollo (commented):
Please consider my suggestions/doubts?

edescalona:
Hi @rrebollo , thanks for the review. I've addressed your comments, please take another look when you have a chance.
@christian-ramos-tecnativa 
@jbaudoux 
@rousseldenis

edescalona:
The test failures are not related to this PR.

rrebollo (approved):
Code Review. LGTM!

edescalona:
Hi @rousseldenis @jbaudoux, do you think it's appropriate to proceed with the merge? Any comments?

PR #2055, opened 2025-08-01
[IMP] stock_move_line_qty_picked: sync `picked` and `qty_picked` fields

sebalix:
Updating the `picked` or `qty_picked` fields directly will now sync the expected values in both fields:

- `picked`: updates `qty_picked` to the full quantity if `True`, reset `qty_picked` to 0 if `False`
- `qty_picked`: set `picked` to  `True` if > 0, set `picked` to `False` otherwise

Also, do not copy these fields by default.

### DIFF
--- a/stock_move_line_qty_picked/models/stock_move_line.py
+++ b/stock_move_line_qty_picked/models/stock_move_line.py
@@ -7,13 +7,36 @@ from odoo.tools import float_compare
 class StockMoveLine(models.Model):
     _inherit = "stock.move.line"
 
-    qty_picked = fields.Float()
+    picked = fields.Boolean(
+        # Override std field
+        inverse="_inverse_picked",
+        copy=False,
+    )
+    qty_picked = fields.Float(inverse="_inverse_qty_picked", copy=False)
+
+    def _inverse_picked(self):
+        if self.env.context.get("move_line_pick_qty"):
+            return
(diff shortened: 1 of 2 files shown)


mmequignon (approved):
:+1:

stock_move_line_qty_picked/models/stock_move_line.py
```diff
@@ -24,7 +46,7 @@ def _pick_qty(self, qty):
             <= 0
         ):
             values["quantity"] = qty
-        self.write(values)
+        self.with_context(move_line_pick_qty=True).write(values)
```
simahawk: we should use `update` here

simahawk:
@sebalix are the values copied? If not I would set copy=False (to be fixed in odoo in case)

mmequignon:
> @sebalix are the values copied? If not I would set copy=False (to be fixed in odoo in case)

Good catch, noticed this as well, but forgot to comment.
Yes they are, that's why we had to do this https://github.com/OCA/stock-logistics-shopfloor/pull/2/files#diff-4ff598dc32d54b1cb987bacadccd749baf8c0471fefa8aca2aa1a63d8572b784R51-R53

sebalix:
Updated

simahawk (approved):

PR #2057, opened 2025-08-04
[16.0][ADD] stock_picking_bill_matching - Work in Progress

rvalyi:
Inspired by Odoo v18 "bill matching" feature, but working at the incoming picking level (and in v16 here; will be ported to v18):

1. you start from a vendor bill (usually EDI imported)
2. you click on the top match button to match it against incoming pickings. If you validated some purchase order for this supplier and these products, you will see the incoming picking listed in the match selection

    - eventually select the incoming pickings lines to match with your vendor bill lines and click on the "match" button
    - if there is no existing picking line line to match,  you can simply click on "add to Picking" and the selected bill lines will be added to an existing picking or to a new picking if you select no existing picking.

The move of the picking will be set to done and eventually the whole picking will be set to done.

In case of mistake, you can "unmatch" any incoming move from a bill line in the form view (so you should work with the line popup form) where you can see the related incoming stock moves and unmatch them eventually. If you do so the vendor bill will be set to be matched again.

Work in progress

### DIFF
--- a/stock_picking_bill_matching/__init__.py
+++ b/stock_picking_bill_matching/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import wizard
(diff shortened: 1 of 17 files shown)


rvalyi:
continued here #2310

PR #2058, opened 2025-08-05
[18.0][FIX] stock_move_line_qty_picked: always align moved qty with (partial) picked qty

sebalix:
- [x] add test

When a backorder is created, `product_uom_qty` moved qty is well updated with the `quantity` picked qty, but this is not the case when backorder are cancelled, leading to inconsistencies when picking a partial qty.

### DIFF
--- a/stock_move_line_qty_picked/tests/test_stock_move_line_qty_picked.py
+++ b/stock_move_line_qty_picked/tests/test_stock_move_line_qty_picked.py
@@ -95,6 +95,8 @@ class TestStockMoveLineQtyPicked(TransactionCase):
         # Ensure 3 were moved on the quant and 2 are reserved on backorder
         self.assertEqual(self.quant.quantity, 97)
         self.assertEqual(self.quant.reserved_quantity, 2)
+        self.assertEqual(move_line.quantity, 3)
+        self.assertEqual(move_line.move_id.product_uom_qty, 3)
 
     def test_move_concurrent_pick_qty(self):
         move = self._create_move(
@@ -119,6 +121,7 @@ class TestStockMoveLineQtyPicked(TransactionCase):
         self.assertEqual(sum(ml.quantity for ml in move.move_line_ids), 10)
         move.picking_id.with_context(skip_backorder=True).button_validate()
         self.assertEqual(move.quantity, 8)
+        self.assertEqual(move.product_uom_qty, 8)
 
     def test_move_unpick_qty(self):
         move = self._create_move(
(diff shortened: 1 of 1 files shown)


simahawk (approved):
LG

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -55,3 +55,18 @@ def _sum_ml_qty_picked(self):
...
+        # Override to ensure 'product_uom_qty' aligns with the picked qty
+        # (especially with partial picked qty) when backorder is cancelled.
+        # Context: Odoo 18 expects to pick all reserved qty when we validate
+        # a move/transfer, which is not what we want anymore with the new
+        # 'qty_picked' field on 'stock.move.line'.
+        # Field 'product_uom_qty' is updated at only one place: when the move is
+        # split, like during the creation of a backorder move.
+        if cancel_backorder:
+            for move in self:
+                if move.state not in ("assigned", "partially_available"):
```
jbaudoux: because call to super may assign
```suggestion
                if move.state in ("done", "cancel"):
```

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -55,3 +55,18 @@ def _sum_ml_qty_picked(self):
...
+                    continue
+                move.product_uom_qty = move.quantity
```
jbaudoux:
```suggestion
                if move.picked:
                    move.product_uom_qty = move.quantity
```

jbaudoux (changes requested):

stock_move_line_qty_picked/models/stock_move.py
jbaudoux: I think it would be safer to split the move line than changing the move. The move line will then be deleted when calling super()

grindtildeath (changes requested):
I find it dangerous to update the demand with the effective quantity. Even if only a partial quantity is moved, I'd still want to know what was the original demand on the move :confused:

grindtildeath:
Can you please check if https://github.com/OCA/stock-logistics-workflow/pull/2076 doesn't solve your issue?

simahawk:
FTR I've simply rebased after https://github.com/OCA/stock-logistics-workflow/pull/2076 got merged and discarded the changes on `_action_done`. Let's see if tests are still happy.

sebalix:
We can close this one then since https://github.com/OCA/stock-logistics-shopfloor/pull/20 is now working well without it?

PR #2059, opened 2025-08-05
[14.0][FIX] delivery_procurement_group_carrier: Use new proc group for returns

mt-software-de:
When a return is created for an outgoing transfer, by default the same procurement group is used.
This isn't good because here https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/delivery/models/stock_picking.py#L317 the carrier gets set to False. 
Which sets the carrier on all other transfers as well to False. 

By creating a procurement.group per return this isn't the case. 
cc @jbaudoux @mmequignon @i-vyshnevska @raumschmiede-joshuaL

### DIFF
--- a/delivery_procurement_group_carrier/wizards/stock_return_picking.py
+++ b/delivery_procurement_group_carrier/wizards/stock_return_picking.py
@@ -10,6 +10,10 @@ class StockReturnPicking(models.TransientModel):
         vals = super()._prepare_move_default_values(return_line, new_picking)
         group = new_picking.move_lines.group_id
         if not group:
-            group = self.env["procurement.group"].create({"name": new_picking.name})
+            group_vals = {"name": new_picking.name, "carrier_id": False}
+            if return_line.move_id.group_id:
+                group = return_line.move_id.group_id.copy(group_vals)
+            else:
+                group = self.env["procurement.group"].create(group_vals)
         vals["group_id"] = group.id
         return vals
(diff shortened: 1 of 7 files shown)


mt-software-de:
FYI @jbaudoux
this happens when `_action_cancel` on a `stock.move` is called, which calls at the https://github.com/OCA/wms/blob/c63f6d29e42dac0573e71ba93e5d77285e8eb054/stock_available_to_promise_release/models/stock_move.py#L640 and creates the return moves via odoo standard. 
IMO it would be better to not use `stock.return.picking` instead use a procurement run. This would also allow to merge moves into one cleanup transfer.

raumschmiede-joshuaL (approved):

jbaudoux (approved):

delivery_procurement_group_carrier/wizards/stock_return_picking.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class StockReturnPicking(models.TransientModel):
+    _inherit = "stock.return.picking"
+
+    def _prepare_move_default_values(self, return_line, new_picking):
+        vals = super()._prepare_move_default_values(return_line, new_picking)
+        group = new_picking.move_lines.group_id
+        if not group:
+            group = self.env["procurement.group"].create({"name": new_picking.name})
```
mt-software-de: Should we maybe change it a little bit?
And use copy? The group would the also include the sale_id. 
```suggestion
            group = return_line.move_id.group_id.copy({"name": new_picking.name})
```

mt-software-de: cc @jbaudoux

PR #2061, opened 2025-08-06
[18.0][MIG] stock_picking_putaway_recompute

grindtildeath:
Migrated from:
 - [ ] https://github.com/OCA/stock-logistics-workflow/pull/1740

TODO:
 - [x] Migrate https://github.com/acsone/stock-logistics-workflow/pull/2 once merged in 16.0 PR


ivantodorovich (approved):
LG

CI issue seems unrelated

stock_picking_putaway_recompute/models/stock_move_line.py
```diff
@@ -0,0 +1,57 @@
...
+        Recompute putaways on operations that:
+
+            - have their picking type configured for that
+            - have their picking not printed (started)
+            - have their picked field set
+        """
+        return self.filtered(
+            lambda line: line.picking_type_id.allow_to_recompute_putaways
+            and not line.picking_id.printed
+            and not line.result_package_id
```
jbaudoux: I would move this limitation in the _compute_can_recompute_putaways. When the action is launched at picking level, it is ok.
Or better, you can still allow it if the recordset of move lines contains all lines with that result package. Meaning that you compute a put away for all lines having that result package.

stock_picking_putaway_recompute/readme/USAGE.md
```diff
@@ -0,0 +1,8 @@
+- In a picking where the recompute operation is enabled (see Configuration),
+  the recompute is available on picking level and on the detailed operations if
+  enabled.
+- If a putaway rule has been changed after product reservation, click on
+  the button 'Recompute putaways'. This will recompute the destination locations
+  on all detailed operations that have no done quantity yet and no result package 
+  (as all operations for the same package should go to the same destination).
```
jbaudoux, line 7: And update this

jbaudoux (commented):
We should be able to recompute the destination of the package. Cc @rousseldenis

stock_picking_putaway_recompute/models/stock_move_line.py
grindtildeath: So do you mean we have to restrict recomputation of putaway if it's called from a move line, and there is another move line (in the same picking?) sharing the same package?

My feeling is it must either be allowed at stock.move.line level if there is no package, or it must be allowed at for a package level. But that's quite tricky to implement, especially since AFAIK package level can include lines from different moves. :grimacing:

jbaudoux: But package level is for the source package, not the destination package

jbaudoux: We need to allow the action at move line for shopfloor. If there is a result package, you could extend the recordset with other lines of the picking having that same result package and that are not picked.

grindtildeath:
@jbaudoux Let's try to get this one merged without package support first, so we can focus on the improvement through https://github.com/OCA/stock-logistics-workflow/pull/2077

jbaudoux:
> @jbaudoux Let's try to get this one merged without package support first, so we can focus on the improvement through #2077

You need to squash your fixups then

jbaudoux:
@grindtildeath you need to move it to stock-logistics-putaway repo

Issue #1882, opened 2025-02-21
[18.0] stock_split_picking - split by quantity - back order vs split order


jbaudoux:
When a transfer is being partially processed by a user and you want to validate only that, you should create a new split order containing the done lines and keep remaining lines in the original transfer.

In (my) business terms, a split order represent an extraction of an original order that is being partially processed. A back order is the remaining of an order that was not entirely available. The back order follows the operation type backorder strategy while a split order do not.

The advantages of this approach are:
* You can have multiple users working on a same transfer. Each user will validate their move lines in a new split order. The original move id and picking id is not changed when a split order is created by a user preventing any impact on the other users (no concurrent update, no lost of move/picking id).
* If the remaining transfer is processed later, it remains the next one in the list of transfers as the id as not changed. This is because transfers are sorted by id. If you split the first transfer from the list, the remaining transfer should stay at the top of the list.

This way of extracting done quantities in a new split order is currently in place in shopfloor:
* https://github.com/OCA/wms/blob/16.0/shopfloor/models/stock_move_line.py#L59
* https://github.com/OCA/wms/blob/16.0/shopfloor/models/stock_move.py#L59

Could we modify `stock_split_picking` for the mode `quantity` so that the method `split_process` works that way?
And modify the posted message to use `split order` instead of `backorder` at https://github.com/OCA/stock-logistics-workflow/blob/18.0/stock_split_picking/models/stock_picking.py#L95

Do you have a reason to keep current behavior? 

Purpose will be also to drop that similar code from `shopfloor` and rely on `stock_split_picking` module.

That remark also applies to other proposed split methods used when creating a batch transfer :
* https://github.com/OCA/stock-logistics-workflow/pull/1741
* https://github.com/OCA/stock-logistics-workflow/pull/1829

@StefanRijnhart @gbrito @bosd @hbrunn @pedrobaeza @sebalix @lmignon @wouitmil

jbaudoux:
Note that the current implementation of `stock_split_picking` by `quantity` seems to be now provided by odoo in standard:
https://github.com/odoo/odoo/blob/18.0/addons/stock/models/stock_picking.py#L1458

pedrobaeza:
[...]

PR #2062, opened 2025-08-07
[18.0] [REF] `stock_split_picking`: Split off moves in a new picking

ivantodorovich:
When a transfer is being partially processed by a user and you want to validate only that, you should create a new split order containing the done lines and keep remaining lines in the original transfer.

In business terms, a split order represent an extraction of an original order that is being partially processed. A back order is the remaining of an order that was not entirely available.

Implements #1882

### DIFF
--- a/stock_split_picking/wizards/stock_split_picking.py
+++ b/stock_split_picking/wizards/stock_split_picking.py
@@ -83,6 +83,9 @@ class StockSplitPicking(models.TransientModel):
             moves_to_recompute_state = self.env["stock.move"]
             for move in picking.move_ids:
                 rounding = move.product_uom.rounding
+                # Do not split moves that are done or cancelled
+                if move.state in ("done", "cancel"):
+                    continue
                 # If there aren't assigned quantities, there's nothing to split
                 if float_is_zero(move.quantity, precision_rounding=rounding):
                     continue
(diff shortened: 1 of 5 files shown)


stock_split_picking/wizards/stock_split_picking.py
```diff
@@ -28,28 +30,125 @@ def _default_picking_ids(self):
...
         """
         new_pickings = self.env["stock.picking"]
-        for picking in self.mapped("picking_ids"):
-            for move in picking.move_ids[1:]:
-                new_pickings += picking._split_off_moves(move)
-        return self._picking_action(new_pickings)
+        for picking in self.picking_ids:
+            if len(picking.move_ids) <= 1:
+                continue
+            new_pickings += picking._split_off_moves(picking.move_ids[0])
```
jbaudoux: not related to your changes but it looks like it's missing to manage cancel/done moves.
```suggestion
        for picking in self.picking_ids:
            first_move = picking.move_ids.browse()
            for move in picking.move_ids:
                if move.state in ("cancel", "done"):
                    continue
                if not first_move:
                   first_move = move
                   continue
                new_pickings += picking._split_off_moves(picking.move_ids[0])
```

stock_split_picking/wizards/stock_split_picking.py
```diff
@@ -28,28 +30,125 @@ def _default_picking_ids(self):
...
+        hand, extracts the selected quantities in a new picking, and keeps the original
+        one as backorder.
+        """
+        self._check_can_split_by_quantity()
+        new_pickings = self.env["stock.picking"]
+        for picking in self.picking_ids:
+            new_moves_vals = []
+            moves_to_split_off = self.env["stock.move"]
+            moves_to_recompute_state = self.env["stock.move"]
+            for move in picking.move_ids:
```
jbaudoux, line 84: same here
```suggestion
            for move in picking.move_ids:
                if move.state in ("cancel", "done"):
                    continue
```

jbaudoux (approved):
Approving the refactor

ivantodorovich:
Thanks @jbaudoux . I pushed a new commit fixing this

PR #2068, opened 2025-08-13
[18.0][FWP][FIX] delivery_procurement_group_carrier: Use new proc group for returns

mt-software-de:
FWP of https://github.com/OCA/stock-logistics-workflow/pull/2059
cc @jbaudoux


jbaudoux (approved):
Thanks for the port

PR #2074, opened 2025-08-14
[18.0] [MIG] `stock_split_picking_dimension`: Migration to 18.0

ivantodorovich:
Includes a refactor (in a separate commit), to adapt for the changes made in:
- https://github.com/OCA/stock-logistics-workflow/pull/2062
(depends on)

And align to https://github.com/OCA/stock-logistics-workflow/issues/1882


stock_split_picking_dimension/wizards/stock_split_picking.py
```diff
@@ -0,0 +1,221 @@
...
+            taken_nbr_lines = 0
+            taken_volume = 0.0
+            taken_weight = 0.0
+            moves_to_split_off = self.env["stock.move"]
+            for move in picking.move_ids:
+                # Stop if we've reached the maximum number of lines, volume, or weight
+                if (
+                    (self.max_nbr_lines and taken_nbr_lines >= self.max_nbr_lines)
+                    or (self.max_volume and taken_volume >= self.max_volume)
+                    or (self.max_weight and taken_weight >= self.max_weight)
```
jbaudoux, line 184: you will never reach this as you increase taken only if it remains under max
Maybe it would be more readable if you move the max nbr lines condition below

stock_split_picking_dimension/wizards/stock_split_picking.py
```diff
@@ -0,0 +1,221 @@
...
+                    not self.max_weight
+                    or float_compare(
+                        move_weight,
+                        self.max_weight - taken_weight,
+                        precision_rounding=weight_uom.rounding,
+                    )
+                    <= 0
+                ):
+                    moves_to_split_off += move
+                    taken_nbr_lines += 1
```
jbaudoux, line 211:
```suggestion
                    taken_nbr_lines += 1
                    if taken_nbr_lines >= self.max_nbr_lines:
                        break
```

jbaudoux (changes requested):

stock_split_picking_dimension/wizards/stock_split_picking.py
ivantodorovich: Not needed IMO, as the same thing occurs immediately with the next iteration

stock_split_picking_dimension/wizards/stock_split_picking.py
ivantodorovich: What do you mean? 🤔 I fail to spot the issue here 😓  can you show with an example?

jbaudoux: You add only if it stays below max. So you will never reach this max

ivantodorovich: In reality here we'd only need an `eq` operator, not `gte`.. but I see no harm 🤔

jbaudoux: Ok I got it now

jbaudoux (approved):

PR #2076, opened 2025-08-18
[18.0][FIX] stock_move_line_qty_picked: Align move quantity with reserved

grindtildeath:
In the previous version of this module the computation of stock.move.line.quantity was modified to reflect the picked quantity.

However, stock.move.line.quantity must reflect the reserved quantity on the move lines since there is an inverse function that is going to increase or decrease the quantity on stock.move.line.quantity accordingly.

The stock.move.quantity must only be updated to reflect the sum of stock.move.line.qty_picked when stock.move._action_done is called, to avoid the creation of a backorder.

### DIFF
--- a/stock_move_line_qty_picked/models/stock_move.py
+++ b/stock_move_line_qty_picked/models/stock_move.py
@@ -1,8 +1,8 @@
 # Copyright 2025 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from collections import defaultdict
 
-from odoo import api, models
+from odoo import models
+from odoo.tools import float_compare
 
 
 class StockMove(models.Model):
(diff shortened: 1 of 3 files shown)


stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -55,3 +24,11 @@ def _sum_ml_qty_picked(self):
                 move_line.qty_picked, self.product_uom, round=False
             )
         return quantity
+
+    def _action_done(self, cancel_backorder=False):
+        for move in self:
+            if move._ml_has_qty_picked():
+                move.with_context(
+                    move_line_pick_qty=True
+                ).quantity = move._sum_ml_qty_picked()
```
jbaudoux: You must loop over move lines that are picked and if qty_picked != quantity, then update quantity=qty_picked. Otherwise you don't know which move line will get its quantity changed and it will likely ben wrong

jbaudoux (changes requested):
Also in _sum_ml_qty_picked you should skip move lines that are not picked

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -55,3 +24,11 @@ def _sum_ml_qty_picked(self):
                 move_line.qty_picked, self.product_uom, round=False
             )
         return quantity
+
+    def _action_done(self, cancel_backorder=False):
+        for move in self:
+            if move._ml_has_qty_picked():
+                move.with_context(
+                    move_line_pick_qty=True
```
simahawk: what is the use of the ctx? Also, if really needed, should be moved to the start of the loop `for move in self.with_contxt(...)`

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -13,36 +12,6 @@ def _get_picked_quantity(self):
...
-        not_picked_moves = self - picked_moves_ids
-        return super(StockMove, not_picked_moves)._compute_quantity()
-
-    def _quantity_sml(self):
-        if self._ml_has_qty_picked():
-            return self._sum_ml_qty_picked()
-        return super()._quantity_sml()
-
     def _ml_has_qty_picked(self):
         self.ensure_one()
```
simahawk: this call to `ensure_one` is useless. This is a private method used in your own loop. It can be removed and save N checks.

stock_move_line_qty_picked/models/stock_move.py
grindtildeath: Context is needed because a second move line might be created in `_set_quantity_done`, and we don't want this move line to have a `qty_picked` set according to its `quantity` when it is created.

EDIT: Not needed anymore after attending @jbaudoux 's comment

stock_move_line_qty_picked/models/stock_move.py
grindtildeath: Ok should be good now (cf test with lots)

jbaudoux (approved):
Now it looks good. Squash the fixups and let's merge this

simahawk (approved):

PR #2079, opened 2025-08-20
[18.0][MIG] stock_grn

rousseldenis:
Improvements:

- Pre-initialize grn_id on stock_move level
- Add a notebook page for shipments
- Add sub-menu to improve readability
- Don't display grn columns if not needed (not incoming) + add optional 

@jbaudoux


jbaudoux (commented):
What about replacing GRN by goods received note. The acronym is not very explicit.
You don't like the new shipment advice module?
I would place this module in transport repo.

rousseldenis:
> You don't like the new shipment advice module?

Why?

rousseldenis:
> What about replacing GRN by goods received note. The acronym is not very explicit.

Yes. But in order to optimize display, it could be kept if we use correct fields helps.

jbaudoux:
> > You don't like the new shipment advice module?
> 
> Why?

It was meant as a replacement

jbaudoux:
> > What about replacing GRN by goods received note. The acronym is not very explicit.
> 
> Yes. But in order to optimize display, it could be kept if we use correct fields helps.

Sure. I meant for the module name

PR #2091, opened 2025-08-29
[18][IMP] stock_move_line_qty_picked: Use register_hook

mt-software-de:
Right now if this addon is installed, it aligns the qty when it is called in the stack. 
Since there are addons which are not depending on this but still overwriting `_action_done`,
we can not ensure when this `_action_done` is called in the stack. It depends on when a addon is loaded, with which dependencies. 

This change ensures, that the alignment happens before `_action_done` at all is called in the stack,
because it hooks in before with `_register_hook` and patching the method. 

The first concerns were raised here. https://github.com/OCA/stock-logistics-workflow/pull/2030#pullrequestreview-3166187551


jbaudoux (approved):
Can you rebase and add a PR description and mark as ready if it does the job

PR #2095, opened 2025-09-04
[18.0] [FIX] stock_picking_batch_creation: support devices without max volume/weight

ivantodorovich:


stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -353,23 +363,11 @@ def _get_remaining_volume(self, partner=False):
...
-                    precision_digits=self._precision_volume(),
-                )
-                > 0
-            ):
-                return device
-        return None
+        for device in self.stock_device_type_ids.sorted("sequence"):
+            if not picking.filtered_domain(self._get_picking_domain_for_device(device)):
+                continue
+            return device
```
florentx: Suggests:
```suggestion
            if picking.filtered_domain(self._get_picking_domain_for_device(device)):
                return device
```

florentx (approved):

henrybackman (approved):

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -353,23 +363,10 @@ def _get_remaining_volume(self, partner=False):
...
-            if (
-                self._volume_condition_for_device_choice(
-                    device.min_volume,
-                    picking.volume,
-                    device.max_volume,
-                )
-                and tools.float_compare(
-                    device.max_weight,
-                    picking.weight,
-                    precision_digits=self._precision_volume(),
```
TDu, line 367: No need to use the precision anymore ?

lmignon (approved):

stock_picking_batch_creation/wizards/make_picking_batch.py
ivantodorovich: IMO it was never needed.

This is weight, we don't really care about precision here. If it's heavier, then it's heavier, it doesn't matter for how much.

Even, the precision that was used before.. made no sense `2 x uom precision`

```
    def _precision_volume(self):
        return max(
            6,
            self.env["decimal.precision"].precision_get("Product Unit of Measure") * 2,
        )
```

TDu (approved):

PR #2104, opened 2025-09-16
[15.0][FIX] stock_valuation_fifo_lot: fix _get_price_unit() function and mrp dependence

Tisho99:
This PR fixes some errors of the module:
- In the price unit calculation, use the default value if the move is a return. [Odoo has an special case for that](https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/stock_account/models/stock_move.py#L47-L55) to calculate it correctly (with this module installed, the result in that case is wrong)
- In the price unit calculation, take into account the related landed costs of the candidate SVL
- Add a new module, stock_valuation_fifo_lot_mrp_landed_cost, to cover the needed mrp_landed_costs dependency. The mrp_production_ids field, used in this module, is created in the mrp_landed_costs module

Unit tests have also been added to cover the case of returning a outgoing shipping

The README has also been improved

T-8760


luis-ron (approved):
I have carried out various tests, which I list below:

Test characteristics:
Product categories have been configured using the FIFO valuation method.
Various products have been created and configured with the following characteristics:


- Storable products.


- With a product category using the FIFO valuation method.


- Products with batch traceability.

- The landed costs will be divided equally among all purchase units.

Test flow:


1. Various purchases of these products have been made, covering various aspects:
- Purchase of PRODUCT A with 1 batch. Price per unit: €2
- Purchase of PRODUCT B with more than 1 batch (2 or more).
- Various purchases of PRODUCT C, with various batches.

In each of these purchases, landed costs have been added, thus creating a more detailed valuation and ensuring the correct behavior of the system. To simplify matters, a landed cost of €10 has been applied to each of the purchases.

The result of the purchase valuation is as follows:
- Purchase 1: PRODUCT A | 10 units | €2 per unit | Landed cost €10 | Lot 1 -> Total cost: €3

- Purchase 2: PRODUCT B | 10 units | €2 per unit | Landed cost €10 | Lot 1 | Lot 2 -> Total cost: €3

- Purchase 3: Product C | 10 units | €2 per unit |  Landed cost  €10 | Lot 1 -> Total cost: €3

- Purchase 4: Product C | 10 units | €3 per unit |  Landed cost  €10 | Lot 2 -> Total cost €4.

- Purchase 5: Product C | 10 units | €4 per unit |  Landed cost  €10 | Lot 3 -> Total cost €5.



I then made a sale of each of the products purchased.

The cost shown on the sales order line is correct.

I then carried out return tests, validating the delivery note for this type of operation.

When validating the delivery note, the cost of the sales order line is correct, maintaining a relationship with the landed cost indicated in the original purchase.

This allows for a correct valuation of the inventory, even when returning products.

It has also been tested with purchases that have more than one landed cost asociated, maintaining correct traceability.

It has been tested with the sale of various products with various landed costs.

The result is correct.

stock_valuation_fifo_lot/__manifest__.py
```diff
@@ -9,7 +9,12 @@
     "license": "AGPL-3",
     "author": "Ecosoft, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/stock-logistics-workflow",
-    "depends": ["stock_account", "stock_no_negative", "stock_landed_costs"],
+    "depends": [
+        "stock_account",
+        "stock_no_negative",
+        "stock_landed_costs",
+        "mrp_landed_costs",
```
ValentinVinagre: why mandatory install mrp?

ValentinVinagre (changes requested):

stock_valuation_fifo_lot/__manifest__.py
Tisho99: The mrp_production_ids field, used in this module, is created in the mrp_landed_costs module.

https://github.com/OCA/stock-logistics-workflow/blob/181401ea5fa461e97a30cba9bdf18cc8b61cada0/stock_valuation_fifo_lot/models/stock_landed_cost.py#L14

https://github.com/odoo/odoo/blob/15.0/addons/mrp_landed_costs/models/stock_landed_cost.py#L13

ValentinVinagre: Absolutely correct, but that comes from the "mrp_landed_costs" module. The dependency on MRP isn't necessary for this module. Several functionalities are being combined in this module that would have to be separated into another module or implemented with a soft dependency.

yostashiro: Just FYI, we have split the module to stock_valuation_fifo_lot and stock_valuation_fifo_lot_mrp_landed_cost in [the 16.0 migration PR](https://github.com/OCA/stock-logistics-workflow/pull/1676).

luis-ron (approved):
2nd Functional review: LGTM 👍🏻

ValentinVinagre (approved):
LGTM 👍🏻

Tisho99:
Hello @newtratip 

This PR includes some fixes for the module you created. It’s ready to merge.
Would you be interested in reviewing it?

Thanks!

Tisho99:
Hello @dreispt 

This PR includes a few fixes for the module you previously merged.
Would you be interested in reviewing it?

Thanks!

PR #2107, opened 2025-09-19
[18.0][MIG] stock_picking_batch_set_quantity: Migration to version 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT57680


stock_picking_batch_set_quantity/models/stock_move.py
```diff
@@ -7,6 +7,8 @@
...
     _inherit = "stock.move"
 
-    def action_set_quantities_to_reservation(self):
-        """Public method to be called from button"""
-        self._set_quantities_to_reservation()
+    def action_set_quantity(self):
+        if self.product_id.free_qty > self.product_uom_qty:
+            self._set_quantity_done(self.product_uom_qty)
+        if self.quantity == self.product_uom_qty:
+            self.picked = True
```
pedrobaeza: Has this to be done manually?

stock_picking_batch_set_quantity/models/stock_move.py
```diff
@@ -7,6 +7,8 @@
 class StockMove(models.Model):
     _inherit = "stock.move"
 
-    def action_set_quantities_to_reservation(self):
-        """Public method to be called from button"""
-        self._set_quantities_to_reservation()
+    def action_set_quantity(self):
+        if self.product_id.free_qty > self.product_uom_qty:
+            self._set_quantity_done(self.product_uom_qty)
+        if self.quantity == self.product_uom_qty:
```
pedrobaeza: In any case, this should be always be done with `float_is_zero`/`float_round`

stock_picking_batch_set_quantity/models/stock_move.py
eduezerouali-tecnativa: I can call `_set_quantity` that would make more sense but for `picked`  i did not found anything. Is compute field that depends on `state` field.

pedrobaeza: `picked` is a computed field, so that's why I don't understand why you are setting it manually:

https://github.com/odoo/odoo/blob/3009efed3d16f6657aaf0a33e4e7b7b4ad83c289/addons/stock/models/stock_move.py#L262

eduezerouali-tecnativa:
@pedrobaeza makes more sense now?

stock_picking_batch_set_quantity/models/stock_move.py
```diff
@@ -7,6 +7,8 @@
...
     _inherit = "stock.move"
 
-    def action_set_quantities_to_reservation(self):
-        """Public method to be called from button"""
-        self._set_quantities_to_reservation()
+    def action_set_quantity(self):
+        self.ensure_one()
+        self._action_assign()
+        if self.quantity:
+            self.state = "done"
```
pedrobaeza, line 14: This is incorrect, as you can't set state manually. In fact, the idea of the module is to assign quantities, but as I see lately, v18 already pre-assign quantities, so this module doesn't make sense for the new version. On the target version, by default, the reserved quantities wasn't set, so this button was useful for setting that quantities, but now, they are pre-set, so no need of it. Can you confirm?

PR #2109, opened 2025-09-22
[18.0] [ADD] stock_dynamic_routing_delivery

ivantodorovich:


stock_dynamic_routing_delivery/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+    def _get_new_picking_values(self):
+        # OVERRIDE to propagate the carrier, depending on the dynamic routing rule
+        # configuration.
+        vals = super()._get_new_picking_values()
+        if (
+            (routing_rule := self.env.context.get("__routing_rule"))
+            and routing_rule.propagate_carrier
+            and not vals.get("carrier_id")
+        ):
+            vals["carrier_id"] = self.group_id.sale_id.carrier_id.id
```
jbaudoux, line 19: I don't know how odoo is doing to propagate the carrier but shouldn't come from the sale. You better take it from the procurement group. There is an oca module adding it there.

stock_dynamic_routing_delivery/tests/test_carrier_propagation.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        cls.carrier = cls.env["delivery.carrier"].create(
+            {
+                "name": "Test carrier",
+                "delivery_type": "fixed",
+                "fixed_price": 50,
+                "product_id": cls.carrier_product.id,
+            }
+        )
+        cls.order = cls.env["sale.order"].create(
```
jbaudoux, line 86: Do not depend on sale if possible.

stock_dynamic_routing_delivery/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2025 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Stock Dynamic Routing Delivery",
+    "summary": "Glue module between stock dynamic routing and delivery",
+    "version": "18.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
```
jbaudoux:
```suggestion
    "author": "Camptocamp, BCIM, Odoo Community Association (OCA)",
```

stock_dynamic_routing_delivery/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "maintainers": ["ivantodorovich"],
```
jbaudoux:
```suggestion
    "maintainers": ["ivantodorovich", "jbaudoux"],
```

stock_dynamic_routing_delivery/tests/test_carrier_propagation.py
ivantodorovich: We have to depend on `stock_delivery`, which ultimately depends on `sale`, because that's where the field is defined:
https://github.com/odoo/odoo/blob/9203b3631f4a200bef5e627597e78eada521af7d/addons/stock_delivery/models/stock_picking.py#L23

stock_dynamic_routing_delivery/models/stock_move.py
ivantodorovich: I tried to copy what Odoo does for their own `propagate_carrier`:
- https://github.com/odoo/odoo/blob/d737e896100cc5c7ffa62c7d13935001715c40f2/addons/stock_delivery/models/stock_move.py#L43

In STD there's no `carrier_id` in the procurement group, we'd need an extra dependency (`delivery_procurement_group_carrier`) but it makes no sense IMO

jbaudoux: Issue is that we do a carrier selection on the delivery. So we can't rely on the SO carrier.
I fixed the odoo rule propagation here https://github.com/OCA/stock-logistics-workflow/pull/2110

ivantodorovich: I added also a glue module: `stock_dynamic_routing_delivery_procurement_group_carrier`.
Does it make sense to you?

jbaudoux (approved):

PR #2110, opened 2025-09-22
[18.0][FIX] delivery_procurement_group_carrier: propagate

jbaudoux:
Propagate the carrier from the group instead of the carrier from the SO

cc @mt-software-de @ivantodorovich @rousseldenis

### DIFF
--- a/delivery_procurement_group_carrier/models/stock_move.py
+++ b/delivery_procurement_group_carrier/models/stock_move.py
@@ -1,4 +1,6 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+
 from odoo import models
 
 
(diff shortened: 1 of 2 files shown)


mt-software-de (approved):
LGTM. Could you add a test case for this?

sebalix (approved):

delivery_procurement_group_carrier/models/stock_move.py
```diff
@@ -8,6 +10,9 @@ class StockMove(models.Model):
     def _get_new_picking_values(self):
         values = super()._get_new_picking_values()
         carrier = self.group_id.carrier_id
-        if carrier and self.picking_type_id.code in ["outgoing", "incoming"]:
+        if carrier and (
+            self.picking_type_id.code in ["outgoing", "incoming"]
```
ivantodorovich: Is this condition still relevant? shouldn't it be fully based on the rule setting?

ivantodorovich: essentially align to odoo here, I mean:
- https://github.com/odoo/odoo/blob/d737e896100cc5c7ffa62c7d13935001715c40f2/addons/stock_delivery/models/stock_move.py#L43

mmequignon (changes requested):
Needs a test

delivery_procurement_group_carrier/models/stock_move.py
jbaudoux: indeed, I'm removing it

jbaudoux:
Test added

jbaudoux:
Comments answered

/ocabot merge patch

PR #2113, opened 2025-09-25
[FIX] stock_no_negative false positive

gurneyalex:
Because of concurrency, it is possible that a negative quant is created because the main quant is locked and Odoo will avoid waiting in that case, and create additional quants for the product in the location.

This fix updates the way the constraint is computed by gathering the quants and checking the overall quantity.

### DIFF
--- a/stock_no_negative/models/stock_quant.py
+++ b/stock_no_negative/models/stock_quant.py
@@ -31,12 +31,22 @@ class StockQuant(models.Model):
             )
             disallowed_by_location = not quant.location_id.allow_negative_stock
             if (
-                float_compare(quant.quantity, 0, precision_digits=p) == -1
-                and quant.product_id.is_storable
+                quant.product_id.is_storable
                 and quant.location_id.usage in ["internal", "transit"]
                 and disallowed_by_product
                 and disallowed_by_location
             ):
+                gathered_quants = self.with_context(quants_cache=None)._gather(
+                    self.product_id,
+                    self.location_id,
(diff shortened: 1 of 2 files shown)


stock_no_negative/models/stock_quant.py
```diff
@@ -31,12 +31,22 @@ def check_negative_qty(self):
...
                 and disallowed_by_product
                 and disallowed_by_location
             ):
+                gathered_quants = self.with_context(quants_cache=None)._gather(
+                    self.product_id,
+                    self.location_id,
+                    lot_id=self.lot_id,
+                    package_id=self.package_id,
+                    owner_id=self.owner_id,
+                    strict=True,
```
simahawk, line 45:
```suggestion
                    quant.product_id,
                    quant.location_id,
                    lot_id=quant.lot_id,
                    package_id=quant.package_id,
                    owner_id=quant.owner_id,
                    strict=True,
```

CasVissers-360ERP (approved):

CasVissers-360ERP:
Thanks for this. It's a nasty issue.

stock_no_negative/models/stock_quant.py
StefanRijnhart: @gurneyalex Can you have a look at this feedback?

mourad-ehm: @gurneyalex Could you have a look at this feedback please ?

PR #2115, opened 2025-09-26
[18.0][FIX] stock_picking_batch_creation: Fix unsupported operand

grindtildeath:
In case device has no attribute max_weight set, _init_counters function will set self._remaining_weight as None, so we need to consider this possibility before decreasing it when adding picking.


sebalix (approved):

henrybackman (approved):

lmignon (approved):
A unittest would be nice...

jbaudoux (commented):
Can you apply the same of _remaining_nbr_picking_lines ? In order that 0 would mean no limit like for the weight

simahawk:
@grindtildeath ping

jbaudoux (approved):

PR #2133, opened 2025-10-13
[18.0][MIG] stock_picking_import_serial_number: Migration to version 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT57280

ping @victoralmau @sergio-teruel 

I change the test as show_reserved is no longer available in 18.0, I did remove that part as IMHO I think it is unnecessary.


stock_picking_import_serial_number/tests/common.py
```diff
@@ -11,17 +11,16 @@ def assertUniqueIn(self, element_list):
...
+                raise Exception(f"Element {element} is not unique in list")
             elements.append(element)
 
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         cls.lot_obj = cls.env["stock.lot"]
         cls.warehouse = cls.env.ref("stock.warehouse0")
-        cls.picking_type_in = cls.env.ref("stock.picking_type_in")
+        cls.picking_type_in = cls.env.ref("stock.picking_type_out")
```
pedrobaeza: Why changing this? We want to try on incoming receptions.

stock_picking_import_serial_number/tests/test_stock_picking_import_serial_number.py
```diff
@@ -135,76 +135,3 @@ def test_import_serial_number_03(self):
         self.assertIn("LOT-2", lot_names)
         self.assertIn("LOT-3", lot_names)
         self.assertFalse(smls[0].result_package_id)
-
-    @mute_logger("odoo.models.unlink")
```
pedrobaeza: Don't remove these tests, but adapt them. We also want to make this work on outgoing pickings.

pedrobaeza (changes requested):

stock_picking_import_serial_number/tests/test_stock_picking_import_serial_number.py
eduezerouali-tecnativa: I removed those tests as I understood that they are checking `show_reserved` and `move_line_nosuggest_ids`, that are no longer in 18. So if I adapt them I will be checking for the same thing as test above. Am I wrong?

stock_picking_import_serial_number/views/stock_picking.xml
```diff
@@ -3,11 +3,11 @@
         <field name="model">stock.picking</field>
         <field name="inherit_id" ref="stock.view_picking_form" />
         <field name="arch" type="xml">
-            <button name="action_toggle_is_locked" position="after">
+            <button name="action_cancel" position="before">
                 <button
                     name="%(action_import_serial_number)d"
                     string="Import S/N"
-                    attrs="{'invisible': ['|', ('state', '!=', 'assigned'), ('picking_type_code', '!=', 'incoming')]}"
+                    invisible="state != 'assigned' or picking_type_code != 'incoming'"
```
eduezerouali-tecnativa: @pedrobaeza it should only work at the moment whit incoming, right? Anyways i do not see why it wont work with outgoings.

pedrobaeza: With outgoing pickings, you have to check also for reservations, so there are so few steps. That's why the extra tests you removed.

stock_picking_import_serial_number/tests/test_stock_picking_import_serial_number.py
```diff
@@ -122,14 +122,14 @@ def test_import_serial_number_03(self):
         wiz.sn_package_column_index = 10
         wiz.action_import()
         smls = self.picking_in_01.move_line_ids.filtered("lot_name")
-        self.assertEqual(len(smls), 3)
+        self.assertEqual(len(smls), 6)
```
pedrobaeza: Why this change in the numbers? This shows that the implementation is not the same...

eduezerouali-tecnativa:
@pedrobaeza I did change test to check for outgoings.

stock_picking_import_serial_number/wizard/import_serial_number_wizard.py
```diff
@@ -115,22 +114,25 @@ def _import_serial_number(self, xl_sheet, stock_move_lines, picking):
...
                         if item[2]:
                             sml.result_package_id = self._search_or_create_package(
                                 picking, item[2]
                             )
                         # Only assign one serial
                         break
             # TODO: Check if product is present on initial demand??
-            # elif product and picking.move_lines.filtered(lambda ln: ln.product_id == product)
+            # elif product and picking.move_lines.filtered(
+            # lambda ln: ln.product_id == product)
```
pedrobaeza, line 135: Remove this commented code.

stock_picking_import_serial_number/wizard/import_serial_number_wizard.py
```diff
@@ -115,22 +114,25 @@ def _import_serial_number(self, xl_sheet, stock_move_lines, picking):
         )
         for item in serial_list:
             product = products.filtered(
-                lambda p: p[self.sn_search_product_by_field] == item[0]
+                lambda p, item=item[0]: p[self.sn_search_product_by_field] == item
             )
-            if picking.picking_type_id.show_reserved:
-                smls = stock_move_lines.filtered(lambda ln: ln.product_id == product)
+            if picking.picking_type_code == "incoming":
```
pedrobaeza, line 119: Is this condition equivalent to the previous one?

eduezerouali-tecnativa: Actually it should be the other way around ( `!=` ), but test keep failing. My question is if there is not `show_reserved` now, so everything is working as if show_reserved was false or am I  missing something ?

mariancuadranetkia:
@eduezerouali-tecnativa @pedrobaeza 
any update on reviews? I can test it if that helps move this forward

eduezerouali-tecnativa:
> @eduezerouali-tecnativa @pedrobaeza any update on reviews? I can test it if that helps move this forward

sorry for the late replay, if you would like to test it would be great, i will do a rebase for runboat

mariancuadranetkia (approved):
Tested and it's working correctly!

pedrobaeza (approved):
/ocabot merge nobump

PR #2135, opened 2025-10-14
[19.0][MIG] stock_no_negative: Migration to 19.0

BhaveshHeliconia:


JordiMForgeFlow (approved):
code + functional 👍🏼

MateuGForgeFlow (approved):
LGTM!!

CasVissers-360ERP (approved):
Functional review

desdelinux:
Hi @rvalyi @pedrobaeza this is ready for merged

desdelinux (approved):

stock_no_negative/models/product.py
```diff
@@ -9,18 +9,22 @@ class ProductCategory(models.Model):
     _inherit = "product.category"
 
     allow_negative_stock = fields.Boolean(
-        help="Allow negative stock levels for the stockable products "
-        "attached to this category. The options doesn't apply to products "
-        "attached to sub-categories of this category.",
+        help=(
```
pedrobaeza: There's no need of this change that only introduces noise in the migration commit. Please let it as it was (and the same for the rest of superfluous changes).

stock_no_negative/views/product_product_views.xml
```diff
@@ -10,9 +10,9 @@
         <field name="model">product.template</field>
         <field name="inherit_id" ref="stock.view_template_property_form" />
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='is_storable']/parent::div" position="after">
-                <field name="allow_negative_stock" invisible="not is_storable" />
-            </xpath>
+            <field name="categ_id" position="after">
+                <field name="allow_negative_stock" invisible="type != 'consu'" />
```
pedrobaeza: This condition is not correct, as you may have a consumable without tracking quantities (non storable).

pedrobaeza (changes requested):
/ocabot migration stock_no_negative

desdelinux:
Hi @pedrobaeza , I see that the requested changes have already been applied, but you weren't notified.

cc @BhaveshHeliconia

stock_no_negative/views/product_product_views.xml
pedrobaeza: You need to put an invisible condition, but the right one, and better to place this below the tracking options

Zar-SW:
When this PR will be merged?

lammas:
Adding another bump to this since it is the season for doing upgrades and we would like to avoid forking this for 19.

CasVissers-360ERP:
I think we should include #2113

rf-ife (approved):

moylop260:
It fails because of:
 - https://github.com/odoo/odoo/pull/243607

Notice I have already created the fix

Only waiting for merge in odoo repo after that this project will be green again

I mean, the red is not related with this PR

So, merging manually for now

PR #2137, opened 2025-10-14
[18.0][IMP] stock_partner_delivery_window: check date

jbaudoux:
Allow to check if a date is in window. Now works with date and datetime. This allows to check a delivery date without time.

cc @grindtildeath @sebalix @mmequignon


sebalix (approved):

simahawk (approved):
LG

PR #2138, opened 2025-10-15
[18.0][FIX] stock_move_line_qty_picked: fix inventory

jbaudoux:
cc @sebalix


stock_move_line_qty_picked/models/stock_move_line.py
```diff
@@ -24,6 +25,13 @@ def _inverse_picked(self):
                 continue
             # Pick full quantity when 'picked = True' and no qty were picked
             if not rec.qty_picked:
+                if not rec.quantity:
```
simahawk, line 28:
```suggestion
if not rec.qty_picked and rec.quantity:
```

sebalix (approved):

PR #2143, opened 2025-10-23
[18.0][FIX] stock_partner_delivery_window: fix tz conversion

jbaudoux:
* Respect daylight saving
  astimezone considers the date in the tz conversion while tz_to_utc_time only looks at the time. Today CEST is +2 but next monday CEST is +1. When the delivery date is next week, proper conversion matters based on the date.
* Add missing conversions
* Fix error message content

cc @grindtildeath @sebalix

### DIFF
--- a/stock_partner_delivery_window/models/res_partner.py
+++ b/stock_partner_delivery_window/models/res_partner.py
@@ -4,11 +4,11 @@
 
 import datetime
 
+import pytz
+
 from odoo import Command, api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.osv import expression
-
-from odoo.addons.partner_tz.tools import tz_utils
 
 WORKDAYS = list(range(5))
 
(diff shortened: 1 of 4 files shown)


gurneyalex (commented):
could we have an automated test?

sebalix (approved):
A test is welcome indeed

stock_partner_delivery_window/models/res_partner.py
```diff
@@ -101,6 +101,9 @@ def is_in_delivery_window(self, date):
         :return: Boolean
         """
         self.ensure_one()
+        tz = pytz.timezone(self.tz or "UTC")
```
sebalix: Question: as you are using `<res.partner>.tz` field, should we not keep `partner_tz` dependency by convenience? It adds the field on partner form view.

jbaudoux: Also, it's not required, so I will add a fallback on the company tz

jbaudoux: done

jbaudoux:
Tests added

jbaudoux:
@gurneyalex good for you now?

PR #2160, opened 2025-11-10
[18.0][FIX] stock_picking_batch_creation: fix stock device order

TDu:
Before this change when searching for a device to use for the batch, the order of the devices was not guaranteed although important. And it could cause some batch split not happening.

Now the possible devices will be sorted by their default order, which is by sequence and name.

### DIFF
--- a/stock_picking_batch_creation/wizards/make_picking_batch.py
+++ b/stock_picking_batch_creation/wizards/make_picking_batch.py
@@ -266,7 +266,7 @@ class MakePickingBatch(models.TransientModel):
         )
 
     def _split_first_picking_for_limit(self, picking):
-        last_device = self.stock_device_type_ids[-1]
+        last_device = self._get_sorted_devices()[-1]
         if last_device.split_mode == "dimension":
             return (
                 self.env["stock.split.picking"]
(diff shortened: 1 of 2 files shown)


simahawk:
no need to update tests?

TDu:
> no need to update tests?

I have added 2 in relation to the changes

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -263,7 +263,7 @@ def _execute_search_pickings(self, domain, limit=None):
         )
 
     def _split_first_picking_for_limit(self, picking):
-        last_device = self.stock_device_type_ids[-1]
+        last_device = self.stock_device_type_ids.sorted()[-1]
```
lmignon: We could use a method to get the list of devices in order to centralize the required logic to ensure the right order etc...

TDu: Good point, I did this in the last fixup.

jbaudoux (approved):

jbaudoux:
rebased and squashed fixup

PR #2161, opened 2025-11-11
[19.0][MIG] stock_picking_mass_action: Migration to 19.0

lef-adhoc:


BhaveshHeliconia (approved):
LGTM!

stock_picking_mass_action/readme/USAGE.md
```diff
@@ -0,0 +1,5 @@
+After selecting several pickings in tree view there are options for: \*
```
carlos-lopez-tecnativa, line 1:
```suggestion
After selecting several pickings in list view there are options for: \*
```

carlos-lopez-tecnativa (changes requested):
I think this module is no longer necessary. Odoo's features already cover these cases when working with multiple pickings from the list view:

- An option to `Check Availability`, which confirms the picking if it is in draft.
- An action to `Validate` the picking, which asks whether a backorder should be created when necessary.

The only missing feature in Odoo core is the cron job, but since the cron is inactive by default, I think many users may never enable it.

What do you think, @pedrobaeza @carlosdauden @sergio-teruel 

https://github.com/odoo/odoo/commit/6ab4b0d4967490ed1ee0559ec296da6637705125

Image

Image

pedrobaeza:
Yes, I agree it seems not needed anymore.

sergio-teruel:
Yes, you are rigth

PR #2163, opened 2025-11-12
[16][FIX] stock_move_propagate_first_move

AdrianaSaiz:
Add index for first_move_id to improve performance

Without indexing first_move_id field, when trying to execute write method on a stock move, for example when we confirm a sale order. The query lasts for at least more than 7 seconds. 

This slows down the system and can even cause PostgreSQL blocks.

As we can see in this profiler, the query:

sql('SELECT "stock_move".id FROM "stock_move" WHERE ("stock_move"."first_move_id" in %s) ORDER BY "stock_move"."id" ') (SELECT "stock_move".id FROM "stock_move" WHERE ("stock_move"."first_move_id" in (25199374)) ORDER BY "stock_move"."id" )

The query is consuming more than 70% of the order confirmation time, when making a return.

image

Whereas if we index the field first_move_id, the query lasts less than a second and all the process is finished in only 1 second

image

### DIFF
--- a/stock_move_propagate_first_move/README.rst
+++ b/stock_move_propagate_first_move/README.rst
@@ -70,6 +70,7 @@ Contributors
 ~~~~~~~~~~~~
 
 * Souheil Bejaoui (ACSONE SA/NV) <souheil.bejaoui@acsone.eu>
+* Adriana Saiz <adriana.saiz@factorlibre.com>
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


xAdrianCif (approved):
Great analysis and I think very useful, thanks for the work!

alvaro-gmz:
LGTM

FlJesusLorenzo:
LGTM

marielejeune (approved):

rousseldenis (approved):

sbejaoui (approved):

BhaveshHeliconia (approved):

PR #2170, opened 2025-11-17
[18.0][MIG] stock_lot_scrap: Migration to 18

Andrii9090-tecnativa:
Straightforward migration to v18

To test this module you need: 
1. to go to **Inventory -> Products -> Lots / Serial Numbers** en select one
image
2. Click **Scrap**
image
3. Click **Validate**
image

@moduon MT-11840


fcvalgar (changes requested):
Thank you so mucho @Andrii9090 for this contrubution,

This is my first time testing this module in Odoo, but some of the tests I performed seem to reveal behaviours that might not be intended. I’m not fully sure whether these issues also occurred in previous versions or if they are new to this one, but I’ve documented them in detail in case they help with the review.

Test 1: Check full scrap of a lot stored in a single internal location – OK
Test 2: Check scrap of a lot with quantities split across multiple internal locations – Not OK: when splitting the lot during receipt Odoo changes quantities on validation, and if we have 50 units of lot 001 in location A and 50 in location B, scrapping A first causes the module to generate an unnecessary draft scrap order for A again when later scrapping B.

image

image

Test 3: Check behaviour when trying to scrap a lot with no internal stock – OK
Test 4: Check chatter message logged on the lot after scrapping – OK
Test 5: Check button visibility and permissions for users with inventory access – OK
Test 6: Check stock consistency after scrapping a lot entirely – OK
Test 7: Check compatibility with the standard scrap form – Not OK: the Lot field cannot be selected even when stock exists, indicating interference with the standard stock.scrap behaviour.

image

Can you review these cases?

Andrii9090-tecnativa:
@fcvalgar thank you for review!

I changed the code and now if `stock.lot quantiy > 0 ` we create the Scrap Order

https://www.loom.com/share/001e9a6be25e492e80d93fd6046592bb

fcvalgar (approved):
LGTM, thank you @Andrii9090 for the great contribution.

stock_lot_scrap/models/stock_lot.py
```diff
@@ -102,7 +102,7 @@ def action_scrap_lot(self):
         )
         result["context"] = self.env.context
         if len(scraps) != 1:
-            result["domain"] = "[('id', 'in', %s)]" % scraps.ids
+            result["domain"] = f"[('id', 'in', {scraps.ids})]"
```
chienandalu: You don't even need to format it. Odoo can parse the python object directly:
```suggestion
            result["domain"] = [("id", "in", scraps.ids)]

File: stock_lot_scrap/models/stock_lot.py [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+                    (
+                        "scrap_location",
+                        "=",
+                        True,
+                    ),
+                ],
+                limit=1,
+            )
+            .id
+        )
```
chienandalu: I'd say you don't need this as the location is precomputed by the scrap object itself: https://github.com/odoo/odoo/blob/8ffa2fe827a607d597134c49b17e651c31720fd1/addons/stock/models/stock_scrap.py#L83-L93

And indeed is causing issues:

Image

stock_lot_scrap/models/stock_lot.py
```diff
@@ -0,0 +1,110 @@
...
+            scraps |= scrap
+        result = self.env["ir.actions.act_window"]._for_xml_id(
+            "stock.action_stock_scrap"
+        )
+        result["context"] = self.env.context
+        if len(scraps) != 1:
+            result["domain"] = f"[('id', 'in', {scraps.ids})]"
+        else:  # pragma: no cover
+            res = self.env.ref("stock.stock_scrap_form_view", False)
+            result["views"] = [(res and res.id or False, "form")]
```
chienandalu:
```suggestion
            result["views"] = [(self.env.ref("stock.stock_scrap_form_view").id, "form")]
```

chienandalu (changes requested):

chienandalu (approved):
Much better 👍 Thanks!

rafaelbn (approved):
Functional review 👍🏼 

/ocabot merge patch

PR #2171, opened 2025-11-17
[18.0][FIX] stock_picking_batch_creation: Imp first picking slection

TDu:
When trying to create a batch picking only with the first availale transfer, there could be some occurence where that picking does not fit the dimension of the device and can not be split.

This can happen when all moves from the transfer are oversize compare to the dimension of the device.

After this change, the module will use the next transfer to create a batch until it is successfull.

### DIFF
--- a/stock_picking_batch_creation/wizards/make_picking_batch.py
+++ b/stock_picking_batch_creation/wizards/make_picking_batch.py
@@ -1,6 +1,7 @@
 # Copyright 2021 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
 import math
 import threading
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


jbaudoux (approved):

stock_picking_batch_creation/wizards/make_picking_batch.py
```diff
@@ -316,23 +319,32 @@ def _get_first_picking(self, raise_if_not_found=False):
...
+                    _logger.warning(
+                        f"The picking {picking.name} could not be split "
+                        "for batch creation."
+                    )
+                    continue
+                selected_picking = split_picking
+            else:
+                selected_picking = picking
+            break
+        return selected_picking
```
mmequignon, line 347:
```suggestion
                return split_picking
            else:
                return picking
```

mmequignon (approved):

stock_picking_batch_creation/wizards/make_picking_batch.py
TDu: If there is no picking in the `pickings`recordset, the function will return None instead of an empty recordset :frowning:

Issue #2116, opened 2025-09-29
[18.0] stock_picking_origin_reference_sale: doesn't work if `client_order_ref` is set on SO


sebalix:
## Module

`stock_picking_origin_reference_sale`

## Describe the bug

As soon as we set a Customer Reference on the SO, the resulting transfer is unable to find it as Origin is set with `SO.name - SO.client_order_ref` value.

## To Reproduce
**Affected versions**: 18.0

1. Create an SO with a Customer Reference
2. Validate it
3. Open the delivery order

**Expected behavior**

A link to the SO available as the source document on transfer.

fsmw:
This is fixed by PR #2322 — extracting the SO name from `picking.origin` by splitting on ` - ` when `client_order_ref` is present.\n\nhttps://github.com/OCA/stock-logistics-workflow/pull/2322

PR #2195, opened 2025-12-08
[18.0][FIX] stock_picking_origin_reference_sale: Handle client_order_ref in origin

fsmw:
Fix issue #2116: Module fails to link pickings to SO when client_order_ref is set.

**Root Cause**: Origin includes 'SO.name - client_order_ref', but search used full origin as name.

**Fix**: Extract SO name by splitting on ' - '.

**Changes**:
- Modified _compute_origin_reference to parse origin correctly.
- Added test for client_order_ref scenario.


fsmw:
Closing in favor of separate PRs per module as required by OCA guidelines. Will re-create:
- stock_picking_tier_validation: fix singleton error (self→rec in loop)
- stock_picking_origin_reference_sale: handle client_order_ref in origin
- stock_picking_group_by_partner_by_carrier: fix _create_backorder serialization

PR #2196, opened 2025-12-08
[18.0] Fix stock_picking_group_by_partner_by_carrier _create_backorder serialization issue

### LINKED ISSUE #2001: [17.0][18.0] stock_picking_group_by_partner_by_carrier: Problem due to the serialization of the super call in `_create_backorder`
Steps to reproduce:

- Install `stock_picking_group_by_partner_by_carrier` and `stock_picking_batch`.
- Create 2 pickings.
- Add both to the same batch.
- Assign and put a partial done quantity.
- Validate the batch picking and do backorders.

Result: The first picking is removed from the batch picking, while it shouldn't.

These steps can be seen in this test reproducing the scenario:

https://github.com/OCA/stock-logistics-workflow/actions/runs/15872247400/job/44751372455?pr=1997#step:8:732

Why is this happening? The module `stock_picking_group_by_partner_by_carrier` is serializing the super call of `_create_backorder`:

https://github.com/Tecnativa/stock-logistics-workflow/blob/63969b7ef7a9d3534c7e4a4c222a292b6882be3a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py#L93

but the module `stock_picking_batch` is performing a condition reviewing if you are backordering the whole batch picking (the part `any(p not in self for p in picking.batch_id.picking_ids - pickings_to_detach)`):

https://github.com/odoo/odoo/blob/d83801027c9a88b6e4db8166b6e2aad41ed760b9/addons/stock_picking_batch/models/stock_picking.py#L178

as we are not passing to super the whole recordset, then each picking except the last are detached from the batch picking.

As I don't know the involved module, I can't suggest a solution for not having to serialize the super call. One I'm thinking on is to prepare a dictionary to pass by context with each picking d True/False as values for being checked in the other methods.

Please check @sebalix, as you were the one introducing the problem in https://github.com/OCA/stock-logistics-workflow/commit/b4bcddfccb6e550cacba68034cf685c38339527a

Other contributors: @jbaudoux @rousseldenis

fsmw:
## Description
Fixes issue #2001: Problem due to the serialization of the super call in `_create_backorder` causing incorrect batch detachment.

### Root Cause
The module was serializing the super call in `_create_backorder`, calling it per picking individually. This caused `stock_picking_batch` to incorrectly detach pickings from the batch because it checks if all pickings in the batch are being backordered, but with serialization, the recordset passed to super was a single picking.

### Solution
- Modified `_create_backorder` to call super once with the full recordset.
- Updated the copy method to use a context dict for per-picking grouping flags.
- Ensured grouping logic is applied to backorders after super call.

### Testing
- Pre-commit checks pass.
- Existing tests should continue to work.
- The fix addresses the reproduction scenario described in the issue.

## Related Issues
- Closes #2001


fsmw:
Closing in favor of #2324 — clean single-commit replacement with proper [18.0][FIX] prefix.

PR #2199, opened 2025-12-09
[IMP]stock_restrict_lot: allow editing of restric_lot_id

danielalvaro1999:
# Refactor: Make `restrict_lot_id` Field Editable in Stock Moves

This PR makes the `restrict_lot_id` field on `stock.move` editable in tree views (picking and move views). Previously, the field was `readonly` to prevent users from modifying it, which could cause issues with reservations and adds validation logic to ensure data consistency when changing lot restrictions.

## Changes Made
- **Removed `readonly="1"` attribute** from the `restrict_lot_id` field in the following views:
  - `stock_move_pick_tree_view` (inherits `stock.view_picking_move_tree`)  
    Displays stock moves inside picking operations
  - `stock_move_tree_view` (inherits `stock.view_move_tree`)  
    Displays stock moves in the general move list
  - `custom_main_stock_picking_form_view` (inherits `stock.view_picking_form`)  
    Displays moves within the picking form view (embedded tree in `move_ids_without_package`)
- **Added domain constraint** to `restrict_lot_id` field:
 
  ```python
  domain="[('product_id', '=', product_id)]"
  ```
- **Added restriction**
  
  ```python
  moves_with_reservation = self.filtered(
      lambda m: m.state in ["assigned", "partially_available"]
      and m.move_line_ids
  )


  if moves_with_reservation:
      raise UserError(
          _(
              "Cannot change the Lot/Serial number restriction on moves "
              "that already have stock reserved.\n\n"
              "Please unreserve the stock first (button 'Unreserve'), "
              "change the Lot/Serial number, and then reserve again "
              "(button 'Check Availability')."
          )
      )
  ```

## Reasoning
- **Improved UX**: Allow users to manually select lots for stock moves before confirmation directly from list views, without needing to open each move in form view
- **Preserve validation logic**: All existing reservation logic and validation implemented in the `stock_restrict_lot` module remains intact
- **Maintain propagation**: Lot changes continue to propagate to chained moves as designed
- **Prevent post-validation changes**: The model-level validation still prevents changes to the lot after moves are validated, ensuring data consistency

## Impact
- **Users** can select or change lots before validation directly in tree views
- **Reservations** are automatically adjusted to reflect selected lots (handled by existing model logic)
- **Tests and functionality** of the `stock_restrict_lot` module remain intact
- **No breaking changes**: Existing workflows continue to function as before, with added flexibility

## Technical Details
The fields were updated to remove the `readonly="1"` attribute while keeping:
- `optional="show"` - field is visible but can be hidden by users
- `groups="stock.group_production_lot"` - restricted to users with lot/serial tracking permissions
- Domain constraint on the field (defined in the model) - limits selection to lots matching the product


alvaro-gmz:
LGTM

PR #2200, opened 2025-12-10
[17.0][IMP] stock_lot_lot: don't reserve locked lots; allow (un)lock of reserved lots

dreispt:
Includes #2250

### DIFF
--- a/stock_lock_lot/tests/test_stock_lock_lot.py
+++ b/stock_lock_lot/tests/test_stock_lock_lot.py
@@ -1,4 +1,5 @@
 # Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2025 Open Source Integrators (http://www.opensourceintegrators.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import exceptions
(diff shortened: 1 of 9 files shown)


pedrobaeza:
As I'm not using this module anymore, I'm not opposing your changes nor being able to review it.

Reyes4711-S73 (approved):
LGTM

eugenios73 (approved):
LGFM

stock_lock_lot/tests/test_stock_lock_lot.py
```diff
@@ -43,3 +57,67 @@ def test_lock_permissions(self):
...
+        # Check initial availability - both should have 10 available
+        self.assertEqual(locked_quant.available_quantity, 10.0)
+        self.assertEqual(unlocked_quant.available_quantity, 10.0)
+
+        # Try to create a stock move that would need to reserve inventory
+        move = self.env["stock.move"].create(
+            {
+                "name": "Test move",
+                "product_id": self.product.id,
+                "product_uom_qty": 5.0,
```
hparfr, line 92: Why not try to reserve 20 ?

dreispt: The test is to reserve a partial quantity of the on hand total.
I don't think testing for larger quantities than the on hand is relevant here.

dreispt:
I can see test errors needing attention...
```

2026-02-11 13:26:17,250 358 ERROR odoo odoo.addons.stock_lock_lot.tests.test_stock_lock_lot: FAIL: TestStockLockLot.test_locked_lot_excluded_from_reservation
  File "/__w/stock-logistics-workflow/stock-logistics-workflow/stock_lock_lot/tests/test_stock_lock_lot.py", line 116, in test_locked_lot_excluded_from_reservation
    self.assertEqual(
AssertionError: 0.0 != 5.0 : Unlocked lot should be reserved
```

stock_lock_lot/tests/test_stock_lock_lot.py
hparfr: Then, how can you be sure odoo can't take the locked lot ?

dreispt:
Thank you for the reviews.
I found cases were we do want to have the locked lot being reserved for operations.
So this PR was modified to have that behaviour configurable.

Reyes4711-S73 (approved):

PR #2204, opened 2025-12-16
[18.0][ADD] stock_cancel_picking_scheduler

rousseldenis:


stock_cancel_picking_scheduler/models/procurement_group.py
```diff
@@ -0,0 +1,51 @@
...
+            ("user_id", "=", False),
+        ]
+
+    @api.model
+    def _cancel_waiting_pickings(self) -> None:
+        pickings = self.env["stock.picking"].search(
+            self._get_cancel_waiting_pickings_domain()
+        )
+        if pickings:
+            pickings.action_cancel()
```
sebalix, line 28: Should we not check if such transfers are linked to others (chained) transfers to be on the safe-side?
A bad configuration on the PACK operation type and it could screw the chain.

PR #2216, opened 2026-01-08
[18.0][IMP] stock_move_line_qty_picked: qty_picked in stock move views

LauraCForgeFlow:


stock_move_line_qty_picked/tests/test_stock_move_line_qty_picked.py
```diff
@@ -65,28 +65,35 @@ def test_move_pick_qty_reservation(self):
         self.assertEqual(move_line.quantity, 5)
         self.assertEqual(move_line.qty_picked, 0)
         self.assertFalse(move_line.picked)
+        self.assertEqual(move.qty_picked, 5)
         self.assertEqual(self.quant.reserved_quantity, 5)
         # Pick partial qty
         move_line.qty_picked = 4
+        move.invalidate_recordset(["qty_picked"])
```
jbaudoux: This shows you forgot the api.depends on the new computed field

jbaudoux (changes requested):
Can you also remove the auto generated files from the commit

LauraCForgeFlow:
Hi @jbaudoux, thanks for reviewing the PR! I have pushed the requested changes.

stock_move_line_qty_picked/tests/test_stock_move_line_qty_picked.py
```diff
@@ -65,20 +65,23 @@ def test_move_pick_qty_reservation(self):
         self.assertEqual(move_line.quantity, 5)
         self.assertEqual(move_line.qty_picked, 0)
         self.assertFalse(move_line.picked)
+        self.assertEqual(move.qty_picked, 5)
```
grindtildeath: It looks wrong to have a qty picked on the move when there isn't any on the move line

stock_move_line_qty_picked/tests/test_stock_move_line_qty_picked.py
```diff
@@ -105,10 +109,12 @@ def test_move_concurrent_pick_qty(self):
         self.assertEqual(move_line.quantity, 10)
         self.assertEqual(move_line.qty_picked, 0)
         self.assertFalse(move_line.picked)
+        self.assertEqual(move.qty_picked, 10)
```
grindtildeath: Same as above, it looks wrong to have a qty picked on the move when there isn't any on the move line

grindtildeath (commented):
Thanks for your contribution, but IMO the compute is not right.

IIRC I didn't define this field in the first place because it was quite tricky to do.

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -1,13 +1,20 @@
...
+from odoo import api, fields, models
 from odoo.tools import float_compare
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    qty_picked = fields.Float(compute="_compute_qty_picked")
+
+    @api.depends("picked", "move_line_ids", "product_uom", "quantity")
```
jbaudoux: This depends is not right

jbaudoux (changes requested):

LauraCForgeFlow:
Hello @jbaudoux, @grindtildeath. Regarding the actual computation of the field, I agree that it could be a bit confusing to have some quantity picked in the SM in some cases where there was no quantity picked in the SML. I originally did it like that to respect the picked quantity computed in the SM by the `_get_picked_quantity` method, however now I have changed the computation to only take into account the actual picked quantities in the SML, and changed the tests accordingly.

Additionally, I've also updated the `@api.depends` decorator for the method. Initially, I hadn't included it since the field is non-stored, but I can understand the importance of having clear dependencies documented. I've redefined it to best reflect the field's actual data sources, even though tracking dependencies for a non-stored field can be a bit complex.

stock_move_line_qty_picked/models/stock_move.py
```diff
@@ -1,13 +1,26 @@
...
+        "move_line_ids.quantity",
+        "product_uom",
+    )
+    def _compute_qty_picked(self):
+        for move in self:
+            move.qty_picked = move._sum_ml_qty_picked()
+
     def _get_picked_quantity(self):
         if self._ml_has_qty_picked():
             return self._sum_ml_qty_picked()
```
jbaudoux:
```suggestion
            return self.qty_picked
```

jbaudoux:
@LauraCForgeFlow Thanks. Can you squash the commits and drop the auto generated files from the commit

LauraCForgeFlow:
Thanks, done! @jbaudoux

grindtildeath (approved):
Nice :ok_hand:

jbaudoux (approved):

PR #2220, opened 2026-01-09
[19.0][MIG] sale_order_global_stock_route: Migration to 19.0

sergioM-S73:
Adapt code to replicate change in odoo base  on sale_order_lines model -> route_id (Many2one) to route_ids (Many2many)
https://github.com/odoo/odoo/commit/b7a9196366eb206201d59c79542194049a890db1#diff-c728b1f95e4d88de6354fa24c888c5a70e7c435f43002b7ee1c87238e6cab80e


Reyes4711-S73 (approved):
LGTM

pablo-cort-s73 (approved):
LGTM

Reyes4711-S73:
@OCA/logistics-maintainers Please, can you review/merge this PR?

sale_order_global_stock_route/models/sale_order.py
```diff
@@ -7,43 +7,51 @@
...
-                    order = self.env["sale.order"].browse(vals["order_id"])
-                    if order.route_id:
-                        vals["route_id"] = order.route_id.id
-            return super().create(vals_list)
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = self.env["sale.order"]
+        for vals in vals_list:
+            res |= super(
+                SaleOrder, self.with_context(route_ids=vals.get("route_ids", False))
```
hparfr, line 39: why this change ?
Did you check if creating a context per record has an impact in perfs ?

hparfr (commented):
Thanks for this work.

sale_order_global_stock_route/models/sale_order.py
sergioM-S73: @hparfr The change was made because of an issue found when running the tests. From the client side, the route_ids are correctly reflected in the sale.order.line thanks to the onchange methods. However, the tests revealed that if sales orders and their lines are created from the backend in this way:

```
new_order = self.env["sale.order"].create(
   [
        {
            "partner_id": self.partner.id,
            "order_line": [
                (
                    0,
                    0,
                    {
                        "name": self.product1.name,
                        "product_id": self.product1.id,
                        "product_uom_qty": 1,
                        "product_uom_id": self.product1.uom_id.id,
                    },
                )
            ],
            "route_ids": [(6, 0, [self.route1.id])],
        },
    ]
)
```


it doesn't work as expected because the lines do not receive the route_ids information from the sale.order.
It was decided to pass the route_ids field through the context to ensure the functionality is covered in all cases.

mav-adhoc:
image
image
After confirmation of the quotation, not only was the route erased from the quotation, but also the 'Test' register i just created.

sergioM-S73:
@mav-adhoc Thanks for the review.

This was not actually caused by the quotation confirmation deleting the route. The issue was that the route could be created directly from this new field in the sale order form.

When a route is created from there, `sale_selectable` is not set to `True` by default, so the newly created route does not match the field domain `sale_selectable = True`. Because of that the route disappears from the field. Confirming the quotation is not required to reproduce it; a simple refresh is enough.

To avoid this behavior, I changed the field to follow the same approach used by the standard sale order line route field, adding `options="{'no_create': True}"`, so routes cannot be created from this shortcut field.

mav-adhoc:
@sergioM-S73 Perfect. Thank you!

LGTM

mav-adhoc (approved):

sale_order_global_stock_route/models/sale_order.py
hparfr: With this design, I think the benefits of model_create_multi is lost for all the orders.

I suggest may be one of the following:
- write the test differently. Like `new_order = env["sale.order"].create({..}) env["sale.order.line"].create({"order_id: new_order.id, ...} `
- refactor the code to split orders by route_ids sets.

May you explicit this constraint / behavior in the readme.

sergioM-S73: Thanks for the review.

I reworked it to keep a single `super().create(vals_list)` call on sale.order. Once the orders are fully created, their routes are propagated to the created lines that do not already have explicit routes.

The `sale.order.line.create()` override covers lines created afterwards on an existing sale order.

Could you please review the new approach? Suggestions are welcome if you think there is a better way to handle it.

sale_order_global_stock_route/models/sale_order.py
```diff
@@ -0,0 +1,59 @@
...
+                lambda line: line.route_ids.ids != vals["route_ids"]
+            )
+            lines.write({"route_ids": vals["route_ids"]})
+        return res
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        orders = super().create(vals_list)
+        for order in orders.filtered("route_ids"):
+            order.order_line.filtered(
```
hparfr, line 38:
```suggestion
            if order.route_ids:
                order.order_line.filtered(
```

There is no need to iterate if there is no route to set

hparfr (commented):
Looks better thanks.

sale_order_global_stock_route/models/sale_order.py
sergioM-S73: This is already handled by orders.filtered("route_ids"), so only orders with routes are iterated.

hparfr: indeed ! :)

PR #2222, opened 2026-01-13
[18.0][IMP] stock_picking_batch_creation: allow no max nbr lines and unlimited volume

jbaudoux:
Similar to previous improvement on max_weight

cc @mmequignon @TDu @lmignon @grindtildeath


lmignon (approved):
a unittest for this usage would be nice

jbaudoux:
> a unittest for this usage would be nice

Test added and applied same change to support unlimited volume.
Added missing test also for unlimited weight related to https://github.com/OCA/stock-logistics-workflow/pull/2115

sebalix (approved):

PR #2236, opened 2026-01-23
[18.0][IMP] stock_restrict_lot: Return picking compatibility

victoralmau:
Return picking compatibility

Please @pedrobaeza and  @pilarvargas-tecnativa can you review it?

@Tecnativa


pedrobaeza (approved):

florian-dacosta (approved):
Seems legit

PR #2177, opened 2025-11-19
[18.0][MIG] stock_picking_auto_create_lot_sequence: Migration to 18.0

AungKoKoLin1997:
@qrtl QT5088

PR #2237, opened 2026-01-23
[IMP] stock_picking_auto_create_lot: add SKU-based lot/serial numbering

tendil:
Improves Stock Picking Auto Create Lot by adding per-product lot/serial generation options.

* Replaced the boolean with `auto_create_lot_option` (Odoo sequence / SKU based).
* Added Inventory setting “SKU Based Numbers Trailing” and applied it to SKU-based numbering.
* Implemented SKU-based lots/serials as `<SKU>-<N>` using an auto-managed per-product `ir.sequence`, with fallback to `stock.lot.serial` when SKU is missing.
* Added migration (bool -> selection), updated docs, and extended tests to keep codecov diff coverage.

### DIFF
--- a/stock_picking_auto_create_lot/README.rst
+++ b/stock_picking_auto_create_lot/README.rst
@@ -39,23 +39,49 @@ create lots for incoming pickings.
 Configuration
 =============
 
-To configure this module, you need to:
+To use the auto-create feature you need to enable tracking by
+lots/serial numbers:
 
-1. Go to a *Inventory > Configuration > Operation Types*.
-2. Set 'auto create lot' option for this operation type.
-3. Go to a *Inventory > Master Data > Products*.
-4. Set 'auto create lot' option for the products you need.
+1. Go to "Inventory > Configuration > Settings" and scroll down to the
+   "Traceability" section;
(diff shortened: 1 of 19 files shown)


yostashiro:
@tendil There is a bridge module stock_picking_auto_create_lot_sequence (18.0 migration PR: https://github.com/OCA/stock-logistics-workflow/pull/2177) to make stock_picking_auto_create_lot work with [product_lot_sequence](https://github.com/OCA/product-attribute/tree/18.0/product_lot_sequence), and I think the improvement you are proposing here overlaps with the existing functionality.

Maybe you could try improving product_lot_sequence if it's lacking something you need?

ivs-cetmix:
Hi @yostashiro, thank you for mentioning this! Yes, we are aware about the [product_lot_sequence](https://github.com/OCA/product-attribute/tree/18.0/product_lot_sequence), however for some reason we missed the bridge module.
Anyway, the product_lot_sequence module flow differs from the one that's being implemented with this PR. Main differences are:
- It defines sequences at the product level.
- It uses the product variant SKU as a basis for the sequence.

So far we will keep an eye on the PR you mentioned and probably will try to merge the flow into it if it's not over-complicated.

PR #2243, opened 2026-01-30
[18.0][MIG] stock_picking_move_line_no_print: Migration to 18.0

Andrii9090-tecnativa:
Migration to 18.0

To prevent displayed picking move lines when Odoo groups lines (the same product), I used the `_merge_moves_fields` method from `stock.move` to check that at least one line has `display_to_report` set to False and set `display_to_report` to False for grouped lines.
https://www.loom.com/share/b5c4cc81136f45bbbaddb548c4b50ee6
@moduon MT-13229


BhaveshHeliconia (approved):
Functional review LGTM!

fcvalgar (approved):
Great work on this contribution @Andrii9090-tecnativa. I have tested the module and the feature works as expected.

The following tests were performed:
Test 1: OK - Delivery slip with one line marked as not printable. The line was correctly hidden from the printed delivery report.
Test 2: OK - Delivery slip with the default printable setting. The line was displayed normally when Display in report was enabled.
Test 3: OK - Delivery slip with multiple lines, one visible and one hidden. The report correctly displayed only the printable line.
Test 4: OK - Hidden line does not affect stock validation. The transfer was validated correctly and stock was moved, while the line remained hidden in the report.
Test 5: OK - Standard behavior without hidden lines. The delivery slip kept the standard Odoo behavior and displayed all printable lines.

Everything looks good to me, so I approve this PR.

PR #2248, opened 2026-02-09
[18.0][ADD] stock_quant_package_reusable: new module

vvrossem:


stock_quant_package_reusable/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    "package_id": package.id,
+                    "picking_id": self.id,
+                    "location_id": False,
+                    "location_dest_id": move_line_ids.location_dest_id.id,
+                    "move_line_ids": [(6, 0, move_line_ids.ids)],
+                    "company_id": self.company_id.id,
+                }
+            )
+        return package
```
vvrossem, line 60: :information_source: To avoid repeating the logic, I've opened this PR https://github.com/odoo/odoo/pull/247952 to extract the package creation into a dedicated _get_put_in_pack_package method. 
If merged, I'll update this current PR

stock_quant_package_reusable/wizards/select_reusable_package.py
```diff
@@ -0,0 +1,26 @@
...
+    warehouse_id = fields.Many2one(
+        related="picking_id.picking_type_id.warehouse_id",
+    )
+    package_id = fields.Many2one(
+        "stock.quant.package",
+        string="Reusable Package",
+        required=True,
+        domain="[('package_use', '=', 'reusable'), "
+        "('location_id.usage', '=', 'internal'), "
+        "('location_id', 'child_of', warehouse_id)]",
```
gurneyalex: Looks incorrect. location_id is a stock.location and warehouse_id is a stock.warehouse. 


```suggestion
        "('location_id', 'child_of', warehouse_id.view_location_id)]",
```

gurneyalex (commented):
I think I found an issue. Can you double check?

stock_quant_package_reusable/wizards/select_reusable_package.py
vvrossem: great catch :pray:

stock_quant_package_reusable/wizards/select_reusable_package.py
```diff
@@ -0,0 +1,31 @@
...
+        domain="[('package_use', '=', 'reusable'), "
+        "('location_id.usage', '=', 'internal'), "
+        "('location_id', 'child_of', warehouse_view_location_id)]",
+    )
+
+    def action_confirm(self):
+        return self.picking_id.with_context(
+            reusable_package_id=self.package_id.id
+        ).action_put_in_pack(
+            move_lines_to_pack=self.env.context.get("move_lines_to_pack")
```
SilvioC2C, line 30: Nitpicking: I would store this info into a proper field for the wizard, to avoid the risk of propagating wrong values from the context (which doesn't have a validation mechanism in place) and prevent data loss.

Not blocking.

SilvioC2C (approved):
LGTM overall; one minor, non-blocking suggestion

PR #2249, opened 2026-02-10
[18.0][MIG] stock_lot_on_hand_first: migrate to 18.0

twalter-c2c:
In Odoo 18, there is no longer the `_name_search` method. I chose to use the `search_fetch` instead (see `stock_lot.py`). I also removed the invisible fields from the views. Keeping them become redundant in newest Odoo versions.


stock_lot_on_hand_first/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Stock Lot On Hand First",
+    "summary": "Allows to display lots on hand first in M2o fields",
+    "version": "18.0.1.0.0",
+    "development_status": "Alpha",
```
dreispt: Is this still Alpha and experimental?

grindtildeath: IMO it could have been promoted to Beta previously and could now even be considered stable

grindtildeath (approved):

stock_lot_on_hand_first/__manifest__.py
twalter-c2c: Based on @grindtildeath suggestion, I updated the `development_status` to stable. Thank you for the remark.

twalter-c2c:
Hi @OCA/logistics-maintainers, could you please take a look and possibly merge this one? 🙏🏻 

CC @jbaudoux @simahawk

stock_lot_on_hand_first/models/stock_lot.py
```diff
@@ -22,34 +22,27 @@ def _name_search(self, name, domain=None, operator="ilike", limit=None, order=No
...
                     order=order,
                 )
-                # _name_search is supposed to return a odoo.osv.query.Query object that
-                # will be evaluated as a list of ids when used in the browse function.
-                # Since we cannot merge the Query objects to respect the intended order
-                # in which we want to display the results,  we return the list of ids
-                # that will be used by browse in the name_search implementation.
-                return (
+                return self.browse(
                     self.browse(with_quantity_ids).ids
```
simahawk, line 38: why browsing twice if you have already ids? 🤔

simahawk (approved):
minor remark

stock_lot_on_hand_first/models/stock_lot.py
```diff
@@ -0,0 +1,48 @@
...
+    _inherit = "stock.lot"
+
+    @api.model
+    def search_fetch(self, domain, field_names, offset=0, limit=None, order=None):
+        """Move lots without a qty on hand at the end of the list"""
+
+        if self.env.context.get("name_search_qty_on_hand_first"):
+            domain = list(domain or [])
+
+            with_quantity_domain = AND([domain, [("product_qty", ">", 0)]])
```
jbaudoux, line 17: Not linked to the migration but this is a computed field that will process all the quants in valued locations from the DB to collect the list all the lots without limit.

stock_lot_on_hand_first/models/stock_lot.py
grindtildeath: IIRC `with_quantity_ids` and `without_quantity_ids` are both `odoo.tools.Query` objects, and there must have been a comment to explain that.

stock_lot_on_hand_first/models/stock_lot.py
```diff
@@ -22,34 +22,27 @@ def _name_search(self, name, domain=None, operator="ilike", limit=None, order=No
...
-                    operator=operator,
+                    offset=0,
                     limit=limit - with_quantity_count,
                     order=order,
                 )
-                # _name_search is supposed to return a odoo.osv.query.Query object that
-                # will be evaluated as a list of ids when used in the browse function.
-                # Since we cannot merge the Query objects to respect the intended order
-                # in which we want to display the results,  we return the list of ids
-                # that will be used by browse in the name_search implementation.
```
grindtildeath, line 43: IMO would be good to keep and adapt the comment as it definitely looks weird to browse twice

stock_lot_on_hand_first/models/stock_lot.py
```diff
@@ -0,0 +1,48 @@
...
+
+            with_quantity_domain = AND([domain, [("product_qty", ">", 0)]])
+            with_quantity_count = self.env["stock.lot"].search_count(
+                with_quantity_domain
+            )
+
+            if with_quantity_count >= limit:
+                domain = with_quantity_domain
+            else:
+                with_quantity_ids = self._search(
```
jbaudoux, line 25: should call `_search_fetch`

stock_lot_on_hand_first/models/stock_lot.py
```diff
@@ -0,0 +1,48 @@
...
+                    domain=with_quantity_domain,
+                    offset=offset,
+                    limit=limit,
+                    order=order,
+                )
+                without_quantity_ids = self._search(
```
jbaudoux, line 31: should call `_search_fetch`

stock_lot_on_hand_first/models/stock_lot.py
jbaudoux: you can just sum the results, no need to browse at all

jbaudoux (commented):
Not related to the migration, but what about doing this in name_search:

```
domain = [("lot_id.display_name", operator, name)]
if args:
  search lots and complete the domain with lot_id in ids

# apply limit on quants
lots_w_qty = self.env['stock.quant']._read_group(domain=domain, groupby=['lot_id'], aggregates=['quantity:sum'], having=[('quantity:sum', '!=', 0)], limit=limit)

# collect missing lots
# by calling super with a domain excluding lots_w_qty and for the remaining limit

# return the sum of the 2 results
```

PR #2258, opened 2026-02-20
[18.0][IMP] scrap_reason_code: hide native scrap_reason_tag_ids field

ThiagoMForgeFlow:
Hide native Odoo field to avoid functional overlap and keep module behavior authoritative. Documented in roadmap for future migration. Minor view adjustments included.

@ForgeFlow


JoanSForgeFlow (approved):
LGTM!

LoisRForgeFlow (approved):
Thanks for adding a roadmap. I think short term, this is the least we can do, hide the field to reduce confusion.

PR #2264, opened 2026-02-27
[18.0][FIX] stock_move_source_relocate: Merge back split moves

grindtildeath:
In case a move was split by the relocation rule, and the move that was reserved gets unreserved we were ending up with two different moves, on the same location (relocated) although it does not make sense to keep both move split, as we want to merge back the moves together.

### DIFF
--- a/stock_move_source_relocate/models/stock_move.py
+++ b/stock_move_source_relocate/models/stock_move.py
@@ -70,6 +70,7 @@ class StockMove(models.Model):
             # move on the specific relocation (for replenishment), so
             # update it's source location
             self.location_id = relocation.relocate_location_id
+            self._action_confirm(merge=True)
             return self
 
         missing_reserved_uom_quantity = self.product_uom_qty - qty_reserved
@@ -84,10 +85,11 @@ class StockMove(models.Model):
         # A part of the quantity could be reserved in the original
         # location, so keep this part in the move and split the rest
         # in a new move, where will take the goods in the relocation
-        new_move = self.create(self._split(need))
-        new_move._action_confirm(merge=False)
-        new_move.location_id = relocation.relocate_location_id
-        self._action_assign()
+        move_vals_list = self._split(need)
+        for move_vals in move_vals_list:
+            move_vals["location_id"] = relocation.relocate_location_id.id
+        new_move = self.create(move_vals_list)
+        new_move._action_confirm()
         return new_move
 
     def _after_apply_source_relocate_rule(self):
--- a/stock_move_source_relocate/models/stock_move.py
+++ b/stock_move_source_relocate/models/stock_move.py
@@ -89,8 +89,7 @@ class StockMove(models.Model):
         for move_vals in move_vals_list:
             move_vals["location_id"] = relocation.relocate_location_id.id
         new_move = self.create(move_vals_list)
-        new_move._action_confirm()
-        return new_move
+        return new_move._action_confirm()
 
     def _after_apply_source_relocate_rule(self):
         # Hook for stock_dynamic_routing
--- a/stock_move_source_relocate/tests/common.py
+++ b/stock_move_source_relocate/tests/common.py
@@ -38,7 +38,7 @@ class SourceRelocateCommon(common.TransactionCase):
             {"name": "Product2", "type": "consu", "is_storable": True}
         )
 
-    def _create_single_move(self, product, picking_type):
+    def _create_single_move(self, product, picking_type, custom_vals=None):
         move_vals = {
             "name": product.name,
             "picking_type_id": picking_type.id,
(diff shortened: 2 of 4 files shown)


stock_move_source_relocate/models/stock_move.py
jbaudoux: And here also in case of partial relocation
```suggestion
new_move.location_id = relocation.relocate_location_id
new_move._action_confirm(merge=True)
```

stock_move_source_relocate/tests/test_source_relocate.py
```diff
@@ -78,6 +78,13 @@ def test_relocate_partial_move(self):
                 }
             ],
         )
+        # Remove stock must remove the reservation and merge back the moves together
```
jbaudoux, line 79:
```suggestion
        # Remove stock must remove the reservation. On _action_assign, this move will also be relocated. Ensure it gets merged back with the previously relocated moves
```

stock_move_source_relocate/tests/test_source_relocate.py
```diff
@@ -78,6 +78,13 @@ def test_relocate_partial_move(self):
...
+        inventory_quant = (
+            self.env["stock.quant"].sudo()._gather(self.product, self.loc_shelf_1)
+        )
+        inventory_quant.inventory_quantity = 0
+        inventory_quant.action_apply_inventory()
+        self.assertEqual(len((move | new_move).exists()), 1)
```
jbaudoux, line 85: Maybe you need to explicitly call `move._action_assign()` here to trigger the relocation?

jbaudoux (changes requested):
You can add another test for the partial case by creating a move from `loc_replenish`.
The relocated `new_move` should be merged with this one.

stock_move_source_relocate/tests/common.py
```diff
@@ -38,7 +38,7 @@ def setUpClass(cls):
             {"name": "Product2", "type": "consu", "is_storable": True}
         )
 
-    def _create_single_move(self, product, picking_type):
+    def _create_single_move(self, product, picking_type, custom_vals=None):
```
jbaudoux, line 41: There is a change in odoo 18.0 that now requires all created moves to have the same `group_id` in order to be in the same picking
See one example fix on another repo https://github.com/OCA/stock-logistics-reservation/pull/48/changes/04b4ff3fa54b419aa79704b818bdedbf6bbcf51e

stock_move_source_relocate/tests/test_source_relocate.py
```diff
@@ -86,6 +86,25 @@ def test_relocate_partial_move(self):
...
+        move._action_assign()
+        new_move = move.picking_id.move_ids - move
+        extra_move = self._create_single_move(
+            self.product,
+            self.wh.pick_type_id,
+            custom_vals={
+                "location_id": self.loc_replenish.id,
+                "picking_id": new_move.picking_id.id,
+            },
+        )
```
jbaudoux: You need to create that extra move before the relocate occurs on `move._action_assign()` otherwise you are not testing the relocation merging but the standard odoo merge on `move._action_confirm()`

jbaudoux (changes requested):

stock_move_source_relocate/tests/test_source_relocate.py
grindtildeath: AFAIU if we create a move before the relocation occurs, such move would then have to be merged with the existing one on action_confirm (since it has to be on the same picking to be 'mergeable' afterwards, it is already 'mergeable' and would be merged then), or am I missing something ❓

stock_move_source_relocate/tests/test_source_relocate.py
grindtildeath: Application of inventory will call action_done that will trigger action confirm and action assign. On which move should we then call this?

stock_move_source_relocate/tests/common.py
grindtildeath: The link you provided is to force the procurement group on the stock.rule, but we don't use rules here 😕

stock_move_source_relocate/models/stock_move.py
grindtildeath: Why?

jbaudoux: Add a test for the partial case by creating a move from loc_replenish.
The relocated new_move should be merged with this one.
(note that's the initial issue)

stock_move_source_relocate/tests/common.py
jbaudoux: Here is another example https://github.com/OCA/delivery-carrier/pull/1130/changes#diff-5e6cd72a54aa9d3f1776868d6959ddc21d6a34387e58e2f1dcd27970cad1d9baR293

stock_move_source_relocate/tests/test_source_relocate.py
jbaudoux: if the test succeed, then it's merged and fine :)

stock_move_source_relocate/tests/test_source_relocate.py
jbaudoux: That extra move has a different location than the move, so it shouldn't be merge before the move is relocated to that same location

jbaudoux (approved):

jbaudoux:
cc @rousseldenis

rousseldenis (approved):

PR #2277, opened 2026-03-12
[18.0] Add module stock_picking_unlink_group

grindtildeath:


stock_picking_unlink_group/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module restricts deletion of stock transfer to a new security group.
```
TDu, line 1: Would be nice to explain, who is allowed to delete stock.picking at installation of the module, only admin it seems.

TDu (approved):

stock_picking_unlink_group/models/stock_picking.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "stock.picking"
+
+    def unlink(self):
+        if not self.env.user.has_group(
+            "stock_picking_unlink_group.group_stock_picking_unlink"
+        ):
+            raise AccessError(
+                self.env._("You are not allowed to delete stock transfers.")
+            )
+        return super().unlink()
```
jbaudoux, line 17: Why do you need this if you already manage it at ACL?
We should always allow to reconfigure ACL and let them play their role

jbaudoux (changes requested):

stock_picking_unlink_group/models/stock_picking.py
grindtildeath: IMO the name of the module and the name of the group are quite explicit in that they need to prevent any deletion of a picking, and modifying the ACL to allow picking deletion would just introduce incoherent data.

If the goal is to allow deactivating that feature without having to uninstall the module, it's easy to add the group to the implied groups of internal users...

simahawk: It would be nice to have res.config.setting flag linked to a group.

grindtildeath: @simahawk Done

PR #2280, opened 2026-03-18
[19.0][ADD] stock_lot_supplier

JasminSForgeFlow:
Adds a **Supplier** to lot/serial numbers.

When a receipt (incoming transfer) is validated, the supplier from the
Picking is automatically written to all tracked lots/serial numbers
received in that transfer (only if no supplier is already set).

The supplier field can be used to filter and group lots in the Lots/Serial
Numbers list view.


stock_lot_supplier/models/stock_lot.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class StockLot(models.Model):
+    _inherit = "stock.lot"
+
+    supplier_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Supplier",
+        index=True,
+        tracking=True,
```
RicardCForgeFlow, line 14:
```suggestion
        tracking=True,
        domain=[("supplier_rank", ">", 0)],
```
Hi @JasminSForgeFlow , would you agree on adding: `domain=[("supplier_rank", ">", 0)],`
So only suppliers can be added as supplier_id?

JasminSForgeFlow: Updated, Thanks.

PR #2281, opened 2026-03-18
[18.0][FIX] stock_picking_return_lot: Add domain to the lot_id field so that only the appropriate lots can be selected

victoralmau:
Add domain to the lot_id field so that only the appropriate lots can be selected

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT61559

### DIFF
--- a/stock_picking_return_lot/wizards/stock_return_picking_line.py
+++ b/stock_picking_return_lot/wizards/stock_return_picking_line.py
@@ -1,4 +1,5 @@
 # Copyright 2024 ACSONE SA/NV
+# Copyright 2026 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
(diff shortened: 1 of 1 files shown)


stock_picking_return_lot/wizards/stock_return_picking_line.py
```diff
@@ -17,12 +18,20 @@ class StockReturnPickingLine(models.TransientModel):
         )
     ]
 
+    domain_lot_id = fields.Binary(compute="_compute_domain_lot_id")
```
pedrobaeza: Better for readibility

```suggestion
    lot_id_domain = fields.Binary(compute="_compute_domain_lot_id")
```

stock_picking_return_lot/wizards/stock_return_picking_line.py
```diff
@@ -17,12 +18,20 @@ class StockReturnPickingLine(models.TransientModel):
...
     )
 
+    @api.depends("move_id")
+    def _compute_lot_id_domain(self):
+        for rec in self:
+            smls = rec.move_id.move_line_ids.filtered(
+                lambda x: x.state == "done" and x.lot_id
+            )
+            domain = [("id", "in", smls.lot_id.ids)]
+            rec.lot_id_domain = domain
```
carlos-lopez-tecnativa, line 33: Why is it so restrictive?
The same as my comment here:  https://github.com/OCA/rma/pull/554#discussion_r2962116400

victoralmau: Similar to what was discussed in the other PR, IMO you should only be able to select lots that have been used; what would be the point of creating a return for a lot that hasn't been used?

PR #2282, opened 2026-03-18
[18.0][FIX] stock_picking_batch_creation: splitting

jbaudoux:
1/ When splitting is enabled, an over limit picking can be elected as first picking. In that case, the last device should be used.
When this occurred, no batch could ever be created as the first picking to process was elected but no suitable device was returned.

2/ An over limit picking must be excluded only when splitting is disabled.
If the splitting is enabled, the over limit picking will be split and elected

3/  When a trying to split a picking with a single move, there will be no split order as there is only one move. In this case, the picking must be used.

cc @TDu @mmequignon @lmignon @phschmidt


lmignon (approved):
It makes sense.

lmignon (approved):
Code review + functional tests

scenario:

2 devices
* device1 max 10 kg sequence 1
* device2 max 20 kg sequence 2
 
1 picking with 35 chairs for a total weight of 350 kg

Results:
-> Launch of the wizard without the split option -> no batch created 👍 
-> Launch of the wizard with the split option -> a batch is created for the picking and the device2 👍 

Common usages still work as expected 👍

PR #2284, opened 2026-03-19
[FIX] product_customerinfo_picking: support variant-level customer info

weinni2000:
## Problem

The `_compute_product_customer_code` method on `stock.move` only looked up
customer info at the **product template** level (`product_tmpl_id.customer_ids`).
It ignored `product.customerinfo` records linked to a specific **product variant**
(`variant_customer_ids`), so the customer reference/name was never filled for
variant-specific entries.

Additionally, the guard condition checked `product_tmpl_id.customer_ids` while
the `@api.depends` tracked `product_id.customer_ids`, causing a mismatch.

## Fix

- Guard condition: check `product_id.customer_ids` (consistent with `@api.depends`)
- Template-level lookup: add `product_tmpl_id` filter to avoid cross-template matches
- Variant-level lookup: after the template-level result, look up
  `variant_customer_ids` filtered by partner **and** product variant; if found,
  override the template-level values so variant-specific info takes precedence

## Test scenario

1. Create a product with variants (e.g. colour A, colour B)
2. Add a customer reference on the template level and a different one on a
   specific variant
3. Create a delivery order for that variant
4. Verify the stock move shows the **variant-level** customer reference,
   not the template-level one

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/product_customerinfo_picking/models/stock_move.py
+++ b/product_customerinfo_picking/models/stock_move.py
@@ -6,6 +6,13 @@ from odoo import api, fields, models
 class StockMove(models.Model):
     _inherit = "stock.move"
 
+    product_customer_code = fields.Char(
+        compute="_compute_product_customer_code",
+    )
+    product_customer_name = fields.Char(
+        compute="_compute_product_customer_code",
+    )
+
     @api.depends(
         "picking_id.partner_id",
         "product_id",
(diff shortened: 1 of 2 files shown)


weinni2000:
@it-fact4lifetimesupport:
Can you review it?

JordiMForgeFlow:
@weinni2000  I have opened PR https://github.com/OCA/stock-logistics-workflow/pull/2413 where I actually use the search method from the base product_customerinfo module. This already considers the variants.

PR #1591, opened 2024-05-13
[14.0][FIX] stock_no_negative: preventing data inconsistencies with the allow negative stock

FrankC013:
This implementation prevents the deactivation of the allow_negative_stock field if any of the fields related to the record (product, category, location) have negative quantities in their current stock operations. Additionally, the product category cannot be modified if this causes inconsistencies with the allow_negative_stock option, that is, if the negative quantities in the quants are not associated with the product or location, but exclusively with the category. Lastly, the raise message has been modified to be more generic, along with its translations.

@eantones Please review

PR #2289, opened 2026-03-26
[18.0][FIX] stock_no_negative: preventing data inconsistencies with the all…

deeniiz:
…ow negative stock

Migration of this PR #1591


deeniiz:
@eantones, can you please review it?

rousseldenis:
@deeniiz I'm not sure about this.

This change will avoid to enable the restriction if there are negative quantities. IMHO, you should be able to do it as you want to empeach new negative quantities usually quite quickly. The household management to remove those quantities can take time.

PR #2290, opened 2026-03-30
[18.0][FIX] stock_dynamic_routing: split move

jbaudoux:
When an already confirmed move is split, do not call `_action_confirm` on the new move. Since odoo v15.0, `_action_confirm` may call `_action_assign`. Also we don't need any procurement creation, assign picking, neg move transformation. I think this approach is better

### DIFF
--- a/stock_dynamic_routing/models/stock_move.py
+++ b/stock_dynamic_routing/models/stock_move.py
@@ -230,7 +230,7 @@ class StockMove(models.Model):
                 # NOTE: starting from Odoo 18.0, '_split' method doesn't check
                 # anymore the move quantity against the qty to split, and performs
                 # a split in all cases, letting an empty move behind. Avoid this.
-                new_move_vals = []
+                new_move_vals_list = []
                 if (
                     float_compare(
                         move.product_qty,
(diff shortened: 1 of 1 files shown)


stock_dynamic_routing/models/stock_move.py
```diff
@@ -239,10 +239,16 @@ def _routing_splits(self, moves_routing):
                     )
                     == 1
                 ):
-                    new_move_vals = move._split(qty)
-                if new_move_vals:
-                    new_move = self.env["stock.move"].create(new_move_vals)
-                    new_move._action_confirm(merge=False)
+                    new_move_vals_list = move._split(qty)
```
mmequignon: why not use a ctx key, and override copy_data to override defaults with `state=move.state, reservation_date=move.reservation_date` when ctx key is set ?

mmequignon (approved):
LG, small no-blocking remark

grindtildeath (approved):
LG 👍

jbaudoux:
rebased

jbaudoux:
it was badly rebased. Fixed now

PR #2291, opened 2026-03-30
[18.0][FIX] stock_move_source_relocate: action_confirm

jbaudoux:
* When a confirmed move is split and/or source location is changed, do not call _action_confirm
* When the unassigned move is relocated in the same picking as the assigned move, merge back the assigned move into the relocated moves. Ensure the current move does not disappear as we are inside _action_assign

### DIFF
--- a/stock_move_source_relocate/models/stock_move.py
+++ b/stock_move_source_relocate/models/stock_move.py
@@ -70,7 +70,7 @@ class StockMove(models.Model):
             # move on the specific relocation (for replenishment), so
             # update it's source location
             self.location_id = relocation.relocate_location_id
-            self._action_confirm(merge=True)
+            # Do not call _action_confirm on a split moves inside _action_assign
             return self
 
         missing_reserved_uom_quantity = self.product_uom_qty - qty_reserved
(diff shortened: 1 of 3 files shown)


stock_move_source_relocate/models/stock_move.py
```diff
@@ -88,9 +88,17 @@ def _apply_source_relocate_rule(self, relocation, quantity, roundings):
         move_vals_list = self._split(need)
```
mmequignon, line 88: same as in https://github.com/OCA/stock-logistics-workflow/pull/2290#discussion_r3044671376
set context key, catch it in `copy_vals` add `state=move.state, reservation_date=move.reservation_date` in default.

mmequignon (approved):
LG, but again, code duplication. Please add a `MOVE_SPLIT_COPY_STATE` key or smth

grindtildeath (approved):
LG 👍

stock_move_source_relocate/models/stock_move.py
jbaudoux: it's before create, it's fine to complete the dict like that.

PR #2294, opened 2026-04-02
[16.0][FIX] stock_move_line_serial_unique: Exclude internal moves from check

AungKoKoLin1997:
[FIX] stock_move_line_serial_unique: Exclude internal moves from validation

The serial number uniqueness constraint was incorrectly blocking valid internal stock moves where existing serial numbers need to be moved between internal locations.

The constraint now uses location-based logic:
- CHECK: Moves from external → internal (receipts, manufacturing)
- SKIP: Moves from internal → anywhere (consumption, internal transfers)

This prevents duplicate serial creation on receipts while allowing normal internal moves with existing serial numbers.

@qrtl QT6552

### DIFF
--- a/stock_move_line_serial_unique/models/stock_move_line.py
+++ b/stock_move_line_serial_unique/models/stock_move_line.py
@@ -22,6 +22,12 @@ class StockMoveLine(models.Model):
                 and record.product_id.tracking == "serial"
                 # Check should be done only in the scenario where new serial is created.
                 and record.picking_type_id.use_create_lots
+                # Only check for moves from non-internal locations (e.g. receipts,
+                # production) to internal locations.
+                # Without this, component flushes in subcontracting receipts would
+                # incorrectly trigger the serial uniqueness check.
+                and record.location_id.usage != "internal"
+                and record.location_dest_id.usage == "internal"
             ):
                 lot_id = record.lot_id
                 if not lot_id:
--- a/stock_move_line_serial_unique/tests/test_stock_move_line_serial_unique.py
+++ b/stock_move_line_serial_unique/tests/test_stock_move_line_serial_unique.py
@@ -104,3 +104,26 @@ class TestStockQuantSerialUnique(TransactionCase):
         moveline.write({"lot_id": self.serial1.id})
         picking_out.action_confirm()
         picking_out.button_validate()
+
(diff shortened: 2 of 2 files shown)


nobuQuartile (approved):
Codre review and functional review: LGTM

stock_move_line_serial_unique/models/stock_move_line.py
```diff
@@ -22,6 +22,10 @@ def _check_serial_number(self):
                 and record.product_id.tracking == "serial"
                 # Check should be done only in the scenario where new serial is created.
                 and record.picking_type_id.use_create_lots
+                # Only check for moves bringing stock IN (destination is internal)
+                and record.location_dest_id.usage == "internal"
+                # Only check for moves from non-internal locations (receipts/production)
+                and record.location_id.usage != "internal"
```
yostashiro: To better clarify the rationale.
```suggestion
                # Only check for moves from non-internal locations (e.g. receipts,
                # production) to internal locations.
                # Without this, component flushes in subcontracting receipts would
                # incorrectly trigger the serial uniqueness check.
                and record.location_id.usage != "internal"
                and record.location_dest_id.usage == "internal"
```

stock_move_line_serial_unique/models/stock_move_line.py
```diff
@@ -32,10 +36,11 @@ def _check_serial_number(self):
...
                         ]
                     )
-                message, dummy = self.env["stock.quant"]._check_serial_number(
-                    record.product_id,
-                    lot_id,
-                    record.company_id,
-                )
-                if message:
-                    raise ValidationError(_(message))
+                if lot_id:
```
yostashiro: I think change can be reverted to make this PR more focused.

stock_move_line_serial_unique/models/stock_move_line.py
AungKoKoLin1997: Updated!

stock_move_line_serial_unique/models/stock_move_line.py
AungKoKoLin1997: Reverted!

yostashiro (approved):
LGTM.

PR #2296, opened 2026-04-03
[18.0][IMP] stock_move_source_relocate: cleanup code

jbaudoux:
cc @grindtildeath @mmequignon @rousseldenis

PR #2297, opened 2026-04-03
[18.0][FIX] stock_move_source_relocate_dynamic_routing: merge moves & perf

jbaudoux:
## stock_dynamic_routing: _action_assign
    
Do not call _action_assign inside _action_assign but use super().
Add hook for relocate before updating next moves.

## stock_move_source_relocate_dynamic_routing:

Use new hooks to ensure relocate is called just after the dynamic routing on the picking and before the dynamic routing on the next moves.  This allows moves to be merged back and prevents chained moves to be split.

Depends on:
* https://github.com/OCA/stock-logistics-workflow/pull/2291

Relates also to:
* https://github.com/OCA/stock-logistics-workflow/pull/2290
* https://github.com/OCA/stock-logistics-workflow/pull/2296

Other related perf improvements:
* https://github.com/OCA/stock-logistics-workflow/pull/2298 
* https://github.com/OCA/stock-logistics-reservation/pull/52

### DIFF
--- a/stock_dynamic_routing/models/stock_move.py
+++ b/stock_dynamic_routing/models/stock_move.py
@@ -7,7 +7,7 @@ from collections import OrderedDict, defaultdict, namedtuple
 
 from psycopg2 import sql
 
-from odoo import models
+from odoo import api, models
 from odoo.tools.float_utils import float_compare
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


stock_dynamic_routing/models/stock_move.py
```diff
@@ -58,18 +58,16 @@ def _chain_apply_routing(self):
...
+            super()._action_assign(force_qty=force_qty)
         else:
-            # these methods will call _action_assign in a savepoint
-            # and modify the routing if necessary
-            moves = self._split_and_apply_routing()
-            return super(StockMove, moves)._action_assign()
+            # These methods will call _action_assign in a savepoint and modify
+            # the routing if necessary. Call to super is done in the method.
+            self._split_and_apply_routing()
+        return
```
grindtildeath, line 70: No need to return here

stock_move_source_relocate_dynamic_routing/models/stock_move.py
```diff
@@ -11,21 +11,14 @@
...
-        result = self._chain_apply_routing()
-        _logger.debug("Dynamic routing applied on relocated moves %s", self.ids)
-        return result
+    def _after_apply_source_relocate_rule(self, merge=True):
+        # Apply dynamic routing on relocated move to potentially route it in
+        # another picking
+        _logger.debug(f"Apply dynamic routing on relocated moves {self}")
+        self._chain_apply_routing()
+        super()._after_apply_source_relocate_rule(merge=merge)
+        return
```
grindtildeath, line 24: Same, bare return not needed at the end of a function

grindtildeath (approved):
I don't have enough knowledge of the module to tell if it's all good, but I guess it is fine if the tests don't break 😅

mmequignon (approved):
Hard to judge, but looks ok

PR #2298, opened 2026-04-03
[18.0][IMP] stock_dynamic_routing: remove logged message

jbaudoux:
Do not log that the initial demand has been updated when the move is split. The demand is still the same, it's just split in 2 moves.

This also increases the performance as logging that message is quite slow.

### DIFF
--- a/stock_dynamic_routing/models/stock_move_line.py
+++ b/stock_dynamic_routing/models/stock_move_line.py
@@ -1,0 +1,14 @@
+# Copyright 2026 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import models
+
+
+class StockMoveLine(models.Model):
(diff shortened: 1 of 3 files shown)


stock_dynamic_routing/models/stock_move_line.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class StockMoveLine(models.Model):
+    _inherit = "stock.move.line"
+
+    def _log_message(self, record, move, template, vals):
+        if self.env.context.get("bypass_log_message"):
+            return
+        super()._log_message(record, move, template, vals)
+        return
```
grindtildeath, line 14: Again, no need for bare returns at the end of a function

grindtildeath (approved):

mmequignon (approved):

stock_dynamic_routing/models/stock_move_line.py
jbaudoux: pre commit complains when there is no return in a method calling super()

PR #2308, opened 2026-04-13
[19.0][MIG] stock_picking_show_return: Migration to 19.0

Andrii9090-tecnativa:
Migration to 19.0
 
image
image

@pedrobaeza @juancarlosonate-tecnativa can you review please?

 TT61976
 @Tecnativa


stock_picking_show_return/models/stock_picking.py
```diff
@@ -28,8 +27,9 @@ def _compute_returned_ids(self):
     def _compute_source_picking_id(self):
         """Get source piking from this picking. Only one origin is possible."""
         for picking in self:
-            picking.source_picking_id = first(
-                picking.move_ids.origin_returned_move_id.picking_id
+            first_move = next(iter(picking.move_ids), False)
+            picking.source_picking_id = (
+                first_move.origin_returned_move_id.picking_id if first_move else False
```
pedrobaeza:
```suggestion
            picking.source_picking_id = (
                picking.move_ids.origin_returned_move_id.picking_id[:1].id

pedrobaeza (approved):
/ocabot migration stock_picking_show_return

christian-ramos-tecnativa (approved):
LGTM

juancarlosonate-tecnativa (commented):
Taking a quick look at it, in my opinion this module no longer makes much sense.

We already have the returns smart button, which essentially provides the same functionality as the tab, it shows multiple records if they exist, or redirects directly when there is only one.

Additionally, to navigate from the return to the original picking, we already have the return_id field in the core (even though it’s not currently shown in the view, it could easily be added), and the original picking is also reflected in the chatter.

pedrobaeza:
Yeah, in that case, indeed this one can be omitted and use the core features. Thanks for checking that. A merge should be added in OpenUpgrade.

PR #2316, opened 2026-04-22
[18.0][ADD] product_stock_all_companies

IriaAlonso:
Show the stock of a product in all the warehouses, even if they are from other companies. If the product has more than one variant, the stock will only be shown in the variants of the product.


IriaAlonso:
@Saran440 @divad1196 Can you review it please?

product_stock_all_companies/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+    def _compute_stock_by_company_ids(self):
+        companies = self.env["res.company"].search([])
+
+        for product in self:
+            result = []
+            for company in companies:
+                product_in_ctx = product.with_context(allowed_company_ids=[company.id])
+                qty = product_in_ctx.qty_available
+
+                result.append(
```
divad1196: Prefer to use the `Command.create`: https://www.odoo.com/documentation/18.0/developer/reference/backend/orm.html#odoo.fields.Command

product_stock_all_companies/models/product_template.py
```diff
@@ -0,0 +1,23 @@
...
+
+    stock_by_company_ids = fields.Many2many(
+        "stock.company",
+        compute="_compute_stock_by_company_template",
+    )
+
+    @api.depends("product_variant_ids.stock_by_company_ids", "product_variant_count")
+    def _compute_stock_by_company_template(self):
+        for template in self:
+            if template.product_variant_count <= 1 and template.type == "consu":
```
divad1196, line 18: Isn't the condition inverted ? Shouldn't `stock_by_company_ids` be False if of type `consu` ?

product_stock_all_companies/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+                    (
+                        0,
+                        0,
```
divad1196: Do you really want to just keep creating? Every time the function runs it will create more `stock.company`, I don't see how the result will be right after the second run.

I feel like the compute should be on `stock.company` instead of the `product.product` for this reason.

divad1196 (changes requested):
@IriaAlonso The module description isn't exact on what the module is doing. There are a few thing that are either buggy or I don't understand: Could you clarify them ?

product_stock_all_companies/models/product_template.py
IriaAlonso: No, because it wouldn't be calculated for any product; `consu` are the only type of product that has stock.

divad1196: Sorry, I don't know why I wrote that.

I think I meant `product_variant_count <= 1` with the consu condition. Shouldn't it be `count >= 1`? Wouldn't it make sense to sum the variants' stock?

IriaAlonso: If a product has more than one variant, the stock will only be displayed for those variants and not for the product itself

IriaAlonso:
@divad1196 done

product_stock_all_companies/models/stock_company.py
```diff
@@ -0,0 +1,38 @@
...
+
+            record = self.search(
+                [
+                    ("product_id", "=", product.id),
+                    ("company_id", "=", company.id),
+                ],
+                limit=1,
+            )
+
+            if record:
```
divad1196, line 29: Works

I would recommend to use [guard clause](https://refactoring.guru/fr/replace-nested-conditional-with-guard-clauses)/pattern or the "[never nesting](https://medium.com/@collinsakuma/never-nesting-bd8c0f2a9ca0)" approach
```python
if record:
    record.quantity_available = qty
    continue

self.create(...)
```

This reduce [cognitive complexity](https://www.baeldung.com/java-cognitive-complexity#2-cognitive-complexity): easier to read/understand and less error-prone.

product_stock_all_companies/models/product.py
```diff
@@ -0,0 +1,22 @@
+# © 2026 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import api, fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    stock_by_company_ids = fields.Many2many(
```
divad1196, line 10: Why make it a `Many2many` ?
It's a `One2many` reverse field on `product_id = fields.Many2one("product.product")` of `stock.company`.

You would be able to remove the computed field.
There are other ways to trigger `_calculate_stock_for_product`.

product_stock_all_companies/tests/test_product_stock_all_companies.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def test_calculate_creates_stock_company_records(self):
+        """_calculate_stock_for_product creates record on
+        stock.company for each company."""
+        StockCompany = self.env["stock.company"].sudo()
+        StockCompany._calculate_stock_for_product(self.product)
+
+        companies = self.env["res.company"].search([])
+        for company in companies:
+            rec = StockCompany.search(
```
divad1196, line 32: Seeing this snippet for the second time, it looks like `_gather` on `stock.quant`.
I think an utility function would help maintainance.

product_stock_all_companies/models/stock_company.py
```diff
@@ -0,0 +1,38 @@
+# © 2026 Solvos Consultoría Informática (<http://www.solvos.es>)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import fields, models
+
+
+class StockCompany(models.Model):
+    _name = "stock.company"
```
divad1196, line 8: Considering that it's a simplified `stock.quant` scoped by company, I would name it `company.stock.quant`

divad1196 (commented):
@IriaAlonso Just the computed `Many2many` thing to review. The rest is just recommendation.

FYI:
`stock.company` is a simplified `stock.quant` scoped by company.
By simplified, I mean that it does not have lot, location, reservation, ...

It does its job, but beware  the current design won't evolve easily laterif you want something more like actual `stock.quant`.

PR #2323, opened 2026-04-29
[18.0][FIX] stock_picking_tier_validation: Fix singleton error in batch transfer validation

fsmw:
Fix singleton error when validating batch transfers with tier validation enabled.

**Root Cause**: Using `self.validation_status` instead of `rec.validation_status` inside a loop iterating over individual records causes a ValueError when multiple pickings are validated together.

**Fix**: Change `self.validation_status` to `rec.validation_status` in the `button_validate` method.

**Before**:
```python
if self.validation_status != "validated":
```

**After**:
```python
if rec.validation_status != "validated":
```

### DIFF
--- a/stock_picking_tier_validation/models/stock_picking.py
+++ b/stock_picking_tier_validation/models/stock_picking.py
@@ -19,7 +19,7 @@ class StockPicking(models.Model):
                 # try to validate operation
                 reviews = rec.request_validation()
                 rec._validate_tier(reviews)
-                if self.validation_status != "validated":
+                if rec.validation_status != "validated":
                     raise ValidationError(
                         _(
                             "This action needs to be validated for at least "


StefanRijnhart (changes requested):
Please remove the version number from the commit message. The version number should only be added to the PR title.

StefanRijnhart (approved):

nurinayati:
I think the native functionality is missing. It should show a warning requiring an approval. when the stock transfer isnt approved yet. Could you please check?

Before update
https://www.awesomescreenshot.com/video/52096682?key=ac1dc8cb7c0de66c0716cbe6eab086a5

after update
https://www.awesomescreenshot.com/video/52096602?key=8631959c429fa5ee3fdc6d1f84e10003

pedrobaeza (approved):
/ocabot merge patch

PR #2324, opened 2026-04-29
[18.0][FIX] stock_picking_group_by_partner_by_carrier: Fix _create_backorder serialization

fsmw:
Fix issue where batch backorder creation causes incorrect detachment.

**Root Cause**: The `_create_backorder` method calls super per picking, which serializes the batch and prevents proper batch-level handling.

**Fix**: Call super once with the full recordset using a context dictionary to track per-picking grouping flags instead of per-picking serialization.

**Changes**:
- Modify `_create_backorder` to compute grouping flags upfront and call super once
- Update `copy` method to use context dict keyed by picking ID
- Apply grouping (merge\_procurement\_groups, update\_merged\_origin) to backorder results after creation

### DIFF
--- a/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
+++ b/stock_picking_group_by_partner_by_carrier/models/stock_picking.py
@@ -86,22 +86,26 @@ class StockPicking(models.Model):
             return super().action_cancel()
 
     def _create_backorder(self, backorder_moves=None):
-        backorders = self.browse()
+        grouping_disabled = {}
         for picking in self:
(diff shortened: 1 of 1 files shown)


fsmw:
Closing this PR as the approach to batch _create_backorder has proven to be too complex and breaks 7 existing tests. The original picking reference approach via move_orig_ids was wrong, and the backorder_id approach also fails in merge scenarios. A simpler fix would require deeper architectural changes to how grouping happens during backorder creation. Re-opening with a different approach if needed.

PR #2325, opened 2026-04-30
[18.0][FIX] product_customerinfo_picking: fallback to commercial partner

### LINKED ISSUE #2272: product_customerinfo_picking - partner selection in _compute_product_customer_code
## Module
product_customerinfo_picking

## Describe the bug
The function _compute_product_customer_code should search for the customer specific code to use for the picking. This will only work if the delivery address is the res.partner record with the customer product codes. As soon as you have for example a delivery address under a res.partner in use - where the customer codes are on the parent_id it will fail

**Expected behavior**
It should search on direct res.partner - then fallback to commercial_partner_id

fsmw:
When the delivery address (`picking_id.partner_id`) is a child contact of the customer, the customer product codes are stored on the `commercial_partner_id` (parent). Since the code only matched against the direct partner, it would not find the customer-specific codes.

**Fix**: After an initial direct match fails, fall back to `partner_id.commercial_partner_id` to find the customer product code/name.

Also added `picking_id.partner_id.commercial_partner_id` to `@api.depends` so the compute triggers when the commercial partner changes.

Closes #2272

### DIFF
--- a/product_customerinfo_picking/models/stock_move.py
+++ b/product_customerinfo_picking/models/stock_move.py
@@ -8,6 +8,7 @@ class StockMove(models.Model):
 
     @api.depends(
         "picking_id.partner_id",
+        "picking_id.partner_id.commercial_partner_id",
         "product_id",
         "product_id.customer_ids.product_code",
         "product_id.customer_ids.product_name",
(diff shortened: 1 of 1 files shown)


BhaveshHeliconia (approved):
LGTM!

product_customerinfo_picking/models/stock_move.py
```diff
@@ -8,6 +8,7 @@ class StockMove(models.Model):
 
     @api.depends(
         "picking_id.partner_id",
+        "picking_id.partner_id.commercial_partner_id",
```
rousseldenis, line 11: @fsmw Don't do that. If you change (for any reason) the commercial partner of a contact, all linked stock moves will be recomputed (which can be harsh on big databases).

JordiMForgeFlow:
@fsmw @rousseldenis I have opened PR https://github.com/OCA/stock-logistics-workflow/pull/2413 where I actually use the search method from the base product_customerinfo module. This already considers the commercial partner.

PR #2326, opened 2026-04-30
[17.0][ADD] stock_picking_origin_state

AndreuOForgeFlow:
This module adds two computed fields on `stock.picking` that summarize the
state of the most upstream not-done origin pickings:

- `origin_state`: aggregated state across the origin pickings at the deepest
  level still active (`waiting`, `partially_available`, `assigned` or `done`).
- `origin_state_label`: a display string combining the origin operation type
  name and the translated state label


stock_picking_origin_state/models/stock_picking.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        Return the partial-aware state of a single picking.
+
+        - assigned when every active move is fully reserved (assigned or
+          done).
+        - partially_available when at least one move is reserved
+          (assigned, done or partially_available) and some other
+          move is not.
+        - waiting otherwise (no reservation yet).
+        """
```
GuillemCForgeFlow, line 89:
```suggestion
```
IHMO, not needed comments, this is already what's happening in odoo's core

stock_picking_predecessor_state/models/stock_picking.py
```diff
@@ -0,0 +1,116 @@
...
+        seen_ids = set(self.ids)
+        frontier = self
+        while True:
+            origin = frontier.mapped("move_ids.move_orig_ids.picking_id")
+            origin = origin.filtered(lambda p: p.id not in seen_ids)
+            if not origin:
+                break
+            seen_ids.update(origin.ids)
+            levels.append(origin)
+            frontier = origin
```
GuillemCForgeFlow:
```suggestion
            picking_orig_ids = frontier.mapped("move_ids.move_orig_ids.picking_id")
            picking_orig_ids = picking_orig_ids.filtered(lambda p: p.id not in seen_ids)
            if not picking_orig_ids:
                break
            seen_ids.update(picking_orig_ids.ids)
            levels.append(picking_orig_ids)
            frontier = picking_orig_ids
```
it is misleading to have a variable called `origin` when it can contain several `stock.picking`s

stock_picking_predecessor_state/models/stock_picking.py
```diff
@@ -0,0 +1,116 @@
...
+        for level in reversed(levels):
+            active = level.filtered(lambda p: p.state not in ("done", "cancel"))
```
GuillemCForgeFlow: same here for the active, there can be multiple transfers

stock_picking_predecessor_state/models/stock_picking.py
```diff
@@ -0,0 +1,116 @@
...
+                return active, self._aggregate_predecessor_state(active)
+        if not levels:
+            return Picking, None
+        done = levels[0].filtered(lambda p: p.state == "done")
+        if done:
+            return done, "done"
+        return Picking, None
+
+    @api.model
+    def _aggregate_predecessor_state(self, pickings):
```
GuillemCForgeFlow:
```suggestion
    def _aggregate_predecessor_state(self):
```
maybe it's better to receive the pickings in self, more clear that way

stock_picking_predecessor_state/models/stock_picking.py
```diff
@@ -0,0 +1,116 @@
...
+        "move_ids.move_orig_ids.picking_id.move_ids.state",
+    )
+    def _compute_predecessor_state(self):
+        for picking in self:
+            level, state = picking._get_predecessor_level()
+            if not level:
+                picking.predecessor_state = False
+                picking.predecessor_picking_type_id = False
+                continue
+            picking_type = level.picking_type_id[:1]
```
GuillemCForgeFlow: why the `[:1]`?

stock_picking_predecessor_state/models/stock_picking.py
```diff
@@ -0,0 +1,116 @@
...
+        worst_index = len(PREDECESSOR_STATE_ORDER)
+        for picking in pickings:
+            atomic = picking._get_atomic_predecessor_state()
+            if atomic in PREDECESSOR_STATE_ORDER:
+                idx = PREDECESSOR_STATE_ORDER.index(atomic)
+                if idx < worst_index:
+                    worst_index = idx
+        if worst_index == len(PREDECESSOR_STATE_ORDER):
+            return "waiting"
+        return PREDECESSOR_STATE_ORDER[worst_index]
```
GuillemCForgeFlow: I don't understand why you need to have the `_aggregate_predecessor_state` and `_get_atomic_predecessor_state` methods. We are hardcoding the move states in a static variable, the ideal would be to already use the states you have in the `state` field for the `stock.move`. They are already declared in the order you are interested in:
```
state = fields.Selection([
        ('draft', 'New'),
        ('waiting', 'Waiting Another Move'),
        ('confirmed', 'Waiting Availability'),
        ('partially_available', 'Partially Available'),
        ('assigned', 'Available'),
        ('done', 'Done'),
        ('cancel', 'Cancelled')]
```

stock_picking_predecessor_state/views/stock_picking_views.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='products_availability']" position="after">
+                <label for="predecessor_state" invisible="not predecessor_state" />
+                <div class="o_row" invisible="not predecessor_state">
+                    <field name="predecessor_picking_type_id" nolabel="1" />
+                    <field name="predecessor_state" nolabel="1" />
+                </div>
+            </xpath>
+        </field>
+    </record>
```
GuillemCForgeFlow: I'd add it to the base tree view instead with optional show next to the base state

Image

GuillemCForgeFlow (changes requested):

stock_picking_predecessor_state/models/stock_picking.py
AndreuOForgeFlow: We can't assign to `predecessor_picking_type_id` a recordset because it is a Many2One. Since all pickings on the chain share the same type we take the first one

stock_picking_predecessor_state/models/stock_picking.py
```diff
@@ -0,0 +1,99 @@
...
+
+    predecessor_state = fields.Selection(
+        selection=[
+            ("waiting", "Waiting"),
+            ("partially_available", "Partially Available"),
+            ("assigned", "Ready"),
+            ("done", "Done"),
+        ],
+        compute="_compute_predecessor_state",
+        compute_sudo=True,
```
GuillemCForgeFlow: is the `compute_sudo` needed? why?

stock_picking_predecessor_state/models/stock_picking.py
GuillemCForgeFlow: then the [:1] should be on the level, not on the `picking_type_id`. that may come also from the variable name `level`, which is not clear to me. it can contain various pickings

stock_picking_predecessor_state/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,11 @@
...
+state of the most upstream not-done predecessor pickings:
+
+- `predecessor_state`: aggregated state across the predecessors at the deepest
+  level still active (`waiting`, `partially_available` or `assigned`).
+- `predecessor_picking_type_id`: the operation type of those predecessors.
+
+Multi-step transfers (for example a 3-step delivery: `PICK -> PACK -> OUT`)
+benefit from seeing on the last picking what the upstream pickings are still
+doing, without having to navigate to them. When several predecessors share the
+same level, the least favorable state is reported.
```
GuillemCForgeFlow: I'd move this part to the CONTEXT.md file, as I understand that this module only makes sense when you have the possibility to have linked transfers, this can tipycally come from two scenarios:
- multiple step outbound / inbound configured in the wh
- linked moves via route / rules logics

stock_picking_predecessor_state/views/stock_picking_views.xml
```diff
@@ -0,0 +1,30 @@
...
+                    readonly="1"
+                />
+                <field
+                    name="predecessor_state"
+                    optional="show"
+                    readonly="1"
+                    widget="badge"
+                    decoration-warning="predecessor_state in ('waiting', 'partially_available')"
+                    decoration-success="predecessor_state == 'assigned'"
+                    decoration-muted="predecessor_state == 'done'"
```
GuillemCForgeFlow: maybe we should follow what's declared in odoo?

```
<field name="state" optional="show" widget="badge"
                           decoration-danger="state=='cancel'"
                           decoration-info="state== 'assigned'"
                           decoration-muted="state == 'draft'"
                           decoration-success="state == 'done'"
                           decoration-warning="state not in ('draft','cancel','done','assigned')"/>
```

stock_picking_predecessor_state/models/stock_picking.py
AndreuOForgeFlow: I had it initially because of user permits. But it isn't need it. I took them out

stock_picking_predecessor_state/readme/DESCRIPTION.md
AndreuOForgeFlow: Moved to CONTEXT.md

stock_picking_predecessor_state/views/stock_picking_views.xml
AndreuOForgeFlow: You're right. Now I'm using the same decoration color badges using the same logic for the predecessors

stock_picking_predecessor_state/models/stock_picking.py
AndreuOForgeFlow: Good point, now I changed the name of the variable to `predecessor_pickings` to be more clear. The first one is now selected on that level field and not `picking_type_id`

GuillemCForgeFlow (approved):
lgtm 👍🏿 code and functional review

JordiMForgeFlow:
@AndreuOForgeFlow @GuillemCForgeFlow this one makes me think about https://github.com/OCA/stock-logistics-warehouse/blob/18.0/stock_picking_show_linked. I wonder if we could add this as a feature in that module, or if we should at least align the wording (predecessor picking vs origin picking)

AndreuOForgeFlow:
> @AndreuOForgeFlow @GuillemCForgeFlow this one makes me think about https://github.com/OCA/stock-logistics-warehouse/blob/18.0/stock_picking_show_linked. I wonder if we could add this as a feature in that module, or if we should at least align the wording (predecessor picking vs origin picking)

@JordiMForgeFlow, thanks for the feedback. I think I'd rather keep it in a separate module because they serve different functionalities (navigation buttons for direct origin/destination pickings vs aggregated state of the upstream chains of stock pickings). I'll align the wording as well as the name of the module.

AndreuOForgeFlow:
Hi @JordiMForgeFlow, I just renamed the wording to align with `stock_picking_origin_state` :+1:

JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
/ocabot merge nobump

PR #2331, opened 2026-05-07
[Fix] [19.0] sale_stock_partner_delivery_window: warning allways trigger

cyrilmanuel:
The warning trigger by the method _onchange_commitment_date_delivery_window return always a true condition because the next_date is returned as partner date tz defined. that mean we pass the commitment_date but we didn't return them if the delivery_time_preference is good. we return the datetime calculated on the timezone of the partner.

`    
    @api.onchange("commitment_date")
    def _onchange_commitment_date_delivery_window(self):
        # OVERRIDE: warn if the commitment date doesn't fit the delivery window
        if not self.commitment_date or not self.partner_id.delivery_time_preference:
            return
        next_date = self.partner_id._next_available_delivery_date(self.commitment_date)
        breakpoint()
        if self.commitment_date != next_date:
            return {
                "warning": {
                    "title": self.env._("Customer delivery preference not met"),
                    "message": self.env._(
                        "The requested date doesn't fit with the customer delivery "
                        "preference. The next available delivery date is "
                        "%(next_date)s.",
                        next_date=format_datetime(self.env, next_date),
                    ),
                }
            }
`

fixed by return the date in the orginal timezone as datetime.

### DIFF
--- a/sale_stock_partner_delivery_window/models/res_partner.py
+++ b/sale_stock_partner_delivery_window/models/res_partner.py
@@ -23,8 +23,8 @@ class ResPartner(models.Model):
         # Pre-compute the from_datetime, in case from_date is a `date`
         # We use the start of the day in the partner's timezone
         tz = pytz.timezone(self.tz or self.env.company.partner_id.tz or "UTC")
-        from_datetime_tz_aware = tz.localize(fields.Datetime.to_datetime(from_date))
-        from_datetime = from_datetime_tz_aware.astimezone(pytz.utc).replace(tzinfo=None)
+        from_datetime = fields.Datetime.to_datetime(from_date)
+        from_datetime_tz_aware = tz.localize(from_datetime)
         # If the delivery is anytime, simply return the from_datetime
         if self.delivery_time_preference == "anytime":
             return from_datetime
--- a/sale_stock_partner_delivery_window/README.rst
+++ b/sale_stock_partner_delivery_window/README.rst
@@ -32,8 +32,44 @@ Sale Stock Partner Delivery Window
 
 |badge1| |badge2| |badge3| |badge4| |badge5|
 
-Use the partner's "Delivery schedule preference" to compute the Sales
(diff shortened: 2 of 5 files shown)


sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -24,7 +24,7 @@ def _next_available_delivery_date(
         # We use the start of the day in the partner's timezone
         tz = pytz.timezone(self.tz or self.env.company.partner_id.tz or "UTC")
         from_datetime_tz_aware = tz.localize(fields.Datetime.to_datetime(from_date))
-        from_datetime = from_datetime_tz_aware.astimezone(pytz.utc).replace(tzinfo=None)
+        from_datetime = fields.Datetime.to_datetime(from_date)
```
Ricardoalso, line 26: As discussed personally with @cyrilmanuel , before this change, “partner’s midnight” was being shifted into UTC, which can change the calendar day/hour (e.g., midnight Europe/Paris becomes 23:00 of the previous day in UTC during winter time).

With this change, switching to fields.Datetime.to_datetime(from_date), the start boundary stays aligned with the provided date (midnight of that date) and avoids the unintended timezone-to-UTC shift that could make the warning fire incorrectly.

Worth adding some additional context to sale_stock_partner_delivery_window/readme/DESCRIPTION.md

Thanks @cyrilmanuel 🙏

Ricardoalso (approved):

sale_stock_partner_delivery_window/models/res_partner.py
ivantodorovich, line 26: nitpicking: if you reverse the variable definition order, you can reuse `from_datetime` and have a single call to `fields.Datetime.to_datetime(from_date)`

```py
        from_datetime = fields.Datetime.to_datetime(from_date)
        from_datetime_tz_aware = tz.localize(from_datetime)
```

ivantodorovich (approved):
Thanks!

PR #2335, opened 2026-05-11
[FIX] purchase_stock_picking_invoice_link: avoid over-linking on partial invoicing

almumu:
Two complementary leaks let a single stock.move end up linked to several invoice lines of the same purchase order line, so each draft invoice ended up displaying pickings that did not belong to it:

- stock.move.write({state: done}) (fires for products with control policy "On ordered quantities") used to push the move into every matching invoice line. It now targets at most one line whose pending quantity is still positive — sum of previously linked moves' qty below line.quantity, with UoM rounding tolerance.

- purchase.order.line.get_stock_moves_link_invoice() (runs for both control policies, on every invoice creation) used to fall through and re-add any move whose to_invoice budget was not yet exhausted, even when that move was already fully claimed by an earlier invoice line. It now nets the qty already claimed by non-cancelled invoice lines (in_invoice add, in_refund subtract) and skips moves whose done qty is fully covered, preserving the legitimate split-into- multiple-invoices case where a single picking is invoiced piecewise.

Two new tests reproduce both shapes
(test_partial_invoice_separate_pickings and
test_sequential_drafts_to_refund_moves); existing tests covering the legitimate accumulation cases keep passing.

### DIFF
--- a/purchase_stock_picking_invoice_link/models/stock_move.py
+++ b/purchase_stock_picking_invoice_link/models/stock_move.py
@@ -2,15 +2,28 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import models
+from odoo.tools import float_compare
 
 
 class StockMove(models.Model):
     _inherit = "stock.move"
 
     def write(self, vals):
-        """Method used to associate the stock.move with the created account.move.line
-        when the invoicing method of the product is 'purchase' and the invoice is done
-        before receiving the products.
+        """Associate the stock.move with the related account.move.line when the
+        invoicing method of the product is 'purchase' and a matching invoice
+        line still has quantity not covered by previously linked moves.
+
+        Each move is linked at most to one invoice line (the oldest one whose
+        pending quantity is still positive). This avoids over-linking the same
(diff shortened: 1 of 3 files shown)


aritzolea (approved):

BhaveshHeliconia (approved):
Code review LGTM!

purchase_stock_picking_invoice_link/models/stock_move.py
rousseldenis, line 31:
```suggestion
            purchase_stock_moves = stock_moves.filtered(...)
            inv_lines = self.env["account.move.line"].search(
                    [
                        ("purchase_line_id", "in", purchase_stock_moves.purchase_line_id.id),
                        ("move_id.move_type", "=", inv_type),
                    ]
                        ("move_id.state", "not in", ("cancel",)),
                    ],
                    order="id",
                )
                
                for purchase_stock_move in purchase_stock_moves:
                    (... filter then the invoice lines on purchase_line_id)
```

rousseldenis (changes requested):
@almumu Could you check suggestion as the current code is not optimized.

PR #2336, opened 2026-05-12
[18.0][ADD] stock_picking_reset_qty

Saran440:
This module adds a "Clear QTY" button on picking operations.

Normally, when a picking is confirmed, Odoo automatically sets the demand quantity equal to the ordered quantity. However, in some cases vendors deliver only partial products, so users need to manually clear unwanted lines by setting the quantity to 0.


BhaveshHeliconia (approved):
Code and functional review LGTM!

Saran440:
Closed because I found `stock_move_reset_quantity` already done
https://github.com/OCA/stock-logistics-warehouse/tree/18.0/stock_move_reset_quantity already

PR #2338, opened 2026-05-12
[IMP] stock_dynamic_routing: apply routing on insert move DRAFT

TDu:
Before this change, when a move was inserted in a chain it was not merged with existing moves with the same characteristics, if possible.

Now when the inserted move can be assigned to an existing picking with a similar move, they will be merged.

### DIFF
--- a/stock_dynamic_routing/models/stock_move.py
+++ b/stock_dynamic_routing/models/stock_move.py
@@ -304,6 +304,9 @@ class StockMove(models.Model):
         routing_to_apply = [
             (move, detail.rule) for move, detail in routing_details.items()
         ]
+        # Keep the moves created by _insert_routing_moves that themselves match
+        # a routing rule, as to try to merge them with similiar moves.
+        inserted_routing_move_ids = set()
         for move, routing_rule in routing_to_apply:
             # Add the routing rule to the context for stock_dynamic_routing_delivery
             move = move.with_context(__routing_rule=routing_rule)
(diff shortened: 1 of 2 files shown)


stock_dynamic_routing/models/stock_move.py
```diff
@@ -379,9 +382,14 @@ def _apply_routing_rule_pull(self, routing_details, assign=False):
...
                     # Add a new routing to apply for this new move
                     routing_to_apply.append((routing_move, routing_rule))
+                    # Mark it as eligible for merging once its rule is applied.
+                    inserted_routing_move_ids.add(routing_move.id)
 
             pickings_to_check_for_emptiness |= move.picking_id
             move._assign_picking()
+            # Try merging inserted move
+            if move.id in inserted_routing_move_ids:
+                move = move._merge_moves()
```
jbaudoux, line 392: I think this section (from pickings_to_check_for_emptiness to _merge_moves) makes only sense when the picking type has changed. Maybe you could build a set in that section above changing the picking type and do this for moves in that set.

In standard, in action_confirm, merge_moves follows assign_picking. What is the impact if you remove your inserted_routing_move_ids condition?
You can test with pick+ship with pick rerouted to Shelving. Then simulate a SO qty increase by recreating a pick+ship in the same procurement group. That new pick should also be rerouted to Shelving and then merged with the first one. Is this currently happening? (without relocate module installed) Because in this case, we don't have any inserted_routing_move_ids but we still _assign_picking.

jbaudoux (approved):

PR #2343, opened 2026-05-18
[14.0][FIX]stock_move_line_auto_fill: skip auto-fill when move line a…

Kev-Roche:
…lready has qty_done

Auto-fill cumulates qty_done when a quantity is manually set and show_reserved is disabled, creating duplicate move lines.

This PR fix auto-fill ignoring existing qty_done on move lines when show_reserved is disabled.

Add tests to cover skip behavior for both show_reserved True and False cases.


rousseldenis:
@Kev-Roche Could you rebase and check tests ?

PR #2345, opened 2026-05-20
[FIX] stock_picking_return_lot: avoid copying return lines in onchange

akmanrisal-sman:
image

When opening the return picking wizard for a picking containing several lots,
the module splits return lines by lot inside `_compute_moves_locations`.

The previous implementation used `line.copy()` to create the extra wizard
lines. In Odoo 18, this creates real transient records with integer ids during
the onchange/form initialization flow. The web client later compares the
one2many snapshot and expects virtual records with an `origin` attribute, which
causes:

AttributeError: 'int' object has no attribute 'origin'

Avoid creating real transient records during the compute. Rebuild the
`product_return_moves` one2many with command values instead, preserving the
same lot split behavior and quantities.

## Problem

Opening the return picking wizard can crash when the original picking contains
more than one lot for the same returned move.

The traceback fails in the web onchange diff:

```text
AttributeError: 'int' object has no attribute 'origin'


sergio-teruel (approved):
LGTM!!

BhaveshHeliconia (approved):
LGTM!

sergio-teruel:
ping  @carlos-lopez-tecnativa

carlos-lopez-tecnativa (changes requested):
Tested locally and it's working, this change fixes the traceback, but I think it needs a minor adjustment.

Try this example:

With a product tracked by lot, create an incoming reception for 5 units, then create two lots:

Lot1 = 2
Lot2 = 3
Validate the picking, then **generate a first return for only Lot1 with 2 units**, and validate the return.

Now try to generate a new return, Lot1 won't be suggested because its quantity is zero. Only Lot2 will be suggested.

Image

sergio-teruel:
@akmanrisal-sman Can you review the @carlos-lopez-tecnativa comments?

PR #2354, opened 2026-05-28
[16.0][ADD] stock_auto_process

AungKoKoLin1997:
This module lets you define rules that automatically process pending stock pickings (confirm, assign, validate) based on configurable conditions.

It is useful when certain pickings are routinely ready to ship as soon as they are reserved, and manually clicking through them adds no value.

@qrtl QT6123


AungKoKoLin1997:
pre-commit issue is not related with my module.

```
Checks for .po[t] files..................................................Failed
- hook id: oca-checks-po
- exit code: 1

****po-python-parse-format****
stock_partner_delivery_window/i18n/gl.po:143 Translation string couldn't be parsed correctly using str.format KeyError('dáche') - [po-python-parse-format]
stock_partner_delivery_window/i18n/gl.po:155 Translation string couldn't be parsed correctly using str.format KeyError('dáche') - [po-python-parse-format]
```

stock_auto_process/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_stock_auto_process_rule_manager,stock.auto.process.rule manager,model_stock_auto_process_rule,stock.group_stock_manager,1,1,1,1
```
nobuQuartile: I think the permission to enable this automated function is very strong, so we should create a new group for this.

stock_auto_process/views/stock_auto_process_rule_views.xml
```diff
@@ -0,0 +1,108 @@
...
+            <form>
+                <sheet>
+                    <widget
+                        name="web_ribbon"
+                        title="Archived"
+                        bg_color="bg-danger"
+                        attrs="{'invisible': [('active', '=', True)]}"
+                    />
+                    <field name="active" invisible="1" />
+                    <field name="company_id" invisible="1" />
```
nobuQuartile:
```suggestion
                    <field name="company_id" groups="base.group_multi_company" />
```

Why is the company field invisible in the form view?
The user cannot update the company field.

stock_auto_process/models/stock_auto_process_rule.py
```diff
@@ -0,0 +1,73 @@
...
+            return [("id", "=", 0)]
+        domain = [
+            ("state", "in", states),
+            ("company_id", "=", self.company_id.id),
+        ]
+        if self.picking_type_ids:
+            domain.append(("picking_type_id", "in", self.picking_type_ids.ids))
+        if self.domain:
+            domain += safe_eval(self.domain)
+        return domain
```
nobuQuartile, line 68: We can remove picking_type_ids because we can use a domain for picking_type.
Why did you add this field?

In my opinion, we should stop automating the validation of all picking. 
When the domain is false, the error should occur.
If the user creates a rule with a false domain, the impact is too huge.

On the other hand, as an idea, it might also make sense to make picking_type_ids mandatory.

nobuQuartile (commented):
Partial review

stock_auto_process/security/ir.model.access.csv
AungKoKoLin1997: If I am not mistaken, only inventory user and admin can use inventory app. So, you want to add the new group because you want to allow the inventory user to create the rules. In my opinion, rules should only be defined by inventory admins.

stock_auto_process/models/stock_auto_process_rule.py
AungKoKoLin1997: The intention of `domain` field is for the adding the extra filters of the picking. But the intention of this module is based on picking types and it is the main factor of the operation. So, I think having that field is better design for the user experience.

stock_auto_process/views/stock_auto_process_rule_views.xml
AungKoKoLin1997: @nobuQuartile Updated

nobuQuartile (approved):
Code review and functional review: LGTM

AungKoKoLin1997:
For our operation, we decided to use automated action with a specific condition. So, we have no time to follow up on generic module and closing this PR.

PR #2353, opened 2026-05-28
[FIX] stock_partner_delivery_window: check delivery windows with env timezone handling

yankinmax:
**Description of the issue/feature this PR addresses:**

Stock picking delivery window checks can evaluate scheduled dates incorrectly when the partner delivery window is configured in a timezone different from UTC.

The module also still used manual context timezone access in warning messages, while Odoo 19.0 recommends using `self.env.tz`.

**Current behavior before PR:**

A picking scheduled date stored as an Odoo naive UTC datetime could be checked as if it were already expressed in the partner timezone.

This could make the delivery window warning appear or disappear incorrectly around window boundaries.

Warning messages also used `self.env.context.get("tz")` instead of the Odoo 19.0 `self.env.tz` timezone reference.

**Desired behavior after PR is merged:**

Picking scheduled dates are converted from Odoo naive UTC to the partner delivery window timezone before checking whether they fit the delivery window.

Warning messages use `self.env.tz` for the current environment timezone reference.

This aligns the module with Odoo 19.0 timezone handling while preserving partner local delivery window semantics.

**Steps to validate expected behavior**

1. Open a test customer.
2. Set `Timezone` to `Europe/Zurich`.
3. Set `Delivery time schedule preference` to `Fixed time windows`.
4. Add one delivery time window:
   - Weekday: `Friday`
   - From: `10:00`
   - To: `18:00`

5. Create or open an outgoing delivery picking for this customer.
6. Set `Scheduled Date` to Friday `09:59` customer local time.
   - Expected: delivery-window warning is shown.

7. Set `Scheduled Date` to Friday `10:00` customer local time.
   - Expected: no delivery-window warning.

8. Set `Scheduled Date` to Friday `18:00` customer local time.
   - Expected: no delivery-window warning.

9. Set `Scheduled Date` to Friday `18:01` customer local time.
   - Expected: delivery-window warning is shown.

10. Set `Scheduled Date` to Thursday `10:00` customer local time.
    - Expected: delivery-window warning is shown.

PR #2355, opened 2026-05-28
[FIX] sale_stock_partner_delivery_window: compute delivery window dates in partner tz

yankinmax:
**Description of the issue/feature this PR addresses:**

When a sale order delivery date is checked against the customer delivery time windows, the next available delivery date can be computed with an incorrect timezone.

Odoo datetime fields are stored as naive UTC datetimes, but `_next_available_delivery_date` was treating the incoming `commitment_date` as if it was already expressed in the partner timezone.

**Current behavior before PR:**

If a customer has delivery windows in a non-UTC timezone, the sale order warning may suggest an incorrect next available delivery time.

Example:
- Partner timezone: Europe/Zurich
- Delivery window: Friday 10:00-18:00
- Requested delivery date is outside the window
- Suggested next available date may be shifted because the UTC datetime is interpreted as partner local time.

**Desired behavior after PR is merged:**

The sale order delivery date is converted from Odoo naive UTC to the partner delivery window timezone before checking the customer delivery preference.

When the next available delivery window is found, it is converted back to Odoo naive UTC before being returned and formatted by Odoo.

This keeps:
- delivery window checks correct in the customer/partner timezone;
- stored datetimes consistent with Odoo UTC conventions;
- warning display handled by Odoo’s normal formatting logic.

**Steps to validate expected behavior**

1. Open a test customer.
  - Set `Timezone` to `Europe/Zurich`.
  - Set `Delivery time schedule preference` to `Fixed time windows`.
  - Add one delivery time window:
   - Weekday: `Friday`
   - From: `10:00`
   - To: `18:00`

2. Create a quotation for this customer.

3. Set `Delivery Date` to Friday `10:00` customer local time.
   - Expected: no “Customer delivery preference not met” warning.

4. Set `Delivery Date` to Thursday `10:00` customer local time.
   - Expected: warning is shown.
   - Suggested next available delivery date should be Friday `10:00` customer/user-visible time.

5. Set `Delivery Date` to Friday `09:59` customer local time.
   - Expected: warning is shown.
   - Suggested next available delivery date should be Friday `10:00`.

6. Set `Delivery Date` to Friday `18:01` customer local time.
   - Expected: warning is shown.
   - Suggested next available delivery date should be the next valid Friday at `10:00`.

7. Confirm the quotation.
   - Expected: delivery picking is created.
   - Its scheduled date should remain consistent with the sale delivery date logic.
   - The picking warning should match the stock check rules from `stock_partner_delivery_window`.

### DIFF
--- a/sale_stock_partner_delivery_window/models/res_partner.py
+++ b/sale_stock_partner_delivery_window/models/res_partner.py
@@ -13,6 +13,18 @@ from odoo.tools.date_utils import start_of
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    def _to_partner_delivery_datetime(self, from_date, tz):
+        """Convert a datetime to partner local datetime.
+
+        :param from_date: date or datetime to convert.
+        :param tz: partner delivery window timezone string.
+        :return: timezone aware datetime in the partner timezone.
+        """
+        timezone = pytz.timezone(tz)
+        if isinstance(from_date, datetime):
+            return pytz.utc.localize(from_date).astimezone(timezone)
+        return timezone.localize(fields.Datetime.to_datetime(from_date))
+
     def _next_available_delivery_date(
         self, from_date: date | datetime | None = None
     ) -> datetime:
@@ -20,11 +32,12 @@ class ResPartner(models.Model):
         # If from_date is not provided, use the current datetime
         if from_date is None:  # pragma: no cover
             from_date = fields.Datetime.now()
-        # Pre-compute the from_datetime, in case from_date is a `date`
-        # We use the start of the day in the partner's timezone
-        tz = pytz.timezone(self.tz or self.env.company.partner_id.tz or "UTC")
+        # get the from_datetime, in case from_date is a `date`.
+        # datetime objects stored in odoo are naive UTC,
+        # while windows are partner local.
+        tz = self.delivery_window_tz
         from_datetime = fields.Datetime.to_datetime(from_date)
-        from_datetime_tz_aware = tz.localize(from_datetime)
+        from_datetime_tz_aware = self._to_partner_delivery_datetime(from_date, tz)
         # If the delivery is anytime, simply return the from_datetime
         if self.delivery_time_preference == "anytime":
             return from_datetime
@@ -67,11 +80,13 @@ class ResPartner(models.Model):
                             continue
                     # Otherwise, since we're looking at days ahead, simply pick the
                     # window's start time
-                    return (
-                        datetime.combine(next_date, start_time)
-                        .astimezone(pytz.utc)
-                        .replace(tzinfo=None)
+                    next_datetime = next_date.replace(
+                        hour=start_time.hour,
+                        minute=start_time.minute,
+                        second=start_time.second,
+                        microsecond=start_time.microsecond,
                     )
+                    return next_datetime.astimezone(pytz.utc).replace(tzinfo=None)
         else:  # pragma: no cover
             raise ValueError(
                 self.env._(
--- a/sale_stock_partner_delivery_window/tests/test_partner_delivery_window.py
+++ b/sale_stock_partner_delivery_window/tests/test_partner_delivery_window.py
@@ -25,6 +25,20 @@ class TestSalePartnerDeliveryWindow(PartnerDeliveryWindowCommon):
             }
         )
 
+    def _set_friday_zurich_window(self):
+        self.customer_time_window.tz = "Europe/Zurich"
+        self.customer_time_window.delivery_time_window_ids.write(
+            {
+                "time_window_start": 10.0,
+                "time_window_end": 18.0,
+                "time_window_weekday_ids": [
+                    Command.set(
+                        self.env.ref("base_time_window.time_weekday_friday").ids
+                    )
+                ],
+            }
+        )
+
     @freeze_time("2020-04-02 10:00:00")  # Thursday
     def test_expected_date_anytime(self):
         """Customer with no delivery preferences.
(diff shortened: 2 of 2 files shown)


sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -13,18 +13,44 @@
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    def _get_delivery_window_timezone(self):
+        """Return the timezone used to interpret partner delivery windows."""
+        self.ensure_one()
+        timezone = self.tz or "UTC"
```
ivantodorovich:
```suggestion
        timezone = self.tz or self.env.company.partner_id.tz or "UTC"
```

sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -13,18 +13,44 @@
...
+
+        :param from_date: date or datetime to convert.
+        :param tz: partner delivery window timezone.
+        :return: timezone aware datetime in the partner timezone.
+        """
+        if isinstance(from_date, datetime):
+            return pytz.utc.localize(from_date).astimezone(tz)
+        return tz.localize(datetime.combine(from_date, time.min))
+
+    def _to_utc_naive(self, datetime_tz):
```
ivantodorovich: I would move this method out from the model.

```py
def to_utc_naive(datetime_tz): ...
    ...


class ResPartner(models.Model):
    ...
```

ivantodorovich (approved):
Thanks!

BhaveshHeliconia (approved):
LGTM!

yankinmax:
Hello @jbaudoux is it ok for you now?

sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -34,7 +52,9 @@ def _next_available_delivery_date(
             weekday = from_datetime_tz_aware.weekday()
             if weekday <= 4:
                 return from_datetime
-            return from_datetime + timedelta(days=7 - weekday)
+            next_date = from_datetime_tz_aware + timedelta(days=7 - weekday)
+            # Convert a datetime back to naive UTC format
+            return next_date.astimezone(pytz.utc).replace(tzinfo=None)
```
jbaudoux: This is not necessary. The convertion to local is for computing the days to add. Those can be added directly on `from_datetime`

sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -67,11 +87,11 @@ def _next_available_delivery_date(
...
                     # window's start time
-                    return (
-                        datetime.combine(next_date, start_time)
-                        .astimezone(pytz.utc)
-                        .replace(tzinfo=None)
+                    next_datetime = tz.localize(
+                        datetime.combine(next_date.date(), start_time)
                     )
+                    # Convert a datetime back to naive UTC format
+                    return next_datetime.astimezone(pytz.utc).replace(tzinfo=None)
```
jbaudoux, line 89: I don't get why this needs to be changed.
`next_date` is in local tz, the time part is replaced, and then converted back to UTC.

sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -13,18 +13,36 @@
 class ResPartner(models.Model):
     _inherit = "res.partner"
 
+    def _get_delivery_window_timezone(self):
+        """Return the timezone used to interpret partner delivery windows."""
+        self.ensure_one()
+        timezone = self.tz or self.env.company.partner_id.tz or "UTC"
+        return self.with_context(tz=timezone).env.tz
```
jbaudoux: If you want to define this, then move it to the base module in your other PR https://github.com/OCA/stock-logistics-workflow/pull/2353
You could also make a property
```suggestion
    @property
    def delivery_window_tz(self):
        """Return the timezone used to interpret partner delivery windows."""
        self.ensure_one()
        return self.tz or self.env.company.partner_id.tz or "UTC"
```
and then you can use it in both modules/PRs

sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -13,18 +13,36 @@
...
+
+    def _to_partner_delivery_datetime(self, from_date, tz):
+        """Convert a datetime to partner local datetime.
+
+        :param from_date: date or datetime to convert.
+        :param tz: partner delivery window timezone.
+        :return: timezone aware datetime in the partner timezone.
+        """
+        if isinstance(from_date, datetime):
+            return pytz.utc.localize(from_date).astimezone(tz)
```
jbaudoux: Keep tz argument as a string
```suggestion
        timezone = pytz.timezone(tz)
        if isinstance(from_date, datetime):
            return pytz.utc.localize(from_date).astimezone(timezone)
        return timezone.localize(datetime.combine(from_date, time.min))


```

sale_stock_partner_delivery_window/models/res_partner.py
```diff
@@ -13,18 +13,36 @@
...
+        return tz.localize(datetime.combine(from_date, time.min))
```
jbaudoux: Use ORM
```suggestion
        return timezone.localize(fields.Datetime.to_datetime(from_date))
```

sale_stock_partner_delivery_window/models/res_partner.py
yankinmax: Agree with you, this change is not needed.
`next_date` is already timezone aware in the partner/local timezone, so I restored the previous approach: keep `next_date`, replace only the time part with the window start, then convert it back to naive UTC before returning.

jbaudoux (approved):

PR #2357, opened 2026-06-02
[16.0][FIX] stock_picking_batch_invoice_frequency: use wrapper method

chienandalu:
Pickings are validated from the batch with the button_validate method.

We want to avoid that other modules using the inner method _action_done are called after the invoices are done, which should be the very last thing to do.

cc @moduon MT-14940

please review @EmilioPascual @Shide

PR #2360, opened 2026-06-05
[18.0][FIX] stock_picking_batch_invoice_frequency: use wrapper method

chienandalu:
fw of: https://github.com/OCA/stock-logistics-workflow/pull/2357

Pickings are validated from the batch with the button_validate method.

We want to avoid that other modules using the inner method _action_done are called after the invoices are done, which should be the very last thing to do.

cc @moduon MT-14940

please review @EmilioPascual @Shide

### DIFF
--- a/stock_picking_batch_invoice_frequency/models/stock_picking.py
+++ b/stock_picking_batch_invoice_frequency/models/stock_picking.py
@@ -8,8 +8,8 @@ from odoo import models
 class StockPicking(models.Model):
     _inherit = "stock.picking"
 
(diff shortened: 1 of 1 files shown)


EmilioPascual (approved):

BhaveshHeliconia (approved):
LGTM!

ivantodorovich (approved):

rafaelbn (approved):
Functional test 👍🏼 

/ocabot merge patch

PR #2361, opened 2026-06-05
[IMP] product_customerinfo_picking: hierarchy-aware lookup and consignment fallback

bosd:
## Summary

- **Partner hierarchy**: Replace the direct equality filter with `product_id._select_customerinfo(partner=partner)`, which resolves delivery addresses to their parent account (same logic used in `product_customerinfo_sale`).
- **Consignment fallback**: When the picking partner yields no match, fall back to `stock.warehouse.partner_id` (the warehouse's linked partner, a standard native field). This covers two consignment scenarios:
  - **Receipt** into a consignment warehouse: picking partner = vendor, code is on the warehouse owner.
  - **Delivery** from a consignment warehouse: picking partner = end customer, code is on the warehouse owner.
- **`@api.depends`** extended with `location_id.warehouse_id.partner_id` and `location_dest_id.warehouse_id.partner_id` to keep computed fields reactive.
- **Tests**: four new cases covering partner hierarchy resolution, consignment receipt, consignment delivery, and the existing direct-partner cases continue to pass.

## Motivation

The existing direct-equality filter (`m.partner_id == picking_id.partner_id`) misses two common real-world situations:

1. A customer code is configured on the main account but the picking goes to a delivery address child — the field stays blank.
2. A consignment warehouse has a `partner_id` set to its owner, but goods arrive from (or leave to) a different partner — the field stays blank even though the code is perfectly defined on the warehouse owner.

This PR aligns `product_customerinfo_picking` with the hierarchy-aware `_select_customerinfo` API that the rest of the `product_customerinfo` family already uses.

## Technical notes

- No model change, no DB migration needed (computed field only).
- `_select_customerinfo` is the OCA public API on `product.product`; it handles `parent_id` and `commercial_partner_id` traversal internally.
- The consignment fallback is a secondary lookup only when the primary returns nothing — existing behaviour for normal pickings is unchanged.


bosd:
Updated with performance optimizations following internal review:

- **Round 1 (cache)**: Replaced per-move  calls with a  cache to avoid repeated queries for duplicate pairs.
- **Round 2 (batch prefetch)**: Further improved to pre-fetch all relevant partners and customerinfo records in a single query before the loop. The in-memory  helper preserves variant-before-template priority and the consignment fallback.

PR #2363, opened 2026-06-10
[19.0][FIX] sale_stock_partner_delivery_window: keep partner tz for future window dates

yankinmax:
**Description of the issue/feature this PR addresses**:

When computing the next available sale order delivery date for partners with fixed delivery time windows, future candidate days can lose the partner timezone information before being converted back to Odoo’s naive UTC datetime format.

Example discovered with:
- Partner timezone: `Europe/Zurich`
- Delivery window: Wednesday `09:00-17:00`
- Requested delivery date: Thursday `2026-06-11 11:00:00` UTC

The module correctly detects that Thursday is outside the Wednesday-only delivery window, but the computed next Wednesday window start must be Wednesday `09:00` Zurich local time, stored as `2026-06-17 07:00:00` UTC in June.

**Current behavior before PR:**

The future candidate date is built with `start_of(...)`, which does not preserve the partner-local timezone-aware datetime. As a result, the module can treat the local window start as if it were already UTC.

For the Zurich case above, the next available date could be computed as:

```text
local Wednesday 09:00 -> 2026-06-17 09:00:00 UTC
```

This is two hours late during summer time.

**Desired behavior after PR is merged:**

Future candidate days are built as partner-local timezone-aware midnights before applying the delivery window start time. The final value is then converted back to Odoo’s naive UTC format.

For the Zurich case above, the next available date is correctly computed as:

```text
local Wednesday 09:00 Europe/Zurich -> 2026-06-17 07:00:00 UTC
```

This keeps partner delivery windows evaluated in the partner timezone while preserving Odoo UTC handling.

### DIFF
--- a/sale_stock_partner_delivery_window/models/res_partner.py
+++ b/sale_stock_partner_delivery_window/models/res_partner.py
@@ -7,7 +7,6 @@ import pytz
 
 from odoo import fields, models
 from odoo.exceptions import UserError
-from odoo.tools.date_utils import start_of
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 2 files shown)


ivantodorovich (approved):
/ocabot merge patch

PR #2365, opened 2026-06-11
[18.0][ADD] stock_move_line_split

LauraCForgeFlow:


JoanSForgeFlow (approved):

stock_move_line_split/tests/test_stock_move_line_split.py
```diff
@@ -0,0 +1,207 @@
...
+from odoo.exceptions import UserError
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestStockMoveLineSplit(BaseCommon):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
BhaveshHeliconia:
```suggestion
```

BhaveshHeliconia: When using BaseCommon, there’s no need to redefine cls.env with tracking_disable=True as these are already define in BaseCommon.

stock_move_line_split/tests/test_stock_move_line_split.py
```diff
@@ -0,0 +1,207 @@
...
+    def _create_picking(cls, specs):
+        picking = cls.env["stock.picking"].create(
+            {
+                "location_id": cls.stock.id,
+                "location_dest_id": cls.customers.id,
+                "picking_type_id": cls.picking_type_out.id,
+                "move_ids": [
+                    (
+                        0,
+                        0,
```
BhaveshHeliconia:
```suggestion
                    Command.create(
```

BhaveshHeliconia (approved):
LGTM!

Just minor comments

rousseldenis:
@LauraCForgeFlow @LoisRForgeFlow Following the module description, why not using batches and maybe https://github.com/OCA/stock-logistics-workflow/tree/18.0/stock_picking_batch_creation ?

LauraCForgeFlow:
Hi @rousseldenis, I think we're solving a different problem at a different granularity.

Your proposal with `stock_picking_batch_creation` works at the picking level: grouping several pickings into a batch and, if "split picking exceeding the limits" is enabled, splitting a whole picking into a new one so it fits the criteria.

What this module does is not creating new pickings. It splits a single stock move line into several move lines that stay under the same stock move and the same picking (no pickings are split here). Quantities are unchanged, we only break one operation into smaller ones (by fixed qty or by max weight). 

A use case could be being able to assign each resulting SML to a different destination package within a single transfer (SM and picking).

rousseldenis:
@LauraCForgeFlow What's your business case ?

LauraCForgeFlow:
Hi @rousseldenis, my particular business case is packing.

Take a SML of 15 units that has to go out in 3 outgoing packages of 5 units. With this module I can split that single SML of 15 units into 3 SMLs of 5 and then assign a different destination package to each (still within the same picking). Without it, we'd have to split the move line manually (5 + 5 + 5) before assigning the packages, which is what we want to avoid.

The split stays within the same SM and the same picking. Quantities are preserved and no new pickings are created.

stock_move_line_split/wizards/stock_move_line_split.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2026 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
```
jbaudoux: should use self.env._ instead of importing _

LauraCForgeFlow: Done! @jbaudoux

PR #2366, opened 2026-06-16
[16.0] [ADD] stock_picking_putaway_deferred: Add module to defer putaway strategy application to operator

lmignon:
* [ ]  depends on #1740


stock_picking_putaway_deferred/models/stock_move_line.py
```diff
@@ -0,0 +1,103 @@
...
+        to_clear = self.env["stock.move.line"]
+        if "location_dest_id" in vals and not self._context.get(
+            "deferred_putaway_apply"
+        ):
+            new_loc_id = vals["location_dest_id"]
+            # Operator explicitly set a *different* destination: treat it as applied.
+            # Do NOT clear when the system propagates the same value (e.g. from
+            # stock.package_level.write/_check_entire_pack rewriting the identical
+            # location on an already-deferred line).
+            to_clear = self.filtered(
```
jbaudoux: if to_clear == self, you could put it in vals

stock_picking_putaway_deferred/models/stock_move_line.py
```diff
@@ -0,0 +1,103 @@
...
+    def _can_recompute_putaway(self):
+        # Allow recomputation for package-level moves (whole package being
+        # relocated): result_package_id == package_id in that case, and the
+        # package_level_id is always set.  Only exclude lines whose
+        # result_package_id represents a *new* package being filled during the
+        # operation (no package_level_id), since the operator has explicitly
+        # chosen that destination and it must never be overridden.
+        if self.picking_type_id.defer_putaway_to_operator:
+            return self.picking_id._can_recompute_putaway() and not (
+                self.result_package_id and not self.package_level_id
```
jbaudoux: Would this be equivalent? To not rely on package level...
```suggestion
                self.result_package_id and self.result_package_id != self.package_id
```

stock_picking_putaway_deferred/models/stock_move_line.py
```diff
@@ -0,0 +1,103 @@
...
+            # putaway computed now (called from _action_assign). Mark them and
+            # leave their location_dest_id at the move's destination.
+            # Lines whose result_package_id is a *new* package being filled
+            # (no package_level_id) are excluded: the operator sets that
+            # destination explicitly and it cannot be recomputed.  Lines that
+            # move a whole existing package (package_level_id is set) are
+            # included in the deferred mechanism.
+            deferred = self.filtered(
+                lambda line: line.picking_type_id.defer_putaway_to_operator
+                and not (line.result_package_id and not line.package_level_id)
```
jbaudoux: would apply here as well then

stock_picking_putaway_deferred/models/stock_move_line.py
```diff
@@ -0,0 +1,103 @@
...
+        if "location_dest_id" in vals and not self._context.get(
+            "deferred_putaway_apply"
+        ):
+            new_loc_id = vals["location_dest_id"]
+            # Operator explicitly set a *different* destination: treat it as applied.
+            # Do NOT clear when the system propagates the same value (e.g. from
+            # stock.package_level.write/_check_entire_pack rewriting the identical
+            # location on an already-deferred line).
+            to_clear = self.filtered(
+                lambda l: l.putaway_deferred and l.location_dest_id.id != new_loc_id
```
jbaudoux: The putaway could end-up with the same location. You should still remove `putaway_deferred`.
The putaway is computed when you set a destination package. Manually or with put-in-pack. This is deferred from the action assign and should be enough. No need to force the user to launch manually the action `_recompute_putaways`

jbaudoux (changes requested):

jbaudoux (approved):

stock_picking_putaway_deferred/models/stock_move_line.py
```diff
@@ -0,0 +1,102 @@
...
+    def _can_recompute_putaway(self):
+        # Allow recomputation for package-level moves (whole package being
+        # relocated): result_package_id == package_id in that case, and the
+        # package_level_id is always set.  Only exclude lines whose
+        # result_package_id represents a *new* package being filled during the
+        # operation (no package_level_id), since the operator has explicitly
+        # chosen that destination and it must never be overridden.
+        if self.picking_type_id.defer_putaway_to_operator:
+            return self.picking_id._can_recompute_putaway() and not (
+                self.result_package_id and self.result_package_id != self.package_id
```
jbaudoux, line 53: The fact that the putaway is deferred should not impact on the condition if it can be recomputed or not.
That should be solved in the base module.
I guess the idea in the base module was to not show the recompute button if there is a result package.

jbaudoux: ping @lmignon

PR #2385, opened 2026-07-04
[18.0][FIX] product_cost_price_avco_sync: Wrong average cost when accumulated quantity is negative

sergio-teruel:
When previous_qty is negative (stock oversold due to a correction on an already validated stock move), a single incoming line's unit_cost was blindly adopted as the new average cost, discarding all previous history. This let one bad price on a purchase/receipt wipe out the product's standard_price entirely.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                      
Now the previous cost is still weighted using abs(previous_qty), so an erroneous price gets blended into the average instead of overwriting it completely.  

cc @Tecnativa TT63558

ping @carlosdauden @CarlosRoca13

### DIFF
--- a/product_cost_price_avco_sync/models/stock_valuation_layer.py
+++ b/product_cost_price_avco_sync/models/stock_valuation_layer.py
@@ -414,13 +414,21 @@ class StockValuationLayer(models.Model):
             # Normal incoming moves
             else:
                 svl_dic["unit_cost_processed"] = True
-                if svl_dic["previous_qty"] <= 0.0:
+                if float_is_zero(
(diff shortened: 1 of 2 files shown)


CarlosRoca13 (approved):

eduezerouali-tecnativa (approved):

rousseldenis:
@sergio-teruel Could you include tests to avoid future regressions as this is part of important flows ? Thanks

sergio-teruel:
@rousseldenis Changes done!

BhaveshHeliconia (approved):
LGTM!

rousseldenis (approved):

PR #2386, opened 2026-07-06
[IMP] stock_dynamic_routing: Add index on canceled_by_routing field

rousseldenis:


jbaudoux (commented):
That looks an overkill index.
What about testing first `not picking.move_ids` ?
Or even just write `picking.state = "cancel"` in `_dynamic_routing_handle_empty` ?

jbaudoux:
Same issue here https://github.com/OCA/stock-logistics-workflow/blob/18.0/stock_picking_group_by_partner_by_carrier/models/stock_picking.py#L52

PR #2311, opened 2026-04-14
[19.0] purchase_stock_picking_invoice_link: Migration to 19.0

vvro:
This PR migrates the `purchase_stock_picking_invoice_link` module to Odoo 19.0.

During the migration, the following Odoo 19.0 specific API changes and test coverage gaps were addressed:

* **UOM Rounding Adaption (`purchase_order.py`)**: Replaced `uom_id.rounding` with `product_uom_id.rounding` during floating point zero evaluations to align with the Odoo 19.0 `purchase.order.line` schema changes.
* **Stock Move Refund Logic (`stock_move.py`)**: Adjusted the evaluation for determining the invoice type (`in_refund` vs `in_invoice`) during stock validation (`_action_done`). In Odoo 19.0, `to_refund` defaults to `True` on incoming shipments, so the check was updated to explicitly evaluate if the destination location's usage is `"supplier"` (`location_dest_id.usage == 'supplier'`).
* **Codecov & Test Coverage enhancements**: Added dedicated tests (`test_purchase_return_after_refund` and `test_over_invoiced_filtered`) to ensure 100% coverage for the return/refund evaluations and the `float_is_zero` filtering branches.
* **Formatting**: Ran the codebase through the latest OCA `pre-commit` hooks to assure `ruff` constraints are fulfilled.

All tests are fully passing natively on Odoo 19.0 and the 100% code coverage requirement is strictly satisfied.

PR #2388, opened 2026-07-06
[19.0][MIG] purchase_stock_picking_invoice_link: Migration to version 19.0

Andrii9090-tecnativa:
Superceed PR #2311

In v19, `stock.move.to_refund` is True by default. To set the invoice type (`in_invoice` or `in_refund`), we need to use location_dest_id.usage  [commit Odoo](https://github.com/OCA/OCB/commit/98fbff3dcab6cba1ea8905311ed1ae012245b5de#diff-ad6229e976ce0bd1592e805b88e9813ac4e3f6fb989d3dfb8dfb8b70af03cd58R14)

@CarlosRoca13 @pedrobaeza ping
@Tecnativa TT61960


purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
```diff
@@ -9,6 +9,15 @@ class TestPurchaseSTockPickingInvoiceLink(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.company = cls.env.ref("base.main_company")
+        cls.company.lc_journal_id = cls.env["account.journal"].create(
```
pedrobaeza: Why do you need this? On post-install, it shouldn't be needed.

pedrobaeza (commented):
I don't get your comments about `to_refund` and the picking type, as I don't see changes in the code due to that.

purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
Andrii9090-tecnativa: > I don't get your comments about `to_refund` and the picking type, as I don't see changes in the code due to that.

In this part of the [code](https://github.com/OCA/stock-logistics-workflow/pull/2388/changes/a521411038ea63c54bc050fc4a8781981ee69722#diff-893744baa51c4c4027f083336ee0dfb04d543579b53eebdbcf6d5073b8b2d4faR22), I changed the logic used to select the `in_refund` or `in_invoice` move type. It no longer depends on `to_refund`.

I used an account journal in the test because the test raises this error https://github.com/OCA/stock-logistics-workflow/blob/19.0/stock_landed_costs_purchase_auto/models/purchase_order.py#L37

pedrobaeza: Then search for an existing one instead of creating it.

purchase_stock_picking_invoice_link/tests/test_purchase_stock_picking_invoice_link.py
```diff
@@ -9,6 +9,11 @@ class TestPurchaseSTockPickingInvoiceLink(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.company = cls.env.ref("base.main_company")
```
pedrobaeza: I don't like too much this workaround. I think you should inhibit instead in the source module the requirement if not testing that module.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #2390, opened 2026-07-07
[19.0][FIX] stock_landed_costs_purchase_auto: Add _test_skip_check_lc_journal_id to the context

Andrii9090-tecnativa:
Use default landed cost account journal

Related with #2388 

@pedrobaeza ping
@Tecnativa

### DIFF
--- a/stock_landed_costs_purchase_auto/models/purchase_order.py
+++ b/stock_landed_costs_purchase_auto/models/purchase_order.py
@@ -33,7 +33,8 @@ class PurchaseOrder(models.Model):
         # We need to use sudo() because only Inventory > Administrator have
         # permissions on stock.landed.cost
         self.ensure_one()
-        if not self.company_id.lc_journal_id:
+        StockLandedCost = self.env["stock.landed.cost"].with_company(self.company_id)
+        if not StockLandedCost._default_account_journal_id():
             raise UserError(
                 self.env._(
                     "Cannot create a landed cost for purchase order %(order)s: "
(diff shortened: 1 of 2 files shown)


stock_landed_costs_purchase_auto/models/purchase_order.py
```diff
@@ -33,6 +34,10 @@ def _create_picking_with_stock_landed_cost(self, picking):
         # We need to use sudo() because only Inventory > Administrator have
         # permissions on stock.landed.cost
         self.ensure_one()
+        if config["test_enable"] and self.env.context.get(
+            "_test_skip_check_lc_journal_id", False
+        ):
+            return
```
pedrobaeza: Put this before the other comment, as the other is contextual to the next lines, and put here also a comment, and the context should be the contrary and injected in these tests, not in the others.

```suggestion
        if config["test_enable"] and not self.env.context.get(
            "test_stock_landed_costs_purchase_auto"
        ):
            return
```

and maybe you can just skip the error.

I don't see that the journal is used in the creation, and this is apart from the test itself, as why is that one needed? Maybe you should remove this error and do it in other place and intercept also the config there.

pedrobaeza (changes requested):

stock_landed_costs_purchase_auto/models/purchase_order.py
Andrii9090-tecnativa: I use this [method](https://github.com/OCA/OCB/blob/19.0/addons/stock_landed_costs/models/stock_landed_cost.py#L26) from `stock.landed.cost` for check account journal

pedrobaeza (approved):
/ocabot merge patch

PR #2393, opened 2026-07-13
[16.0][FIX] stock_move_propagate_first_move: ensure first move is well propagated during splits

nicolas-delbovier-acsone:
- Change `first_move_id` definition in case you are the first move of a chain: before=yourself, now=False (this allows to split and thus copy fields without side effects when the move is not inside a picking)
- When splitting the first move, ensure the moves that were linked to it are then linked to the newly created split move (to ensure the link is preserved during chained moves split)

### DIFF
--- a/stock_move_propagate_first_move/models/stock_move.py
+++ b/stock_move_propagate_first_move/models/stock_move.py
@@ -38,27 +38,51 @@ class StockMove(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         # Extract the tracking flag before calling super so it doesn't break standard create
-        split_first_move_ids = {}
+        split_first_move_id_by_move_index = {}
         for i, vals in enumerate(vals_list):
             if "_split_first_move_id" in vals:
-                split_first_move_ids[i] = vals.pop("_split_first_move_id")
+                split_first_move_id_by_move_index[i] = vals.pop("_split_first_move_id")
 
         moves = super().create(vals_list)
 
-        # Update the upstream moves to point to the newly created backorder
-        for i, move in enumerate(moves):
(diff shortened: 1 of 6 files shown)


rousseldenis (approved):
LGTM. Indeed

BhaveshHeliconia (approved):
LGTM!

nicolas-delbovier-acsone:
This is not good as is, after further investigations, the copy is expected in case you split a move from a chain like pick->pack->ship.

For example, if you partially process the "pick" picking and then create a backorder, the backorder moves should inherit the "first_move_id" from their parent (otherwise, we break the chain).

stock_move_propagate_first_move/models/stock_move.py
```diff
@@ -22,10 +21,44 @@ class StockMove(models.Model):
...
+            if not split_first_move_id:
+                continue
+            # Find all moves that were tied to the split first_move
+            # and have not yet been completed or cancelled.
+            moves_to_update = self.env["stock.move"].search(
+                [
+                    ("first_move_id", "=", split_first_move_id),
+                    ("state", "not in", ("done", "cancel")),
+                ]
+            )
```
lmignon: @nicolas-delbovier-acsone  a search into a loop is far from optimal....  You must get all the moves to update in one call to `search`  if `split_first_move_ids` and then iterate over the created moves to update the moves to update

lmignon (changes requested):

stock_move_propagate_first_move/models/stock_move.py
nicolas-delbovier-acsone: Ah indeed, I'll fix this

lmignon (approved):

PR #2397, opened 2026-07-16
[18.0][ADD] stock_move_vendor_comment: New module to add vendor comment to stock moves and stock move lines

CarlosRoca13:
cc @Tecnativa TT57673

ping @carlosdauden @pilarvargas-tecnativa

PR #2398, opened 2026-07-16
[18.0][ADD] procurement_purchase_no_grouping_comment: New module to avoid grouping lines when the comment is different

CarlosRoca13:
cc @Tecnativa TT57640

ping @carlosdauden @sergio-teruel 

This module depends of:

- [ ] https://github.com/OCA/stock-logistics-workflow/pull/2397
- [ ] https://github.com/OCA/purchase-workflow/pull/3128
- [ ] https://github.com/OCA/sale-workflow/pull/4427


procurement_purchase_no_grouping_comment/models/stock_rule.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @api.model
+    def _get_procurements_to_merge_groupby(self, procurement):
+        """Do not group purchase order line if they are linked to different
+        vendor comment in sale order line.
+        """
+        return (
+            procurement.values.get("vendor_id"),
+            procurement.values.get("vendor_comment"),
+            super()._get_procurements_to_merge_groupby(procurement),
```
carlosdauden: Return one tuple with all elements instead of one tuple with 2 elements + third element is tuple with x elements.
```suggestion
            *super()._get_procurements_to_merge_groupby(procurement),
```

carlosdauden (changes requested):

procurement_purchase_no_grouping_comment/models/stock_rule.py
CarlosRoca13: Change done

PR #2399, opened 2026-07-16
[18.0][FIX] stock_owner_restriction: Owner restriction on chained moves

carlosdauden:
Odoo 18.0 _action_assign prefetches a quants cache for chained moves through stock.quant._read_group and reserves them by calling _update_reserved_quantity_vals directly:

- The _read_group override injected the raw context value (a res.partner recordset set by stock.move._action_assign) into the domain, which core now rejects (AssertionError: Invalid value res.partner(...) in domain term), so assigning any chained move under picking_partner or partner_or_unassigned restriction crashed.
- The _update_reserved_quantity override was dead code on the chained moves path; move it to _update_reserved_quantity_vals, which covers both the regular and the chained reservation paths.

[FIX] stock_owner_restriction: restrict manual quant selection

The Pick From selector (quant_id, detailed operations) let the user pick a quant regardless of the picking type owner restriction, and _copy_quant_info then carried that owner to the move line, reserving another partner's stock. Extend the selector domain in both detailed operation views according to the restriction, backed by two helper fields on stock.move.line.

The sml_x2_many widget builds its own quant domain in onAdd (product and location only), so the Add line dialog of the detailed operations offered quants from any owner regardless of the restriction. Patch the widget to append the owner clause, fed by the same helper fields now also exposed on stock.move.

@Tecnativa TT63741

### DIFF
--- a/stock_owner_restriction/__manifest__.py
+++ b/stock_owner_restriction/__manifest__.py
@@ -11,7 +11,17 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "installable": True,
-    "data": ["views/stock_picking_type_views.xml", "views/stock_picking_views.xml"],
+    "data": [
+        "views/stock_move_line_views.xml",
+        "views/stock_move_views.xml",
+        "views/stock_picking_type_views.xml",
+        "views/stock_picking_views.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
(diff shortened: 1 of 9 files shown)


carlos-lopez-tecnativa (approved):
LGTM

sergio-teruel (approved):

CarlosRoca13 (approved):

BhaveshHeliconia (approved):
LGTM!

PR #2300, opened 2026-04-07
[MIG] stock_picking_return_empty_package: Migration to 18.0

rousseldenis:

PR #2401, opened 2026-07-20
[19.0][ADD] stock_picking_return_force_unpack: force unpack on returns

Ricardoalso:
Add a Force Unpack on Return option on the Operation Type. When enabled on the operation type a return or exchange resolves to, any source and destination package assignment is automatically stripped from that return's stock moves (including nested packages), preventing validation errors when the original package still holds other, non-returned units at the customer.


rousseldenis:
@Ricardoalso There is already a module for that case. Could you review https://github.com/OCA/stock-logistics-workflow/pull/2300 instead and migrate it ?

Ricardoalso:
Duplicated

PR #2404, opened 2026-07-21
[18.0][FIX] product_cost_price_avco_sync: Restore core negative-stock vacuum for average cost product

carlosdauden:
## Problem

`_run_fifo_vacuum` was overridden to unconditionally no-op for **every** product (not just FIFO ones, as the pre-refactor version filtered by `cost_method`), since 8a36b333 ("Remove FIFO stuff", for performance reasons).

Core's `_run_fifo_vacuum` is the mechanism that re-prices a negative-stock deficit once a real receipt arrives, and it works for `average` costing too — not only FIFO, despite the name. Core's own `remaining_qty`/ `remaining_value` bookkeeping is maintained for **every** cost method (`_prepare_in_svl_vals`/`_prepare_out_svl_vals` in `stock_account`), and `stock_account`'s own test suite (`test_average_negative_5`) proves the vacuum correctly recomputes `standard_price` as `value_svl / quantity_svl` for an average-cost product once the deficit is covered.

With the override in place, nothing in this module covers that case either: `_cost_price_avco_sync` only reacts to a `write()` on an existing `stock.valuation.layer`, and `stock_picking._action_done` sets `skip_avco_sync=True` during normal picking validation. So a plain receipt validated while on-hand quantity is negative falls straight through the raw weighted-average formula in `product_price_update_before_done` (`stock_account` core), which has no guard for a negative previous quantity.

Observed in production: a product ended up with `standard_price = -277 EUR/kg` after a chain of receipts/returns while stock was negative, even though every real purchase price was ~21-24 EUR/kg.

## Fix

Remove the `_run_fifo_vacuum` override so the core vacuum runs again for every cost method. The module no longer keeps its own competing remaining-qty bookkeeping (it was deleted in the same 8a36b333 commit that added this override), so there is no double-processing risk.

Added a regression test that reproduces the scenario through the normal picking flow (receive, oversell into negative stock, receive again at a different cost) and asserts the product ends up with the vacuum-corrected average instead of the runaway blended value.

## Open question

@pedrobaeza — 8a36b333 mentions "performance reasons" for dropping this without further detail. If there was a specific profiling case behind it (large instance, many SVLs) that isn't visible from the commit message, please flag it here — `_run_fifo_vacuum` already short-circuits via a `remaining_qty < 0` read_group when there's nothing to vacuum, so the
common case (stock never negative) shouldn't pay extra cost, but happy to be corrected.

@Tecnativa TT63558

ping @sergio-teruel @CarlosRoca13 @carlos-lopez-tecnativa


sergio-teruel (approved):

carlos-lopez-tecnativa (approved):
The changes solve the problem, but I would like to understand the reason behind completely overriding the `_run_fifo_vacuum` method. This change comes from [OCA/stock-logistics-workflow#1697](https://github.com/OCA/stock-logistics-workflow/pull/1697), but I couldn't find any tests covering it.

@pedrobaeza, could you explain the reason for completely overriding this method? Is there a specific edge case or behavior that requires this implementation? I also couldn't find anywhere that `force_run_fifo_vacuum` is set.

pedrobaeza:
AFAIK, that was removed being part of the FIFO mechanism, having `fifo` in its name, but according to what you mention, it's used for negative values. I wonder: if `remaining_value` / `remaining_quantity` are not synchronized, does this affects to this method?

carlosdauden:
Replaced by deep refactor: https://github.com/OCA/stock-logistics-workflow/pull/2416

PR #2408, opened 2026-07-23
14.0 backport stock scrap tier validation from 15.0

hitrosol:
standard migration


hitrosol:
The error in pre-commit caused by pkg_resources, not sure why this error and Missing newline in other module.

PR #2409, opened 2026-07-23
[18.0][FIX] stock_owner_restriction: owner restriction on _set_quantity_done reservation path

carlosdauden:
stock.move.quantity's inverse (_set_quantity -> _set_quantity_done -> _set_quantity_done_prepare_vals) reserves quants directly, entirely bypassing _action_assign, the only method this module protects. Any code writing that field directly - a stock.picking import via move_ids_without_package/quantity, typing a quantity in "Immediate Transfer" without reserving first, MRP component consumption, subcontracting - could silently reserve another partner's owned stock.

_gather()'s existing context-based filter also silently no-op'd here because its early-return guarded on the incoming `owner_id` parameter being non-None, which this core call path never normalizes, instead of relying only on the `force_restricted_owner_id` context the module itself controls (already the pattern used by the sibling _get_available_quantity override).

Add a _set_quantity_done override mirroring _action_assign's per-move owner-restriction split, and drop the owner_id-parameter dependency in _gather's guard so the filter can no longer be skipped by a caller that doesn't normalize it.

@Tecnativa TT63741

ping @sergio-teruel @CarlosRoca13 @carlos-lopez-tecnativa


carlos-lopez-tecnativa (approved):
LGTM.

stock_owner_restriction/tests/test_stock_owner_restriction.py
```diff
@@ -236,6 +236,48 @@ def test_chained_moves_picking_partner(self):
...
 
+    def test_set_quantity_done_import_unassigned_owner(self):
+        """Writing move.quantity directly must still respect the restriction.
+
+        This is what a stock.picking import does via
+        move_ids_without_package/quantity (as opposed to product_uom_qty
+        alone): it reserves through stock.move._set_quantity's inverse,
+        entirely bypassing _action_assign. Reproduces the WH/OUT/723826
+        incident, where that path grabbed a quant owned by a different
+        partner because it was older (lower in_date) than the free stock.
```
carlos-lopez-tecnativa, line 247:
```suggestion
        """Writing move.quantity directly must still respect the restriction.

        This is what a stock.picking import does via
        move_ids_without_package/quantity (as opposed to product_uom_qty
        alone): it reserves through stock.move._set_quantity's inverse,
        entirely bypassing _action_assign. 
```

carlos-lopez-tecnativa (commented):
Odoo allows users to edit a picking after it has been validated and directly modify the quantity field. I'm not sure whether the reservation will be handled correctly in this case. Have you tested this case without needing to import anything from a file?

image

sergio-teruel (approved):

CarlosRoca13 (approved):

rousseldenis:
@carlosdauden Could you check tests ?

PR #2416, opened 2026-07-30
[18.0][REF] product_cost_price_avco_sync: Support valuation by lot and fix the cost on negative stock

carlosdauden:
Odoo keeps one average cost per product, or one per lot when the product has "Valuation by Lot/Serial number" ticked, and both kinds of product live in the same database. This module only knew about the first one: it replayed the valuation chain per (product, company) ignoring the lot, so on a product valuated by lot it re-priced the outgoing layers of every lot with the average of all of them, silently undoing the only thing lot valuation buys. Forcing a full resync after enabling it turned a correct per lot valuation into a product-average one and inflated the stock value by more than half.

The replay now runs per valuation chain, `(product, company, lot)` when the product is valuated by lot and `(product, company)` when it is not, so a cost correction only re-prices what left that very lot and the other lots keep their own cost. The product cost stops being replayed and is derived from the layers, which is what Odoo itself does in `stock_lot._change_standard_price` and in `purchase_stock`. The layers a product already had before being switched to valuation by lot carry no lot and keep replaying as the single chain they were, so both models coexist and a product can be switched at any time.

An average cost product whose accumulated quantity is negative ends up with a cost that is not an average of anything, and that can be negative. From there the damage compounds: every outgoing move copies that negative cost, so its layer gets a positive value and the stock valuation runs away. Seen in production on a product permanently at -1.5 units: receiving 3 units at 25 EUR moved its cost from 23.22 to 19.65 and then to -1379.38, over a value_svl that earlier occurrences of the same problem had already inflated to 2069 EUR for a negative quantity.

Three formulas in `stock_account` divide by that negative quantity, guarded only against it being exactly zero:
`stock.move.product_price_update_before_done`, which does it once for the product and once per lot; the tail of `product._run_fifo_vacuum`, again for both; and `stock.move._product_price_update_after_done`, which only fires for the products valuated by lot. All three are neutralised, and whoever is oversold, product or lot, takes the cost of the receipt: there are no units left to average against, and it is the same cost the vacuum uses to settle the deficit once enough real stock arrives, so both ends agree and the price is always a real purchase price. That rule lives in `_get_avco_svl_price` alone and is shared by the picking validation and by the replay, which used to disagree.

The `_run_fifo_vacuum` override that made the vacuum do nothing at all is gone. It was meant for FIFO but no longer filtered by cost method, and Odoo maintains its remaining_qty/remaining_value bookkeeping for every one of them; its own `test_average_negative_5` proves the vacuum correctly re-prices an average cost deficit once a real receipt arrives. Since this module no longer keeps a competing bookkeeping of its own, letting it run again carries no risk of double processing.

Correcting the quantity of an already validated move keeps restating its layer rather than appending a correction one, which is what makes a stock valuation asked for a date before the correction come out right, and what re-prices the outgoing moves valued in between so the margin follows. What the restatement never did was update `remaining_qty` and `remaining_value`, leaving them understated by the corrected amount and so desynchronised from the two mechanisms that read them, the negative stock vacuum and the vendor bill price difference. They are now kept in step, the correction is applied to the layer of the right lot instead of to every layer of the move, and the vacuum is run afterwards just like Odoo does.

The inventory adjustment branch is gone: it rewrote the quantity and swapped the locations of the underlying stock move through the `keep_avco_inventory` context, when an inventory layer only needs to be re-priced with the running cost, like a return. That is what `_is_avco_valued_at_current_cost` now expresses for the three cases that share it.

A layer that only carries value raises or lowers the cost of what is in stock, and the sign of the layer it hangs from says whether it belongs to that stock at all. A child of an incoming layer is a landed cost or a vendor bill price difference, part of what those units cost. A child of an outgoing layer is one of core's `Revaluation of ... (negative inventory)` settlements, written by `_run_fifo_vacuum` for units that have already left the company, which say nothing about what is still on hand. Spreading the latter over it divided a value by a quantity that had nothing to do with it: seen in production, a revaluation of -1120 EUR landing when 0.22 units were held gave -5062 EUR/unit, and every later layer of the chain inherited it.

A negative average cost is not a low valuation, it is a corruption. It flips the sign of the value of every layer after it, so outgoing moves start adding value and the totals of the product cancel out to something plausible while every single layer inside is wrong, which is what hides the damage from any report that reads the totals: the same product summed to 0.28 EUR while holding seventeen layers at -5062 EUR/unit. So the spread is skipped when it would take the cost below zero, and the settlements of units already gone no longer touch it at all.

The stock revaluation wizard writes a value-only layer with no parent to hang from, so it was never spread either, and a later correction left the product with a cost contradicting what its own layers were worth. It names no target cost because it just adds value to what is on hand, so it is now spread the same way a landed cost is.

The same reasoning applies to the price Odoo writes at the end of `_run_fifo_vacuum`, `standard_price = value_svl / quantity_svl`. By then `value_svl` carries the layers the vacuum has just created to settle the deficit, which have no quantity because they correct units that already left the company. When the history is internally consistent both figures agree and nothing changes, which is why every existing test stays green. When it is not, and repairing exactly that is what this module is for, the ratio wins over the price taken from the real incoming cost. Seen in production: a manufacturing order left a product at 6.6567 EUR/unit and the vacuum rewrote it to 8.3811, which is 1141.67 / 136.220 once three `Revaluation of ... (negative inventory)` layers had added 210.41 EUR settling 52.18 units delivered a month and a half earlier. The units on hand became 26% more expensive because of goods that were no longer there. The override now keeps the price for every product and lot the vacuum settles, not only for those left oversold, so the live path and the replay agree on the cost.

@Tecnativa TT63558


product_cost_price_avco_sync/models/stock_lot.py
rrebollo, line 1: It looks like there's room for a "DRYer" approach here—there are common patterns in `product_cost_price_avco_sync/models/product_product.py`.

product_cost_price_avco_sync/models/stock_valuation_layer.py
```diff
@@ -1,49 +1,110 @@
...
+        """
+        entry_points = {}
+        for svl in self.sorted(lambda x: (x.create_date, x.id)):
+            if (
+                svl.stock_valuation_layer_id
+                or svl.product_id.with_company(svl.company_id).cost_method != "average"
+            ):
+                continue
+            entry_points.setdefault(svl._get_avco_sync_key(), svl)
+        return self.browse([svl.id for svl in entry_points.values()])
```
rrebollo, line 53: Why do you need to call `browse()` when you already have the SVL record?

rrebollo (approved):
**Code Review:** LGTM! I've left a few comments you might want to address for clarification. 👍 for the deep documentation update and the improve on code coverage. Great.

product_cost_price_avco_sync/models/stock_lot.py
carlosdauden: Changes done

product_cost_price_avco_sync/models/stock_valuation_layer.py
carlosdauden: Good question, and it's actually the safer choice here on purpose.

`browse()` itself is O(N) with no query (just wraps a tuple of ids), so accumulating ids and calling it once at the end is linear.

The tempting alternative — unioning the recordsets with `|`/`+` (e.g. `functools.reduce(operator.or_, entry_points.values(), self.browse())`, or a `result |= svl` loop) — looks equivalent but isn't: each `|`/`+` call rebuilds the *entire* left-hand-side ids list from scratch and re-wraps it in an `OrderedSet` (see `union`/`concat` in `models.py`, which are explicitly documented as linear only for a single variadic call with all arguments at once). Doing that N times in a loop is O(1+2+...+N) = O(N²), which becomes noticeable on the large batches `action_force_avco_sync` is meant to handle.

The variadic form (`self.browse().union(*entry_points.values())`) would be Big-O equivalent to the current code (plus an unnecessary dedup pass, since `entry_points.values()` are already unique), so there's no real upside to changing this — keeping the accumulated-ids + single `browse()` avoids the O(N²) trap entirely rather than relying on always remembering to call union variadically.

product_cost_price_avco_sync/models/stock_lot.py
rrebollo: Really nice work on the mixin approach.

CarlosRoca13 (approved):

PR #2417, opened 2026-07-30
[18.0][ADD] stock_owner_restriction_mrp: New module

carlosdauden:
Manufacturing did not follow the owner restriction, and the way it did not quietly destroys the valuation of what it produces.

Odoo declines to value the consumption of goods belonging to a partner -- `stock.move._is_out()` skips any line carrying an `owner_id` -- and then values the finished product anyway, at whatever the components cost, which by then is nothing. Every such order books stock out of thin air. Seen on a production database: a product sitting at 20.00 per unit takes in 10 units made from a partner's material and drops to 16.00, and everything made from it afterwards inherits the collapse.

Naming the owner on the way out stops it at the source. The finished move produces no valuation layer at all rather than a zero-cost one, and the product's cost is left untouched. Both fields are needed and they do different things: `owner_id` on the lines keeps the quantity out of the valuation layer, `restrict_partner_id` on the move keeps it out of the journal entries. This is what `stock.picking._action_done` already does for a picking with an owner.

What each restriction means for an order is not the same, so the module reads them apart:

- Unassigned owner says no component may belong to anybody. An owned one points at something that went around the reservation, and blessing it would hand the finished product to a partner nobody chose. Refused.
- Picking partner says every component comes from that partner. Company stock among them is equally anomalous and has no answer. Refused.
- Picking partner or unassigned owner is the one mode where mixing is the designed behaviour -- it reserves from the partner and completes with unowned stock -- so a mix is expected and what comes out goes to the partner. Leaving it unowned would put their share back at zero cost, which is the problem this module exists to stop.
- Components of two different partners have no answer under any of them.

It also gives the raw material of an order a partner to be restricted to. `_get_owner_for_assign` reads it off the picking, and a manufacturing order has none, so it resolved to nothing -- which under Picking partner does not mean "unrestricted" but "restricted to nobody": the order reserved zero and manufacturing stopped with nothing on screen explaining why. A production carries no partner of its own either (`mrp.production` has no `partner_id` or `owner_id`, only the responsible `user_id`), so it follows the same chain the base method already follows for the finished product: whoever the order ends up delivering to. Manufacturing to stock chains to nothing and keeps resolving to no partner, which is right for it.

A bridge rather than code in `stock_owner_restriction` because this needs `mrp` in earnest -- two behaviours, a validation that refuses to post, and tests that build orders -- and guarding all of it behind field-presence checks read worse than declaring the dependency.

@Tecnativa


yostashiro:
@carlosdauden There is a preceding PR. Maybe you could review this one instead? https://github.com/OCA/manufacture/pull/1518

carlosdauden:
Thanks @yostashiro, you are right — I had not found
OCA/manufacture#1518 before opening this.

Closing in favour of it. Having read that module, it already covers what this one
was for, including assigning the finished product to the owner of what it was
made from, which is what stops the order from booking stock at zero cost. On top
of that it handles unbuild and MO merging, which this one did not.

I have left a review there with two things that look like they need attention
before the migration lands, and offered the one piece of this module that does
not overlap: refusing to validate when the components come from more than one
source, since there is no way to express partial ownership of what comes out.

PR #2302, opened 2026-04-07
[18.0][MIG] stock_picking_variable_qty: Migration to 18.0 

chienandalu:
Superseds:

- https://github.com/OCA/stock-logistics-workflow/pull/2183

@moduon MT-12702

PR #2421, opened 2026-08-05
[18.0][FIX] stock_picking_variable_qty: keep chained moves in sync

chienandalu:
This was meant to be part of the supersed made in #2302 but the superseded PR got merged and now we need to amend it so the module it's functional

Keep chained move quantities aligned when variable quantities are processed across multi-step routes. Partial backorders now split downstream moves correctly, including when several origins feed the same destination, while keeping the existing propagate_variable_qty setting unchanged.

Also refreshes the module documentation and translations and adds regression coverage for the backorder and multi-origin flows.

cc @moduon MT-12702

please review @EmilioPascual @Shide


Shide (approved):
Code and Functional Review

chienandalu:
ping @EmilioPascual

EmilioPascual (approved):

rafaelbn (approved):
/ocabot merge minor
Thank you David!

PR #2425, opened 2026-08-12
[FIX] stock_move_source_relocate: reserve after the source is relocated

divad1196:
PR #2291 removed the `_action_confirm` call that used to re-run the reservation on the relocated moves.
=> Relocated moves stay unreserved.

Fix: Re-run _action_assign on the relocated moves instead. Ignore the relocation rules to avoid infinite recursion.

Add test to reproduce the bug and confirm its resolution

### DIFF
--- a/stock_move_source_relocate/models/stock_move.py
+++ b/stock_move_source_relocate/models/stock_move.py
@@ -37,6 +37,12 @@ class StockMove(models.Model):
                 relocated_moves |= relocated
         if relocated_moves:
             relocated_moves._after_apply_source_relocate_rule()
+            # The source location changed
+            # => Reservation must be tried again on it.
+            relocated_moves.exists().with_context(
+                # Skip the relocation rules to avoid recursing.
+                exclude_apply_source_relocate=True
+            )._action_assign()
 
     def _apply_source_relocate_rule(self, relocation):
         """Perform the source location relocation.
--- a/stock_move_source_relocate/tests/test_source_relocate.py
+++ b/stock_move_source_relocate/tests/test_source_relocate.py
@@ -137,6 +137,30 @@ class TestSourceRelocate(SourceRelocateCommon):
             ],
         )
 
+    def test_relocate_reserve_in_new_location(self):
+        # the rule applies on the parent location, so the relocation location
+        # is not searched by the first reservation attempt made on Shelf 1
+        self._create_relocate_rule(
+            self.loc_shelf, self.loc_replenish, self.wh.pick_type_id
+        )
+        self._update_qty_in_location(self.loc_replenish, self.product, 10)
+        move = self._create_single_move(
+            self.product,
+            self.wh.pick_type_id,
+            custom_vals={"location_id": self.loc_shelf_1.id},
+        )
+        self.assertRecordValues(
+            move,
+            [
+                {
+                    "state": "assigned",
+                    "product_qty": 10.0,
+                    "quantity": 10.0,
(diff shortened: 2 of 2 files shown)


stock_move_source_relocate/models/stock_move.py
```diff
@@ -37,6 +37,12 @@ def _apply_source_relocate(self):
                 relocated_moves |= relocated
         if relocated_moves:
             relocated_moves._after_apply_source_relocate_rule()
+            # The source location changed
+            # => Reservation must be tried again on it.
+            relocated_moves.exists().with_context(
+                # Skip the relocation rules to avoid recursing.
+                exclude_apply_source_relocate=True
+            )._action_assign()
```
TDu, line 45: Makes sense, but should it not be done after L19 in the `_action_assign ?

divad1196: @TDu `_action_assign` must be re-called whenever `_apply_source_relocate` is called. Even if `_apply_source_relocate` is used elsewhere.

the `if relocated_moves` code is already there to call `_after_apply_source_relocate_rule`. It makes sense to put the `_action_assign` re-trigger there as well

jbaudoux (commented):
Relocation occurs when no reservation can be made. 
A new reservation after the relocation makes sense if the new location is not a child of the initial location. I don't know if you can configure it that way. Looking t yours test, apparently yes. However, we don't use it the module that way as far as I know.

Cc @rousseldenis

divad1196:
@jbaudoux We experienced a regression following PR #2291

> if the new location is not a child of the initial location

Reservation retry must be done even if the new location is a child
We typically have this scenario with reservation rules: https://github.com/OCA/stock-logistics-reservation/tree/18.0/stock_reserve_rule

If a relocation happens, it's because we expect the new location to be able to provide.
Otherwise, there is no reason for the relocation to happen in the first place.

jbaudoux:
> If a relocation happens, it's because we expect the new location to be able to provide. Otherwise, there is no reason for the relocation to happen in the first place.

No it's not the purpose. It's to place the demand on a child location of the initial location to trigger a replenishment on that child location.

jbaudoux (changes requested):
The location must be a child of the initial location. So this doesn't make sense. According to me, this PR on this module should be closed.

See https://github.com/OCA/stock-logistics-workflow/blob/18.0/stock_move_source_relocate/models/stock_source_relocate.py#L55

divad1196:
> this PR on this module should be closed.


### Customer flow
1. Picking created -> first attempt to reserve -> fail (due to reservation rule)
2. Reservation failure triggers relocation -> change the source location of the move
3. Reservation runs again (This step got removed in PR #2291)

Step 3 is now broken due to #2291, which looked like a bug fix `[FIX]`, not a behavior change, which is a breaking change.
This PR restores a behavior that was lost by mistake in PR #2291 


### Question

Is there a reason, within this module, to not attempt a reservation?
Was removing this behavior actually done on purpose?


### To clarify my point
> If a relocation happens, it's because we expect the new location to be able to provide.
> Otherwise, there is no reason for the relocation to happen in the first place.

Relocation only happens when the first reservation fails. 
https://github.com/OCA/stock-logistics-workflow/blob/8f65ce3441e7fb1758a8240e22a55a19e7c84329/stock_move_source_relocate/models/stock_move.py#L26

If the reservation was done, there would be no need to relocate since we already have the stock.

PR #2427, opened 2026-08-13
[16.0][FIX] stock_move_line_change_lot: don't break mrp

hbrunn:
when mrp [posts](https://github.com/odoo/odoo/blob/16.0/addons/mrp/models/mrp_production.py#L1587) an inventory, a [recordset](https://github.com/odoo/odoo/blob/16.0/addons/mrp/models/mrp_production.py#L2350) is written on stock.move.line#lot_id, which while uncommon, is [supported](https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L3046) by the ORM.

browse however doesn't support this, so we need to convert this to a type it understands


lmignon (approved):
LGTM (Code review)

PR #2431, opened 2026-08-14
[FIX] stock_restrict_lot: avoid MissingError in _check_lot_consistent_with_restriction

myrrkel:
## Problem

When `_action_done` converts a product into a kit, the original move is unlinked and replaced by its component moves. The `self` recordset that flows through `_action_done` still references those deleted moves when `_check_lot_consistent_with_restriction` iterates over it, raising a `MissingError`.

## Reproduction

This is reproduced by the standard MRP test:

```
mrp_account.tests.test_mrp_account.TestMrpAccount.test_delivery_validate_after_product_converted_to_kit
```

The test delivers a product that gets converted to a kit, then validates the delivery. With `stock_restrict_lot` installed, the post-`_action_done` consistency check iterates over already-unlinked moves and fails with `MissingError`.

## Fix

Iterate only over the moves that still exist (`self.exists()`) in `_check_lot_consistent_with_restriction`, skipping the records unlinked during `_action_done`.

### DIFF
--- a/stock_restrict_lot/models/stock_move.py
+++ b/stock_restrict_lot/models/stock_move.py
@@ -121,7 +121,11 @@ class StockMove(models.Model):
         Check that the lot set on move lines
         is the same as the restricted lot set on the move
         """
-        for move in self:
+        # ``_action_done`` may unlink some of the moves it processed (e.g. when a
+        # product is converted to a kit, the original move is replaced by its
+        # component moves), so ``self`` can still reference deleted records here.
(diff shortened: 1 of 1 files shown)


lmignon (changes requested):
Thank you for the FIX.  I would prefer make the check on exists prior to the call to `_check_lot_consistent_with_restriction`-> `self.exists()._check_lot_consistent_with_restriction()` This check is required for this specific context only.

PR #2436, opened 2026-08-24
[19.0][FIX] stock_move_line_expiration_date_required

MarinaAForgeFlow:
Two fixes:

- [FIX] stock_move_line_expiration_date_required: adapt _generate_serial_move_line_commands signature

Core takes (field_data, location_dest_id=False, origin_move_line=None) and
callers do pass location_dest_id by keyword. The override still carried the
16.0 signature, so any such call raises a TypeError once this module is installed.

- [FIX]  stock_move_line_expiration_date_required: empty date on generated lots

Since https://github.com/odoo/odoo/commit/4f339301179c product_expiry defaults
on the lines built by the 'Generate/Import Lots or Serials' dialog. For a product
that uses expiration dates with no expiration time configured, that date is
the receipt date itself, which is precisely what this module exists to avoid.

### DIFF
--- a/stock_move_line_expiration_date_required/tests/test_expiration_date_required.py
+++ b/stock_move_line_expiration_date_required/tests/test_expiration_date_required.py
@@ -1,6 +1,8 @@
 # Copyright 2024 Moduon Team S.L. <info@moduon.team>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 
+from datetime import datetime, timedelta
+
 from odoo import fields
 from odoo.exceptions import UserError
 from odoo.tests import Form, common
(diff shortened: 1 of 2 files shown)


stock_move_line_expiration_date_required/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Stock Move Line Expiration Date Required",
     "Summary": "Expiration date is required to enter manually on Move Lines.",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
```
AaronHForgeFlow: don't change the version number, the bot will do it

stock_move_line_expiration_date_required/tests/test_expiration_date_required.py
```diff
@@ -234,3 +258,46 @@ def test_sanity_check_raises_when_missing_expiration_date(self):
             self.product.display_name,
         ):
             self.picking.with_context(skip_sanity_check=False)._sanity_check()
+
```
AaronHForgeFlow, line 261: Perhaps we can consolidate the new tests, 1 test for each fix, that would be easier to read.

AaronHForgeFlow (commented):
code LG, some minor comments

stock_move_line_expiration_date_required/tests/test_expiration_date_required.py
lmignon: I prefer 1 test by case to test ;-)

lmignon (approved):
Except the manual bump of the version, the code LGTM. Thank you for the fix

PR #2438, opened 2026-08-27
[16.0][IMP] stock_picking_group_by_partner_by_carrier: avoid fetching all

lmignon:
group pickings to check closed state

_merge_procurement_groups() loaded every picking of the procurement group through the ORM (up to a hundred for the busiest "merged" groups, in a stock_picking table with several million rows) just to filter on printed/state in Python afterwards and detect if one is closed.... Replace this with an indexed search_count limited to 1 result, which only performs an existence check: the cost stays roughly constant regardless of the group size, instead of growing with how long-lived a customer's group is.


jbaudoux (approved):

BhaveshHeliconia (approved):

lmignon:
ping @jbaudoux

lmignon:
thank you @jbaudoux !

PR #2441, opened 2026-09-01
[16.0][IMP] stock_picking_batch_start: allow starting/cancelling batch when partially eligible

nicolas-delbovier-acsone:
Previously, `action_start_allowed` and `action_cancel_start_allowed` required ALL pickings
in the batch to satisfy the start or cancel condition (`all()`), and calling `action_start` or
`action_cancel_start` would attempt to process all pickings unconditionally (raising a `UserError`
if any individual picking was not allowed to transition).

### DIFF
--- a/stock_picking_batch_start/tests/test_stock_picking_batch_start.py
+++ b/stock_picking_batch_start/tests/test_stock_picking_batch_start.py
@@ -2,7 +2,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import Command
-from odoo.exceptions import UserError
 from odoo.tests.common import TransactionCase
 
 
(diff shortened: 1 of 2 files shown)


lmignon (approved):

stock_picking_batch_start/tests/test_stock_picking_batch_start.py
```diff
@@ -92,28 +91,23 @@ def _update_qty_in_location(cls, location, product, quantity):
     def test_00(self):
         """test action start and user are propagated"""
         self.env.company.stock_picking_assign_operator_at_start = True
+
```
BhaveshHeliconia, line 94: avoid adding extra empty lines, as they create unnecessary diff.

BhaveshHeliconia (approved):
LGTM!

Just minor suggestion

stock_picking_batch_start/tests/test_stock_picking_batch_start.py
nicolas-delbovier-acsone: Yeah sorry, when working on this PR, I felt like adding those empty lines helped with the readability of the test but I can remove them if you prefer so.

PR #2442, opened 2026-09-01
[19.0][FIX] stock_picking_invoice_link: restrict the invoice link to invoicing users

jcadhoc:
## Problem

A user with **Inventory / User** access and no accounting access cannot open an
internal transfer once this module is installed. The form fails with:

```
AccessError: Failed to read field stock.picking.invoice_ids
You are not allowed to access 'Journal Entry' (account.move) records.
```

The picking form injects `invoice_ids` (m2m to `account.move`) and the *Invoice*
stat button with no `groups`. Reading a many2many triggers a search on the
comodel, and warehouse profiles have no read access on `account.move`, so an
accessory piece of information ends up blocking the core operation of the role.

The same happens on the **stock.move** form with `invoice_line_ids`
(`account.move.line`).

## Why `groups` on the field was not enough

`invoice_line_ids` already carried `groups="account.group_account_invoice"`, but
the surrounding `<group>` has an `invisible="invoice_line_ids == []"` modifier.
View postprocessing adds the field back as an invisible node so the modifier can
be evaluated:

```xml
<field name="invoice_line_ids" invisible="True" readonly="True"
       data-used-by="invisible='invoice_line_ids == []' (group,invoice_line)"/>
```

so the field is read anyway and the form still breaks. The fix puts `groups` on
the node that carries the modifier, which removes both the modifier and the
field.

## Change

`views/stock_view.xml` only:

- picking form: `groups="account.group_account_invoice"` on the `invoice_ids`
  field and on the stat button (the button holds the
  `invisible="invoice_count == 0"` modifier, so it goes away with it);
- move form: `groups` moved from the `invoice_line_ids` field to the enclosing
  `<group name="invoice_line">`.

Nothing changes for users who can already see invoices.

## Test plan

On a 19.0 database with this module installed, a user whose only groups are
`base.group_user` + `stock.group_stock_user`:

- **before**: `picking.read(<fields exposed by stock.view_picking_form>)` raises
  `AccessError` on `account.move`; the same on `stock.view_move_form` raises on
  `account.move.line`;
- **after**: `invoice_ids` / `invoice_count` / `invoice_line_ids` are no longer
  present in the arch returned by `get_view()` for that user, and reading every
  field the forms expose succeeds;
- a user with `account.group_account_invoice` still gets the stat button, the
  invoice lines group and `invoice_count`, and reaches the invoice from the
  picking.

The list views (`stock.vpicktree`, `stock.view_move_tree`) never exposed these
fields, and were verified not to raise for the restricted user.

The same asymmetry is present on 18.0; a backport can follow once this is
reviewed.


mav-adhoc (approved):

les-adhoc (approved):

stock_picking_invoice_link/__manifest__.py
```diff
@@ -7,7 +7,7 @@
 
 {
     "name": "Stock Picking Invoice Link",
-    "version": "19.0.1.0.1",
+    "version": "19.0.1.0.2",
```
BhaveshHeliconia, line 10: No need to bump the version manually; it will be bumped automatically when the PR is merged.

BhaveshHeliconia (approved):
LGTM!

les-adhoc (approved):
LGTM!

ajimenez-pyming (commented):
LGTM!

stock_picking_invoice_link/__manifest__.py
jcadhoc: Ready for merge!

PR #2443, opened 2026-09-02
[18.0][ADD] stock_valuation_layer_removal_date

AungKoKoLin1997:
This module adds a Removal Date field to stock valuation layers, taken from the lots/serials of the related stock move.

Companies handling perishable goods need to see how much of their stock value is about to expire, and how much of the value they consumed came from stock close to its removal date. `product_expiry` stores the removal date on the quants of a lot, which covers the stock on hand, but the valuation layers keep no trace of it once the goods have moved. Reaching it from a layer means walking through the stock move and its move lines up to the lots, which cannot be filtered or grouped on.

Design notes for review:

- The lots are read from the stock move rather than from the layer, because the standard Lot/Serial Number of a layer is only filled for products with lot valuation enabled. Layers of a lot valuated product do carry their own lot, and take the date from it.
- When a move holds several lots, the earliest removal date is kept. That is the date that matters for the exposure the field is meant to report; a single date per layer cannot express more.
- The field is a stored computed one, so it realigns itself when the removal date of a lot changes or when the lots of the move change. `compute_sudo=True` is needed because valuation layers are only writable by stock managers while lot dates are maintained by any stock user, and it also lets the realignment reach the layers of the other companies that a company-neutral lot can have.
- Adding the column would otherwise make the ORM compute the field for every existing layer on installation, so a `pre_init_hook` creates and fills it in SQL. `test_fill_removal_date_hook` checks that the SQL and the compute agree.

Assisted-by: Claude Opus 5

@qrtl QT7121


stock_valuation_layer_removal_date/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,29 @@
...
+        "lot_id.removal_date",
+        "stock_move_id.move_line_ids.lot_id.removal_date",
+        "stock_move_id.move_line_ids.quantity",
+    )
+    def _compute_removal_date(self):
+        for layer in self:
+            lots = layer.lot_id or layer.stock_move_id.lot_ids
+            layer.removal_date = min(
+                filter(None, lots.mapped("removal_date")), default=False
+            )
```
nobuQuartile, line 29: If layer.lot_id is set, it should use layer.lot_id.removal_date.
```suggestion
def _compute_removal_date(self):
    for layer in self:
        if layer.lot_id:
            layer.removal_date = layer.lot_id.removal_date
        else:
            removal_dates = layer.stock_move_id.lot_ids.mapped("removal_date")
            layer.removal_date = min(
                filter(None, removal_dates),
                default=False,
            )
```

nobuQuartile (commented):
Partial review

stock_valuation_layer_removal_date/models/stock_valuation_layer.py
nobuQuartile: It's my misunderstanding.

stock_valuation_layer_removal_date/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="stock_valuation_layer_view_search" model="ir.ui.view">
+        <field name="name">stock.valuation.layer.search</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.view_inventory_valuation_search" />
+        <field name="arch" type="xml">
+            <filter name="group_by_lot_id" position="after">
```
yostashiro, line 18: Enable filtering, not only group-by.

stock_valuation_layer_removal_date/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.depends(
+        "lot_id.removal_date",
+        "stock_move_id.move_line_ids.lot_id.removal_date",
+        "stock_move_id.move_line_ids.quantity",
+    )
+    def _compute_removal_date(self):
+        for layer in self:
+            lots = layer.lot_id or layer.stock_move_id.lot_ids
+            layer.removal_date = min(
```
yostashiro, line 27: Perhaps this min() deserves a comment in ROADMAP.md?

stock_valuation_layer_removal_date/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, fields, models
+
+
+class StockValuationLayer(models.Model):
+    _inherit = "stock.valuation.layer"
+
+    removal_date = fields.Datetime(
+        compute="_compute_removal_date",
+        store=True,
+        index=True,
```
yostashiro: Since there can be many records with no value in this field.
```suggestion
        index=btree_not_null,
```

stock_valuation_layer_removal_date/models/stock_valuation_layer.py
AungKoKoLin1997: Done!

stock_valuation_layer_removal_date/views/stock_valuation_layer_views.xml
AungKoKoLin1997: Done!

stock_valuation_layer_removal_date/models/stock_valuation_layer.py
AungKoKoLin1997: Done!

stock_valuation_layer_removal_date/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="stock_valuation_layer_view_search" model="ir.ui.view">
+        <field name="name">stock.valuation.layer.search</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.view_inventory_valuation_search" />
+        <field name="arch" type="xml">
+            <filter name="outgoing" position="after">
+                <separator />
+                <filter name="month" string="Removal Date" date="removal_date" />
```
yostashiro: This filter won't be so useful for users' operations as is, as it won't allow filtering for future periods. Perhaps pass attributes such as end_month and end_year?

yostashiro (approved):
Code review. LGTM.

nobuQuartile (approved):
LGTM: Code review and functional review

PR #2446, opened 2026-09-03
[ADD] stock_location_validate_inventory

jbaudoux:
This module adds a technical user group "Location Validate Inventory" to allow the validation of the inventory of a location.

In standard, only users in the "Inventory Manager" group can perform this and only if there is a quant on the location as you need to apply a counted quantity.

This module provides an action on a location to validate that the current inventory is right. This will reset any counted differences and update the location last inventory date.

cc @lmignon 

Implementation note: As there is no header in the stock.location form view to add a button, I added a contextual Action.


lmignon (approved):
LGTM (Code review + Functional tests)

stock_location_validate_inventory/models/stock_location.py
```diff
@@ -0,0 +1,24 @@
...
+        location last inventory date.
+        """
+        if not (
+            self.env.user.has_group(
+                "stock_location_validate_inventory.group_stock_location_can_validate_inventory"
+            )
+        ):
+            raise UserError(_("You are not allowed to validate the location inventory"))
+        self.quant_ids.action_set_inventory_quantity_to_zero()
+        self.sudo().last_inventory_date = fields.Date.today()
```
BhaveshHeliconia:
```suggestion
        self.sudo().last_inventory_date = fields.Date.today()
        date_by_location = {loc: loc._get_next_inventory_date() for loc in self}
        for quant in self.quant_ids:
            quant.inventory_date = date_by_location[quant.location_id]
            quant.user_id = False
```
Since quant.inventory_date is stored, updating the location doesn't recompute it. Updating the quants directly and clearing user_id prevents them from staying stuck as overdue in Inventory Adjustments.

stock_location_validate_inventory/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+
+{
+    "name": "Location Validate Inventory",
+    "summary": """Validate the inventory of a location""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "BCIM, ACSONE SA/NV, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/stock-logistics-workflow",
+    "depends": ["stock"],
+    "data": [
```
BhaveshHeliconia, line 13: Please add yourself as a maintainer.

stock_location_validate_inventory/models/stock_location.py
lmignon: The goal is to track the inventory date of empty locations, even when there is no quant.

BhaveshHeliconia: Understood! If the intention is strictly to record the inventory date at the location level for empty locations, then the current implementation is good to go

jbaudoux: @BhaveshHeliconia Thanks for the feedback. I added the case when the location is not empty to update the quants last count date and next inventory date.

jbaudoux:
I need to add some tests :)

stock_location_validate_inventory/models/stock_location.py
BhaveshHeliconia: Great!

stock_location_validate_inventory/__manifest__.py
jbaudoux: done

jbaudoux:
> I need to add some tests :)

Test added

BhaveshHeliconia (approved):
LGTM!

Thanks for the module!

lmignon (approved):

PR #2449, opened 2026-09-04
[16.0][ADD] stock_quant_lock: add lock/unlock support for stock quants

lmignon:
This addon lets you lock individual stock quants to prevent them from being consumed or allocated by stock operations. It is designed for use cases where some quantities must remain untouched and protected from automatic replenishment or reservation logic.


lmignon:
Ping @jbaudoux. The first part is done, but I'm keeping this in draft until I complete the second part of our discussion and make sure everything integrates smoothly..

stock_quant_lock/models/stock_quant.py
```diff
@@ -0,0 +1,230 @@
...
+            return False
+
+    def _lock_with_picking_type(self, picking_type):
+        """Lock the quant using the specified picking type.
+
+        This will create a stock move of the specified picking type
+        and reserve the remaining available quantity for the quant.
+        """
+        self._check_is_lock_with_picking_type_allowed(picking_type, raise_error=True)
+        lock_move_vals = self._prepare_lock_move_vals(picking_type)
```
jbaudoux, line 168: If you create first a picking with the src/dest of the picking type and use it in the move creation, then all moves will be in the same picking.
I think it's a bit better when you lock a lot on several locations.

jbaudoux (approved):
Thanks for making this generic module out of old code.

I would tend to prefer to have one transfer per action with multiple moves in it instead of having multiple transfers.

jbaudoux:
@sbidoul names to reserve on pypi: stock_quant_lock, stock_picking_operation_loss_quantity, shopfloor_loss_quantity

lmignon:
@OCA/pypi-support names to reserve on pypi: stock_quant_lock, stock_picking_operation_loss_quantity, shopfloor_loss_quantity :smirk:

sbidoul:
I reserved odoo_addon_stock_quant_lock. Ping if the merge on others fails.

PR #2452, opened 2026-09-10
[19.0][ADD] stock_picking_portal_tracking_reference: see tracking information to the portal view

arielbarreiros96:
This module helps visualizing the tracking information of a picking, and easily accessing the tracking references and URL from the portal when avaialable.

In a nutshell, it adds these two buttons for easy access to tracking (if the picking does not provide url is just a label without link or button):
Captura de pantalla 2026-09-10 a las 14 59 01
Captura de pantalla 2026-09-10 a las 14 59 26


pedrobaeza:
Isn't that something of `delivery`? And in `delivery_status` I think there's something about that, or it's a good place to put it.

arielbarreiros96:
Hello @pedrobaeza this module basically exposes the fields carrier_tracking_ref and carrier_tracking_url on the portal as well. I've placed the PR on stock-ligistics-workflow because the module it uses and extends (stock_picking_portal) is also under this repo. I thought also that delivery-carrier was a good place, but had to make a decision 😅.

> Isn't that something of `delivery`? And in `delivery_status` I think there's something about that, or it's a good place to put it.

pedrobaeza:
OK, I would got to not using `stock_picking_portal` at all, as sales order portal view already list the pickings and let them to print, and you can add the information there (if not present already).

arielbarreiros96:
> OK, I would got to not using `stock_picking_portal` at all, as sales order portal view already list the pickings and let them to print, and you can add the information there (if not present already).

True, checked and it does all that, even opens the tracking link. In that sense is redundant, It was more for expanding the stock_picking_portal to also include that capability when checking the pickings from the portal as an standalone feature. But the info lives and works on the sales portal view so this module is technically redundant 😅

pedrobaeza:
OK, as said, I feel `stock_picking_portal` as redundant in this case.

arielbarreiros96:
> OK, as said, I feel `stock_picking_portal` as redundant in this case.

I understand the argument, though `stock_picking_portal` is the only module I know of, that allows to see all stock.picking records on a single place regardless of their sale order, even if they don't have a sale order attached, this could be useful for other workflows like some parts of 3PL. The alternative would be deprecating this feature right?

Captura de pantalla 2026-09-10 a las 20 08 20

pedrobaeza:
If you have a case where you need to expose directly the pickings without sales orders, it's OK then.

PR #2456, opened 2026-09-16
[18.0][FIX] stock_move_line_expiration_date_required: skip pending lines

chienandalu:
Partial receipts can retain detailed lines for quantities that have not been received yet. Requiring expiry dates on those lines prevents users from validating the quantities they actually received.

Check only processed lines after receipt processing starts, while preserving expiry validation before any line is marked as processed.

Also:

- Fixed overriden docstring
- Fixed missing parameter in call to super!

cc @moduon MT-15942

please review @Shide @EmilioPascual


Shide (approved):
Code review

EmilioPascual (approved):

PR #2461, opened 2026-09-22
[16.0][FIX] stock_quant_lock: action_unlock_quant on empty recordset unlock…

lmignon:
…ed all quants

_get_lock_move_domain() falls back to a domain matching every locked move when self is empty, so calling action_unlock_quant() with no records ended up cancelling the lock moves of all quants instead of doing nothing. Guard against the empty recordset case.


lmignon:
ping @jbaudoux @nicolas-delbovier-acsone

nicolas-delbovier-acsone (approved):
Code review, LGTM

